PR #1, opened 2018-11-27
Add license

seanpmorgan:
@karmel @ewilderj 

Can you confirm this is the correct license (copied from tensorflow/tensorflow)?


armando-fandango:
Sounds like this is the correct one :-)

PR #14, opened 2019-01-15
Implementation of Normalizations

Smokrow:
First draft of a file for Normalizations implementing #6 . 
I have Implemented GroupNorm, InstanceNorm and LayerNorm and a first testcase for GroupNorm( I will add a few more and will also implement some for Layer/Instance norm).

Could you give me a quick feedback on the implementation or if something is missing?

Thx in advance

PR #18, opened 2019-01-16
Rename WeightNorm -> WeightNormalization

lgeiger:
This renames the `WeightNorm` wrapper to `WeightNormalization` in order to be consistent with  [`tf.keras.layers.BatchNormalization`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/layers/normalization.py#L774).

See https://github.com/tensorflow/addons/pull/14#discussion_r248224896


seanpmorgan:
Thanks. I have another PR in the works to optimize this for TF2, but no need to wait to rename

PR #19, opened 2019-01-17
Implement Maxout layer

### LINKED ISSUE #9: Implement Maxout Layer
Per the [RFC](https://github.com/tensorflow/community/blob/393dfc9e03dd4150a839de6357a26e0259aa97d6/rfcs/20181214-move-to-addons.md), we need to move [maxout](https://github.com/tensorflow/tensorflow/blob/v1.12.0/tensorflow/contrib/layers/python/layers/layers.py#L3097) from contrib to addons:

This will involve restructuring in an OO format by inheriting from base Keras Layer, modifying the code to match those APIs, and modifying test cases to run in TF2.

https://arxiv.org/abs/1302.4389

facaiy:
Fix #9

### DIFF
--- a/tensorflow_addons/layers/python/maxout_test.py
+++ b/tensorflow_addons/layers/python/maxout_test.py
@@ -21,58 +21,50 @@ from __future__ import print_function
 import numpy as np
 
 from tensorflow.python.keras import testing_utils
-from tensorflow.python.keras.utils import generic_utils
 from tensorflow.python.platform import test
 from tensorflow_addons.layers.python.maxout import Maxout
 
 
 class MaxOutTest(test.TestCase):
     def test_simple(self):
-        # TODO: more simple way to deserialize the layers in addons.
-        with generic_utils.custom_object_scope({'Maxout': Maxout}):
-            testing_utils.layer_test(
-                Maxout, kwargs={'num_units': 3}, input_shape=(5, 4, 2, 18))
+        testing_utils.layer_test(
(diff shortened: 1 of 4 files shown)


tensorflow_addons/layers/python/maxout_test.py
```diff
@@ -0,0 +1,79 @@
...
+from tensorflow.python.keras import testing_utils
+from tensorflow.python.keras.utils import generic_utils
+from tensorflow.python.platform import test
+from tensorflow_addons.layers.python.maxout import Maxout
+
+
+class MaxOutTest(test.TestCase):
+    def test_simple(self):
+        # TODO: more simple way to deserialize the layers in addons.
+        with generic_utils.custom_object_scope({'Maxout': Maxout}):
```
facaiy: tf.keras cannot recognize custom layers, I think there are at least two ways to mitigate the problem:
1. We inject our custom layers into `_GLOBAL_CUSTOM_OBJECTS` of keras (or use `CustomObjectScope.__enter__()`) when importing. Users don't need do anything.
2. We create a `custom_objects` for addons.layers, and ask users to pass them to `Model.from_config(config, custom_objects)` explicitly.

facaiy: Do we need to create an issue in addons to discuss the problem?

seanpmorgan: At first glance I lean toward option 1. Perfectly okay with an issue to discuss it though. I started #23 so we could formalize this requirement once a decision is made

PR #28, opened 2019-02-07
Implement triplet_semihard_loss

facaiy:
Related with: #26

### DIFF
--- a/tensorflow_addons/losses/BUILD
+++ b/tensorflow_addons/losses/BUILD
@@ -8,6 +8,7 @@ py_library(
         "__init__.py",
         "python/__init__.py",
         "python/triplet.py",
+        "python/metric_learning.py",
     ],
     srcs_version = "PY2AND3",
 )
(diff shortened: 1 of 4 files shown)


facaiy:
I'd like to create a new PR to move the custom object register, what do you think? 

And I agree that we can decorate all the methods with `@tf.function`. I'm not sure whether it's right time to do it, however, because [keras.losses](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/losses.py) hasn't decorated its functions. By the way, if the base class `Loss` uses `tf.function` for its `__call__` method in the future, it seems not necessary?

seanpmorgan:
No issue with a separate PR for moving the register.. it'll need to be applied to all our keras objects anyways.

For the function decorator, I don't see a reason we should hold back just because it's not in tf.keras yet. We're building against tf2-nightly and adapting to TF2 best practices so I think we should include it. The base class implementing @tf.function on `__call__` is a good point... but then are all of these other methods such as `pairwise_distance` never meant to be called outside of Triplet Loss? If so, I would prefer them to have am underscore prefix (But I can imagine it could be used outside of this loss function)

facaiy:
Make sense, I agree with you. And I'll decorate the methods with tf.function in a few days :-) I need to attend a wedding tomorrow.

facaiy:
Fixed. Thank you :-)

PR #31, opened 2019-02-08
Update repository description

seanpmorgan:
I don't think `bleeding edge` contributions should be a requirement for Addons. Many things in this repo have been around for a while, but are useful to a subset of the community and shouldn't be installed with every TF installation.

Thoughts? Certainly happy to re-word.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ not available in core TensorFlow. TensorFlow natively supports
 a larger number of operators, layers, metrics, losses, and optimizers.
 However, in a fast moving field like ML, there are many interesting new
 developments that cannot be integrated into core TensorFlow
-(because their significance is not yet clear, or it is mostly used by a smaller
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,10 +1,11 @@
-**Addons** is a repository of bleeding edge contributions that conform to
+**Addons** is a repository of contributions that conform to
 well-established API patterns, but implement new functionality
 not available in core TensorFlow. TensorFlow natively supports
 a larger number of operators, layers, metrics, losses, and optimizers.
 However, in a fast moving field like ML, there are many interesting new
 developments that cannot be integrated into core TensorFlow
-(because they are experimental, or their significance is not yet clear).
+(because their significance is not yet clear, or it is mostly used by a smaller
```
karmel: "significance" seems so judgmental-- "scope"? Broad applicability? Just a suggestion-- fine as is, too.

PR #36, opened 2019-02-13
Add cpu testing script

seanpmorgan:
Wanted to start this PR to discuss a few things that need fixing before merge.

1. There appears to be no `tf-nightly-2.0-preview` for python3.4. Ubuntu 14.04 which our docker image is built from has this as the pre-installed python3. Would it be possible to get this the preview packaged on PyPi?

2. Is splitting it up into a cpu / tpu / gpu file the best way to do this. I figured that it's better than a single file because not all testing instances will have all hardware.

3. Is it okay to run multiple bazel test in a single script? I'm unsure how the CI instance reads test outputs.


karmel:
1. The Preview build for 3.x is failing right now. This should get fixed shortly.
2. Good question. I don't think we can set up CI for TPUs for Github repos right now, so we can at least take that off the table. For CPU/GPU, we should just bear in mind that the testing is combinatorial-- so, py2/3 * cpu/gpu. I think doing those four is reasonable, but I wouldn't want to go beyond that. The other option is to run both CPU and GPU on a GPU-enabled node by masking the GPUs for the CPU portion, but I don't know that that's better.
3.  IIUC, that should work. I think there may need to be some funneling of exit status, though, to make sure the tests report failure if any of  the tests fail. We do something like that [here](https://github.com/tensorflow/models/blob/master/official/utils/testing/scripts/presubmit.sh) with Python scripts. CC @av8ramit to correct me if I am wrong.

av8ramit:
So you can run the bazel test command twice, but I don't think the internal CI will be able to handle the logs  well. Based on exit status it will only determine pass/fail. Will do some more digging to confirm.

av8ramit:
Confirmed with @yifeif who may have more insight.

yifeif:
I think you do need to propagate the exit status.

karmel:
@yifeif -- there is an example of a testing setup (build.sh) that looks like this: 

```
chmod +x "bazel_wrapper.py"
"bazel_wrapper.py" \
    --host_jvm_args=-Dbazel.DigestFunction=SHA256 \
    test \
    --test_timeout="300,450,1200,3600" \
    --build_tests_only \
    --test_output=errors \
    --verbose_failures=true \
    --keep_going \
    --remote_accept_cached=true \
    --remote_local_fallback=false \
    --remote_timeout=600 \
    --spawn_strategy=remote \
    --strategy=Javac=remote \
    --strategy=Closure=remote \
    --genrule_strategy=remote \
    -- //TESTS_TO_RUN/...
```
Would that work in this case, with the listed tests in that last line?

karmel:
Spoke offline with @yifeif -- it sounds like what we want to do here is run the bazel wrapper from our side, and target this script. If you remove the set -e, you can run multiple tests, and just funnel the exit code such that you return one code. Without the -e, failed tests will not stop all processing, and the next lines will still run.

@yifeif -- is that a correct restatement?

ci_testing/run_cpu.sh
```diff
@@ -0,0 +1,52 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# ==============================================================================
+
+set -e
```
karmel: As noted in the main comment thread, we may want to remove this.

ci_testing/run_cpu.sh
```diff
@@ -0,0 +1,52 @@
...
+
+export CC_OPT_FLAGS='-mavx'
+export TF_NEED_CUDA=0 # TODO: Verify this be used in GPU custom-op?
+
+######### Python3 Tests
+# TODO: tf-nightly-2.0-preview is unavailable for python3.4
+
+
+#export PYTHON_BIN_PATH=`which python3`
+#/bin/bash configure.sh --py_version=3
```
karmel: I think it will be easier if we set the python from the script that calls this one. That way, we have a few scripts here, and we can run the same script for py2/3. We can also install the tf preview from there-- but up to you; maybe it's better to pip install here, so that if you want to change TF versions (eg when it is no longer the preview), you can control that from here. That would also work presumably if we have separate GPU testing scripts, as you can install the GPU package in the same way.

ci_testing/run_cpu.sh
```diff
@@ -0,0 +1,52 @@
...
+## Run bazel test command. Double test timeouts to avoid flakes.
+#bazel test -c opt -k \
+#    --jobs=${N_JOBS} --test_timeout 300,450,1200,3600 \
+#    --test_output=errors --local_test_jobs=8 \
+#    //tensorflow_addons/...
+
+
+######### Python2 Tests
+export PYTHON_BIN_PATH=`which python2`
+/bin/bash configure.sh --py_version=2
```
karmel: You can assume you're in a virtualenv with either py2 or 3 depending on the test that is running. You can also assume you are at the addons project root.

ci_testing/run_cpu.sh
```diff
@@ -0,0 +1,52 @@
...
+
+## Run bazel test command. Double test timeouts to avoid flakes.
+bazel test -c opt -k \
+    --jobs=${N_JOBS} --test_timeout 300,450,1200,3600 \
+    --test_output=errors --local_test_jobs=8 \
+    //tensorflow_addons/...
```
karmel: Might be worthwhile to create a dummy test file with one passing test and one failing test, so that when we hook up the pieces we can see what the output looks like.

ci_testing/run_cpu.sh
seanpmorgan: Because our BUILD depends on a `.bazelrc` file that is created by `configure.sh` it's probably best that it's called in the script. I also like the idea of controlling the TF version because of issues like https://github.com/tensorflow/addons/pull/37 where it's important that we pin to a specific version once they're available.

ci_testing/run_cpu.sh
seanpmorgan: Sounds good. I originally wanted a script that would test py2 + py3 in the docker image for local development. I'm okay with simplifying it though because we can depend on the CI for testing new PRs.

ci_testing/run_cpu.sh
seanpmorgan: Added a shell script `run_tests_failing.sh`  that causes a failure by adding some text. Is that sufficient to debug with?

PR #38, opened 2019-02-18
Contribution docs

seanpmorgan:
Major doc changes...
* Added project core concepts to central README
* Added table of contents to central README
* Began separate contribution guidelines for each sub-package API

Just a note.. there is a planned PR coming which will place all "custom op" sub-packages under a single API type so that we can have contribution guidelines similar to `layers`, `losses`, and `optimizers`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ developments that cannot be integrated into core TensorFlow
 (because their broad applicability is not yet clear, or it is mostly
  used by a smaller subset of the community).
 
-## Content
+## Contents
 | Sub-Package    | Addon  | Reference                                  |
 |:----------------------- |:----------- |:---------------------------- |
 | addons.image | Transform |                                           |
--- a/README.md
+++ b/README.md
@@ -1,43 +1,82 @@
-**Addons** is a repository of contributions that conform to
+# TensorFlow Addons
+
+TensorFlow Addons is a repository of contributions that conform to
 well-established API patterns, but implement new functionality
 not available in core TensorFlow. TensorFlow natively supports
-a larger number of operators, layers, metrics, losses, and optimizers.
+a large number of operators, layers, metrics, losses, and optimizers.
 However, in a fast moving field like ML, there are many interesting new
 developments that cannot be integrated into core TensorFlow
-(because their broad applicability is not yet clear, or it is mostly used by a smaller
-subset of the community).
+(because their broad applicability is not yet clear, or it is mostly
+ used by a smaller subset of the community).
 
-# Scope
-The tensorflow/addons repository, will contain additional functionality fitting the following criteria:
+## Content
+| Sub-Package    | Addon  | Reference                                  |
+|:----------------------- |:----------- |:---------------------------- |
+| addons.image | Transform |                                           |
+| addons.layers | Maxout | https://arxiv.org/abs/1302.4389             |
+| addons.layers | PoinareNormalize | https://arxiv.org/abs/1705.08039  |
+| addons.layers | WeightNormalization | https://arxiv.org/abs/1602.07868 |
+| addons.losses | TripletLoss | https://arxiv.org/abs/1503.03832       |
+| addons.optimizers | LazyAdamOptimizer | https://arxiv.org/abs/1412.6980 |
+| addons.text | SkipGrams | https://arxiv.org/abs/1301.3781 |
 
-* The functionality is not otherwise available in TensorFlow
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -1,43 +1,82 @@
...
 developments that cannot be integrated into core TensorFlow
-(because their broad applicability is not yet clear, or it is mostly used by a smaller
-subset of the community).
+(because their broad applicability is not yet clear, or it is mostly
+ used by a smaller subset of the community).
 
-# Scope
-The tensorflow/addons repository, will contain additional functionality fitting the following criteria:
+## Contents
+| Sub-Package    | Addon  | Reference                                  |
```
facaiy: The table seems redundant, and also might become a too long list to maintain in the future.

tensorflow_addons/losses/README.md
```diff
@@ -1,4 +1,19 @@
...
+| TripletLoss | https://arxiv.org/abs/1503.03832 |
 
-## Standard API
\ No newline at end of file
+
+## Contribution Guidelines
+#### Standard API
+In order to conform with the current API standard, all losses
+must:
+ * Inherit from `keras.losses.LossFunctionWrapper`.
```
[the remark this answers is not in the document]
seanpmorgan: Hmmm, no issue with adding that but want to clarify why it's necessary. My understanding of why we need to do this is so that Keras can properly serialize/deserialize objects. Are losses / optimizers saved as part of keras models? If so we should probably it to layers, losses, and optimizers

README.md
seanpmorgan: I agree this is not sustainable in the long term. I would like a single list of whats available so new users can quickly see if what they need is available, without having to dig through the repo. Are you okay with keeping this until we have setup our Python API docs to show package contents.

For example: https://pytorch.org/docs/master/torchvision/

tensorflow_addons/losses/README.md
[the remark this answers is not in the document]
seanpmorgan: Updated so we register all keras objects.

karmel:
@dynamicwebpaige -- it would be great to have your review here given your work on this for TF.

dynamicwebpaige:
😄 This is a great start! I'd also like to see sections on:

* **Governance and support** (ex: the `addons` repo will periodically run CI tests and if `package` fails to meet criteria for `x` time period, it will be deprecated).

* **Proxy maintainership** (responsibilities and expectations).

* **Preferred methods of communication** (ex: the Gitter channel, mailing list, SIG Addons meeting).

 * A link to [**the 1:1 symbols map**](https://docs.google.com/spreadsheets/d/1FLFJLzg7WNP6JHODX5q8BDgptKafq_slHpnHVbJIteQ/edit#gid=0) (though ideally a markdown version).

* How users can **file requests for packages to migrate**.

* A link to the [`tf.contrib` analyzer](https://tf-contrib-analyzer.herokuapp.com/) for ideas on which packages would be useful to target next.

Does that make sense? I'm not sure if this would fit here, or elsewhere in Addons. Am also happy to share a copy of the contribution guide we've been working on internally for TensorFlow.

seanpmorgan:
Thanks so much for the review @dynamicwebpaige 

1) Great idea, at the upcoming meeting I'd like to discuss an RFC so we can clarify with the community our method of deprecating / removing.

2) Proxy maintainership I'd also like to be tied into that RFC. For example if there is no proxy maintainer that will have implications on the removal

3) I'm planning on adding an image to the gitter channel after our nightly CI is set and it can be included in those images. [Example here](https://github.com/tensorpack/tensorpack/blob/master/README.md)

4) I'm open to the idea of having that somewhere in the repo for a temporary amount of time. I think this repo has a vested interest in becoming an independent repo not a guide on TF2.0 conversion.

5) Great idea. I've [created an issue](https://github.com/tensorflow/addons/issues/44) to create a template for issue submissions.

6) Also open to this idea for a temporary time. I also think this repository has an interest in becoming more than the new location of tf.contrib. I'm also concerned about the last update to that page being ~Nov 2018. Do we know if we can ask for an updated pull?

seanpmorgan:
Merging this as its blocking a new repository structure I'd like to submit. Still entirely open to discussing the comments though and I think the repo plans on many more iterations of our documentation.

seanpmorgan:
Also would love to tie in those new contribution guidelines @dynamicwebpaige

README.md
facaiy: Sorry for the delay. Great, that's what I think. We can remove it after API docs is set up in the future.

PR #43, opened 2019-02-20
minor fixes to configure.sh

armando-fandango:
- added version check for TensorFlow
- added check for existence of .bazelrc before deleting it

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -24,7 +24,7 @@ function write_action_env_to_bazelrc() {
 [ -f .bazelrc ] && rm .bazelrc
 
 # TODO: Verify the tensorflow version here...
-TENSORFLOW_VERSION=$(python -c 'import tensorflow as tf; print(tf.__version__)')i
+TENSORFLOW_VERSION=$(python -c 'import tensorflow as tf; print(tf.__version__)')
 if [[ $TENSORFLOW_VERSION == 2* ]] 
 then
     echo 'using installed tensorflow'
--- a/configure.sh
+++ b/configure.sh
@@ -21,7 +21,7 @@ function write_action_env_to_bazelrc() {
   write_to_bazelrc "build --action_env $1=\"$2\""
 }
 
-rm .bazelrc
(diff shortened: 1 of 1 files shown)


configure.sh
```diff
@@ -21,10 +21,12 @@ function write_action_env_to_bazelrc() {
   write_to_bazelrc "build --action_env $1=\"$2\""
 }
 
-rm .bazelrc
+[ -f .bazelrc ] && rm .bazelrc
 
 # TODO: Verify the tensorflow version here...
```
seanpmorgan: Thanks for knocking this out. Mind removing this TODO?

configure.sh
```diff
@@ -21,10 +21,12 @@ function write_action_env_to_bazelrc() {
...
-if python -c "import tensorflow" &> /dev/null; then
+TENSORFLOW_VERSION=$(python -c 'import tensorflow as tf; print(tf.__version__)')
+if [[ $TENSORFLOW_VERSION == 2* ]] 
+then
     echo 'using installed tensorflow'
 else
     pip install tf-nightly-2.0-preview
```
facaiy: Perhaps we don't need to check tf version, just always `pip install --upgrade  tf-nightly-2.0-preview`?

seanpmorgan: Good point, that's valid while we're still in preview. Once we have releases I think having a check will be a good idea though so I'm okay with adding it now.

facaiy: You're right, I agree! How about creating requirements.txt? with `pip install -r requirements.txt`?

seanpmorgan: I like that idea, its much more pythonic. @armando-fandango Could you modify the script so it always runs a pip install from `requirements.txt`, and then just specify the version in the requirements file.

The check for `.bazelrc` is still a good addition

armando-fandango: Yes, thanks for the review and suggestions. Let me move pip install to requirements.txt.

seanpmorgan:
This is now stale and is included in #62

PR #48, opened 2019-02-22
[DO NOT MERGE] Testing PRs.

av8ramit:

PR #49, opened 2019-02-22
ENH: Restructure custom ops

seanpmorgan:
Starting from a discussion in our last monthly call... It makes the repository significantly easier to maintain if each contribution is binned into an API format. Some contributions are not derived from something as standardized as a Keras API though so we want a category where we could build more fluid guidelines. It also makes the directory structure more approachable from the root `tensorflow_addons` directory. That was a problem in the tf.contrib directory. 

I would imagine this being the final directory layout for our first release. That's not to say we won't have more additions, but that these 4 directories will round out the APIs and we can start building more thorough docs for all of them


seanpmorgan:
@av8ramit could you confirm that the python2 is failing because of CI debugging? It's passing for me locally and I know you're working with #48

tensorflow_addons/custom_ops/image/cc/kernels/image_projective_transform_op.cc
```diff
@@ -19,7 +19,7 @@ limitations under the License.
 #define EIGEN_USE_GPU
 #endif  // GOOGLE_CUDA
 
-#include "tensorflow_addons/image/cc/kernels/image_projective_transform_op.h"
+#include "tensorflow_addons/custom_ops/image/cc/kernels/image_projective_transform_op.h"
```
facaiy: how about put it in the last include line, like gpu.cu.cc below?

tensorflow_addons/custom_ops/text/python/skip_gram_ops_test.py
```diff
@@ -24,8 +24,8 @@
 import os
 
 from tensorflow.python.framework import test_util
-from tensorflow_addons.text.python import skip_gram_ops
-from tensorflow_addons import text
+from tensorflow_addons.custom_ops.text.python import skip_gram_ops
```
facaiy: swap those two lines?

tensorflow_addons/custom_ops/image/__init__.py
```diff
@@ -12,13 +12,10 @@
...
+"""Image manipulation ops"""
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-# Skip Gram Sample
-from tensorflow_addons.text.python.skip_gram_ops import skip_gram_sample
-from tensorflow_addons.text.python.skip_gram_ops import skip_gram_sample_with_text_vocab
+# Transforms
+from tensorflow_addons.custom_ops.image.python.transform import transform
```
facaiy: Although It's convenient for user, I'm a little worried about this usage, which will initialize all sub-modules as soon as user import it. Perhaps `tf.export` is a better solution, but anyway, I think it's accepted for now. Thanks :-)

karmel:
The test error is indeed in the testing setup ( @av8ramit , for your reference: `chmod: cannot access ‘/.../bazel_wrapper.py’: No such file or directory`) -- I think it's safe to trust your local testing for now.

tensorflow_addons/custom_ops/image/__init__.py
seanpmorgan: Good point. IMO user experience is more important until it becomes a major issue, but I could be persuaded otherwise with some performance numbers

Issue #25, opened 2019-01-18
Implement Image functions


seanpmorgan:
Per the [RFC](https://github.com/tensorflow/community/blob/393dfc9e03dd4150a839de6357a26e0259aa97d6/rfcs/20181214-move-to-addons.md), we will be moving several of the [image modules](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/image
) from contrib to addons:

- [ ] dense_image_warp
- [ ] random_hsv_in_yiq
- [ ] adjust_hsv_in_yiq
- [ ] rotate
- [ ] translate
- [ ] angle_to_projective_transforms
- [ ] translations_to_projective_transforms
- [ ] transform
- [ ] compose_transforms
- [ ] flat_transforms_to_matricies
- [ ] connected_components

This will involve moving the code base and modifying test cases to run in TF2.

tabshaikh:
Cool i will began with this then :)

facaiy:
@tabshaikh Thanks, Tabish. Don't put all in one PR. You can solve those functions one by one, say, `rotate` at first.

tabshaikh:
Okay sure

facaiy:
@tabshaikh Hi, Tabish. Which op is you working on?

tabshaikh:
I have started with rotate

WindQAQ:
Hi @tabshaikh, I'm going to work on `dense_image_warp` and I hope that you have not started on this one.

hksonngan:
Is anyone working on connected_components? If not I will start with it?

facaiy:
@hksonngan Great, go head! All contributions are welcome, thanks :-)

tabshaikh:
@seanpmorgan i was just checking random_hsv_in_yiq does not have any tests in contrib https://github.com/tensorflow/tensorflow/search?q=random_hsv_in_yiq&unscoped_q=random_hsv_in_yiq ?

facaiy:
@tabshaikh  Yeah, Tabish, that's a problem. I think @WindQAQ are working on `random_hsv_in_yiq` in #67. 

At the first glance, it seems that `test_adjust_random_value_in_yi` are used for both `random_hsv_in_yiq` and `adjust_hsv_in_yiq` implicitly:
https://github.com/tensorflow/addons/pull/67/files#diff-9e079d46248934679daeb7fd5f3a5835R71

[...]

PR #52, opened 2019-02-25
Migrate dense_image_warp op

WindQAQ:
Address #25. CI test passed on my local machine. Note that the test of `_interpolate_bilinear` in `dense_image_warp` will fail if I use the decorator `tf.function` on it (that's why I haven't added it yet). I'm still trying to figure out why.


WindQAQ:
Forget to create a new branch. Recreate a PR later.

Issue #16, opened 2019-01-15
Standardize formatting


seanpmorgan:
Would like to standardize repository style using a code formatter. @facaiy suggested https://github.com/google/yapf which looks like it would work very nicely. 

In the future this should be tied in with our CI testing

facaiy:
yapf cannot handle document correctly, I once used docformatter tools: https://pypi.org/project/docformatter/ 
- currently automatically formats docstrings to follow a subset of the PEP 257 conventions.

lgeiger:
One other very popular code formatter to consider would be https://github.com/ambv/black

seanpmorgan:
Thanks, I've used black extensively on other projects and love it. I didn't mention it because I didn't want to argue for its departures from PEP8, but it's certainly an option.

facaiy:
@karmel Hi, Karmel. Do you have any suggestion?

karmel:
There are two pieces here-- autoformatting, and linting when we run tests to block merges with lint errors.

I think you should pick the tool you all prefer, though it's complicated by the fact that TF adheres to the [Google Python style guide](http://google.github.io/styleguide/pyguide.html), which deviates slightly from PEP8 (ie, two-space tabs). Ideally, then, a formatter that is configurable would be best.

For TF Models, we didn't set up an autoformatter, but we did set up a [pylint check](https://github.com/tensorflow/models/blob/master/official/utils/testing/scripts/presubmit.sh#L44) with a [custom rcfile](https://github.com/tensorflow/models/blob/master/official/utils/testing/pylint.rcfile) that approximated Google's style guidelines. We can copy that here if you want, or use whatever you think best.

facaiy:
@karmel Thank your for the input. I think it's better to stick with PEP 0008, which is supported by nearly all python toolchains (IDE, text editor, etc). 

Yes, both yapf and pylint are configurable. I'll submit a shell script later for the issue.

PR #64, opened 2019-03-03
use code_format.sh to format all files

facaiy:
Related with #16 

Fix all codes style.

```
# Format all code files:
bash tools/ci_build/code_format.sh --in-place
```


tools/ci_build/install/ci_requirements.txt
```diff
@@ -1,4 +1,4 @@
-pylint >= 1.6.4, < 2.0; python_version < '3.4'
-pylint >= 2.0; python_version >= '3.4'
-docformatter >= 1.1
-yapf >= 0.26.0
+# TODO: Upgrade pylint to 2.0 when python >= 3.4
+pylint == 1.6.4
```
facaiy: 1.6.4 is too out-of-date. Hope that we could upgrade pylint to 2.0 later.

PR #66, opened 2019-03-04
DOC: LazyAdamOptimizer example

seanpmorgan:
Simple example using LazyAdamOptimizer tested after installing the dev pip package. After we publish a 0.1 release we can install it in colab and have interactive examples.


seanpmorgan:
cc @karmel . Do you mind giving this a once over? I'd like to have a small example in the first release branch.

Going forward we'll strive for colab notebooks as well, and more interesting examples.

karmel:
@qlzh727 -- any chance you can take a look? I am unlikely to have time before the Dev Summit.

qlzh727:
Ack, I will take a quick look.

seanpmorgan:
Ah, sorry I forgot today is probably quite hectic. Probably doesn't need more than a glance its just MNIST with the LazyAdamOptimizer. I tested it and it converges fine

tensorflow_addons/examples/tfa_optimizer.py
```diff
@@ -0,0 +1,85 @@
...
+
+    :return: Dictionary of data split into train/test/val
+    """
+    dataset = {}
+
+    # Load MNIST dataset as NumPy arrays
+    (x_train, y_train), (x_test, y_test) = tf.keras.datasets.mnist.load_data()
+
+    # Preprocess the data
+    x_train = x_train.reshape(60000, 784).astype('float32') / 255
```
qlzh727: The number of example "60000" is a magic number, which probably doesn't worth to be exposed to user (they actually don't care about how many example in MNIST). How about change this to reshape(-1, 784)?

tensorflow_addons/examples/tfa_optimizer.py
```diff
@@ -0,0 +1,85 @@
...
+    x_test = x_test.reshape(10000, 784).astype('float32') / 255
```
qlzh727: same for 10000 here.

tensorflow_addons/examples/tfa_optimizer.py
```diff
@@ -0,0 +1,85 @@
...
+
+    # Reserve 10,000 samples for validation
+    dataset['x_val'] = x_train[-10000:]
```
qlzh727: Let's put the variable in train/eval/test which is a more standard ordering.

qlzh727: Let's extract the magic number 10000 as a const and put it on the top, so that we only have one number to modify if we change our mind in future.

tensorflow_addons/examples/tfa_optimizer.py
```diff
@@ -0,0 +1,85 @@
...
+    dataset['y_train'] = y_train[:-10000]
+
+    dataset['x_test'] = x_test
+    dataset['y_test'] = y_test
+
+    return dataset
+
+
+if __name__ == "__main__":
+    data = generate_data()
```
qlzh727: Can we move those logic into a method and call it here? The __main__ body should just invoke main() and not include concrete logic.

tensorflow_addons/examples/tfa_optimizer.py
```diff
@@ -0,0 +1,85 @@
...
+    # Train the network
+    history = dense_net.fit(
+        data['x_train'],
+        data['y_train'],
+        batch_size=64,
+        epochs=10,
+        validation_data=(data['x_val'], data['y_val']))
+
+    # Evaluate the network
+    print('\n# Evaluate on test data')
```
qlzh727: Just curious why it need the leading '\n'. Alternatively you can use tf.logging, which will print with same format as other TF logs generated during training.

tensorflow_addons/examples/tfa_optimizer.py
```diff
@@ -0,0 +1,85 @@
...
+import tensorflow_addons as tfa
+
+
+def build_mnist_model():
+    """Build a simple dense network for processing MNIST data.
+
+    :return: Keras `Model`
+    """
+    inputs = tf.keras.Input(shape=(784,), name='digits')
+    net = tf.keras.layers.Dense(64, activation='relu', name='dense_1')(inputs)
```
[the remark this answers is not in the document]
seanpmorgan: Yeah, I have no real preference, but feel that the functional form is more familiar to other frameworks.

tensorflow_addons/examples/tfa_optimizer.py
seanpmorgan: Thanks, done!

tensorflow_addons/examples/tfa_optimizer.py
seanpmorgan: Agree, thanks

tensorflow_addons/examples/tfa_optimizer.py
seanpmorgan: All of the small implementation details were mostly copied from this:
https://github.com/tensorflow/docs/blob/master/site/en/r2/guide/keras/training_and_evaluation.ipynb

seanpmorgan:
@qlzh727 thanks for the review!

PR #70, opened 2019-03-04
move ci_testing into tools dir

facaiy:
@av8ramit @karmel If we change the location of CI script, how will it affect you?


seanpmorgan:
My understanding is that the code from the ci script is copied internally for now so this change should be okay. Will wait for confirmation though.

karmel:
This should be fine-- thanks.

av8ramit:
We are now looking to just invoke a script so that way no internal intervention is always required. So the path change is the only thing that may need to be modified internally.

PR #81, opened 2019-03-08
CLN: remove tf_test_util and keras.test_util

facaiy:


tensorflow_addons/custom_ops/image/python/transform_test.py
```diff
@@ -64,7 +64,6 @@ def test_extreme_projective_transform(self):
                 [[1, 0, 0, 0], [0, 0, 0, 0], [1, 0, 0, 0], [0, 0, 0, 0]],
                 image_transformed)
 
-    @tf_test_util.run_all_in_graph_and_eager_modes
```
facaiy: I'm looking into why the test case cannot run in graph mode.

PR #84, opened 2019-03-10
Added import instructions for addons.

dynamicwebpaige:


README.md
```diff
@@ -62,6 +62,13 @@ To install the latest version, run the following:
 pip install tensorflow-addons
 ```
 
+You will also need TensorFlow 2.0 or higher. To use addons:
+
+```python
+import tensorflow as tf
+import tensorflow_addons
```
seanpmorgan: In the examples and README we use `import tensorflow_addons as tfa` Are you okay with adding that alias? Similar to tensorflow_probability as tfp

dynamicwebpaige: `tfa` works, as well! 🙂 Will make the mod now.

Issue #86, opened 2019-03-12
ImageProjectiveTransform should be named ImageProjectiveTransformV2


andresusanopinto:
**Describe the bug**
- It looks like ImageProjectiveTransform is based on ImageProjectiveTransformV2 present in tf.contrib in TF-1.x.

**Describe the expected behavior**
- Op names should not colide between 1.x and 2.x environments or saved models won't be shareable across them.

andresusanopinto:
Links to ops:

* [tf.contrib.image](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/image/ops/image_ops.cc#L110)
* [tensorflow_addons.image](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/custom_ops/image/cc/ops/image_ops.cc#L95)

Looking at code it really seems that the op "ImageProjectiveTransformV2" from tf.contrib.image was renamed to "ImageProjectiveTransform" (which was a name already taken by other op). I guess this was un-intentional or in an attempt to cleanup the names, but in reality it can break moving models across TF-1.x and TF-2.x.

seanpmorgan:
Thanks @andresusanopinto. I think you are correct that this is a problem even though we don't support 1.x because of saved models.

@karmel @martinwicke Should we maintain all V2 op naming just to be safe?

karmel:
Yes-- we continue to support v1 graphdef/ops in v2, so maintaining separate namespaces will help avoid collisions. Eventually, we will be able to prefix op names with appropriate namespaces (ie, tensorflow_addons:OP_NAME), but we don't do that yet.

seanpmorgan:
Thanks! Adding help wanted, and this bug fix would require reviewing all Ops that we've moved over.

jharmsen:
Looks like there are only three ops in addon so far.

Two match:
- AdjustHsvInYiq [[contrib](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/image/ops/distort_image_ops.cc) | [addon](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/custom_ops/image/cc/ops/distort_image_ops.cc)]
[...]

PR #90, opened 2019-03-13
rename ImageProjectiveTransform to ImageProjectiveTransformV2

jharmsen:
Rename addon op ImageProjectiveTransform to ImageProjectiveTransformV2.

This ensures the op names are consistent between addon and contrib (see #86).

### DIFF
--- a/tensorflow_addons/custom_ops/image/cc/kernels/image_projective_transform_op.cc
+++ b/tensorflow_addons/custom_ops/image/cc/kernels/image_projective_transform_op.cc
@@ -50,12 +50,12 @@ using generator::INTERPOLATION_NEAREST;
 using generator::ProjectiveGenerator;
 
 template <typename Device, typename T>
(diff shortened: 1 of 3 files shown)


jharmsen:
Thank you very much!  What would the next steps be for rebuilding the pip package with this change? 
(We're hoping to use this new package in a Udacity module next week.)

seanpmorgan:
We can build a 0.1.1 version and publish it this weekend if that works. Just a note we're still working on our MacOS and Windows compatibility so that might need to be a disclaimer in the course. (Hoping to have that figured out by a 0.2 or 0.3 release)

jharmsen:
Thanks for the quick review!  Building a new 0.1.1 version would be excellent.  Let me know if I should open an issue or anything to track it.

seanpmorgan:
@jharmsen 0.1.1 has been released to pypi. Thanks for your help!

PR #91, opened 2019-03-14
Restructure custom ops

seanpmorgan:
Based on the upcoming maintainability RFC and feedback about directory structure, I think it's in our best interest to have all sub-packages exist in `tensorflow_addons/` and if they contain custom_ops, only the cc code will be housed in `custom_ops`. This PR has a number of benefits:

* Removes un-necesary "python" folder structure
* Remove the over-extending table of contents
* Sets up RFC for maintainability 
    * Makes image and text a clear "sub-package"
    * Places our maintainer status on the front-page README
* Removes a private API call for resource gathering


facaiy:
> if they contain custom_ops, only the cc code will be housed in custom_ops

Good idea. I have a question, do we expect that `custom_ops` acts like [tensorflow/core](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/core) directory? If that, do we need to create sub-directories (custom_ops/image, custom_ops/text) in the custom_ops?

BUILD
```diff
@@ -7,9 +7,10 @@ sh_binary(
         "setup.py",
         "//tensorflow_addons",
         "//tensorflow_addons/activations:activations_py",
-        "//tensorflow_addons/custom_ops:custom_ops_py",
+        "//tensorflow_addons/image:images_ops_py",
```
facaiy: How about removing all `_py` suffix? Perhaps  `images_ops`, `activations`, `layers` looks better. What do you think, Sean?

tensorflow_addons/custom_ops/image/BUILD
```diff
@@ -35,47 +36,4 @@ cc_binary(
...
-    srcs = [
-        "python/transform_test.py",
-    ],
-    main = "python/transform_test.py",
-    srcs_version = "PY2AND3",
-    deps = [
-        ":images_ops_py",
-    ],
-)
+)
```
facaiy: Please add a line break

tensorflow_addons/custom_ops/text/BUILD
```diff
@@ -18,41 +19,4 @@ cc_binary(
...
-    srcs = [
-        "python/skip_gram_ops_test.py",
-    ],
-    main = "python/skip_gram_ops_test.py",
-    srcs_version = "PY2AND3",
-    deps = [
-        ":text_py",
-    ],
-)
+)
```
facaiy: ditto

tensorflow_addons/image/README.md
```diff
@@ -0,0 +1,23 @@
...
+ * Add the addon to the `py_library` in this sub-package's BUILD file.
+
+#### Testing Requirements
+ * Simple unittests that demonstrate the custom op is behaving as
+    expected.
+ * When applicable, run all unittests with TensorFlow's
+   `@run_in_graph_and_eager_modes` (for test method)
+   or `run_all_in_graph_and_eager_modes` (for TestCase subclass)
+   decorator.
+ * Add a `py_test` to the custom-op's BUILD file.
```
facaiy: sub-package's BUILD file?

tensorflow_addons/text/README.md
```diff
@@ -0,0 +1,23 @@
...
+ * Add the addon to the `py_library` in this sub-package's BUILD file.
+
+#### Testing Requirements
+ * Simple unittests that demonstrate the custom op is behaving as
+    expected.
+ * When applicable, run all unittests with TensorFlow's
+   `@run_in_graph_and_eager_modes` (for test method)
+   or `run_all_in_graph_and_eager_modes` (for TestCase subclass)
+   decorator.
+ * Add a `py_test` to the custom-op's BUILD file.
```
facaiy: ditto

tensorflow_addons/utils/resource_loader.py
```diff
@@ -0,0 +1,38 @@
...
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os
+
+
+def get_project_root():
+    """Returns project root folder."""
+    return os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```
facaiy: Could you create an unit test ? Say, assert `get_path_to_datafile('tf_dependency')` directory exists.

BUILD
seanpmorgan: Agree looks great.

tensorflow_addons/utils/resource_loader.py
seanpmorgan: I would really like a test here because this is specific to the file staying where it exists today.  The problem is I think the unittest will have issue verifying a directory

```
import os
import tensorflow as tf
from tensorflow_addons.utils.resource_loader import get_path_to_datafile


class ResourceLoaderTest(tf.test.TestCase):
    def test_get_project_root(self):
        dependancy_dir = get_path_to_datafile('tf_dependency')
        self.assertTrue(os.path.exists(dependancy_dir))


if __name__ == '__main__':
    tf.test.main()
```
Fails.

And if I force a comparison of the string `dependancy_dir` I see:
```
AssertionError: '/root/.cache/bazel/_bazel_root/e051f2f195071208ea7f081f88730f4f/execroot/__main__/bazel-out/k8-fastbuild/bin/tensorflow_addons/utils/resource_loader_test.runfiles/__main__/tensorflow_addons/tf_dependency' != 'test'
```

facaiy: Oh, I get it. Does it work to check in runtime, I mean, in the `get_project_root` function?

seanpmorgan:
> > if they contain custom_ops, only the cc code will be housed in custom_ops
> 
> Good idea.
> 
> I have a question, do we expect that `custom_ops` acts like [tensorflow/core](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/core) directory? There are only `kernels` and `ops` directories for tensorflow/core.
> If that, do we need to create sub-directories (custom_ops/image, custom_ops/text) in the custom_ops?
> 
> @karmel @dynamicwebpaige Would you mind taking a look? Thanks :-)

I prefer having sub-directories as the BUILD file for core just feels overwhelming. Also it makes removing a subpackage more surgical.

tensorflow_addons/utils/resource_loader.py
seanpmorgan: Hmmm at runtime 'tf_dependecy' won't exist in the package as the pip build is just `tensorflow_addons` and what's below it.  That reminded me that we should have been looking for say `examples` dir, but that also failed in the unit test. 

If `get_path_to_datafile` fails at runtime it'll throw a file not found error on the op load... so while bad it'll be an understandable error and obvious break. Open to adding a test but can't think of one that helps much.

facaiy:
> Also it makes removing a subpackage more surgical.

Oh, it's a good point!

tensorflow_addons/utils/resource_loader.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def get_path_to_datafile(path):
+    """Get the path to the specified file in the data dependencies. The path is
```
facaiy: Please split the docstring:

```python
"""Get the path to the specified file in the data dependencies.

The path is xxx
```

facaiy:
Looks good to merge :-)

tensorflow_addons/text/BUILD
```diff
@@ -0,0 +1,40 @@
...
+    ]),
+    data = [
+        "//tensorflow_addons/custom_ops/text:_skip_gram_ops.so",
+        "//tensorflow_addons/utils:utils",
+    ],
+    srcs_version = "PY2AND3",
+)
+
+py_library(
+    name = "text",
```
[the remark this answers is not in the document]
seanpmorgan: Sounds good to me. Which one would we like to keep? Image currently is called `images_ops`, but I don't know if I like that better than just `image` and `text`. Not super important so if there's one way you like better let me know.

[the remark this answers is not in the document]
facaiy: Both are fine :-) Just `image` and `text`?

PR #93, opened 2019-03-15
DOC: use logo for SIG/addons

facaiy:
Thank @ewilderj :-)


facaiy:
Totally agreed. How about the new one? Dose it look better?

seanpmorgan:
@facaiy I clipped some of the whitespace around the text and uploaded a local file so we don't need to worry about the link. If you like the image better, should we merge it into community or is the static file in repo okay?

facaiy:
I prefer public url to static file in the repo, but both are fine. And I'm afraid svg format cannot play well with markdown, while png format perhaps is better for web :-)

facaiy:
Sean, I'd like to leave it up to you :-)

seanpmorgan:
So I think we'll meet somewhere in the middle. I do think SVG is preferred format for web, but to be fair that wasn't a true svg graphic, just an image imported into svg. 

I'm uncomfortable with the URL link as it's outside of our control and could be broken without our knowledge. So PNG in repo I think is a good middle ground. I could live with the URL link, but I think it needs to be cropped anyway.

PR #103, opened 2019-03-26
Migrate contrastive loss

WindQAQ:
As discussed in #26, `contrastive_loss` is designed as two positional arguments `y_true` and `y_pred`.

Also should we include something like `euclidean_distance()` in addons or add an example script to show how to use `ContrastiveLoss`?

Feel free to ping me if necessary. Thanks!


tensorflow_addons/losses/contrastive.py
```diff
@@ -0,0 +1,79 @@
...
+    the samples of the same label and the embedding to be far apart at least
+    by the margin constant for the samples of different labels.
+
+    See: http://yann.lecun.com/exdb/publis/pdf/hadsell-chopra-lecun-06.pdf
+
+    Args:
+      y_true: 1-D integer `Tensor` with shape [batch_size] of
+        binary labels indicating positive vs negative pair.
+      y_pred: 1-D float `Tensor` with shape [batch_size] of
+        distances between two embedding matrices.
```
facaiy: I'm afraid that we have to explain how to calculate the embedding distance `y_pred` in details, otherwise users might not get it correctly.

tensorflow_addons/losses/contrastive_test.py
```diff
@@ -0,0 +1,111 @@
...
+        self.assertAlmostEqual(loss, 0.0, 3)
+
+    def test_unweighted(self):
+        cl_obj = contrastive.ContrastiveLoss()
+        y_true = tf.constant([0, 0, 1, 1, 0, 1], dtype=tf.dtypes.int64)
+        y_pred = tf.constant([0.1, 0.3, 1.3, 0.7, 1.1, 0.5],
+                             dtype=tf.dtypes.float32)
+        loss = cl_obj(y_true, y_pred)
+        loss = self.evaluate(loss)
+        self.assertAlmostEqual(loss, 0.6216, 3)
```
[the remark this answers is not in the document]
WindQAQ: Sure, I'll elaborate how to compute it. As for tolerance, it seems that keras loss tests adopt 1e-3 tolerance. But indeed, we can use 1e-5 or more precision if needed.

tensorflow_addons/losses/contrastive_test.py
```diff
@@ -0,0 +1,111 @@
...
+# pylint:disable=not-callable
+@test_utils.run_all_in_graph_and_eager_modes
+class ContrastiveLossTest(tf.test.TestCase):
+    def test_config(self):
+        cl_obj = contrastive.ContrastiveLoss(
+            reduction=tf.keras.losses.Reduction.SUM, name='cl')
+        self.assertEqual(cl_obj.name, 'cl')
+        self.assertEqual(cl_obj.reduction, tf.keras.losses.Reduction.SUM)
+
+    def test_all_correct_unweighted(self):
```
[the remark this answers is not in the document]
WindQAQ: If `y_pred = 1`, the loss would try to minimize the distance to 0 (closer); otherwise, the loss try to maximize the distance to `margin`. That's why I call it `all_correct`, meaning that we can obtain minimum loss in this test case. (a quite confusing though).

WindQAQ:
@facaiy Except for `all_correct_unweighted` case (see comments above), I've made some changes. Many thanks for the review!

tensorflow_addons/losses/contrastive_test.py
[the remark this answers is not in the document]
facaiy: Good point! It might be better to keep consistent with [keras: losses_test.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/losses_test.py): all_correct means that  y_true = y_pred (distance). Or choose a new name?

tensorflow_addons/losses/contrastive_test.py
```diff
@@ -0,0 +1,162 @@
...
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow as tf
+from tensorflow_addons.losses import contrastive
+from tensorflow_addons.utils import test_utils
+
+
+# pylint: disable=not-callable
```
facaiy: @WindQAQ Hi, don't use it for the whole block, it's too broad and dangerous.  

```python
cl_obj = conxxxxx.Conxxxx(   # pylint: disable=not-callable
```

tensorflow_addons/losses/contrastive_test.py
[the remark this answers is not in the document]
WindQAQ: Seems that test cases in keras called `*all_correct*` coincide with zero loss. I'd vote for a new name, say `test_zero_loss`, for distinguishing.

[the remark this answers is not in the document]
facaiy: Good name!

facaiy:
Thank you, Tzu-Wei . Please ping me after renaming the test method. I'd like to merge the change, great work :-)

tensorflow_addons/losses/contrastive_test.py
```diff
@@ -0,0 +1,160 @@
...
+        # Weighted loss = [0.81 * 1.2, 0.49 * 0.8, 1.69 * 0.5,
+        #                  0.49 * 0.4, 0 * 1.5, 0.25 * 1.0]
+        #               = [0.972, 0.392, 0.845, 0.196, 0, 0.25]
+        # Reduced loss = (0.972 + 0.392 + 0.845 + 0.196 + 0 + 0.25) / 6
+        #              = 0.4425
+
+        self.assertAllClose(loss, 0.4425)
+
+    def test_zero_weighted(self):
+        cl_obj = contrastive.ContrastiveLoss()
```
facaiy: Oh, we forget it, the name has been used

tensorflow_addons/losses/contrastive.py
```diff
@@ -0,0 +1,97 @@
...
+@tf.function
+def contrastive_loss(y_true, y_pred, margin=1.0):
+    """Computes the contrastive loss between `y_true` and `y_pred`.
+
+    This loss encourages the embedding to be close to each other for
+    the samples of the same label and the embedding to be far apart at least
+    by the margin constant for the samples of different labels.
+
+    The euclidean distances `y_pred` between two embedding matrices
+    `a` and `b` with shape [batch_size, hidden_size] can be computed
```
facaiy: two l2-normalized embedding matrices `a` and `b` ?

WindQAQ: In [keras example](https://github.com/keras-team/keras/blob/master/examples/mnist_siamese.py), the two embedding seems not to be 12-normalized. I also check original paper and find nothing about L2 normalization.

WindQAQ:
Hi, @facaiy, all is done now :-)

tensorflow_addons/losses/contrastive.py
facaiy: I find it in the document of original implementation:
> Embeddings should be l2 normalized.

https://github.com/tensorflow/tensorflow/blob/3cd3277b81406c9de1a02b401b1fd0a7ad72f0b3/tensorflow/contrib/losses/python/metric_learning/metric_loss_ops.py#L99

but it's test case doesn't conform the restriction:
https://github.com/tensorflow/tensorflow/blob/3cd3277b81406c9de1a02b401b1fd0a7ad72f0b3/tensorflow/contrib/losses/python/metric_learning/metric_loss_ops_test.py#L69

And in the original paper
> No assumption is made about the function G_W except that it is differentiable with respect to W

So I think we don't need to require all embedding l2-normalized.

facaiy:
Thank you, Tzu-Wei!

WindQAQ:
It's my pleasure to contribute to addons 😄

PR #95, opened 2019-03-18
Enhance configure: add supporting for customize pip index URL

howl-anderson:
User can set customized  pip index URL (pypi mirror) in configure process for accelerate python package downloading (more then 100M packages need be downloaded)

PR #107, opened 2019-03-28
Add docs for how using PyPi mirror

howl-anderson:
Using default Python Package Index (PyPI, https://pypi.org/) maybe pretty slow due to the network connection. Developers can choose using a PyPI mirror. This PR will add docs for telling developers how to set python using a PyPI mirror as default package source.

Related to  #95

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,6 +63,17 @@ Try those commands below:
 2. Run unit test: `make unit-test`
 3. All of the above: `make`
 
+### Tips
+Using default Python Package Index (PyPI, https://pypi.org/) maybe pretty slow due to the network connection.
+Developers can choose using a PyPI mirror, search engine can help you find nearest and fast PyPI mirror.
+
+Setting python using a PyPI mirror as default package source is easy. For pip version >=10.0.0 (if not, please using `pip install pip -U` to upgrade pip):
+```bash
+pip config set global.index-url https://pypi.tuna.tsinghua.edu.cn/simple
+```
+
+in this example, `https://pypi.tuna.tsinghua.edu.cn/simple` is a PyPI mirror for China.
+
 ## Coding style
 
 Addons provides `make code-format` command to format your changes
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,7 +63,10 @@ Try those commands below:
 2. Run unit test: `make unit-test`
 3. All of the above: `make`
 
-**Tip**: If having trouble with running this script due to network connection, users may find want to consult [this documentation](https://mirror.tuna.tsinghua.edu.cn/help/pypi/) to properly set their pypi repository.
+**Tip**: If having trouble with running this script due to network connection, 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -63,6 +63,17 @@ Try those commands below:
 2. Run unit test: `make unit-test`
 3. All of the above: `make`
 
+### Tips
+Using default Python Package Index (PyPI, https://pypi.org/) maybe pretty slow due to the network connection.
+Developers can choose using a PyPI mirror, search engine can help you find nearest and fast PyPI mirror.
+
+Setting python using a PyPI mirror as default package source is easy. For pip version >=10.0.0 (if not, please using `pip install pip -U` to upgrade pip):
```
facaiy: `pip install -U pip`?

facaiy:
Thanks, Xiaoquan.  

@seanpmorgan Sean, could you take a look and comment?

CONTRIBUTING.md
howl-anderson: It seems `pip install -U pip` looks better than `pip install pip -U`. I will modify it.

CONTRIBUTING.md
```diff
@@ -63,6 +63,17 @@ Try those commands below:
 2. Run unit test: `make unit-test`
 3. All of the above: `make`
 
+### Tips
+Using default Python Package Index (PyPI, https://pypi.org/) maybe pretty slow due to the network connection.
+Developers can choose using a PyPI mirror, search engine can help you find nearest and fast PyPI mirror.
+
+Setting python using a PyPI mirror as default package source is easy. For pip version >=10.0.0 (if not satisfied, please using `pip install -U pip` to upgrade pip):
+```bash
+pip config set global.index-url https://pypi.tuna.tsinghua.edu.cn/simple
```
seanpmorgan: So, I have a small issue including this command in the Tips section for all users. Realistically `configure.sh` is only run once per build and I don't really think the speed of the pypi server is the issue here.  I worry that people will just lazily read through this guide and set their global pypi repo to the university simpy because it was a tip.

I would be okay with merging this if we simplified this to a single line saying something like:

**Tip**: If having trouble with running this script due to network connection, users may find want to [consult this documentation](https://mirror.tuna.tsinghua.edu.cn/help/pypi/) to properly set their pypi repository.

howl-anderson: That's a good idea, more concise.

facaiy: Note the mirror site is designed for programmers in China, how about?

Tip: If having trouble with running this script due to network connection in China, users ...

howl-anderson: Sounds good to me! I already update the doc according to this.

seanpmorgan:
@howl-anderson Thank you very much for the contribution here... but the more I think about it the more I have trouble adding this into the docs. This is very much a global pip issue and really has nothing to do with Addons. If users are getting this error they probably have experienced this in the past and should know the fix.

PR #110, opened 2019-03-28
Implemented bazel python version flags

Smokrow:
I implemented python interpreter flags to make it easier to switch between Python2 and Python3 while developing. 
I added two Bazel Interpreter rules in a BUILD file in the tools directory which points to the specific interpreter in the docker container that we use for development. 
In order to run both test I had to update the ./configure.sh script aswell to download tensorflow with pip3 as well. 

In the end I added a little description to CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,6 +94,16 @@ bazel test -c opt -k \
 --test_output=errors \
 //tensorflow_addons/...
 ```
+This test will test Python2 by default.
+
+To test Python2 or Python3 specifically please run:
+```bash 
+# Python 3
+bazel test --python_top=//tools:python3.4.3 //...
+# Python 2
+bazel test --python_top=//tools:python2.7.6 //...
+```
+
 
 ## Code Reviews
 
--- a/configure.sh
+++ b/configure.sh
@@ -36,6 +36,7 @@ function write_action_env_to_bazelrc() {
 
 [[ -f .bazelrc ]] && rm .bazelrc
 pip install $QUIET_FLAG -r requirements.txt
+pip3 install $QUIET_FLAG -r requirements.txt
 
 TF_CFLAGS=( $(python -c 'import tensorflow as tf; print(" ".join(tf.sysconfig.get_compile_flags()))') )
 TF_LFLAGS=( $(python -c 'import tensorflow as tf; print(" ".join(tf.sysconfig.get_link_flags()))') )
--- a/tools/BUILD
+++ b/tools/BUILD
@@ -1,0 +1,13 @@
+py_runtime(
+	name = "python2.7.6",
+	files = [],
+	interpreter_path = "/usr/bin/python",
+	visibility = ["//visibility:public"]
+)
+
+py_runtime(
+	name = "python3.4.3",
+	files = [],
+	interpreter_path = "/usr/bin/python3",
+	visibility = ["//visibility:public"]
+)


CONTRIBUTING.md
```diff
@@ -94,6 +94,16 @@ bazel test -c opt -k \
 --test_output=errors \
 //tensorflow_addons/...
 ```
+This test will test Python2 by default.
```
armando-fandango: Since Python2 is being sunset in favor of Python3, @seanpmorgan @facaiy @Smokrow What is your opinion on making Python3 testing by default ?

facaiy:
Thanks, good work, Moritz. 
But I'm afraid the requirements are too strict: developers have to install both python 2 (by default) and python 3, and the distribution must be in the specified location (/usr/bin/python, /usr/bin/python3). It might not make sense for all people, say, [anaconda](https://anaconda.org/anaconda/python) users.

As far as I known, there are at least two solutions:
1. Create py2 and py3 virtual environments (or docker images). Test them separately like what we do in the [Continuous build](https://github.com/tensorflow/addons#official-builds).
2. Detect python bin location and let user choose which one to run:  https://github.com/tensorflow/tensorflow/blob/f089b3180ec7ddecd503d6d499cd6977c22b8f11/configure.py#L202

What do you think :-)

CONTRIBUTING.md
facaiy: Hi, Armando, I think we test both py2 and py3. It seems not necessary to make a decision here, which depends on the user's environment

Smokrow:
Yes this sounds really good.  The idea behind this was to have a flag to use inside our development container for quick and easy development. That’s why I placed the rules in the tool directory.

tools/BUILD
```diff
@@ -0,0 +1,13 @@
+py_runtime(
+	name = "python2.7.6",
+	files = [],
+	interpreter_path = "/usr/bin/python",
```
karmel: The flag name here seems overly specific, right? For any given user we are not guaranteed that bin/python == 2.7.6. In fact, it may be py3, right?

configure.sh
```diff
@@ -36,6 +36,7 @@ function write_action_env_to_bazelrc() {
 
 [[ -f .bazelrc ]] && rm .bazelrc
 pip install $QUIET_FLAG -r requirements.txt
+pip3 install $QUIET_FLAG -r requirements.txt
```
karmel: We run ./configure from inside a virtualenv with either py3 or py2 as default, meaning the line on 38 will work appropriately, and this line should in theory work, if redundantly, but also could theoretically result in some strange interactions. Maybe this switching should be handled on the user side with virtualenvs rather than in the script itself? Ie, we provide instructions on setting up a py2 and py3 virtual env with TF (or rather, link to the same in the TF docs). That's closer to best practice for TF in any case.

CONTRIBUTING.md
seanpmorgan: Yeah until 2020 we have to support and test both.

Smokrow:
@seanpmorgan @facaiy how should we proceed? I really like the idea of detecting the current python2 and python3 env automatically. 

Of course we could also setup multiple docker images. What would be the best way to do that?

facaiy:
Moritz, could you reply to Karmel's comments? And could you give more details of the idea of detecting the current python2 and python3 env automatically? I'm not sure if I understand it correctly :-)

tools/BUILD
Smokrow: the container we are currently using uses these python versions. I am not sure what the plan of tensorflow/tensorflow:nightly-custom-op is and how they update the versions. But yea i can change this to specify the version.

configure.sh
Smokrow: Sounds good! I will see would i can do

tools/BUILD
facaiy: Moritz, have you read tensorflow script and know how it solve the problem ? https://github.com/tensorflow/tensorflow/blob/f089b3180ec7ddecd503d6d499cd6977c22b8f11/configure.py#L202

configure.sh
facaiy: Let's remove pip3, Moritz ?

seanpmorgan:
I think the goal here is to have it so developers can quickly test py2 and py3 locally. Because our dev docker image has both we should be able to enable that, but this current implementation will break our CI. Maybe we could re-work this as a script run under something like `tools/dev_testing/addons_py2py3_cpu.sh` which runs the tests 2x. Or just as a separate script in `tools/ci_testing`?

tools/BUILD
Smokrow: Now i have. Thank you for the hint :+1:

Smokrow:
Sounds great.
Are there any plans yet to have our own docker registry for multiple images? 

When we have these up and running we could just do something like "docker run -rm tensorflow/tensorflow-addons-py3 "bazel test //...""   https://www.tensorflow.org/install/docker

facaiy:
We can run the same script in different images / virtual environments, it looks unnecessary to create a mixed script. If developers really need it, they can use two commands like:
```bash
docker run --rm -it -v ${PWD}:/addons -w /addons py2-image make unit-test
docker run --rm -it -v ${PWD}:/addons -w /addons py3-image make unit-test
```

seanpmorgan:
> We can run the same script in different images / virtual environments, it looks unnecessary to create a mixed script. If developers really need it, they can use two commands like:
> 
> ```shell
> # For docker
> docker run --rm -it -v ${PWD}:/addons -w /addons py2-image make unit-test
> docker run --rm -it -v ${PWD}:/addons -w /addons py3-image make unit-test
> # For virtual environment
> # change to python 2 environment
> make unit-test
> # change to python 3 environment
> make unit-test
> ```

Yeah, I think that's just the more practical answer. Thanks for the contribution though Moritz, we could consider an addition to the CONTRIBUTING guidelines showing a succinct way to do this.

PR #120, opened 2019-03-30
Added prompt before install

### LINKED ISSUE #117: Confirm user wishes to install tf2 in build script
Currently if a user decides to build from source we instruct them to run `configure.sh`. During this script we pip install tf2-nightly, though this may not be understood to the user before running. We should add a `yes/no` prompt so this potentially problematic command is understood by the user

https://github.com/tensorflow/addons/blob/master/configure.sh#L38

shreyashub:
Resolves #117. Added yes/no prompt in `configure.sh`

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -35,21 +35,21 @@ function write_action_env_to_bazelrc() {
 }
 
 function ask_before_install() {
-	req_name=`cat requirements.txt`
-	while true; do
-		echo "Installing $req_name.."
-	    read -p "Proceed ?(Y/N) " yn
-	    case $yn in
-	        [Yy]* ) echo "Installing.."; break;;
-	        [Nn]* ) echo "Exiting.."; exit;;
-	        * ) echo "Please answer Y or N.";;
-	    esac
-	done
+	while read -u 3 -r req_name || [[ -n "$req_name" ]]; do
+		while true; do
+			echo "Installing $req_name.."
+		    read -p "Proceed ?(Y/N) " yn
+		    case $yn in
+		        [Yy]* ) pip install $QUIET_FLAG $req_name; break;;
+		        [Nn]* ) echo "Exiting.."; exit;;
+		        * ) echo "Please answer Y or N.";;
+		    esac
+		done
+	done 3< requirements.txt
 }
 
 [[ -f .bazelrc ]] && rm .bazelrc
 ask_before_install
-pip install $QUIET_FLAG -r requirements.txt
 
 TF_CFLAGS=( $(python -c 'import tensorflow as tf; print(" ".join(tf.sysconfig.get_compile_flags()))') )
 TF_LFLAGS=( $(python -c 'import tensorflow as tf; print(" ".join(tf.sysconfig.get_link_flags()))') )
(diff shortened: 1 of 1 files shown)


configure.sh
```diff
@@ -34,7 +34,21 @@ function write_action_env_to_bazelrc() {
...
+	    case $yn in
+	        [Yy]* ) echo "Installing.."; break;;
+	        [Nn]* ) echo "Exiting.."; exit;;
+	        * ) echo "Please answer Y or N.";;
+	    esac
+	done
+}
+
 [[ -f .bazelrc ]] && rm .bazelrc
+ask_before_install
```
armando-fandango: If I say yes to 5 requirements but then say no to 6th, the script will just terminate.

shreyashub: I have made a new commit that fixes that issue. Is there anything else that needs to be done?

seanpmorgan:
Before merging this I just want to make sure that our CI scripts will not break here. @facaiy it looks like we already pass `yes` and `y` to configure.sh?

facaiy:
Thank you for the change, @shreyashub !

@seanpmorgan Sean, I'll take a look at weekend :-)

configure.sh
facaiy: Great work. Can we make it simple? like

```bash
read -r -p "Tensorflow will be upgraded to 2.0. Are You Sure? [Y/n] " reply
case $reply in
    [yY]) echo "Install...";;
    *) echo "Goodbye!"; exit;;
esac
```

facaiy:
> Before merging this I just want to make sure that our CI scripts will not break here. @facaiy it looks like we already pass `yes` and `y` to configure.sh?

Yes,  https://github.com/tensorflow/addons/blob/83195c98032409aeb6ae9a27b4047597a5b04a2b/tools/ci_testing/addons_cpu.sh#L45

configure.sh
shreyashub: @facaiy But won't the script break if we change the `requirements.txt` file later on? If that is fine, then I'll make the changes.

facaiy: Yeah, I agree with you. However it looks kind of boring to ask developers to check each item in the requirements file. That would be great if pip supports dry run, or ask users for confirmation when upgrading, https://stackoverflow.com/questions/29531094/how-to-make-pip-dry-run

Because configure.sh is only used by developers, and we might make docker image for addons itself in the future (we don't need to install python library any more), that's why I want to make the script as simple as possible. Anyway, both solutions are fine with me. I'd like to leave the decision to Sean.

@seanpmorgan Sean, could you talk a look? Thank you :-)

armando-fandango: I agree with Yan that for now we can make it simple to only check for TensorFlow Nightly install instead of every requirement in requirements.txt and then later burn all the requirements in the docker container anyway.

facaiy: Thank you, Armando. @seanpmorgan Hi, Sean, what do you think?

facaiy:
@shreyashub Hi, Shreyas, apologized for the delay. We'll handle the PR as soon as possible, thank you for your patience :-)

configure.sh
seanpmorgan: Yes, I think the just prompting the question so it's understood is good enough. The more simple we can make config.sh the better

shreyashub: @facaiy I have made the prompt simpler as requested.

PR #130, opened 2019-04-02
FIX: correctly link new so

seanpmorgan:
As of https://github.com/tensorflow/tensorflow/pull/22797, the tensorflow shared library name is now versioned. This is the cause of the nightly failures today.

Unfortunately, the so version is still 1.13.1 even for tf2-nightly-preview. I spoke with SIG BUILD and it seems like this is understood and the best short term policy is just to link to so.1, but going forward I will open an issue to see if we can get the shared library name stored in the pip package. Otherwise we will have another breaking nightly once the 2.0 release comes around.


seanpmorgan:
Another thing is that with the on-going push for modularity we should create some bazel wrappers that all can use in the TF community. This relates to the discussion in https://github.com/tensorflow/custom-op/issues/8

seanpmorgan:
Also worth noting that this has implications for when we build against tf2-alpha. It'll have to get switched back on our release branches.

tf_dependency/BUILD.tpl
```diff
@@ -9,8 +9,8 @@ cc_library(
 
 cc_library(
     name = "libtensorflow_framework",
-    srcs = [":libtensorflow_framework.so"],
-    #data = ["lib/libtensorflow_framework.so"],
+    srcs = [":libtensorflow_framework.so.1"],
+    #data = ["lib/libtensorflow_framework.so.1"],
```
karmel: Maybe add comments linking to the explanation in this issue, or summarizing in the comment itself.

PR #131, opened 2019-04-02
Improve AttentionWrapper compatibility with TF 2.0

### LINKED ISSUE #128: Make tfa.seq2seq.AttentionWrapper 2.0 compatible
The `AttentionWrapper` class is currently not compatible with TF2.0:

* it extends `tf.compat.v1.nn.rnn_cell.RNNCell` instead of `tf.keras.layers.AbstractRNNCell`
* it uses `tf.compat.v1.layers.Dense` instead of `tf.keras.layers.Dense`
* attention layers are saved in a `tuple` instead of a `list` which prevent them to be tracked

I can contribute to this change, unless this is a WIP by @qlzh727.

guillaumekln:
* Inherit from `tf.keras.layers.AbstractRNNCell`
* Replace `tf.compat.v1.layers.Dense` by `tf.keras.layers.Dense`
* Store sublayers in `list` instead of `tuple`

Closes #128

### DIFF
--- a/tensorflow_addons/seq2seq/attention_wrapper.py
+++ b/tensorflow_addons/seq2seq/attention_wrapper.py
@@ -1369,7 +1369,8 @@ class AttentionWrapperState(
         Example:
 
         ```python
-        initial_state = attention_wrapper.zero_state(dtype=..., batch_size=...)
+        initial_state = attention_wrapper.get_initial_state(
+            batch_size=..., dtype=...)
         initial_state = initial_state.clone(cell_state=encoder_state)
         ```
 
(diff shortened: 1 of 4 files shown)


qlzh727:
Thanks for the update. Looks great.

qlzh727:
I will run the tests again to filter out any flaky test.

seanpmorgan:
I believe we'll need to wait for #135 to merge for tests to pass

seanpmorgan:
@guillaumekln @qlzh727 CI tests are back. Looks good but some minor linting issues if you could please fix and we can merge

guillaumekln:
The lint issues are from #72. Do we want to fix them in this PR or in another one?

qlzh727:
I think somehow the linter failed to detect that the object is a layer now (callable). We can address the issue in separate PR I think.

seanpmorgan:
So trying to understand why this didn't fail after #72 then? Bazel caching? I'm hesitant to merge this without a fix PR first because all subsequent PRs in Addons will fail for lint and we'll need to review the logs. 

Could we prioritize a PR to fix the issue ahead of this one? I'm okay with a pylint-ignore as long as the linter is incorrect.

guillaumekln:
Ok, let's pass another PR first that silences these warnings.

PR #135, opened 2019-04-03
FIX: Correctly link the new soname for tf2

seanpmorgan:
Bit of a moving target here... but at least the soname is correct for tf2-preview now

### DIFF
--- a/tf_dependency/BUILD.tpl
+++ b/tf_dependency/BUILD.tpl
@@ -8,13 +8,13 @@ cc_library(
 )
 
 # FIXME: Read shared library name from the installed python package.
-# See https://github.com/tensorflow/addons/pull/130 for why we're linking
-# to framework.so.1  in the meantime.
+# See https://github.com/tensorflow/tensorflow/issues/27430 for why we're
+# linking with hardcoded framework.so.2
 
 cc_library(
     name = "libtensorflow_framework",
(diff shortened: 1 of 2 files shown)


seanpmorgan:
Also a question for @av8ramit and @karmel :

Last night the nightlies failed for 3/4 of our builds. The one that passed ran on `tf_nightly_2.0_preview-2.0.0.dev20190402-cp34-cp34m-manylinux1_x86_64.whl`,

 while the others ran on `tf_nightly_2.0_preview-2.0.0.dev20190403-cp27-cp27mu-manylinux1_x86_64.whl`

Is there anyway to delay the jobs we currently have by say 20min-1hr?

qlzh727:
Also @gunan

karmel:
I will push back the nightlies by 1 hour.

karmel:
Sorry, to clarify-- do we want to delay the jobs or make them run earlier? The one that passed was the one that ran with the prior day's TF build (ie, the earliest of the addons runs).

seanpmorgan:
I believe delay so that they all would have failed. I'm not positive but my read was that one of the jobs got launched before the nightly was finished pushing to pypi

karmel:
The time delay was submitted, and should get picked up with the next build.

Issue #77, opened 2019-03-06
Build Addons for MacOS


seanpmorgan:
Currently our template for custom-ops is designed for linux builds. We'll need to adapt our build setup to build for all operating systems.

https://github.com/tensorflow/custom-op

@yifeif @karmel When time allows after the summit, could we discuss the steps needed.

hksonngan:
in Windows in tf.sysconfig.get_link_flags() return empty
and truncate if my username have space at line [38](https://github.com/tensorflow/custom-op/blob/6cacf3adfcb9b42c46619e221a85899fdb8bfaac/configure.sh#L35)
I think we should check  compile flags by build_ext

seanpmorgan:
@yifeif bumping for your feedback, now that the summit is over and GPU op is in place on the custom-op repo. I noticed there is a crosstool build for nvcc on windows so I figured there's been some thought on this.

Without having dug too deeply on tensorflow/tensorflow BUILD, my first question is [how does it work to lookup .so files](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/text/python/ops/skip_gram_ops.py#L35)? Shouldn't it be dependent on OS (i.e. dll/dylib/so)

armando-fandango:
Doesn't autconf or ./configure figure out the dll/dylib/so dependencies ?

armando-fandango:
pip install tf-nightly-2.0-preview fails on mac with the following error:
Could not find a version that satisfies the requirement tf-nightly-2.0-preview (from versions: )
No matching distribution found for tf-nightly-2.0-preview

seanpmorgan:
Had some time to look at the tensorflow build wrappers. Seems like we'll need to do something along these lines for the multi OS build

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tensorflow.bzl#L599

seanpmorgan:
> Doesn't autconf or ./configure figure out the dll/dylib/so dependencies ?

Good point. Yeah the `readlink` or whatever we build with should discover the `libtensorflow_framework`, but the C++ ops we compile ourselves are currently hardcoded as `.so` files.

ageron:
Tried installing from source on my Mac, as instructed on [home page](https://github.com/tensorflow/addons#installing-from-source), but it failed, first because the .so names are hardcoded. I renamed them to dylib. But then I ran into another issue:

```
[...]

PR #139, opened 2019-04-04
[bugfix] unit test on mac

howl-anderson:
addons_cpu.sh cannot work correctly on Mac OS X system due to cannot get CPU core number correctly. this PR fix it.

Also, related with #77 , I build and unit test tensorflow-addons on Mac based on [commit a3fd8ef](https://github.com/tensorflow/addons/commit/a3fd8efd977accec0a68c96c7a49358369f641a1) (not the master branch which is CI test failed when I write this PR): building is OK and unit test passed. 

As a reminder for other developers on Mac, due to the issue bazelbuild/bazel#7241 with bazel, developers should not using virtualenv to test addons on Mac for now.

### DIFF
--- a/tools/ci_testing/addons_cpu.sh
+++ b/tools/ci_testing/addons_cpu.sh
@@ -24,7 +24,14 @@ fi
 
 set -x
 
-N_JOBS=$(grep -c ^processor /proc/cpuinfo)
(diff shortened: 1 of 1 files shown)


seanpmorgan:
Thx!

PR #148, opened 2019-04-07
DOC: Add CODEOWNERS

seanpmorgan:


CODEOWNERS
```diff
@@ -0,0 +1,17 @@
+# TensorFlow Addons Codeowners
+
+# SIG maintainers are the catch all for reviews
+* @tensorflow/sig-addons-maintainers
+
+
+# Subpackage Owners
+seq2seq/* @qlzh727
```
facaiy: `/tensorflow_addons/seq2seq`?

CODEOWNERS
```diff
@@ -0,0 +1,17 @@
...
+
+
+# Submodule Owners
+# These may not automatically trigger a review because submodule owners do not
+# always have write access. But still adding these here for
+# repository understanding.
+
+sparsemax* @AndreasMadsen
```
facaiy: `/tensorflow_addons/activations/sparsemax*` ?

CODEOWNERS
seanpmorgan: So seq2seq is also located in custom_op which is why I used this syntax. https://help.github.com/en/articles/about-code-owners#example-of-a-codeowners-file

CODEOWNERS
seanpmorgan: Same reason... Sparsemax also located in activations, layers, and losses

seanpmorgan:
> Let's move it to `.github` directory?

Agree... moved and updated syntax a bit

CODEOWNERS
facaiy: This synatx will not match further nested files, say `custom/seq2seq/cc/kernel`. Perhaps we can list them line by line: 

```
/tensorflow_addons/seq2sqe qlzh727
/tensorflow_addons/custom_ops/seq2seq qlzh727
```

seanpmorgan: Ah you're correct I mis-read that example. I'll submit a change in a bit here... I think verbosely listing out all managed directories is fine.

PR #157, opened 2019-04-10
updated style guide

armando-fandango:
added link to TensorFlow coding style guide

Issue #156, opened 2019-04-10
Unify styles of custom operations


WindQAQ:
Currently, custom operations in addons have no unified styles. For example
the `name` argument in
https://github.com/tensorflow/addons/blob/463112e6e7c8d3623d3d97e6b1f365d1c403f2c4/tensorflow_addons/image/dense_image_warp.py#L161
and
https://github.com/tensorflow/addons/blob/463112e6e7c8d3623d3d97e6b1f365d1c403f2c4/tensorflow_addons/image/transform_ops.py#L38-L42
are not consistent.

As per [coding style for python operation in Tensorflow](https://www.tensorflow.org/community/contribute/code_style#python_operations), the last argument should be `name` with a default value of `None`. Though not all operations in core Tensorflow conform to this (see `tf.image` source code), I'd suppose we should unify the API/style of python operations in addons (like what we do for losses and layers). Maybe we can try to follow the link above first.

facaiy:
Good catch, I think `None` is better for default value, because it's easy to maintain compatibility.

WindQAQ:
Got it, I'll create a PR for image ops first.

seanpmorgan:
I think this is complete? Are there any other ops that need modifying?

WindQAQ:
Just find we miss some:
https://github.com/tensorflow/addons/blob/c2faddd818280815bd5e47b81742efdfc90e7aaf/tensorflow_addons/image/distance_transform.py#L34-L36
https://github.com/tensorflow/addons/blob/c2faddd818280815bd5e47b81742efdfc90e7aaf/tensorflow_addons/image/transform_ops.py#L320

Some ops like `compose_transforms` in [transform_ops.py](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/transform_ops.py) do not have `name` argument (even when they are in contrib though). Would it be better to add `name` to them for API compatibility?

facaiy:
1. For `rotate` op, we should replace `str(name)` by `(name or 'rotate')`. 
2. There exists an known annoying issue about `name`:  if we want to add new attribute, we have to append it after `name` for ensuring backward compatibility.  And all new attributes are forced to be optional. But I agree to conform tensorflow convention: add `name` for all public API in image module. 
 cc @martinwicke Hi, Martin, do you have any good idea?

@seanpmorgan Sean, what do you think?

martinwicke:
The name arg problem is annoying. 

[...]

PR #158, opened 2019-04-10
unify style for image operations

WindQAQ:
Addresses #156 and corrects some docstrings. It's better to merge this after merging #157.

### DIFF
--- a/tensorflow_addons/image/dense_image_warp.py
+++ b/tensorflow_addons/image/dense_image_warp.py
@@ -22,10 +22,7 @@ import tensorflow as tf
 
 
 @tf.function
-def interpolate_bilinear(grid,
-                         query_points,
-                         name="interpolate_bilinear",
-                         indexing="ij"):
+def interpolate_bilinear(grid, query_points, indexing="ij", name=None):
     """Similar to Matlab's interp2 function.
 
     Finds values for query points on a grid using bilinear interpolation.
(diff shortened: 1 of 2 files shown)


tensorflow_addons/image/dense_image_warp.py
```diff
@@ -189,7 +186,9 @@ def dense_image_warp(image, flow, name="dense_image_warp"):
       ValueError: if height < 2 or width < 2 or the inputs have the wrong
         number of dimensions.
     """
-    with tf.name_scope(name):
+    with tf.name_scope(name or "dense_image_warp"):
+        image = tf.convert_to_tensor(image)
```
facaiy: I'm not sure whether we need to use convert_to_tensor everywhere. cc @karmel @seanpmorgan can you give us some input?

seanpmorgan: I would say that it matches the [pattern seen in core](https://github.com/tensorflow/tensorflow/blob/943a11faa9ec2ca06d9c0185a132c0c5b3e65b3f/tensorflow/python/ops/image_ops_impl.py#L322) so I think it's probably best to keep it unified.

Whether we should be throwing an error on non-tensors instead of just converting them I think is a fair question... but for now this is good IMO

facaiy: Yeah, I have no objection. Just want to figure out when to use it, and when not (or unnecessary).

seanpmorgan: [From the style guide](https://www.tensorflow.org/community/contribute/code_style#python_operations): 
> Operations that take tensors as arguments should call convert_to_tensor to convert non-tensor inputs into tensors if they are using C++ operations. Note that the arguments are still described as a Tensor object of a specific dtype in the documentation.

Not sure if there's a reason there other than to avoid some unfriendly C++ crash?

seanpmorgan:
> Do we change arguments for every python function or only functions decorated with @tf.function ?

All python ops. Most likely are decorated, but don't have to be.
> A Python operation is a function that, given input tensors and parameters, creates a part of the graph and returns output tensors.

PR #160, opened 2019-04-10
Fix minor error in contrastive_loss

WindQAQ:
Minor patch: use `convert_to_tensor` to make sure `dtype` attribute exists.


WindQAQ:
Edit: since keras might support a (nested) list as input, `convert_to_tensor` becomes necessary in this use case.
https://github.com/tensorflow/tensorflow/blob/5991a12e8c717f49caba2cc5673e6c72b0e301f5/tensorflow/python/keras/losses.py#L207-L210

PR #178, opened 2019-04-16
Unify styles of ops

### LINKED ISSUE #156: Unify styles of custom operations

WindQAQ:
Closes #156. Feel free to ping me if there is anything missing.

### DIFF
--- a/tensorflow_addons/image/distance_transform.py
+++ b/tensorflow_addons/image/distance_transform.py
@@ -31,9 +31,7 @@ ops.RegisterShape("EuclideanDistanceTransform")(
 
 
 @tf.function
-def euclidean_dist_transform(images,
-                             dtype=tf.float32,
-                             name="euclidean_distance_transform"):
+def euclidean_dist_transform(images, dtype=tf.float32, name=None):
     """Applies euclidean distance transform(s) to the image(s).
 
     Args:
@@ -54,7 +52,7 @@ def euclidean_dist_transform(images,
         rank 3 or 4.
     """
 
-    with tf.name_scope(name):
+    with tf.name_scope(name or "euclidean_distance_transform"):
         image_or_images = tf.convert_to_tensor(images, name="images")
 
         if image_or_images.dtype.base_dtype != tf.uint8:
--- a/tensorflow_addons/image/transform_ops.py
+++ b/tensorflow_addons/image/transform_ops.py
@@ -123,14 +123,15 @@ def transform(images,
 
 
 @tf.function
-def compose_transforms(*transforms):
+def compose_transforms(transforms, name=None):
     """Composes the transforms tensors.
 
     Args:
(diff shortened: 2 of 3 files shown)


WindQAQ:
Edit: since python2 does not accept this kind of signature:
```python
def foo(*x, name=None):
    pass
```
while python3 does.

Therefore, I modidy `compose_transforms` a little bit.

tensorflow_addons/image/median_filter_2d.py
```diff
@@ -21,7 +21,7 @@
 
 
 @tf.function
-def median_filter_2D(image, filter_shape=(3, 3)):
+def median_filter_2D(image, filter_shape=(3, 3), name=None):
```
WindQAQ: Consider `tf.nn.conv2d`, `scipy.ndimage.median_filter` and `scipy.ndimage.median_filter2d`, should this be one of `median_filter` or `median_filter2d`?

https://www.tensorflow.org/api_docs/python/tf/nn/conv2d
https://docs.scipy.org/doc/scipy/reference/ndimage.html

seanpmorgan: @Mainak431 are you okay with this change? I think its the right change and we can make it with ease before the 0.3 release

Mainak431: Yes. i am ok.

seanpmorgan: @WindQAQ would you like to squeeze this into this PR? I believe we'd prefer `median_filter2d` as there may be a 1d coming in the future

WindQAQ: Another PR would be better for me. In fact, I suppose we could create a `filters.py` or a `filters` folder to put all these filters into one file/folder if many useful filtering-based operations would be coming up. Maybe I can help this in that separate PR.

seanpmorgan: Sounds perfect. That follows the style seen in keras modules.

PR #180, opened 2019-04-16
removed layernorm

### LINKED ISSUE #172: Graduate LayerNormalization layer to TF core.
**System information**
- TensorFlow version (you are using):
- TensorFlow Addons version:
- Is it in the tf.contrib (if so, where):
- Are you willing to contribute it (yes/no):
- Are you willing to maintain it going forward? (yes/no):

LayerBatchNormalization has been moved to core here: https://github.com/tensorflow/tensorflow/blob/43a255d44035f2a68585288876e2ccbb77cb70de/tensorflow/python/keras/layers/normalization.py#L807

Can we graduate it from Addons?

Yes, this will delete the LayerBatchNormalization API in Addons.

**Any Other info.**

Smokrow:
I removed LayerNorm from layers/Normalizations.py since it moved to core. This resolves #172


Smokrow:
damn I knew I forgot something :)

PR #181, opened 2019-04-16
CLN: Simplify Build Script

seanpmorgan:
* Remove unnecessary check for cmake build. We only offer bazel build
* Remove os dependent call to readlink. All this was doing was gathering an absolute directory path

### DIFF
--- a/build_pip_pkg.sh
+++ b/build_pip_pkg.sh
@@ -21,6 +21,7 @@ PIP_FILE_PREFIX="bazel-bin/build_pip_pkg.runfiles/__main__/"
 function abspath() {
   cd $(dirname $1)
   echo $PWD/$(basename $1)
+  cd $OLDPWD
 }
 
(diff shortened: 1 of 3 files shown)


build_pip_pkg.sh
```diff
@@ -16,44 +16,26 @@
...
 PIP_FILE_PREFIX="bazel-bin/build_pip_pkg.runfiles/__main__/"
 
-if [[ ${PLATFORM} == "darwin" ]]; then
-    READLINK="greadlink"  # from pkg `coreutils` in brew/macports in Mac OS X
-else
-    READLINK="readlink"
-fi
-
+function abspath() {
+  cd $(dirname $1)
```
facaiy: could you change back the directory before returning result?

seanpmorgan: Done. thx.

PR #182, opened 2019-04-16
FIX: Explicitly cast minDistance to float

### LINKED ISSUE #177: Euclidean distance op doesn't compile on Mac OS X
It appears that the Euclidian distance op fails to compile on OSX using clang. Throwing this error:

```
euclidean_distance_transform_op.cc:22:
./tensorflow_addons/custom_ops/image/cc/kernels/euclidean_distance_transform_op.h:55:11: error: no viable conversion from 'Eigen::half' to 'float'
    float minDistance = std::numeric_limits<T>::max();
```

The possible types are:
```
struct std::numeric_limits<_Tp>
struct std::numeric_limits<Eigen::half>
struct std::numeric_limits<const Eigen::half>
struct std::numeric_limits<volatile Eigen::half>
struct std::numeric_limits<const volatile Eigen::half>
class std::numeric_limits<tensorflow::qint8>
class std::numeric_limits<tensorflow::quint8>
class std::numeric_limits<tensorflow::qint16>
class std::numeric_limits<tensorflow::quint16>
class std::numeric_limits<tensorflow::qint32>
class std::numeric_limits<absl::uint128>
```

My hunch is that clang wants some explicit conversion, while gcc was okay to just ignore. This would have been caught by our CI, but we do not have it setup to run on OSX.

@Squadrick do you mind taking a look? I have a [branch setup](https://github.com/seanpmorgan/addons/commits/build-macos) for debugging if it helps.

seanpmorgan:
Closes #177

### DIFF
--- a/tensorflow_addons/custom_ops/image/cc/kernels/euclidean_distance_transform_op.h
+++ b/tensorflow_addons/custom_ops/image/cc/kernels/euclidean_distance_transform_op.h
@@ -52,7 +52,7 @@ class EuclideanDistanceTransformGenerator {
 
     if (input_(coords) == T(0)) return T(0);
 
-    float minDistance = std::numeric_limits<T>::max();
+    float minDistance = static_cast<float>(std::numeric_limits<T>::max());
 
     for (int h = 0; h < height_; ++h) {
       for (int w = 0; w < width_; ++w) {


facaiy:
@Squadrick could you take a look? Thanks!

PR #185, opened 2019-04-17
Adding Mean_filter_2D after restructuring

Mainak431:


tensorflow_addons/image/__init__.py
```diff
@@ -23,5 +23,6 @@
 from tensorflow_addons.image.distort_image_ops import adjust_hsv_in_yiq
 from tensorflow_addons.image.distort_image_ops import random_hsv_in_yiq
 from tensorflow_addons.image.filters import median_filter2d
+from tensorflow_addons.image.filters import mean_filter2d
```
seanpmorgan: Please alphabatize mean above median.

tensorflow_addons/image/filters.py
```diff
@@ -109,3 +109,90 @@ def func2():
...
+               C is the second value in the filter is the number of columns
+               in the filter. This creates a filter of shape (R,C) or RxC
+               filter. Default value = (3,3)
+
+     Returns:
+         A 3D mean filtered image tensor of shape [rows,columns,channels] and
+         type 'int32'. Pixel value of returned tensor ranges between 0 to 255
+    """
+
+    def _normalize(li):
```
seanpmorgan: Can we pull this out as two functions now use the same normalize func? Probably should keep `_` prefix and its `tf.function` decorator

tensorflow_addons/image/filters_test.py
```diff
@@ -19,6 +19,7 @@
 
 import tensorflow as tf
 from tensorflow_addons.image import median_filter2d
+from tensorflow_addons.image import mean_filter2d
```
seanpmorgan: ditto: alphabetize

tensorflow_addons/image/filters_test.py
```diff
@@ -109,3 +110,89 @@ def test_three_channels(self):
 
 if __name__ == "__main__":
     tf.test.main()
+
```
seanpmorgan: Looks to be a duplicated `if __name__: test` call here. Can we remove the first one so it's only done once.

tensorflow_addons/image/filters_test.py
```diff
@@ -109,3 +110,89 @@ def test_three_channels(self):
...
+            mean_filter2d(tf_img, (None, 3))
+
+    def testfiltervalue(self):
+        tf_img = tf.zeros([3, 4, 3], tf.int32)
+
+        with self.assertRaises(ValueError):
+            mean_filter2d(tf_img, (4, 3))
+
+    def testDimension(self):
+        tf.compat.v1.disable_eager_execution()
```
seanpmorgan: Why does this need to be disabled? and Could you re-write this test so it tests against eager and graph? The if-else method you've used in other tests is acceptable. Could you please do that for the median filter as well.

tensorflow_addons/image/__init__.py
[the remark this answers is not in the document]
Mainak431: solved

tensorflow_addons/image/filters_test.py
Mainak431: The eager mode was only deactivated for the particular test of checking dimension. We had to deactivate eager mode for this test because placeholder is not supported in the Eager mode of execution. I needed placeholder to create tensor with None Dimension. I found no other way to create None dimensional tensor. Hence, for this particular test eager mode is deactivated.

tensorflow_addons/image/filters.py
```diff
@@ -109,3 +109,90 @@ def func2():
...
+        raise TypeError('Size of the filter must be Integers')
+    (row, col, ch) = (image.shape[0], image.shape[1], image.shape[2])
+    if row != None and col != None and ch != None:
+        (row, col, ch) = (int(row), int(col), int(ch))
+    else:
+        raise TypeError(
+            'All the Dimensions of the input image tensor must be Integers.')
+    if row < filter_shapex or col < filter_shapey:
+        raise ValueError(
+            'No of Pixels in each dimension of the image should be more \
```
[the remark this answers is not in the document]
Mainak431: updated

tensorflow_addons/image/filters_test.py
Mainak431: solved

seanpmorgan:
When time allows please update as shown in https://github.com/tensorflow/addons/pull/202 described in #199 .

Sorry for the several iterations here...

Mainak431:
@seanpmorgan Please review and merge.

seanpmorgan:
@Mainak431 Looks like tests are failing... almost ready to merge though

Mainak431:
@seanpmorgan Test cases fixed. Please, see #223 also.

seanpmorgan:
Thanks @Mainak431 !

PR #190, opened 2019-04-18
Pin to old nightly version

qlzh727:
Until https://github.com/tensorflow/tensorflow/issues/27955 is fixed.


qlzh727:
Ack, will do tomorrow morning.

PR #202, opened 2019-04-23
Fix test cases

### LINKED ISSUE #199: Bugs in filters_test.py
In `image/filters_test.py`, there are some buggy test cases:

1. There are some patterns like
https://github.com/tensorflow/addons/blob/6bb0d4bfb814df9d9e6d5b200128f25120e09e21/tensorflow_addons/image/filters_test.py#L60-L63
However, this is not safe because each context manager will catch only one exception, and then exit. That it, the following code snippet will pass the test:
```python
def foo(x):
    if x < 5:
        raise ValueError("fail " + str(x))


class Test(tf.test.TestCase):

    def test_context_manager(self):
        with self.assertRaisesRegexp(ValueError, "fail 3"):
            foo(3)
            foo(2)
            foo(1)

        with self.assertRaises(ValueError):
            foo(5)
            foo(4)
            foo(3)
```
2. One test case disables eager execution, but this will affect **all test cases** (in the same file) executed after it.
https://github.com/tensorflow/addons/blob/6bb0d4bfb814df9d9e6d5b200128f25120e09e21/tensorflow_addons/image/filters_test.py#L71-L75
In other words, `test_check_eager` will fail in this toy test case.
```python
class Test(tf.test.TestCase):

    # Make it execute first
    def test_a(self):
        tf.compat.v1.disable_eager_execution()
        tf.compat.v1.placeholder(tf.float32, shape=(1, 2, None))

    def test_check_eager(self):
        if not tf.executing_eagerly():
            raise Exception("Not in eager mode")
```
This might not be what we want. In order to utilize `tf.compat.v1.placeholder`, which can be used to test `None` shape, in tf2.0, maybe `@run_deprecated_v1` is helpful enough.

WindQAQ:
Closes #199.

### DIFF
--- a/tensorflow_addons/image/filters_test.py
+++ b/tensorflow_addons/image/filters_test.py
@@ -22,72 +22,56 @@ from tensorflow_addons.image import median_filter2d
 from tensorflow_addons.utils import test_utils
 
 
-@test_utils.run_all_in_graph_and_eager_modes
 class MedianFilter2dTest(tf.test.TestCase):
     def _validate_median_filter2d(self,
                                   inputs,
                                   expected_values,
                                   filter_shape=(3, 3)):
+        output = median_filter2d(inputs, filter_shape)
+        self.assertAllClose(output, expected_values)
 
-        values_op = median_filter2d(inputs)
-        with self.test_session(use_gpu=False) as sess:
-            if tf.executing_eagerly():
-                expected_values = expected_values.numpy()
-                values = values_op.numpy()
-            else:
-                expected_values = expected_values.eval()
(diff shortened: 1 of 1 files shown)


seanpmorgan:
cc @Mainak431 could you adjust mean_filter tests to match these. I may have gave some bad advice on running all tests in graph/eager for example the None shape dimension would only occur in graph mode so you can decorate that test with compat.v1 instead of disable eager which affects the subsequent tests. The other fixes should be described in #199

Mainak431:
Thanks @WindQAQ

PR #204, opened 2019-04-23
[WIP] FIX: WeightNorm Build

### LINKED ISSUE #203: WeightNormalization test are failing without symbolic tensors in build
Since https://github.com/tensorflow/tensorflow/commit/9df99aafc3480638cd8d2bf6a0caeab111b68302 our keras_layer test is failing on weightnorm. 

This is the block that fails from the [test uti](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/testing_utils.py#L207)l:
```
 # test serialization, weight setting at model level
  model_config = model.get_config()
  recovered_model = keras.models.Sequential.from_config(model_config)
  if model.weights:
    weights = model.get_weights()
    recovered_model.set_weights(weights)
    output = recovered_model.predict(input_data)
    np.testing.assert_allclose(output, actual_output, rtol=2e-3
```
So our model config isn't properly serializing the data. 

@omalleyt12 could you take a look and provide some insight. I'm not overly familiar with `ops.init_scope` and the check for `_keras_style`

Here is our weight norm layer:
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/layers/wrappers.py#L24

seanpmorgan:
Closes #203 

Though I don't think this is the right thing to do from my understanding of `keras_style`:
https://www.tensorflow.org/api_docs/python/tf/layers/experimental/keras_style_scope

### DIFF
--- a/tensorflow_addons/layers/wrappers.py
+++ b/tensorflow_addons/layers/wrappers.py
@@ -67,6 +67,7 @@ class WeightNormalization(tf.keras.layers.Wrapper):
 
         super(WeightNormalization, self).__init__(layer, **kwargs)
         self._track_trackable(layer, name='layer')
+        self._keras_style = False
 
     def _compute_weights(self):
         """Generate weights by combining the direction of weight vector with
@@ -127,7 +128,6 @@ class WeightNormalization(tf.keras.layers.Wrapper):
                 trainable=True,
                 aggregation=tf.VariableAggregation.MEAN)
 
-            # TODO: Check if this needs control deps in TF2 graph mode
             self.layer.g.assign(self._init_norm(self.layer.v))
             self._compute_weights()
 


qlzh727:
I don't think that's the correct fix. Basically the change in https://github.com/tensorflow/tensorflow/commit/9df99aafc3480638cd8d2bf6a0caeab111b68302 is trying to prevent build() to contain any logic of actual tensor calculation, which should happen within call(). 

IMHO, the right fix is to move those tensor ops to call() instead of putting them in build().

qlzh727:
By looking at the implementation of WeightNormalization, i feel the implementation is quite anti-pattern, eg the wrapper will adding/editing the weights for the wrapped layer, I feel those created/updated weights should just tie to wrapper itself. 

I will try to do fix for this today.

PR #205, opened 2019-04-24
Add LayerNormLSTMCell to RNN cell

qlzh727:
1. The change was based on contrib.LayerNormBasicLSTMCell, which is popular based on https://tf-contrib-analyzer.herokuapp.com.
2. The cell has been updated to align with keras and v2 standard, eg use keras LSTMCell as base class, get rid of variable scope, etc.
3. The tests have also been ported and updated with graph/eager mode.

### DIFF
--- a/tensorflow_addons/rnn/cell.py
+++ b/tensorflow_addons/rnn/cell.py
@@ -208,3 +208,180 @@ class NASCell(keras.layers.AbstractRNNCell):
         }
         base_config = super(NASCell, self).get_config()
         return dict(list(base_config.items()) + list(config.items()))
+
+
+class LayerNormLSTMCell(keras.layers.LSTMCell):
+    """LSTM cell with layer normalization and recurrent dropout.
+
+    This class adds layer normalization and recurrent dropout to a LSTM unit.
+    Layer normalization implementation is based on:
+
+      https://arxiv.org/abs/1607.06450.
+
+    "Layer Normalization" Jimmy Lei Ba, Jamie Ryan Kiros, Geoffrey E. Hinton
+
+    and is applied before the internal nonlinearities.
+    Recurrent dropout is base on:
+
+      https://arxiv.org/abs/1603.05118
+
+    "Recurrent Dropout without Memory Loss"
+    Stanislau Semeniuta, Aliaksei Severyn, Erhardt Barth.
+    """
+
+    def __init__(self,
+                 units,
+                 activation='tanh',
+                 recurrent_activation='sigmoid',
+                 use_bias=True,
+                 kernel_initializer='glorot_uniform',
+                 recurrent_initializer='orthogonal',
+                 bias_initializer='zeros',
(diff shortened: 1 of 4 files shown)


tensorflow_addons/rnn/cell.py
```diff
@@ -208,3 +208,183 @@ def get_config(self):
...
+
+    This class adds layer normalization and recurrent dropout to a LSTM unit.
+    Layer normalization implementation is based on:
+
+      https://arxiv.org/abs/1607.06450.
+
+    "Layer Normalization" Jimmy Lei Ba, Jamie Ryan Kiros, Geoffrey E. Hinton
+
+    and is applied before the internal nonlinearities.
+    Recurrent dropout is base on:
```
WindQAQ: base -> based

qlzh727:
@WindQAQ, thanks for the question. My understanding is that the weights between each gates are separate and we only fuse them together so that we can do large matmul to improve performance. Since each gate carries different mathematical meaning (input, carry, forget, output), if we do normalization for the fused gates, it doesn't make sense for math, whereas the normalization for individual gates makes more sense to me.

That said, since it does seems to be different from the original paper, let me check with the engineer who implemented it originally.

qlzh727:
As a side note, from the implementation in https://github.com/ryankiros/layer-norm/blob/master/torch_modules/LSTM_LN.lua#L30, which is from the original paper author, only the output is normalized, instead of individual gates. This is also different from the original formula in the paper, which makes me quite confused.

qlzh727:
@ebrevdo and @drpng

WindQAQ:
@qlzh727, Thanks for the investigation!

> As a side note, from the implementation in https://github.com/ryankiros/layer-norm/blob/master/torch_modules/LSTM_LN.lua#L30, which is from the original paper author, only the output is normalized, instead of individual gates. This is also different from the original formula in the paper, which makes me quite confused.

I found that the theano implementation in the same repo indeed applies layer norm on fused gate. To be more clear, it seems to match the formula stated in the paper.
https://github.com/ryankiros/layer-norm/blob/master/layers.py#L457-L475

qlzh727:
Done. I have update the implementation to align with the original paper, even it doesn't make much sense to me. I don't have a e2e model to verify this, but since this matches to the reference in https://github.com/ryankiros/layer-norm/blob/master/layers.py#L457-L475, I will follow that implementation.

Also fixed the issue that bias is ignored when norm was on. I think that was a bug introduced by the original implementation, and have confirmed that with original author.

qlzh727:
Sorry, seems that I was incorrectly rebase and pull my master branch, which somehow include my other PR into here. Should I revert this one and start with a new PR?

seanpmorgan:
No worries, but yeah new PR is probably easier

qlzh727:
Done. See https://github.com/tensorflow/addons/pull/210. I am closing this PR now.

PR #208, opened 2019-04-25
Update and fix WeightNorm Wrapper.

qlzh727:
1. Move the weights v and g to the wrapper itself, since they are
more tie to the concept of weight norm, not the wrapped layer.
2. Update the build() method to only create weights, and not do
weights manipulation. Setting the init value based on the inputs
are done in call().
3. Remove the tf.function annotation on call, for debugability.
4. Renamed few function and add some comments for readability.
5. Added get_config() for serialization.

### DIFF
--- a/tensorflow_addons/layers/wrappers.py
+++ b/tensorflow_addons/layers/wrappers.py
@@ -56,49 +56,10 @@ class WeightNormalization(tf.keras.layers.Wrapper):
     """
 
     def __init__(self, layer, data_init=True, **kwargs):
-        if not isinstance(layer, tf.keras.layers.Layer):
-            raise ValueError(
-                'Please initialize `WeightNormalization` layer with a '
-                '`Layer` instance. You passed: {input}'.format(input=layer))
-
-        self.initialized = True
-        if data_init:
-            self.initialized = False
-
         super(WeightNormalization, self).__init__(layer, **kwargs)
+        self.data_init = data_init
+        self._initialized = False
         self._track_trackable(layer, name='layer')
-
-    def _compute_weights(self):
-        """Generate weights by combining the direction of weight vector with
-        its norm."""
-        with tf.name_scope('compute_weights'):
-            self.layer.kernel = tf.nn.l2_normalize(
-                self.layer.v, axis=self.kernel_norm_axes) * self.layer.g
-
-    def _init_norm(self, weights):
-        """Set the norm of the weight vector."""
-        with tf.name_scope('init_norm'):
-            flat = tf.reshape(weights, [-1, self.layer_depth])
-            return tf.reshape(
-                tf.linalg.norm(flat, axis=0), (self.layer_depth,))
-
-    def _data_dep_init(self, inputs):
-        """Data dependent initialization."""
-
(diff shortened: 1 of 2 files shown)


qlzh727:
https://github.com/tensorflow/addons/issues/203

tensorflow_addons/layers/wrappers.py
```diff
@@ -118,34 +78,77 @@ def build(self, input_shape):
...
-            self.layer.v = self.layer.kernel
-            self.layer.g = self.layer.add_variable(
+            self.v = self.layer.kernel
+            self.g = self.add_variable(
                 name="g",
                 shape=(self.layer_depth,),
                 initializer=tf.keras.initializers.get('ones'),
                 dtype=self.layer.kernel.dtype,
-                trainable=True,
-                aggregation=tf.VariableAggregation.MEAN)
```
seanpmorgan: Should we be dropping the variable aggregation? We have no way to test this is distributed fashion, but [based on this PR](https://github.com/seanpmorgan/tensorflow/pull/1) I believe its useful.

seanpmorgan:
My only other thought would be, is it documented anywhere that build method shouldn't contain tensor computations? The only reason I was able to pin a specific issue was because I had a nightly break so it limited it to 2 or 3 commits. Otherwise this would have been really challenging to figure out why serialization wasn't properly working (no error was thrown)

tensorflow_addons/layers/wrappers.py
qlzh727: For trainable weights, usually the aggregation is handled by the distribution strategy scope, which is why we always leave the aggregation to be default value. If you check the code base, the aggregation is only specified for non-trainable weights, eg in metric and batch norm. Those non trainable weights are not updated in backprop, and handled specially in forward pass. 

Having aggregation for trainable variable is likely to cause issue with the combination of distribution strategy, and the code owner is better to be sure what they are trying to do.

seanpmorgan: Thank you that makes sense. So regarding the error the user mentioned in that PR... should they have set a global aggregation strategy or something? I would have assumed it'd default to something instead of erroring like they mentioned?

seanpmorgan:
Merging so we can get working CI back and I've slated the questions about keras objects for the monthly meeting. This PR is not gated by an answer.

tensorflow_addons/layers/wrappers.py
qlzh727: Its unclear to me what the error/warning message is from that PR. The default aggregation is None when no distribution strategy is specified. For individual distribution strategy, the reduction is different, and the default value should be available in the API doc.

qlzh727:
For the original failure, I don't think it is related to from_config/get_config(). Having said that, the get_config/from_config() is a public contract that custom layer should implement.

For the question about build() method contains actual tensor computation, I don't think there is an enforcement for that. https://keras.io/layers/writing-your-own-keras-layers only state that its recommended to create weights in build() and do computation in call(). I think the change from Tom was trying to make that more explicit and expose the issue if user does so.

PR #210, opened 2019-04-25
 Add LayerNormLSTMCell to RNN cell

qlzh727:
1. The change was based on contrib.LayerNormBasicLSTMCell, which is
popular based on https://tf-contrib-analyzer.herokuapp.com.

2. The cell has been updated to align with keras and v2 standard,
eg use keras LSTMCell as base class, get rid of variable scope, etc.

3. The implementation has been updated to match the original paper.
There are 3 norm performed. One for input kernel, second one for
recurrent kernel, and finally for the carry state before activation.
More importantly, the first 2 norm are based on fused gates, rather
than the existing implementation (individual gates separately).

4. Adding the bias when normalization is on. I think missing the bias
was an oversight for the original implementation, and I have confirmed
this with original author of the contrib code.

5. Update the unit test with the correct value.

### DIFF
--- a/tensorflow_addons/rnn/cell_test.py
+++ b/tensorflow_addons/rnn/cell_test.py
@@ -176,10 +176,10 @@ class NASCellTest(tf.test.TestCase):
 
 
 @test_utils.run_all_in_graph_and_eager_modes
-class LayerNormBasicLSTMCellTest(tf.test.TestCase):
+class LayerNormLSTMCellTest(tf.test.TestCase):
 
     # NOTE: all the values in the current test case have been calculated.
-    def testBasicLSTMCell(self):
+    def testCellOutput(self):
         x = tf.ones([1, 2], dtype=tf.float32)
         c0 = tf.constant(0.1 * np.asarray([[0, 1]]), dtype=tf.float32)
         h0 = tf.constant(0.1 * np.asarray([[2, 3]]), dtype=tf.float32)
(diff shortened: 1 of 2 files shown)


tensorflow_addons/rnn/cell_test.py
```diff
@@ -175,5 +175,161 @@ def test_NASCell_config(self):
         self.assertEqual(config, restored_config)
 
 
+@test_utils.run_all_in_graph_and_eager_modes
+class LayerNormBasicLSTMCellTest(tf.test.TestCase):
```
WindQAQ: Should be LayerNormLSTMCell since we have renamed it?

seanpmorgan:
> Thanks! Looks good to me! Just want to know whether we should remove `layer_norm` argument. If users want to use legacy LSTM, then using `LSTMCell` directly. @seanpmorgan and @facaiy what do you feel about it?

Agree. Unless there is a scenario when you'd want to turn layernorm off and on I don't see the need.

tensorflow_addons/rnn/cell_test.py
qlzh727: Good catch. Done.

qlzh727:
Remove the layer_norm flag and update the test accordingly.

WindQAQ:
Thank you @qlzh727!

seanpmorgan:
@qlzh727 When time allows could update the rnn init and README

qlzh727:
@seanpmorgan, thanks for catching that. See https://github.com/tensorflow/addons/pull/217.

Issue #174, opened 2019-04-15
ABI settings in BUILD files causing issues


mfojtak:
**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow):
- OS Platform and Distribution: Ubuntu 18.04 
- TensorFlow installed from (source or binary): binary
- TensorFlow version (use command below): tf-nightly-2.0-preview==2.0.0.dev20190405
- TensorFlow Addons installed from (source, PyPi): source
- TensorFlow Addons version: latest
- Python version and type (eg. Anaconda Python, Stock Python as in Mac, or homebrew installed Python etc): 3.7
- Bazel version (if compiling from source): 0.24.1
- GCC/Compiler version (if compiling from source): 7.3.0
- Is GPU used? (yes/no): yes
- GPU model (if used): Tesla P100

**Describe the bug**

When running an app using addons build from source, it crashes with undefined symbol error.
```
undefined symbol: _ZN10tensorflow8internal21CheckOpMessageBuilder9NewStringEv
```
There are BUILD files under /tensorflow_addons/custom_ops
When I remove *-D_GLIBCXX_USE_CXX11_ABI=0* from these files, it starts working again.

Why are these lines present in the BUILD files?
Should I raise PR and get them removed?

seanpmorgan:
Hi Michal,

Thanks for bringing this up. So the build process is a bit hacky while we wait for a build toolchain which is manylinux2010 compatible (Should be by the end of the month). Since TF is compiled with gcc 4.8 we need these ABI flags to maintain compatibility in case addons is compiled with gcc>=5. As described:
https://github.com/calid/docs/blob/master/site/en/extend/adding_an_op.md#build-the-op-library

I noticed you're running python3.7 which is a broken whl as far as I know (they had to use ubuntu16 so it's unlike all of the other TF whls). Could you try this again using the py36 or any other version?

If you really want to run py37 then dropping the ABI flags as you have done and compiling from source should be the fix. Let me know if neither of these options is suitable for you. In either case this should all be cleaned up in the coming weeks.

seanpmorgan:
Some more info:
https://github.com/tensorflow/tensorflow/issues/27067

kyleabeauchamp:
FYI I've been able to (~2 weeks ago, around the time of my last commit) use py37 (conda-forge miniconda) by flipping the CXX11_ABI flag in all relevant places in the bazel files.

kyleabeauchamp:
[...]

PR #211, opened 2019-04-26
Add bzl file to control CXX11_ABI

kyleabeauchamp:
So for my miniconda3 builds to link properly, I find that I constantly have to flip `D_GLIBCXX_USE_CXX11_ABI=1` in the custom_ops files.  I'd love some way to make this easier.  

Here is one possible approach, which involves factoring out this flag into a variable stored in a `.bzl` file.  

I'm open to better approaches, but please provide explicit links to SO posts or docs that explain the usage, as I've found the bazel docs to be a bit...sparse.


kyleabeauchamp:
See #174 for an example of the relevant issue.

seanpmorgan:
I think it's a great idea, and we've briefly discussed a shared bazel wrapper library. See https://github.com/tensorflow/custom-op/issues/8

Let me think about the implementation for a few days, but I think something that is more or less what you're proposing.

kyleabeauchamp:
Anyone else have thoughts on how to implement this?  The current code structure is kinda painful because any addons development I do requires having a branch with 3 different BUILD files edited...

seanpmorgan:
This will be brought up in the monthly meeting this Friday 1pm PDT:
https://docs.google.com/document/d/1kxg5xIHWLY7EMdOJCdSGgaPu27a9YKpupUz2VTXqTJg/edit

I'm leaning toward a library of bazel wrappers but want to ask if we should house it internally or if there should be an external library for all custom-op repos.

seanpmorgan:
Update from the monthly meeting: We're going to start building some bazel macros for Addons and as it matures we can split it out into its own repository. Possibly managed by SIG Build.

So for this PR could you make the file name more generic so it can hold more macros than just this one? Additionally, there is a flag `tf.sysconfig.CXX11_ABI_FLAG` so we can have this automatically configured in `configure.sh` Its up to you if you want to include that in this PR or we can do it separately. Thanks!

kyleabeauchamp:
I've renamed the macro file as `addons_bazel_macros.bzl`

kyleabeauchamp:
I'm trying to understand `tf.sysconfig`---isn't that set by the upstream TF build that's installed?  In my case, I'm using the pip-installed TF build but modifying the ABI flag for my addons build, which is built from source.  Thus, if my mental model is right, propagating `tf.sysconfig.CXX11_ABI_FLAG` won't help me.

kyleabeauchamp:
I've implemented the first half of the sysconfig setup.  However, I'm not going to be able to finish it because I literally cannot find any similar usage of bazelrc in the bazel docs.  So I guess my recommendation is review as-is and finish in a follow-up PR?

seanpmorgan:
Thanks @kyleabeauchamp I'll be able to run and test tomorrow, but I did want to ask... When do we imagine the compiled ABIs should not match TF core? Is there only a mismatch right now because TF2 preview in not on conda?

kyleabeauchamp:
I think I mis-spoke about that particular point and have deleted the comment.  I would imagine we do want them in sync.

configure.sh
```diff
@@ -45,7 +45,10 @@ pip install $QUIET_FLAG -r requirements.txt
 TF_CFLAGS=( $(python -c 'import tensorflow as tf; print(" ".join(tf.sysconfig.get_compile_flags()))') )
 TF_LFLAGS=( $(python -c 'import tensorflow as tf; print(" ".join(tf.sysconfig.get_link_flags()))') )
 TF_SHAREDLIB=( $(python -c 'import tensorflow as tf; print(tf.sysconfig.get_link_flags()[-1])') )
+TF_CXX11_ABI_FLAG=( $(python -c 'import tensorflow as tf; print(tf.sysconfig.CXX11_ABI_FLAG)') )
 
 write_action_env_to_bazelrc "TF_HEADER_DIR" ${TF_CFLAGS:2}
 write_action_env_to_bazelrc "TF_SHARED_LIBRARY_DIR" ${TF_LFLAGS:2}
 write_action_env_to_bazelrc "TF_SHARED_LIBRARY_NAME" ${TF_SHAREDLIB:3}
+write_action_env_to_bazelrc "TF_CXX11_ABI_FLAG" ${TF_CXX11_ABI_FLAG:1}
```
seanpmorgan: Just `${TF_CXX11_ABI_FLAG}` as it's just one character 0 or 1

kyleabeauchamp: Fixed

PR #220, opened 2019-04-30
FIX: x_test scaling in weightnormalization example

WindQAQ:
Should be 255 instead of 25.


seanpmorgan:
@WindQAQ good catch. Thanks!

Issue #193, opened 2019-04-19
Packaged Addons can't run on TF2-nightly if built against TF2-alpha


seanpmorgan:
We're in the process of getting ready to release a 0.3 version, but there are several breaking changes in the tf2-nightly vs. tf2-alpha (as there should be). 

To start the shared libraries are now named with versions and OS syntax. We can get around this by using the link flag that has now been added to tf.sysconfig though.

API changes are harder to work around. For example:  https://github.com/tensorflow/addons/pull/192 -- which means on the 0.3 release branch we'll need to roll-back some changes if we continue to build against tf2-alpha. I think this is the best course of action while we wait for a tf2-beta, but wanted to discuss.

We can also point users to this issue if they are building from source and running against tf2-alpha.

cc @karmel @martinwicke

martinwicke:
We did make some tf.image name changes to eliminate duplicate "image"s.

In theory you could use a try block to ensure the code works for alpha and
nightly, which could be eliminated once the 2.0 is in RC (or beta) and
people stop caring about alpha0.

How much trouble would that be?

seanpmorgan:
Now that we have [nightlies being published](https://pypi.org/project/tfa-nightly/#files), the recommended way to run against `tf-nightly-2.0-preview` is to use `tfa-nightly`

```
pip install tfa-nightly
```

Once we publish 0.4 we'll do our best to keep it compatible with tf2-nightly but that is not something we can guarantee.

PR #223, opened 2019-04-30
fixing extract_image_patches

Mainak431:
tensorflow==2.0.0.a0 does not have any function extract_patches in tf.image. That will be extract_image_patches. I have installed the latest public version of tensorflow==2.0.0.a0.

@seanpmorgan Please restore this.

### DIFF
--- a/tensorflow_addons/image/filters.py
+++ b/tensorflow_addons/image/filters.py
@@ -91,7 +91,7 @@ def median_filter2d(image, filter_shape=(3, 3), name=None):
         for a in range(ch):
             img = image[:, :, a:a + 1]
             img = tf.reshape(img, [1, row, col, 1])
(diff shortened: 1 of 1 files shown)


seanpmorgan:
See this issue https://github.com/tensorflow/addons/issues/193

The packaged 0.3 release (today or tomorrow) will have this and a few other changes so it's compatible with 2.0.0.a0. If you're building from source you must use tf-nightly as it says in the central README.

This problem mismatch issue will be ~fixed once we start getting some stable 2.0 releases. As Martin mentioned we could also add an if/else type of check, but there are are incompatibilities that'll fail (e.g.shared library names)

Mainak431:
Ok. updating #185 accordingly.  Thanks.

PR #229, opened 2019-05-02
Add median filter 1d

Mainak431:


tensorflow_addons/image/filters.py
```diff
@@ -112,6 +109,66 @@ def mean_filter2d(image, filter_shape=(3, 3), name=None):
...
+                more than the filter size. Got filter_shape %s' %
+                filter_shapex +
+                ' Image Shape (%s)' % image.shape)
+        if filter_shapex % 2 == 0 :
+            raise ValueError('Filter size should be odd. Got filter_shape '
+                             '(%s)' % filter_shapex)
+        image = tf.cast(image, tf.float32)
+        tf_i = tf.reshape(image, [row * col * ch])
+        ma = tf.math.reduce_max(tf_i)
+        image = _normalize(image, ma)
```
WindQAQ: Wonder why we have to normalize the image?

tensorflow_addons/image/filters.py
```diff
@@ -112,6 +109,66 @@ def mean_filter2d(image, filter_shape=(3, 3), name=None):
...
+            top = tf.nn.top_k(slic, mid, sorted=True)
+            li = tf.slice(top[0], [0, 0, 0, mid - 1], [-1, -1, -1, 1])
+            li = tf.reshape(li, [row, col, 1])
+            listi.append(li)
+        y = tf.concat(listi[0], 2)
+        for i in range(len(listi) - 1):
+            y = tf.concat([y, listi[i + 1]], 2)
+
+        y *= 255
+        y = tf.cast(y, tf.int32)
```
WindQAQ: ditto

tensorflow_addons/image/filters.py
```diff
@@ -112,6 +109,66 @@ def mean_filter2d(image, filter_shape=(3, 3), name=None):
         return y
 
 
+@tf.function
+def median_filter1d(image,filter_shape=3, name=None):
```
WindQAQ: I think `median_filter1d(image, filter_shape=3)` should have the same functionality with `median_filter(image, filter_shape=(3, 1))`? If so, can we just call `median_filter2d` in `median_filter1d`?

tensorflow_addons/image/filters.py
```diff
@@ -112,6 +109,66 @@ def mean_filter2d(image, filter_shape=(3, 3), name=None):
...
+    """This method performs 1D Median Filtering on images.Filter shape can be user given.
+       This method takes both kind of images where pixel values lie between 0 to 255 and where it lies between 0.0 and 1.0
+       Args:
+           input: A 3D `Tensor` of type `float32` or 'int32' or 'float64' or 'int64 and of shape`[rows, columns, channels]`
+
+           filter_shape: Optional Argument.Single Integer. Default value = 3
+
+        Returns:
+            A 1D median filtered image 3D tensor of shape [rows,columns,channels] and type 'int32'. Pixel value of returned 
+            tensor ranges between 0 to 255
```
WindQAQ: should be `backticks` instead of 'quotation marks' if you want to highlight the word in docstring. Also, some spaces between words and punctuations are missing. Please check it.

tensorflow_addons/image/filters.py
Mainak431: 1d here just mean we are using a 1d filters. Hence the filter-shape is a single odd integer.  Not a tuple.  1d filter will improve time complexity im case of batch processing.

tensorflow_addons/image/filters.py
Mainak431: In real life there are some images which are in the range of 0 to 255 and some images pixel values are in the floating range 0.0 to 1.0. Image viewers take care of both the images.  Hence,  my functanatilies also takes care of that.  And normalize is not a new addition in this pull request.  This is already present in the master file.

tensorflow_addons/image/filters.py
Mainak431: Ditto

tensorflow_addons/image/filters.py
Mainak431: I don't want to highlight any word. Solved.

Mainak431:
@WindQAQ The objective of 1d filter is just to make fast the batch processing with respect to 2d filters as already discussed with Sean. Axis operation cannot be supported as it is an image operation. As we are performing median filtering of the entire image (all channels). Only, the difference is that the filters are 1D filters.  I have tested the filter with a real image. And its performing fine in removing salt pepper noise which is the main objective of the median filter.

tensorflow_addons/image/filters.py
WindQAQ: So shouldn't it be:
- if input ranges from 0 to 1, then output ranges from 0 to 1.
- if input ranges 0~255, then output ranges 0 to 255.

:smile: ?

Mainak431: I wanted uniformity in the output pattern, i.e, integer values.

tensorflow_addons/image/filters.py
WindQAQ: Okay seems the functionality is quite different with the one in my understanding.

WindQAQ:
Seems that I have missed some discussions previously. @seanpmorgan Hi Sean, could you take over this PR?

tensorflow_addons/image/filters.py
WindQAQ: Ok, but still adreess some issues of spaces

1. 
This method performs 1D Median Filtering on images.Filter shape can be user given.
=>
This method performs 1D Median Filtering on images. Filter shape can be user given.

2. 
input: A 3D `Tensor` of type `float32` or 'int32' or 'float64' or 'int64 and of shape`[rows, columns, channels]`
=>
input: A 3D `Tensor` of type `float32` or 'int32' or 'float64' or 'int64 and of shape `[rows, columns, channels]`

3.
filter_shape: Optional Argument.Single Integer. Default value = 3
=>
filter_shape: Optional Argument. Single Integer. Default value = 3

4.
A 1D median filtered image 3D tensor of shape [rows,columns,channels]
=>
A 1D median filtered image 3D tensor of shape [rows, columns, channels]

seanpmorgan: @Mainak431 Please address the comments of the reviewers. The PR reviewers get to decide when an issue is resolved.

seanpmorgan:
@Mainak431 I'm sorry but I meant to ask this earlier. Instead of adding 1D filters I think it's best we wait and follow the community use of the 2D filters. What would be great would be to create an example notebook `image_filters.ipydb` that highlites what this does to a given image and how it can improve training. Does that make sense?

tensorflow_addons/image/filters.py
Mainak431: Ok . @WindQAQ Now i got what you mean by issue of spaces.

Mainak431: @seanpmorgan Yes of course.

Mainak431:
Ok. @seanpmorgan . Then, please close this PR. I will add the batch processing of the 2D filters as that will only make it useful. I thought, at first, I will add all the filters and then go for batch processing.

seanpmorgan:
@Mainak431 Thanks for understanding. If you could also start that examples notebook that would really help as I think this repo is most in need of examples prior to a TF2 release

seanpmorgan:
@Mainak431 could you create an issue describing the need for batch processing before submitting the PR. Now that the repo is maturing it makes it managable and easier on contributors if we discuss before working on stuff.

Mainak431:
@seanpmorgan At first, I will implement the batch processing and then write the notebook. Without batch processing, these modules can't be used with direct CNN architecture. This functionality can then directly be used by other tf dependent modules also.

Mainak431:
@seanpmorgan Issue means github issue?

seanpmorgan:
Yes github issue. Just to describe why its needed and what you plan to do.

PR #235, opened 2019-05-03
Speed up mean_filter2d with depthwise_conv2d

### LINKED ISSUE #234: Speed up mean_filter2d with depthwise convolution
**System information**
- TensorFlow version (you are using): **tf-nightly-2.0-preview**
- TensorFlow Addons version: **source**
- Is it in the tf.contrib (if so, where): **no**
- Are you willing to contribute it (yes/no): **yes**
- Are you willing to maintain it going forward? (yes/no): **yes**

Currently, `tfa.image.mean_filter2d` is implemented by `tf.image.extract_patches`, while its functionality is the same with the one applying 2-D convolution with box filter (or uniform filter) channel by channel. This can be easily done by `tf.nn.depthwise_conv2d` with significant speed improvement.

Here is the [notebook example](https://colab.research.google.com/drive/1jJksZAkTfdjJIzPgdw9MhDkNQkFlUM6i) to compare the performance of these two implementations. On colab platform, **it is 4x faster than the original one for either single image or multiple images**. Note that I extend the original implementation with `tf.map_fn` to make it support 4-D input, and some trivial normalization is omitted.

Still wondering why some normalization and casting are performed in the original implementation in both `mean_filter2d` and `median_filter2d`. (no offense, just want to know why)

- For [opencv](https://docs.opencv.org/4.1.0/d4/d86/group__imgproc__filter.html#gad533230ebf2d42509547d514f7d3fbc3),
> dst: destination array of the same size and type as src.
- For [scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.uniform_filter.html), 
> output : array or dtype, optional
The array in which to place the output, or the dtype of the returned array. By default an array of the same dtype as input will be created.
- For [matlab](https://www.mathworks.com/help/images/ref/medfilt2.html),
> Output image, returned as a numeric matrix of the same class as the input image I.

As far as I'm concerned, there is no need to force the computation limited in range [0, 1], and it's not necessary to transform the output back to `uint8` range even if the input is not of type `uint8`. **All we need to do is to compute the output (may first cast image to float because `depthwise_conv2d` does not accept non-float kernel and input) and cast it back to the original data type.** I believe that for median and average operation, there will not result in any over/underflow situation. Hence, it might be very safe for casting the output back to the original data type, though I think it's better to discuss this first.

Moreover, if users want to do some post processing on the intermediate feature maps from CNNs, these feature maps might not range in either [0, 1] or `uint8`. They are more likely to tensors of real number depending on what the activation function is used.

`mean_filter2d(image, filter_shape=(3, 3), padding="REFLECT", constant_values=0, name=None)`

The new API is similar to [scipy.ndimage.uniform_filter](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.uniform_filter.html) which supports padding mode. The default padding mode is "REFLECT" because not only scipy but also [opencv](https://docs.opencv.org/4.1.0/d4/d86/group__imgproc__filter.html#gad533230ebf2d42509547d514f7d3fbc3) set REFLECT padding as a default (while some toolboxes in matlab adopt zero-padding by default). Besides, [tf.image.sobel_edges](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/image_ops_impl.py#L3183), a filter-based application, also uses REFLECT padding mode.

People who want to process images.

**Any Other info.**

To conclude:

1. Implement with `tf.nn.depthwise_conv2d`, which can speed up a lot and support batch-wise computation.
2. Cast the input image to float if needed and cast the output back to `image.dtype`.
3. Support padding modes.

Some references for API design:
- [tf.pad](https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/pad)
- [scipy.ndimage.uniform_filter](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.uniform_filter.html)

WindQAQ:
Closes #234. Will create another PR to clean up median_filter2d.

### DIFF
--- a/tensorflow_addons/image/filters.py
+++ b/tensorflow_addons/image/filters.py
@@ -103,13 +103,16 @@ def mean_filter2d(image,
             image, filter_shape, mode=padding, constant_values=constant_values)
 
         # Filter of shape (filter_width, filter_height, in_channels, 1)
-        # has value 1 / (filter_width * filter_height) for each element.
-        area = filter_shape[0] * filter_shape[1]
+        # has the value of 1 for each element.
+        area = tf.constant(
+            filter_shape[0] * filter_shape[1], dtype=image.dtype)
         filter_shape = filter_shape + (tf.shape(image)[-1], 1)
-        kernel = tf.ones(shape=filter_shape, dtype=image.dtype) / area
+        kernel = tf.ones(shape=filter_shape, dtype=image.dtype)
 
         output = tf.nn.depthwise_conv2d(
             image, kernel, strides=(1, 1, 1, 1), padding="VALID")
+
+        output /= area
 
         # Squeeze out the first axis to make sure
         # output has the same dimension with image.
--- a/tensorflow_addons/image/filters.py
+++ b/tensorflow_addons/image/filters.py
@@ -67,7 +67,7 @@ def mean_filter2d(image,
         padding mode.
       name: A name for this operation (optional).
     Returns:
-      3-D or 4-D float `Tensor`, as per the input.
+      3-D or 4-D `Tensor` of the same dtype as input.
     Raises:
       ValueError: If `image` is not 3 or 4-dimensional,
         if `padding` is other than "REFLECT", "CONSTANT" or "SYMMETRIC",
@@ -94,6 +94,7 @@ def mean_filter2d(image,
 
         # Keep the precision if it's float;
         # otherwise, convert to float32 for computing.
+        orig_dtype = image.dtype
         if not image.dtype.is_floating:
             image = tf.dtypes.cast(image, tf.dtypes.float32)
 
@@ -115,7 +116,7 @@ def mean_filter2d(image,
         if rank == 3:
             output = tf.squeeze(output, axis=0)
 
-        return output
+        return tf.dtypes.cast(output, orig_dtype)
 
 
 @tf.function
--- a/tensorflow_addons/image/filters.py
+++ b/tensorflow_addons/image/filters.py
@@ -18,6 +18,7 @@ from __future__ import division
 from __future__ import print_function
 
 import tensorflow as tf
+from tensorflow_addons.utils import keras_utils
 
 
 @tf.function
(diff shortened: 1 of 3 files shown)


tensorflow_addons/image/filters.py
```diff
@@ -34,82 +35,88 @@ def func2():
...
+
+        output = tf.nn.depthwise_conv2d(
+            image, kernel, strides=(1, 1, 1, 1), padding="VALID")
+
+        # Squeeze out the first axis to make sure
+        # output has the same dimension with image.
+        if rank == 3:
+            output = tf.squeeze(output, axis=0)
+
+        return tf.dtypes.cast(output, orig_dtype)
```
facaiy: I'm afraid we should let users do the conversion if they'd like, because the operation might cause loss of precision (say, uint8)

tensorflow_addons/image/filters.py
```diff
@@ -34,82 +35,88 @@ def func2():
...
+        # Explicitly pad the image
+        image = _pad(
+            image, filter_shape, mode=padding, constant_values=constant_values)
 
-        return y
+        # Filter of shape (filter_width, filter_height, in_channels, 1)
+        # has value 1 / (filter_width * filter_height) for each element.
+        area = filter_shape[0] * filter_shape[1]
+        filter_shape = filter_shape + (tf.shape(image)[-1], 1)
+        kernel = tf.ones(shape=filter_shape, dtype=image.dtype) / area
```
facaiy: To avoid loss of precision, you might want to divide the `output` by `area` after using depthwise_conv2d.

tensorflow_addons/image/filters.py
```diff
@@ -34,82 +35,88 @@ def func2():
...
-        y = tf.cast(y, tf.int32)
+        # Explicitly pad the image
+        image = _pad(
+            image, filter_shape, mode=padding, constant_values=constant_values)
 
-        return y
+        # Filter of shape (filter_width, filter_height, in_channels, 1)
+        # has value 1 / (filter_width * filter_height) for each element.
+        area = filter_shape[0] * filter_shape[1]
+        filter_shape = filter_shape + (tf.shape(image)[-1], 1)
```
facaiy: Could you add an unit test for channels = None?

tensorflow_addons/image/filters.py
```diff
@@ -34,82 +35,88 @@ def func2():
...
+def _pad(image, filter_shape, mode="CONSTANT", constant_values=0):
+    """Explicitly pad a 4-D image."""
+    assert mode in ["CONSTANT", "REFLECT", "SYMMETRIC"]
+    filter_height, filter_width = filter_shape
+    pad_top = (filter_height - 1) // 2
+    pad_bottom = filter_height - 1 - pad_top
+    pad_left = (filter_width - 1) // 2
+    pad_right = filter_width - 1 - pad_left
+    paddings = [[0, 0], [pad_top, pad_bottom], [pad_left, pad_right], [0, 0]]
+    return tf.pad(image, paddings, mode=mode, constant_values=constant_values)
```
facaiy: Can we add an unit test for even shape, say (4, 4), and make sure the result is the same with scipy?

tensorflow_addons/image/filters.py
```diff
@@ -34,82 +35,88 @@ def func2():
...
-        filter_shape: Optional Argument. A tuple of 2 integers (R,C).
-               R is the first value is the number of rows in the filter and
-               C is the second value in the filter is the number of columns
-               in the filter. This creates a filter of shape (R,C) or RxC
-               filter. Default value = (3,3)
+@tf.function
+def mean_filter2d(image,
+                  filter_shape=(3, 3),
+                  padding="REFLECT",
+                  constant_values=0,
```
facaiy: Also add an unit test for non-zero constant_values?

tensorflow_addons/image/filters_test.py
```diff
@@ -23,73 +23,129 @@
...
+        expected = tf.stack([expected, 2. * expected, 3. * expected], axis=-1)
+        # Squeeze shape from (3, 3, 1, 3) to (3, 3, 3)
+        image = tf.squeeze(image, axis=-2)
+        expected = tf.squeeze(expected, axis=-2)
+
+        output = mean_filter2d(image)
+        self.assertAllClose(output, expected)
+
+    def test_4d_image(self):
+        # Test shape (2, 3, 3, 1)
```
facaiy: Can we add more unit tests, say:
1. channes_num is 3
2. uint8 images
3. float64 images

facaiy:
@Mainak431 Mainak, as the author, would you like to take a look? Thanks.

tensorflow_addons/image/filters.py
WindQAQ: Do you mean that we should return the float image? I'm fine with that though.

facaiy: Let's return float image. And add an argument later if anyone requests it. What do you think :-)

Mainak431: If float image is retuned, the floating range should be in between 0.0 and 1.0. Because many image showing tools like matplotlib can not create the image properly if the values are in the range 0.0 to 255.0. If you wish to return between 0 to 255 range,  i think int32 will be a better choice.  In the original code int range was used.  I also think there should be uniformity in the output type.

WindQAQ: I have two concerns here:

1. Since median operation might not require explicitly casting, `median_filter2d` would be suitable to return the image of original dtype. This would result in the mismatched API for `mean_filter2d` and `median_filter2d`.

2. Scipy returns the image of original dtype

Mainak431:
@windqaq.  I think we should only take 4D tensor.  If anybody wants to process a single image.  He/she should reshape the tensor to (1, image-shape). This pattern is followed in many tf operations that supports both single and batch processing.  I think that trend should be followed here too.

WindQAQ:
@Mainak431 But most image `tf.image` functions support both 3-D and 4-D input
https://github.com/tensorflow/tensorflow/blob/r1.13/tensorflow/python/ops/image_ops_impl.py#L259
https://github.com/tensorflow/tensorflow/blob/r1.13/tensorflow/python/ops/image_ops_impl.py#L281
https://github.com/tensorflow/tensorflow/blob/r1.13/tensorflow/python/ops/image_ops_impl.py#L351
https://github.com/tensorflow/tensorflow/blob/r1.13/tensorflow/python/ops/image_ops_impl.py#L423
https://github.com/tensorflow/tensorflow/blob/r1.13/tensorflow/python/ops/image_ops_impl.py#L523
https://github.com/tensorflow/tensorflow/blob/r1.13/tensorflow/python/ops/image_ops_impl.py#L553
etc

tensorflow_addons/image/filters.py
Mainak431: We can explicitly cast the both filter operation to same type.  Let it be int32 or float64 or anything else.  That will only ensure uniformity in output.

Mainak431:
Yes.  Image operations support this.  But,  many other operations i have come through does not support this.  It seems there is no rule for the same.  Ok.

Mainak431:
Can you please try your function with conv2d layer,  activation layer,  dropout and basic layer by creating a small tf architecture.  And ensure that it is working as expected even in the training of nn.  A colab notebook will be helpful.  As, the batch processing operation aims at directly usage of this in the nn architecture.  Thanks.

WindQAQ:
> @WindQAQ. I think we should only take 4D tensor. If anybody wants to process a single image. He/she should reshape the tensor to (1, image-shape). This pattern is followed in many tf operations that supports both single and batch processing. I think that trend should be followed here too.

Ok, @facaiy What's your feel about this. I'd vote for that the input can be both 3-D and 4-D

> Can you please try your function with conv2d layer, activation layer, dropout and other basic layers by creating a small tf architecture. And ensure that it is working as expected even in the training of nn. A colab notebook will be helpful. As, the batch processing operation aims at direct usage of this in the nn architecture. Thanks.

I'll run a toy model on MNIST dataset when I'm back to the office.

Mainak431:
Just out of curiosity,  what happens to a 2D grayscale image.  Batch of this kind of images will look like a 3D image.  I think either this should be explicitly mentioned in the documentation.  Or we should consider taking only 4D tensors.

WindQAQ:
> Just out of curiosity, what happens to a 2D grayscale image. Batch of this kind of images will look like a 3D image. I think either this should be explicitly mentioned in the documentation. Or we should consider taking only 4D tensors.

I think we have already mentioned this in the documentation :-) Users have to make sure that the last axis is the channel axis as the one in the core TF.
https://github.com/tensorflow/addons/blob/51b18f626adce8724acfca713268bad8218c7fda/tensorflow_addons/image/filters.py#L58-L60

Mainak431:
Yes.  But this thing that user have to set last axis as 1 for 2d grayscale
image is not explicitly mentioned. it will be useful for first time users.
I remember while i was using it for the first time,  i ran into similar
errors and i had to stackoverflow it. Or,  the other option is to take only
4D tensors.  Whatever you think more suitable.  Thanks.

tensorflow_addons/image/filters.py
facaiy: OK, let's keep it simple, and don't return float image. I'm fine with the current implementation: returns the image of original dtype.

facaiy:
Sorry for  the delay, @WindQAQ @Mainak431 

I think we should keep consistency with tf.image: function works on either a single image (image is a 3-D Tensor), or a batch of images (image is a 4-D Tensor).

facaiy:
@WindQAQ Tzu-Wei, could you update the PR when time allows :-)

WindQAQ:
Thanks for reminding, facai. I'm currently workingon this :-)

tensorflow_addons/image/filters.py
WindQAQ: Should we import scipy for test here?

WindQAQ:
@facaiy, hi facai, I make some changes:

1. Avoid loss of precision
2. Refactor test cases according to https://github.com/tensorflow/tensorflow/blob/b14c390fc869b63fd2b1a4e6f4477ce410b9383e/tensorflow/python/kernel_tests/conv_ops_3d_test.py
3. Add a test case with channels of None

As for even-sized kernel test, should we use or import scipy for test here?

tensorflow_addons/image/filters.py
WindQAQ: I could add a test case to check whether the output of explicitly zero padding is the same with the one of native "SAME" padding supported by depthwise_conv2d if we really do not want to compare the result produced by other packages.

facaiy: Hi, Tzu-Wei, perhaps we don't have to import scipy. Can we rename test_reflect_padding to test_reflect_padding_with_3x3_filter, and create an new test_reflect_padding_with_4x4_filter in the same way?

tensorflow_addons/image/filters_test.py
```diff
@@ -23,73 +23,145 @@
...
-                [[42, 57, 56], [58, 106, 77], [27, 82, 49]]]
-        expt = tf.convert_to_tensor(value=expt)
-        self._validate_mean_filter2d(tf_img, expt)
+            mean_filter2d(image, padding="TEST")
+
+    def test_none_channels(self):
+        # 3-D image
+        fn = mean_filter2d.get_concrete_function(
+            tf.TensorSpec(dtype=tf.dtypes.float32, shape=(3, 3, None)))
+        fn(tf.random.uniform(shape=(3, 3, 1)))
```
facaiy: How about using tf.ones? We might not be able to reproduce random data.

tensorflow_addons/image/filters.py
WindQAQ: Hi, Facai, after a rough test, I found the padding method (or the anchor of even-sized kernel) of tensorflow seems different from the one of scipy. Here is the [testing notbook](https://colab.research.google.com/drive/17tB18pyGqdlV16wsJ3YKe_mfPo4d4t7q). Take 4x4 filter for example. The explicit (zero) padding could yield the same result with implicit SAME padding supported by depthwise_conv2d. However, I have to adjust `origin` argument in [scipy's implement](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.uniform_filter.html) to (-1, -1) so that it could produce the same result. In this case, should we follow the scipy's implementation or match the implicit padding in tensorflow?

WindQAQ: BTW, 
scipy's reflect padding => (c b a | a b c d | d c b)
TF's reflect padding => (d c b | a b c d | c b a)
TF's reflect padding = scipy's mirror padding
scipy's reflect padding = TF's symmetric padding

Seems that they have different meaning...
https://www.tensorflow.org/api_docs/python/tf/pad

facaiy: Hi, Tzu-Wei, 
1) If it's not easy to be compatible with scipy's implementation, we can give out all details about how we calculate it in the document.
2) As for padding mode, I think [`tf.pad`](https://www.tensorflow.org/api_docs/python/tf/pad) keeps consistent with [`np.pad`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.pad.html). And similarity, we can either keep compatible with scipy, or write a detailed document.
Both are fine for me. cc @seanpmorgan Sean, what's your opinion?

WindQAQ: Sure :-) In either way, it should be more docs there so let me write some details in this version first.

seanpmorgan:
> You are owner of some files modified in this pull request.
> Would you kindly review the changes whenever you have the time to?
> Thank you very much.

@gabrieldemarmiesse this doesn't happen often but wondering if you see any reason this happens?

gabrieldemarmiesse:
It's the first time I hear about it. Do you have other examples so I can see if there is a pattern?

In the yaml, the only condition starting the bot is:
```
on:
  pull_request:
    types: [opened]
```

So that makes me quite perplex that something happens after a PR is closed.

gabrieldemarmiesse:
Sorry @seanpmorgan  I've seriously no idea what causes this. If that happens too many time I'll take a much longer look.

Issue #237, opened 2019-05-08
Nightly Tests have stopped or are lagging


seanpmorgan:
Checking the nightly tests is looks as though the last test ran on May 5th. When I run locally there are several failing tests and getting this information the day of makes troubleshooting problematic commits from TF core a lot easier.

https://source.cloud.google.com/results/invocations/7f864214-a363-4ccd-9887-28f752d0e4a4/targets/tensorflow_addons%2Fubuntu%2Fcpu%2Fpy2%2Fcontinuous/log

Could you take a look when you're free @av8ramit @karmel

seanpmorgan:
FYI it looks like today's fail is from missing SelectV2 gradient and it got rolledback so build should pass tomorrow.
https://github.com/tensorflow/tensorflow/pull/15982

karmel:
@qlzh727 -- is this related to the changes you made to the build frequency?

qlzh727:
Humm, the change I made for changing the project type is submitted on Apr 30, and I still can see new build is triggered after May 5, eg https://source.cloud.google.com/results/invocations/ee86873d-b266-40e6-8ebd-64e8a8273cd1/targets.

qlzh727:
Btw, the tests are triggered now by change commit to the master branch on tfa, so that any change that is breaking tfa can be quickly picked up. That said, if there isn't any change, then it will not run.

karmel:
Hmm, yes, it looks like tests are running again? Was it just the SelectV2 that was the problem, @seanpmorgan ?

seanpmorgan:
Yeah there is no issue with the build at the moment. So if I understand correctly tests will only be ran daily if there are changes to our master branch? I'm not sure that's sufficient as a large portion of our code breaks have been because of the daily changes to TF core master branch.

qlzh727:
I changed it to only run after any change in the tfa repo, and skip the nightly run. Let me update it to run with in both CI fashion and with nightly trigger.

qlzh727:
Should be fixed now.

seanpmorgan:
Thanks!

PR #238, opened 2019-05-08
CLN: Remove outdated python RegisterShape

seanpmorgan:
Removing private API and I believe these are leftover from a TF version from long ago.

0.11 Release mentions removing the python SetShape, and I see no uses of this in TF core repository. I belive the C++ SetShape is all that is required:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/ops.py#L2798

### DIFF
--- a/tensorflow_addons/image/distance_transform.py
+++ b/tensorflow_addons/image/distance_transform.py
@@ -18,16 +18,12 @@ from __future__ import division
 from __future__ import print_function
 
 import tensorflow as tf
-from tensorflow.python.framework import common_shapes
-from tensorflow.python.framework import ops
 from tensorflow_addons.utils.resource_loader import get_path_to_datafile
 
 _image_ops_so = tf.load_op_library(
     get_path_to_datafile("custom_ops/image/_image_ops.so"))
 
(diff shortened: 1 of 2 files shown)


seanpmorgan:
Failing tests will need to wait for tomorrows nightly. See #237

facaiy:
>   Shape functions must now be registered via the SetShapeFn on the
  original Op specification in C++.

I think so. The shape of those ops has been registered in C++ side: https://github.com/tensorflow/addons/blob/06a85e505b68b295f4c486c8f26cd31a803f49d3/tensorflow_addons/custom_ops/image/cc/ops/image_ops.cc#L110

@karmel Karmel, could you please confirm if we understand it correctly?

karmel:
That is correct-- SetShapeFn is sufficient.

mels630:
@gabrieldemarmiesse perhaps some old commits got picked up when the bot was activated?

gabrieldemarmiesse:
The bot code is here https://github.com/tensorflow/addons/blob/master/.github/workflows/notify_codeowners.yml 

It seems that it should only activate when pull request are opened. If you have any idea how to fix the bot, please suggest :)

PR #240, opened 2019-05-10
CLN: Rework bazel build setup

seanpmorgan:
Rework build setup so that addons is linked to the TF CXX11 ABI by default. If for some reason you don't want it to match you can edit the `.bazelrc` file after running `.configure.sh`

Also matched the shared library parsing that was added to custom-op so that we only call `tf.sysconfig.get_link_flags` once.

cc @kyleabeauchamp this look okay to you?


kyleabeauchamp:
Code changes seem reasonable (modulo my limited bazel knowledge).  I can check that it runs this evening.

kyleabeauchamp:
I can confirm that the build works for me.  I'd have one more pair of eyes for the review as I'm fairly new to bazel.

configure.sh
```diff
@@ -43,12 +43,13 @@ esac
 pip install $QUIET_FLAG -r requirements.txt
 
 TF_CFLAGS=( $(python -c 'import tensorflow as tf; print(" ".join(tf.sysconfig.get_compile_flags()))') )
-TF_LFLAGS=( $(python -c 'import tensorflow as tf; print(" ".join(tf.sysconfig.get_link_flags()))') )
-TF_SHAREDLIB=( $(python -c 'import tensorflow as tf; print(tf.sysconfig.get_link_flags()[-1])') )
+TF_LFLAGS="$(python -c 'import tensorflow as tf; print(" ".join(tf.sysconfig.get_link_flags()))')"
 TF_CXX11_ABI_FLAG=( $(python -c 'import tensorflow as tf; print(tf.sysconfig.CXX11_ABI_FLAG)') )
```
PatWie: On my version `tf.sysconfig.CXX11_ABI_FLAG` does not exist. 
From TF v1.9 to 1.13 [I use `tf.__cxx11_abi_flag__`](https://github.com/PatWie/tensorflow-cmake/blob/dc0b4eeee274849abc1effdb29df6b3cd5f39d06/cmake/modules/FindTensorFlow.cmake#L65)

facaiy: Thank you, Patrick. Addons is expected to be compatible with tf 2.x only.

kyleabeauchamp:
Would love to get a review / merge on this one, as I think it will finally make building on my system possible without source code changes.

Issue #243, opened 2019-05-13
Make image transforms compatible with TF Dataset


seanpmorgan:
From #242 we found that the tfa.image transforms are not compatible with dataset mapping. This is a very likely usecase for tfa.image so we need to address this. Here is a minimal example showing the differences in tfa.rotate(pi/2) and tf.image.rot90:
https://colab.research.google.com/drive/1ZDhnGrorvSf04wzS-0utztZ6Nnzusvl7

Problems seem to occur because Dataset mapping runs in graph mode and this check fails:
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/transform_ops.py#L323

By comparison here is tf.image.rot90:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/image_ops_impl.py#L514

kyleabeauchamp:
Will think about this more tonight

alew3:
Acording to https://github.com/tensorflow/tensorflow/issues/27811, TF2 transforms Data mappings to subgraphs for better performance.

Here is a workaround to force Eager Mode and use tfa.image until there is a solution:
image_ds = path_ds.map(lambda path: tf.py_function(func=load_and_preprocess_image,inp=[path],Tout=tf.float32))

(supposing the output is tf.float32)

seanpmorgan:
Thanks for the temporary workaround! Yeah gpu kernels are on the way #118

kyleabeauchamp:
Two thoughts:

1.  I opened #244 to have a (failing) test case for this issue
2.  The requirement of fixed and static image rank is not only in rotate, but upstream in `transform`: https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/transform_ops.py#L76  Is there a strong technical reason why this restriction exists?

kyleabeauchamp:
3.  The `rot90()` code seems to assume that if no shape information is available, one can assume that the rank is 3.  Do we think that's safe for our transformations as well?

WindQAQ:
> 1. I opened #244 to have a (failing) test case for this issue

Thank you, @kyleabeauchamp!

> 2. The requirement of fixed and static image rank is not only in rotate, but upstream in `transform`: https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/transform_ops.py#L76  Is there a strong technical reason why this restriction exists?

[...]

PR #244, opened 2019-05-14
[WIP]  Add failing test case for image dataset transforms

kyleabeauchamp:
Initial attempt at getting test coverage for #243


kyleabeauchamp:
OK, so I've rebased this PR off of #330.  I can confirm that the failing test case is now passing.

kyleabeauchamp:
For sanity, we probably don't want to review this until #330 is merged.

facaiy:
Looks good, Kyle. As we add test case for unknown shape tensor, is it necessary to retain
 tf.data.Dataset test case? What do you think?

kyleabeauchamp:
Yeah it's probably a bit redundant, but the main value is adding a test
that covers the reported use case.

facaiy:
Make sense to me as well, but I prefer to make unit test simple according to Occam's Razor. How about letting us add test case with tf.data.Dataset later when we think it necessary?

kyleabeauchamp:
Sgtm, I'm fine with closing this ticket.

PR #245, opened 2019-05-15
[WIP] Add random image augmentations

kyleabeauchamp:
So it would be nice to have a "batteries-included" suite of random image augmentations that all get compiled as `@tf.function` operators.  I have implemented two here, a random rotation and a random 90 degree rotation.  IMHO, we should assess this PR by answering (in order) the following questions:

1.  Is this in scope?
2.  Are we happy with the API?
3.  Are we happy with the tests?

Finally, we can't really proceed on this PR until we address #243 , but I still wanted to open this for commentary.


kyleabeauchamp:
The CLA warning is a false positive IMHO.

seanpmorgan:
Hi Kyle, thanks for the contribution and solid set of questions for the review. 

1a) First look I don't see any reason why this falls out of scope. In practice I think most people just want augmented images and are less concerned with having full control of that. Would want to confirm this with @facaiy and @WindQAQ  though.

2a) Without a full review I think it looks more or less solid. For the random 90 turn though I think `max_turns` parameter is a bit of API bloat. If someone is choosing random90 turns I struggle to see a time when they would say only 1 or 2 turns.

2b) Should random90 utilize tf.image.rot90. I imagine there are speed ups that could be done on 90 degree turns that may not be accounted for in angle transform, but I haven't dug too far into it.

3a) Seems good. I think the average rot90 test is a little strange though. There are already tests to ensure that random.uniform is correct in TF core and I'm not sure that this would catch anything?

seanpmorgan:
The CLA check is failing because there are commits from several users in this PR I believe. You could rebase your branch with master so this doesn't happen... but for this PR its no big deal we can over-rule the CLA check

kyleabeauchamp:
Regarding 2a, I'm open to changing this.  I guess my goal was to support hypothetical folks who might want either zero or one turn of rotation as part of their augmentation.  I agree that it makes most sense to do arbitrary numbers of turns but I figured I'd keep it general in case there's some great use case that I'm missing.  It's very easy for us to maintain the extra couple of lines for generalization, but it would be hard for us to change in the future if we later decided that the minimalist approach wasn't sufficiently general.  

Regarding 2b, I believe I went with the `rotate` solution because the code in `tf.image.rot90` seems to support only a scalar number of rotations, but in our case we want a random vector of rotation numbers: https://github.com/tensorflow/tensorflow/blob/39357a6f3d6cc75bb8eee002bd7c05b117a05968/tensorflow/python/ops/image_ops_impl.py#L488:L496.  I also noticed that `tf.image.rot90` doesn't seem to support rank 2 input, though we could work around that via reshaping.  I imagine there a ways to work around this by using a for loop of `tf.image.rot90`, but it's not clear to me which approach wins in terms of speed and readability.  

3a:  I agree the test is strange, it was the only obvious way to implement a positive control without relying on a specific random seed.  I'm open to better approaches.

facaiy:
Great idea, Kyle, I woder if [tf.keras.preprocessing.image.ImageDataGenerator](https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/keras/preprocessing/image/ImageDataGenerator) works for you however.

kyleabeauchamp:
I think the Keras ImageDataGenerator is great.  My main concern with it is that it seems to rely a lot on numpy code, rather than using TF image operations (https://github.com/keras-team/keras-preprocessing/blob/master/keras_preprocessing/image/image_data_generator.py), which seems like a more natural tool if we're pipelining Dataset pipelines.  Let me know if I'm mis-reading that code, however.

facaiy:
I'm just worried the users might only want an image augmentation functionality, rather than a series of small ops.

kyleabeauchamp:
I totally agree, we might want to build a pure tf approach for custom
augmentation pipelines.

facaiy:
Let me know if you want to take a try, Kyle. Or we can create an issue for keras in tensorflow/tensorflow.

kyleabeauchamp:
Revisiting this, I think that ideally we provide users with both a full-featured augmentation function as well as composable individual components for folks who are doing something more specialized.  For example, I've encountered situations where I wanted something slightly outside the scope of the keras function---but hit a roadblock because the keras function didn't get me the right "legos" to play with.  

FWIW, here is the augmentation tool I've used internally---I'd love to generalize and optimize it at some point as well.  https://gist.github.com/kyleabeauchamp/41b218d940ca8340b0b5dad5da7ff2ec

facaiy:
Hi, Kyle, I'm not against individual components. Just a little worried that we have to create (and maintain) too many small ops (anyone can implement it by themselves quickly). Anyway, it makes sense to have an efficient image_augmentation op.

@WindQAQ @seanpmorgan Tzu-Wei, Sean, what do you think?

WindQAQ:
Though I'm not in favor of small ops (the combination of addons' ops), we could have this one after there're are some requests from the community. But anyway, thanks for the PR, Kyle :P

seanpmorgan:
I'd defer to you two on what type of ops the subpackage should have. One thing I would like to add is that it'd be nice to benchmark any new image_augmentation op against the numpy/keras equivalent to make sure there is truly a performance improvement.

kyleabeauchamp:
I've refactored this to use the nice new helper utilities that are available in master.

karmel:
Coincidentally, @Dref360 and @fchollet have been working on some of this functionality for [keras-preprocessing](https://github.com/keras-team/keras-preprocessing). Can we combine efforts here?

kyleabeauchamp:
I'm happy to combine efforts, particular if it helps us agree on the right API and functionality.

seanpmorgan:
I think to avoid conflict here it's better to close this PR and see where the keras preprocessing ends up. Happy to re-open if there are continued calls for this and existing functionality isn't there or is slow.

Thanks again Kyle.

Dref360:
Hello!
Sorry for the late response. 
We have a meeting tomorrow to discuss the future of Keras Prepro. I'll post a summary and see how it goes.

Dref360:
Summary:
- The RFC is in good shape and should get approved quickly.
- Adding ops to TF is part of this RFC.
- The new K-Preprocessing API will support tf.data.Dataset.

So we should combine our efforts, but I think we should wait for the RFC to be merged and see what ops should be added to tf or to tf.addons.

facaiy:
@Dref360  Which RFC, Frédéric?

Dref360:
This one: https://github.com/keras-team/governance/pull/6

PR #252, opened 2019-05-18
Implement sparse image warp

kyleabeauchamp:
As requested in #25.  

The main reason this is WIP is that I had to skip a handful of tests that were incompatible with eager mode.  I still need to wrap my head around some of the eager testing / gradient mechanics, so I might not be able to finish this myself.


tensorflow_addons/image/interpolate_spline.py
```diff
@@ -0,0 +1,304 @@
...
+      v: `[b, d, k]` weights on each input dimension
+    Raises:
+      ValueError: if d or k is not fully specified.
+    """
+
+    # These dimensions are set dynamically at runtime.
+    b, n, _ = tf.unstack(tf.shape(train_points), num=3)
+
+    d = train_points.shape[-1]
+    if tensor_shape.dimension_value(d) is None:
```
WindQAQ: I'd suppose we could access the static shape directly. Like
```python
if train.shape[-1] is None:
```

tensorflow_addons/image/interpolate_spline_test.py
```diff
@@ -0,0 +1,365 @@
...
+
+                target_interpolation = tp.HARDCODED_QUERY_VALUES[(order,
+                                                                  reg_weight)]
+                target_interpolation = np.array(target_interpolation)
+                with self.cached_session() as sess:
+                    interp_val = sess.run(interpolator)
+                    self.assertAllClose(interp_val[0, :, 0],
+                                        target_interpolation)
+
+    def test_nd_linear_interpolation_unspecified_shape(self):
```
WindQAQ: To use placeholder in the test case, we could test it with decorator `test_utils.run_deprecated_v1`.
Another way to test the shape of None at graph build time is https://github.com/tensorflow/addons/issues/260#issuecomment-494886404.

tensorflow_addons/image/interpolate_spline_test.py
```diff
@@ -0,0 +1,365 @@
...
+
+        with self.assertRaises(ValueError):
+            _ = interpolate_spline(train_points_ph_invalid, train_values_ph,
+                                   query_points_ph, order, reg_weight)
+
+        with self.assertRaises(ValueError):
+            _ = interpolate_spline(train_points_ph, train_values_ph_invalid,
+                                   query_points_ph, order, reg_weight)
+
+    def test_interpolation_gradient(self):
```
WindQAQ: I wonder if [tf.test.compute_gradient](https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/test/compute_gradient) can help you.

kyleabeauchamp:
Thanks for the comments, I believe I have addressed all three.  On the third comment, I ended up switching to use `GradientTape`, which seems like the most natural thing to do moving forward in TF2.  However, let me know if I'm mistaken there...

tensorflow_addons/image/sparse_image_warp_test.py
```diff
@@ -0,0 +1,255 @@
...
+# ==============================================================================
+"""Tests for sparse_image_warp."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import numpy as np
+import tensorflow as tf
+import tensorflow.compat.v1 as tf1  # TODO: port TF1 test files?
+from tensorflow_addons.image import _get_boundary_locations
```
seanpmorgan: Hi Kyle, thanks for all your great work! Do you mind replacing these with for example:
`from tensorflow_addons.image.sparse_image_warp import _get_boundary_locations`

It's probably best we keep the underscore prefix methods out of the __init__

seanpmorgan:
@yhliang2018 Could you give this a review when time allows? Thanks.
cc @parth-p

tensorflow_addons/image/sparse_image_warp_test.py
kyleabeauchamp: done

yhliang2018:
@seanpmorgan Sure, I will take a look.

tensorflow_addons/image/sparse_image_warp.py
```diff
@@ -0,0 +1,201 @@
...
+                            edge (not including the corners).
+                            The total number of points added is
+                            4 + 4*(boundary_points_per_edge).
+
+    Returns:
+      merged_control_point_locations: augmented set of control point locations
+      merged_control_point_flows: augmented set of control point flows
+    """
+
+    batch_size = tensor_shape.dimension_value(control_point_locations.shape[0])
```
yhliang2018: Use `tf.compat.dimension_value`: https://github.com/tensorflow/tensorflow/blob/5fed792127f6b9ad34a7be5f6a2b17ce90c50edd/tensorflow/python/framework/tensor_shape.py#L98

tensorflow_addons/image/interpolate_spline_test.py
```diff
@@ -0,0 +1,365 @@
...
+        query_points = tf.constant(query_points_np)
+        query_values = self.np_function(query_points_np)
+
+        return query_points, query_values, train_points, train_values
+
+
+class _QuadraticPlusSinProblem1D(_InterpolationProblem):
+    """1D interpolation problem used for regression testing."""
+    DATA_DIM = 1
+    HARDCODED_QUERY_VALUES = {
```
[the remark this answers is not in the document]
kyleabeauchamp: I guess my preference is to leave them where they are so that the the tests and the test constants are co-located...also that's how the code was in TF.

kyleabeauchamp:
OK I think I've addressed all the comments.

kyleabeauchamp:
So I had to add scipy to the requirements.  In previous commits, even in this PR, it was transitively included from the tensorflow install.  However, it was missing in the latest couple of commits.  For now, the easiest solution was to include it in the requirements.

kyleabeauchamp:
OK, I resolved all conflicts again on this PR, so it should be all synced up for review/merge.

Issue #218, opened 2019-04-29
Add recompute_grad decorator


olesalscheider:
**System information**
- TensorFlow version (you are using): 2.0 alpha
- TensorFlow Addons version: 1.2.1
- Is it in the tf.contrib (if so, where): Yes, tf.contrib.layers.recompute_grad
- Are you willing to contribute it (yes/no): Yes
- Are you willing to maintain it going forward? (yes/no): No

**Describe the feature and the current behavior/state.**
The tf.contrib.layers.recompute_grad decorator allows to recompute the decorated function on the backward pass. This way the activation maps do not have to be kept in memory.
It allows to train larger models that normally would not fit into the VRAM but with higher computational costs. The current implementation works fine for me.

**Will this change the current api? How?**
No.

**Who will benefit with this feature?**
Anyone who wants to train a very large network with limited GPU memory.

**Any Other info.**
I know that the current RFC for addons (https://github.com/tensorflow/community/blob/master/rfcs/20181214-move-to-addons.md) proposes not to move the decorator, but it is marked "pending objections". I would like to ask you to reconsider your decision. I have used the decorator in the past whenever necessary (i. e. when I would run out of memory). The reason why you don't find OSS projects using it by default might be that you do not want the annotation if it is not necessary because it slows down the training process. But it is still a good alternative if you do not have a GPU with lots of VRAM available.

seanpmorgan:
Hi @olesalscheider welcome and thanks for bringing this up. I agree this is a useful thing to have, but I do want to make sure we're not duplicating functionality. 

I haven't looked tpo closely at the details, but is the same capability provided by:
https://github.com/openai/gradient-checkpointing

It looks like it's set for tf1.x, but perhaps upgrading that repo is another option.

olesalscheider:
The implementation that you linked to existed before the decorator. It replaces tf.gradients and actually offers more than the decorator in TensorFlow: It also comes with some heuristics to decide without manual annotation which activation maps can be recomputed.
I think TF has something like that in grappler's memory optimizer nowadays but it will only recompute very cheap operations: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/grappler/optimizers/memory_optimizer.cc#L60

The code that you linked to has not been updated in some time and it is not compatible with TF 2.
[...]

PR #256, opened 2019-05-21
WIP Add the recompute_grad decorator from tf.contrib

olesalscheider:
Fix https://github.com/tensorflow/addons/issues/218

I tried to port the [recompute_grad](https://www.tensorflow.org/versions/r1.12/api_docs/python/tf/contrib/layers/recompute_grad?hl=en) decorator to TF 2.0.
However, there are still three TODOs in the code where I don't know how to handle them.
Also, the tests still fail in graph mode...

### DIFF
--- a/tensorflow_addons/layers/decorators_test.py
+++ b/tensorflow_addons/layers/decorators_test.py
@@ -1,0 +1,176 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Tests for Decorators."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import numpy as np
+import tensorflow as tf
+
+from tensorflow_addons.layers import decorators
+from tensorflow_addons.utils import test_utils
+
+
+@test_utils.run_all_in_graph_and_eager_modes
+class RecomputeGradTest(tf.test.TestCase):
+
+  def testRecomputeGrad(self):
+
+    m = tf.keras.Sequential()
+    for _ in range(3):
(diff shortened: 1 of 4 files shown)


tensorflow_addons/layers/decorators_test.py
```diff
@@ -0,0 +1,176 @@
...
+    with tf.GradientTape() as tape:
+      x = tf.ones((2, 4), tf.float32)
+      y = layer_with_recompute(x)
+      loss = tf.math.reduce_sum(y)
+    tvars = dense.trainable_variables + \
+            batch_normalization.trainable_variables
+    tape.gradient(loss, tvars)
+
+    # TODO: What to do with this?
+    #update_ops = ops.get_collection(ops.GraphKeys.UPDATE_OPS)
```
facaiy: Hi, Niels, I'm not sure what do you mean here. If we want to collect update_ops, perhaps we could build [tf.keras.Sequential](https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/keras/Sequential) or create a subclass of [tf.keras.Model](https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/keras/Model). And then collect all update_ops by `[layer.updates for layer in model.layers]`, does it work for you?

facaiy: You might want to read: https://www.tensorflow.org/alpha/guide/keras/functional

olesalscheider: I mean what this check is supposed to do the following: Because of the recompute_grad decorator the function is evaluated multiple times. But we want to only update the batch normalization statistics on the first run, because otherwise we get wrong statistics.
So far in graph mode the batch normalization updates were added to the UPDATE_OPS collection. Therefore the code above could remove the duplicate update ops before they were carried out.

But now in eager mode the updates are carried out in place. So we can't take them back anymore.
Therefore we need a way to avoid that the update happens in the first place in eager mode.
I don't know how to do that without modifying the BatchNormalization layer so this still is a TODO.

facaiy: Got it, thank you for the clarification, Niels. 

I'll be out of town in the next two weeks, @WindQAQ @kyleabeauchamp Tzu-Wei, Kyle, could you take a look and help Niels out when time allows? Thank you :-)

facaiy: @pavithrasv @qlzh727 @mihaimaruseac Hi, Pavithra, Qianli, Mihai, do you have any good idea to control the updating behavior of layer? Thank you very much, cc @fchollet

seanpmorgan:
Again thank you very much for your work here... apologies for the miscommunication w.r.t this being added to tf core. Its on the agenda for today's meeting to discuss.

PR #257, opened 2019-05-21
Fix distort_image_ops_test

WindQAQ:
Fix failed builds because of update in upstream.

### DIFF
--- a/tensorflow_addons/image/distort_image_ops.py
+++ b/tensorflow_addons/image/distort_image_ops.py
@@ -138,6 +138,10 @@ def adjust_hsv_in_yiq(image,
     """
     with tf.name_scope(name or "adjust_hsv_in_yiq"):
         image = tf.convert_to_tensor(image, name="image")
+
+        if image.shape.rank < 3:
+            raise ValueError("input must be at least rank 3.")
+
         # Remember original dtype to so we can convert back if needed
         orig_dtype = image.dtype
         flt_image = tf.image.convert_image_dtype(image, tf.dtypes.float32)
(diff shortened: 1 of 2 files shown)


kyleabeauchamp:
Is there a nightly build link or something that shows the failure?  I trust you, but for the sake of reproducibility might be nice to have some sort of link to the exception.

WindQAQ:
@kyleabeauchamp, hi Kyle, here it is
https://source.cloud.google.com/results/invocations/0388ed59-b166-4bdd-85e3-cb05ac762e22/targets

kyleabeauchamp:
Thanks!

WindQAQ:
Shape checking within `tf.function` is somewhat not robust in either c++ or python API... @facaiy, hi facai, is there any suggested way to do this.

Btw, thanks for asking and review, Kyle 😄

seanpmorgan:
Thanks for the fix! Just trying to understand what happend here... typically the RegEx is supposed to catch the error thrown by this I believe:
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/custom_ops/image/cc/ops/distort_image_ops.cc#L33

But at first pass I can't find the commit that affected this. LGTM to merge if @facaiy is good with it.. but I think something has changed with shape checking in our op?

seanpmorgan:
One option would be to merge this fix and create a separate issue to look into why this happend.

WindQAQ:
> Thanks for the fix! Just trying to understand what happend here... typically the RegEx is supposed to catch the error thrown by this I believe:
> https://github.com/tensorflow/addons/blob/master/tensorflow_addons/custom_ops/image/cc/ops/distort_image_ops.cc#L33
> 
> But at first pass I can't find the commit that affected this. LGTM to merge if @facaiy is good with it.. but I think something has changed with shape checking in our op?

Hmm, the original regex text and the type of exception seem wrong while the test passed previously... When checking core TensorFlow, I find this commit on Apr 10:
https://github.com/tensorflow/tensorflow/commit/4b4a39e7120b1c7744f9686bc6cce9363846d7e6#diff-68b5e47db1d9389c8d12852996845819

> Op raising `InvalidArgumentError` unless `x` is all negative. This can be
      used with `tf.control_dependencies` inside of `tf.function`s to block
      followup computation until the check has executed.

According to the doc, does it encourage us to use `tf.control_dependencies` inside `tf.function` to do something like shape checking?

PR #261, opened 2019-05-22
FIX: LazyAdam init

seanpmorgan:
Fixes a break caused by change in upstream:
https://github.com/tensorflow/tensorflow/commit/611db3490b1dcf22c13ffcb64f50b75faa5e0ecc


seanpmorgan:
Only question would be should the name_scope for this optimizer be LazyAdam instead of Adam?

facaiy:
I agree, we should override `__init__` method to modify the default name .

PR #266, opened 2019-05-31
Refactor loss to inherit from tf.keras.losses.Loss

### LINKED ISSUE #259: Refactor losses to inherit from base keras.Loss
Per recommendation in https://github.com/tensorflow/tensorflow/issues/28601 we should aim to inherit from base `Loss`. I believe this only requires a minor refactoring so replicate the results of `LossWrapper`

WindQAQ:
Closes #259.

### DIFF
--- a/tensorflow_addons/losses/sparsemax_loss.py
+++ b/tensorflow_addons/losses/sparsemax_loss.py
@@ -115,6 +115,14 @@ class SparsemaxLoss(tf.keras.losses.Loss):
             raise ValueError('from_logits must be True')
 
         super(SparsemaxLoss, self).__init__(name=name, reduction=reduction)
+        self.from_logits = from_logits
 
     def call(self, y_true, y_pred):
         return sparsemax_loss_from_logits(y_true, y_pred)
+
+    def get_config(self):
+        config = {
+            "from_logits": self.from_logits,
+        }
+        base_config = super(SparsemaxLoss, self).get_config()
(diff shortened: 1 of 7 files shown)


tensorflow_addons/losses/sparsemax_loss.py
```diff
@@ -114,5 +114,7 @@ def __init__(self,
         if from_logits is not True:
             raise ValueError('from_logits must be True')
 
-        super(SparsemaxLoss, self).__init__(
-            sparsemax_loss_from_logits, name=name, reduction=reduction)
+        super(SparsemaxLoss, self).__init__(name=name, reduction=reduction)
+
+    def call(self, y_true, y_pred):
+        return sparsemax_loss_from_logits(y_true, y_pred)
```
seanpmorgan: Should this also have it's own `get_config` method? The `from_logits` parameter doesn't seem like the best way to handle that situation, but for consistency if we have parameters we should serialize them.

WindQAQ: Sure, I'm fine with this one. Also cc @AndreasMadsen.

AndreasMadsen: `from_logits` Is a standardized parameters, that is often False by default. However for sparsemax there is no way of handling that situation. Thus it will just throw.

In isolation it looks pretty wired to have the parameter, but it appears in Softmax and other classification losses which is why it is there.

PR #270, opened 2019-06-04
Update sparsemax to use where v2

### LINKED ISSUE #253: Update sparsemax to use tf.where V2
As described in #250 and temporarily patched in #251 sparsemax has one instance of tf.where that needs the broadcasting dimensions changed to match numpy and TF2 style.

seanpmorgan:
Closes #253 cc @AndreasMadsen


AndreasMadsen:
Hmm, I don't think the `broadcast_to` necessary according to the docs? https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/where

Otherwise, it looks fine.

seanpmorgan:
Thanks Andreas... yeah the broadcast_to wasn't doing anything there.

WindQAQ:
Thank you!

Issue #119, opened 2019-03-30
Make whls manylinux2010 compatible


seanpmorgan:
Currently our manylinux1 packages are breaking spec. This is the same issue for TensorFlow core and many other pip packages with C++ linking. Creating this issue to track the problem and we'll make the switch once the new BUILD toolchain is finalized.

Tracking:
https://groups.google.com/a/tensorflow.org/forum/#!topic/build/Xo190xAPyW8

seanpmorgan:
Re-opening as there is an issue loaded ops that are manylinux2010 repaired. Waiting on this issue:
https://github.com/tensorflow/tensorflow/issues/31807

Issue #75, opened 2019-03-06
Package Addons for python3.7 


seanpmorgan:
Using the current docker image, we're unable to install / build TensorFlow in python3.7 with the following error:
```
ImportError: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by /root/miniconda2/envs/py37/lib/python3.7/site-packages/tensorflow/python/_pywrap_tensorflow_internal.so)
```

It seems like we need to compile using Ubuntu 16.04 libs, which will require a bump in the custom-op docker image. @yifeif would it be possible to bump the base image in the container? If so we can rebuild our release as 0.1.1 and include py37.

Alternatively, we could create a new docker image for builds.

facaiy:
@dynamicwebpaige Paige, tf-2.0-preview has supported python 3.7 in https://github.com/tensorflow/tensorflow/issues/25429 . Could you tell us how did you do it?

seanpmorgan:
cc @gunan. Is ubuntu 14.04 being used for the build or some other docker image?

gunan:
For python 3.7, we had to base our builders off of ubuntu 16.04.
Everything else uses ubuntu 14.04

yifeif:
We will definitely need a new image for python3.7.

seanpmorgan:
Addons 0.2.0 has a py37 available on pypi

seanpmorgan:
Bumping to ubuntu16+GCC5 caused issues issues as seen here in the most recent builds of TF:
https://github.com/tensorflow/tensorflow/issues/27067

This will remain unsoved until the manylinux2010 toolchain is finalized. See #119

hartikainen:
What's the ETA for tf addons python 3.7 support?

seanpmorgan:
@perfinion is there an updated build environment and date when TF pip packages will be built in it? Just for reference python 3.8 release is soon.

seanpmorgan:
Looks like this is almost ready:
https://github.com/tensorflow/tensorflow/commit/ec83c8a8292aa881b701928a7ffbe2305cc0e4e5

@av8ramit Could you comment on the above question?

av8ramit:
Ideally worst case within two weeks we will transition over. Waiting on one last issue with GPU custom ops.

PR #271, opened 2019-06-07
ENH: Automate build

### LINKED ISSUE #76: Automate Build Process
Currently we have no automated process for building Addons across python version and operating systems. Going forward we'll want this process to be automated.. but it may be challenging for us to start builds without access to the Google internal tooling.

We could conceivably use Travis... but if we can keep consistent CI that would be ideal.

seanpmorgan:
Closes #76 

Nightlies are being pushed to:
https://pypi.org/project/tfa-nightly/#files

Will set up a nightly cron once this is merged into master. This is also a big help for the 0.4 release which should be happening shortly after TF2 release candidate

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,5 +38,4 @@ wheels/
 /artifacts
 
 # Addons
-.*
 /docs/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,43 @@
+dist: xenial
+language: minimal
+
+git:
+  quiet: true
+
+services:
+- docker
+
+install:
+- curl https://bootstrap.pypa.io/get-pip.py | sudo -H python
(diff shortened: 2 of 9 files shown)


seanpmorgan:
I'll also be adding a `RELEASE.md` in tensorflow/community with the now manageable steps needed for a release

seanpmorgan:
@WindQAQ great question. So we can create a python3.7 for macos right now, but py37 will not install on the custom-op docker image for ubuntu 14.04. We could patch the docker image similar to how SIG-IO has done but I wanted to stick with the custom-op image provided so if there is any ABI issues etc. then we will be able to push that issue upstream.

SIG-Build has mentioned they will be providing build environments which will solve #75 and #119 which is great and from doesn't seem to be too far away.

PR #272, opened 2019-06-07
Pass kwargs to wrapped cell in AttentionWrapper

guillaumekln:


seanpmorgan:
Thank you very much @guillaumekln !

PR #273, opened 2019-06-08
FIX: Use all keras optimizers

seanpmorgan:
Fix broken nightlies and ensure we're only using tf.keras optimizers in test cases


seanpmorgan:
Thx for the review. Looking into it now. It's a build failure... thinking its related to a bazel issue but it might be due to the new virtual pip package

PR #278, opened 2019-06-11
FIX: Use keras loss

seanpmorgan:
Fixes today's broken nightly.

Just curious -- I guess the decision has been made to remove the tf.* aliases? I think its a nice API clean-up personally but I wasn't aware this was happening.


qlzh727:
Thanks for the fix.

PR #284, opened 2019-06-12
f1 metric - micro/macro/weighted

SSaishruthi:
This PR contains a script for adding f1-macro metric.

Colab notebook: https://colab.research.google.com/drive/1LJ1yb8cUgisNP3ETfMLRYPQEyJivzDv-


SSaishruthi:
Please let me know what will be my next steps. To do as far as I know will be to add a test for the same.

tensorflow_addons/metrics/f1_macro.py
```diff
@@ -0,0 +1,63 @@
+import tensorflow as tf
```
Squadrick: Add license and future import like [here](https://github.com/AakashKumarNain/addons/blob/0c31db81ac4d69d63e9fcf0a79aae9feed8e2176/tensorflow_addons/metrics/cohens_kappa.py#L1)

tensorflow_addons/metrics/f1_macro.py
```diff
@@ -0,0 +1,63 @@
...
+        self.false_positives_col.assign_add(tf.cast(tf.math.count_nonzero(
+            y_pred * (y_true - 1), axis=0), tf.float32))
+        # false negative across column
+        self.false_negatives_col.assign_add(tf.cast(
+            tf.math.count_nonzero((y_pred - 1) * y_true, axis=0), tf.float32))
+
+    def result(self):
+        p_sum = tf.cast(self.true_positives_col + self.false_positives_col,
+                        tf.float32)
+        precision_macro = tf.cast(tf.compat.v1.div_no_nan(
```
Squadrick: Replace: `tf.compat.v1.div_no_nan` -> [`tf.math.divide_no_nan`](https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/math/divide_no_nan)

tensorflow_addons/metrics/f1_macro.py
```diff
@@ -0,0 +1,63 @@
...
+            self.true_positives_col, p_sum), tf.float32)
+
+        r_sum = tf.cast(self.true_positives_col + self.false_negatives_col,
+                        tf.float32)
+        recall_macro = tf.cast(tf.compat.v1.div_no_nan(
```
Squadrick: ditto

tensorflow_addons/metrics/f1_macro.py
```diff
@@ -0,0 +1,63 @@
...
+            self.true_positives_col, r_sum), tf.float32)
+
+        mul_value = 2 * precision_macro * recall_macro
+        add_value = precision_macro + recall_macro
+        f1_macro = tf.cast(tf.compat.v1.div_no_nan(mul_value, add_value),
```
Squadrick: ditto

SSaishruthi:
Thanks @Squadrick 
Will update soon

SSaishruthi:
@Squadrick 
Updated with F1 micro, macro and weighted
Test colab notebook: https://colab.research.google.com/drive/1qSq0SsYkPqjdKUgM1RM4kKM67X75ocFj

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -0,0 +1,166 @@
...
+import tensorflow as tf
+from tensorflow.keras.metrics import Metric
+import numpy as np
+
+
+class F1_micro(Metric):
+    """
+    Calculates F1 micro score
+    """
+    def __init__(self, name='f1-score'):
```
Squadrick: Nit: f1-score -> f1_score

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -0,0 +1,166 @@
...
+        super(F1_micro, self).__init__(name=name)
+        self.true_positives = self.add_weight('TP-class', shape=[],
```
Squadrick: Nit: TP-class -> TP_class

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -0,0 +1,166 @@
...
+                                              initializer='zeros',
+                                              dtype=tf.float32)
+        self.false_positives = self.add_weight('FP-class', shape=[],
```
Squadrick: Nit: FP-class -> FP_class

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -0,0 +1,166 @@
...
+                                               initializer='zeros',
+                                               dtype=tf.float32)
+        self.false_negatives = self.add_weight('FN-class', shape=[],
```
Squadrick: Nit: FN-class -> FN_class

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -0,0 +1,166 @@
...
+        # false negative across column
+        self.false_negatives.assign_add(tf.cast(
+            tf.math.count_nonzero((y_pred - 1) * y_true, axis=None),
+            tf.float32))
+
+    def result(self):
+        p_sum = tf.cast(self.true_positives + self.false_positives,
+                        tf.float32)
+        # precision calculation
+        precision_micro = tf.cast(tf.math.divide_no_nan(
```
Squadrick: No need to cast. `p_sum` and `self.true_positives` are both `tf.float32`

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -0,0 +1,166 @@
...
+            self.true_positives, p_sum), tf.float32)
+
+        r_sum = tf.cast(self.true_positives + self.false_negatives,
+                        tf.float32)
+        # recall calculation
+        recall_micro = tf.cast(tf.math.divide_no_nan(
```
Squadrick: ditto

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -0,0 +1,166 @@
...
+            self.true_positives, r_sum), tf.float32)
+
+        mul_value = 2 * precision_micro * recall_micro
+        add_value = precision_micro + recall_micro
+        f1_micro = tf.cast(tf.math.divide_no_nan(mul_value, add_value),
```
Squadrick: No need to cast. `mul_value` and `add_value` are both `tf.float32`.

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -0,0 +1,166 @@
...
+        self.false_positives.assign(0)
+        self.false_negatives.assign(0)
+
+
+class F1_macro_and_weighted(Metric):
+    """
+    Calculates F1 macro or weighted based on the user's choice
+    """
+    def __init__(self, num_classes, average,
+                 name='f1-macro-and-weighted-score'):
```
Squadrick: Nit: f1-macro-and-weighted-score -> f1_macro_and_weighted_score

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -0,0 +1,166 @@
...
+        super(F1_macro_and_weighted, self).__init__(name=name)
+        self.num_classes = num_classes
+        self.average = average
+        self.true_positives_col = self.add_weight('TP-class',
```
Squadrick: Nit: TP-class -> TP_class

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -0,0 +1,166 @@
...
+                                                  shape=[self.num_classes],
+                                                  initializer='zeros',
+                                                  dtype=tf.float32)
+        self.false_positives_col = self.add_weight('FP-class',
```
Squadrick: ditto

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -0,0 +1,166 @@
...
+                                                   shape=[self.num_classes],
+                                                   initializer='zeros',
+                                                   dtype=tf.float32)
+        self.false_negatives_col = self.add_weight('FN-class',
```
Squadrick: ditto

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -0,0 +1,166 @@
...
+                                                   shape=[self.num_classes],
+                                                   initializer='zeros',
+                                                   dtype=tf.float32)
+        self.weights_intermediate = self.add_weight('weights-int-f1',
```
Squadrick: ditto

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -0,0 +1,166 @@
...
+            (y_pred - 1) * y_true, axis=0), tf.float32))
+        # variable to hold intermediate weights
+        self.weights_intermediate.assign_add(tf.cast(
+            tf.reduce_sum(y_true, axis=0), tf.float32))
+
+    def result(self):
+        p_sum = tf.cast(self.true_positives_col + self.false_positives_col,
+                        tf.float32)
+        # calculate precision
+        precision_macro = tf.cast(tf.math.divide_no_nan(
```
Squadrick: Remove cast.

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -0,0 +1,166 @@
...
+            self.true_positives_col, p_sum), tf.float32)
+
+        r_sum = tf.cast(self.true_positives_col + self.false_negatives_col,
+                        tf.float32)
+        # calculate recall
+        recall_macro = tf.cast(tf.math.divide_no_nan(
```
Squadrick: Remove cast.

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -0,0 +1,166 @@
...
+            self.true_positives_col, r_sum), tf.float32)
+
+        mul_value = 2 * precision_macro * recall_macro
+        add_value = precision_macro + recall_macro
+        f1_macro_int = tf.cast(tf.math.divide_no_nan(mul_value, add_value),
```
Squadrick: Remove cast.

tensorflow_addons/metrics/f1_scores.py
WindQAQ: A suggestion here: maybe "true_positives" is a better name to match core TF.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/metrics.py#L1160

Squadrick: Yup, it's better we match core TF.

tensorflow_addons/metrics/f1_scores.py
WindQAQ: I think `F1Micro` should be a better Python class name.

BTW, my thought on different average scheme for f1 score:
```python
class F1Score(Metric):
    def __init__(self, name="f1_score", average="micro"):
        # average could be one of "micro", "macro" or "weighted"
```
This would be much close to sklearn API.
https://scikit-learn.org/stable/modules/generated/sklearn.metrics.f1_score.html

I'm also fine with seperated `F1Micro`, `F1Macro` and `F1Weighted` though. What do you free about this @Squadrick.

SSaishruthi:
@Squadrick 
Updated scripts. Please let me know if there is any change to be made.

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -0,0 +1,177 @@
...
+                                             r_sum)
+
+        mul_value = 2 * precision_macro * recall_macro
+        add_value = precision_macro + recall_macro
+        f1_macro_int = tf.math.divide_no_nan(mul_value, add_value)
+        # f1 macro score
+        f1_score = tf.reduce_mean(f1_macro_int)
+        # condition for weighted f1 score
+        if self.average == 'weighted':
+            f1_int_weights = tf.cast(tf.math.divide_no_nan(
```
Squadrick: Remove cast, not needed.

SSaishruthi:
Hi @Squadrick @WindQAQ 

I have updated code to calculate F1 scores under one class.
Now, we have four options:

1. Micro
2. Macro
3. Weighted
4. None (This will produce F1 scores for each class)

Colab: https://colab.research.google.com/drive/1qSq0SsYkPqjdKUgM1RM4kKM67X75ocFj#scrollTo=hZSxoW4eYqYW

I ran code format to check scripts. I encountered an installation error in `clang` and removed that step because I noticed that is only for C++.

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -23,124 +23,85 @@
...
+        if average not in (None, 'micro', 'macro', 'weighted'):
             raise ValueError("Unknown average type. Acceptable values "
-                             "are: [macro, weighted]")
+                             "are: [micro, macro, weighted]")
         else:
             self.average = average
-        self.true_positives_col = self.add_weight('true_positives',
+            if self.average == 'micro':
+                self.axis = None
+            if self.average == 'macro' or self.average == 'weighted':
```
Squadrick: Use `else` here, no need for an `if`.

tensorflow_addons/metrics/f1_scores.py
```diff
@@ -23,124 +23,85 @@
...
     """
 
     def __init__(self, num_classes, average=None,
-                 name='f1_macro_and_weighted_score', dtype=tf.float32):
-        super(F1MacroAndWeighted, self).__init__(name=name)
+                 name='f1_score', dtype=tf.float32):
+        super(F1Score, self).__init__(name=name)
         self.num_classes = num_classes
-        if average not in (None, 'macro', 'weighted'):
+        if average not in (None, 'micro', 'macro', 'weighted'):
```
Squadrick: What happens when `average` is set to `None`?

Squadrick: Because it passes the check, but the calculation doesn't consider a case when `average=None`.

SSaishruthi: When the average is set to `None`, F1-score is calculated for each class. This condition is handled in https://github.com/SSaishruthi/addons/blob/9348474f998e3bf0438f541832d98e31ea3fc2db/tensorflow_addons/metrics/f1_scores.py#L173

I got reference for `None` from sklearn. Computed is similar to `Macro` but mean will not be computed.

SSaishruthi: I did not add `None` condition before. Added that today. So we have four conditions now:

Micro/ Macro / weighted /None

SSaishruthi: Also, added a test condition for this.

Squadrick:
@SSaishruthi Can you pull from `master` and update this PR?

SSaishruthi:
@Squadrick 

I have updated the branch and resolved conflicts.

tensorflow_addons/metrics/f1_test.py
```diff
@@ -0,0 +1,125 @@
...
+        f1_obj = F1Score(num_classes=3, average='micro')
+        f1_obj1 = F1Score(num_classes=3, average='macro')
+        f1_obj2 = F1Score(num_classes=3, average='weighted')
+
+        self.evaluate(tf.compat.v1.variables_initializer(f1_obj.variables))
+        self.evaluate(tf.compat.v1.variables_initializer(f1_obj1.variables))
+        self.evaluate(tf.compat.v1.variables_initializer(f1_obj2.variables))
+        return f1_obj, f1_obj1, f1_obj2
+
+    def initialize_vars_none(self):
```
Squadrick: Combine `f1_none` into `initialize_vars`. It's better than having a separate method that does the exact same thing.

tensorflow_addons/metrics/f1_test.py
```diff
@@ -0,0 +1,125 @@
...
+        # Check save and restore config
+        f1_obj2 = F1Score.from_config(f1_obj.get_config())
+        self.assertEqual(f1_obj2.name, 'f1_score')
+        self.assertEqual(f1_obj2.dtype, tf.float32)
+        self.assertEqual(f1_obj2.num_classes, 3)
+
+    def initialize_vars(self):
+        f1_obj = F1Score(num_classes=3, average='micro')
+        f1_obj1 = F1Score(num_classes=3, average='macro')
+        f1_obj2 = F1Score(num_classes=3, average='weighted')
```
Squadrick: Nit: Rename `f1_obj` -> `f1_micro`, `f1_obj1` -> `f1_macro`, `f1_obj2` -> `f1_weighted`.

tensorflow_addons/metrics/f1_test.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def update_obj_states(self, f1_obj, f1_obj1, f1_obj2, actuals, preds):
+        update_op1 = f1_obj.update_state(actuals, preds)
+        update_op2 = f1_obj1.update_state(actuals, preds)
+        update_op3 = f1_obj2.update_state(actuals, preds)
+        self.evaluate(update_op1)
+        self.evaluate(update_op2)
+        self.evaluate(update_op3)
+
+    def update_obj_states_none(self, f1_none, actuals, preds):
```
Squadrick: Refactor to include this in `update_obj_states`.

tensorflow_addons/metrics/f1_test.py
```diff
@@ -0,0 +1,125 @@
...
+        # Initialize
+        f1_obj, f1_obj1, f1_obj2 = self.initialize_vars()
+        # Update
+        self.update_obj_states(f1_obj, f1_obj1, f1_obj2, actuals, preds)
+        # Check results
+        self.check_results(f1_obj, 0.6666666)
+        self.check_results(f1_obj1, 0.6555555)
+        self.check_results(f1_obj2, 0.6777777)
+
+    def test_f1_none_score(self):
```
Squadrick: Rather make making a separate test case for `average=None`, include it in the other `test_f1_*` methods. So we can check for `perfect`, `random` and `worst` for the `None` case.

tensorflow_addons/metrics/f1_test.py
Squadrick: Do the renaming through out the file.

SSaishruthi:
@Squadrick 

Inputs to `None` condition should be multiclass but not multilabel. That's why I created a separate function for `None` as they require a different set of inputs. Can we have it like this or if you can suggest me an alternative that would be great.

Thanks

Squadrick:
@SSaishruthi Oh, my bad. In that case, leave it as is.

tensorflow_addons/metrics/f1_test.py
Squadrick: We can merge once as soon as this is done.

Squadrick:
Triggering the test cases.

SSaishruthi:
@Squadrick updated the scripts. Can you please trigger the test again?

Issue #230, opened 2019-05-02
Batch Processing for Tensorflow_addons.image filters functionalities


Mainak431:
Batch processing is very much required for all the image functionalities. Batch processing will enable direct use of these features with layers like conv2d. Hence no preprocessing steps will be required for CNN applications. Currently, this is required. Directly, the user can create the network and feed the images for training. I wish to implement the batch processing for median_filter2d and mean_filter2d using the currently available functions. Please provide feedback on the same.

Mainak431:
@seanpmorgan CC.

seanpmorgan:
Closing this as it'll be address with #234

PR #288, opened 2019-06-14
Support batch input for median_filter2d

WindQAQ:
Addresses #230. Also meets the interface of mean_filter2d.

A general random test case is down below. Maybe there are some ways to speed up. It's now much more slower than scipy's implementation even for the input of shape [1, h, w, 1].
https://colab.research.google.com/drive/1UkNXxjTMHifKmHB2N5ojOHFojtXrOUIG

### DIFF
--- a/tensorflow_addons/image/filters.py
+++ b/tensorflow_addons/image/filters.py
@@ -189,6 +189,10 @@ def median_filter2d(image,
                 strides=[1, 1, 1, 1],
                 rates=[1, 1, 1, 1],
                 padding="VALID")
+
+            # Note the returned median is casted back to the original type
+            # Take [5, 6, 7, 8] for example, the median is (6 + 7) / 2 = 3.5
+            # It turns out to be int(6.5) = 6 if the original type is int
             top = tf.nn.top_k(patches, k=ceil).values
             median = (top[:, :, :, floor - 1] + top[:, :, :, ceil - 1]) / 2
             return tf.dtypes.cast(median, x.dtype)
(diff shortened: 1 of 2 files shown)


tensorflow_addons/image/filters.py
```diff
@@ -142,80 +128,80 @@ def mean_filter2d(image,
...
+            x = tf.expand_dims(x, axis=-1)
+            patches = tf.image.extract_patches(
+                x,
+                sizes=[1, filter_shape[0], filter_shape[1], 1],
+                strides=[1, 1, 1, 1],
+                rates=[1, 1, 1, 1],
+                padding="VALID")
+            top = tf.nn.top_k(patches, k=ceil).values
+            median = (top[:, :, :, floor - 1] + top[:, :, :, ceil - 1]) / 2
+            return tf.dtypes.cast(median, x.dtype)
```
facaiy: Do we need to document the behavior? Say, return 6, rather than 6.5, for [5, 6, 7, 8]

facaiy:
@Mainak431 Hi, would you mind taking a look?

tensorflow_addons/image/filters.py
WindQAQ: Sure, it makes sense

Mainak431:
@WindQAQ Thanks for the excellent contribution. Everything looks pretty fine.
Can you please add a small training example using this layer with a conv2d layer and dropout layer in tensorflow with maybe mnist dataset? So, that it can be directly used in the network itself?

WindQAQ:
@Mainak431 Here is the example [colab notebook](https://colab.research.google.com/drive/1ZOAFTRdxKYDBpG75ZITukIr8MjUR2qDA). I will include this in addon's example in maybe the next PR.

Mainak431:
Nice contribution.

Squadrick:
@WindQAQ Thanks for the contribution.

PR #291, opened 2019-06-15
Make tests compatible with graph mode

WindQAQ:
1. Do TODO : make tests compatible with graph mode in distort_image_ops.
2. Revert #257. 
3. Have to use `self.evaluate()` to make exception raised in C++.

### DIFF
--- a/tensorflow_addons/image/distort_image_ops.py
+++ b/tensorflow_addons/image/distort_image_ops.py
@@ -139,9 +139,6 @@ def adjust_hsv_in_yiq(image,
     with tf.name_scope(name or "adjust_hsv_in_yiq"):
         image = tf.convert_to_tensor(image, name="image")
 
-        if image.shape.rank < 3:
-            raise ValueError("input must be at least rank 3.")
-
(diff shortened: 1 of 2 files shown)


kyleabeauchamp:
I think this all LGTM.  The only question I have is whether we need to do anything more along the lines of  to #260 / #290

WindQAQ:
Hi Kyle, initial codes migrated from contrib have only dynamic shape checking (check in graph building time). When decorating with `tf.function`, they randomly failed the CI testing (I don't know the reason...). As suggested in https://github.com/tensorflow/addons/issues/260#issuecomment-494902359, doing static shape checking and dynamic shape checking seem to be the most robust and most correct way (see https://github.com/tensorflow/addons/issues/260#issuecomment-494886404 for more details).

facaiy:
Tzh-Wei, what will happen without self.evaluate?

WindQAQ:
Hi, Facai, without `self.evaluate`, the test will fail in graph mode because the OpError is not raised.

Squadrick:
@WindQAQ Looks good, let's get this merged.

Issue #118, opened 2019-03-30
Implement GPU kernels


seanpmorgan:
Currently we've omitted the gpu kernels from addons while we got everything setup. Now that CI testing is setup for GPU we should begin adding these to addons. There's a lot of things required so I wanted to start this issue so we can track / discuss.

Some of the TODO:
- [x] Add gpu kernels and compile them as shown in the [custom_op template](https://github.com/tensorflow/custom-op) which now includes a GPU op : PR #294
- Configure [BUILD scripts to use cuda tests](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/BUILD#L33)
   - [x] image/image_projective_transform_op_gpu.cu.cc  #294
   - [x] distance_transform_ops
   - [x] image/adjust_hsv_in_yiq_op_gpu.cu.cc
   - [x] seq2seq/beam_search_ops_gpu.cu.cc
- [x] Update CONTRIBUTING documentation and add `tensorflow/tensorflow:custom-op-gpu` as the docker image 
-  [x] Specify the `TF_NEED_CUDA` variable appropriately in our scripts

seanpmorgan:
@gunan @martinwicke Is there anyway we can see how https://github.com/tensorflow/community/pull/2 is progressing?

We've been discussing how we would like to package an Addons that includes GPU kernels and I believe we'll be depending on this in order to have a single pip package that loads either gpu or non gpu kernels. Is that correct? Publishing an `addons-gpu` does not seem like a good idea if this is on the horizon

gunan:
@yifeif and I are working towards this
We are hoping to deliver this in Q2

seanpmorgan:
Small update: dynamic kernels are now implemented on the tensorflow master branch so we can begin work on this.

Squadrick:
@seanpmorgan Finally got myself an nvidia GPU. I can work on this.

Squadrick:
@gunan Hey, do you have some sample code on how this is supposed to be used?

In the RFC, you mention a new API: `tf.load_kernel_library`, but I can't seem to find in the `master` branch of TensorFlow, I did find `tf.load_library` though.

gunan:
During implementation, we decided to make it more general and name it "load_library" instead.
Please use that function instead.

facaiy:
[...]

PR #294, opened 2019-06-17
set up GPU environment

facaiy:
First step of  #118 .
Refer to GPU op example in https://github.com/tensorflow/custom-op


configure.sh
```diff
@@ -53,3 +53,19 @@ write_action_env_to_bazelrc "TF_HEADER_DIR" ${TF_CFLAGS:2}
...
+write_to_bazelrc "build --spawn_strategy=standalone"
+write_to_bazelrc "build --strategy=Genrule=standalone"
+write_to_bazelrc "build -c opt"
+write_action_env_to_bazelrc "TF_NEED_CUDA" ${TF_NEED_CUDA}
+# TODO(yifeif): do not hardcode path
+if [[ "$TF_NEED_CUDA" == "1" ]]; then
+  write_action_env_to_bazelrc "CUDNN_INSTALL_PATH" "/usr/lib/x86_64-linux-gnu"
+  write_action_env_to_bazelrc "TF_CUDA_VERSION" "10.0"
+  write_action_env_to_bazelrc "TF_CUDNN_VERSION" "7"
+  write_action_env_to_bazelrc "CUDA_TOOLKIT_PATH" "/usr/local/cuda"
```
seanpmorgan: How do we feel about using $CUDA_HOME here?

tensorflow_addons/custom_ops/image/BUILD
```diff
@@ -35,11 +78,11 @@ cc_binary(
     copts = [
         "-pthread",
         "-std=c++11",
-        D_GLIBCXX_USE_CXX11_ABI,
-    ],
+        "-D_GLIBCXX_USE_CXX11_ABI=0",
```
seanpmorgan: Hmmm why remove the configurable ABI flag here?

tensorflow_addons/utils/test_utils.py
```diff
@@ -26,6 +26,7 @@
 from tensorflow.python.framework.test_util import run_all_in_graph_and_eager_modes
 from tensorflow.python.framework.test_util import run_deprecated_v1
 from tensorflow.python.framework.test_util import run_in_graph_and_eager_modes
+from tensorflow.python.framework.test_util import use_gpu
```
seanpmorgan: This seems fairly straightforward to just fork?
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/test_util.py#L1420

configure.sh
seanpmorgan: Hmmm seems custom-op-gpu doesn't have this ENV variable though I do think it's pretty standard practice:
https://numba.pydata.org/numba-doc/dev/cuda/overview.html#setting-cuda-installation-path

facaiy:
@seanpmorgan Yeah, Sean. The PR is a draft, and most of codes are just copied from the gpu example of custom-op. I want to use it to check whether our GPU environment is ready. If all tests pass, then I'll like to reconstruct all the codes and invite everyone to take a look :-)

configure.sh
facaiy: Agree, I'll fix them later after we fix the environment issues :-)

facaiy:
@av8ramit Amit, can you compare internal image with tensorflow/tensorflow:custom-op-gpu? Does they share the same CUDA setting?

> Cuda Configuration Error: No library found under: /usr/lib/x86_64-linux-gnu/lib64/libcudnn.so.7, /usr/lib/x86_64-linux-gnu/lib64/stubs/libcudnn.so.7, /usr/lib/x86_64-linux-gnu/lib/powerpc64le-linux-gnu/libcudnn.so.7, /usr/lib/x86_64-linux-gnu/lib/x86_64-linux-gnu/libcudnn.so.7, /usr/lib/x86_64-linux-gnu/lib/x64/libcudnn.so.7, /usr/lib/x86_64-linux-gnu/lib/libcudnn.so.7, /usr/lib/x86_64-linux-gnu/libcudnn.so.7

facaiy:
cc @yifeif @gunan

seanpmorgan:
Friendly bump as this is blocking one our of milestone issues.
Is cuDNN installed on the GCP image? If so we can set an [ENV variable in our GPU CI script](https://github.com/tensorflow/addons/blob/master/tools/ci_testing/addons_gpu.sh) 

cc @karmel

yifeif:
I assume there are some minor differences between the setup, and we are also in the middle of moving everything to centos (cc: @av8ramit).
Does addon use tensorflow/tensorflow:custom-op-gpu for release, GCP image for build? Or GCP image for both?

seanpmorgan:
> I assume there are some minor differences between the setup, and we are also in the middle of moving everything to centos (cc: @av8ramit).
> Does addon use tensorflow/tensorflow:custom-op-gpu for release, GCP image for build? Or GCP image for both?

So our current setup is to use travis and custom-op images for build/release and GCP for CI testing (travis does not have GPU servers).

Writing this out though we'll need to confirm we can compile with nvcc on a no GPU server (I believe it's possible but haven't tried.)

av8ramit:
I'm currently in the process of moving everything to CentOS. As for the current Ubuntu images I'm not sure of any significant differences. As I understand it they are designed to be fairly similar to each other. Any commands or tests I can run on the GCP instances for you?

facaiy:
I think we'd better use the same image for both build/release and CI testing. It seems that tensorflow/tensorflow:custom-op-gpu works well right now. As for GCP image, is it possible that we can get it from somewhere? Or, can we use tensorflow/tensorflow:custom-op-gpu for CI testing?

facaiy:
I find it really difficult to debug if we cannot get access to the CI testing image.

av8ramit:
Hey @facaiy yeah I agree. Unfortunately this is for security reasons as well as a route to provide easy debugging internally. I'm not opposed to having your entire setup run on  tensorflow/tensorflow:custom-op-gpu.

gunan:
Unfortunately, I cannot give you our GCP images.
However, the custom op docker image has exactly the same toolchains set up.

As far as I can see, for addons we only have ubuntu/linux builds set up.
I am OK with changing the build setup for addons repository so every build runs inside the docker container.

seanpmorgan:
I have no objections to using the custom-op docker images for travis and kokoro. ITbh that'd be ideal so we have a full understanding of our build env.

facaiy:
@av8ramit @gunan Thanks for the input, Amit, Gunhan. We agree to switch to custom-op-gpu, so what do we need to do?

facaiy:
Gently ping @av8ramit :-)

cc @karmel for visibility.

av8ramit:
I'm not sure what you mean by your question. Do you mean what to adjust to use the docker container?

I guess the next step would be to make `addons_gpu.sh` invoke and run tests in the corresponding docker container. That's what is being invoked on our side internally.

We can also modify how it's being invoked internally. Right now it's being called in a virtualenv.

facaiy:
Yeah, I want to know what to do if we'll use custom-op-gpu image for CI testing. 

> I guess the next step would be to make addons_gpu.sh invoke and run tests in the corresponding docker container.

Agree, let's run `addons_gpu.sh` in custom-op-gpu container. Can you set up it, please?

> Right now it's being called in a virtualenv.

Do you mean the script is invoked in a virtual environment inside a docker container? It sounds good, because it's convenient to support both python 2 and 3 when using custom-op image. cc @seanpmorgan any thoughts, Sean?

gunan:
I think what Amit wanted to say was, kokoro directly calls `addons_gpu.sh`
You are free to make any changes to it, and the CI will just pick that up.
You can edit that script right here:
https://github.com/tensorflow/addons/blob/master/tools/ci_testing/addons_gpu.sh

facaiy:
>  kokoro directly calls addons_gpu.sh

I agree, and I think it's a good idea. But how to run it in the corresponding docker image? I think we cannot do it without help from Amit. Or do you suggest to call `docker run` command in the addons_gpu.sh by ourselves? It looks unsafe to me.

gunan:
Yes, i do recommend running docker run from addons_gpu. As long as it is
properly reviewed, i think it should be fine.

facaiy:
I see, either way is fine. Thanks for your input, Gunhan.

@seanpmorgan @av8ramit @karmel Hi, Sean, Amit, Karmel, what do you think of them?

seanpmorgan:
> I see, either way is fine. Thanks for your input, Gunhan.
> 
> @seanpmorgan @av8ramit @karmel Hi, Sean, Amit, Karmel, what do you think of them?

I have no issues calling the docker image. Though I would prefer we also call custom-op in our cpu testing script as well for consistency.

seanpmorgan:
@facaiy tests are now passing in the GPU containers. We can start working on this again.

facaiy:
> @facaiy tests are now passing in the GPU containers. We can start working on this again.

Good news, Sean, I'll start working on the PR, and tell you when ready.

tensorflow_addons/custom_ops/image/BUILD
facaiy: restored

tensorflow_addons/utils/test_utils.py
facaiy: done :-)

configure.sh
facaiy: I leave a TODO comment here, and want to solve the problem in the following PRs. What do you think?

facaiy:
@seanpmorgan @yifeif Sean, Yifei, can you take a look? I prefer to solve  https://github.com/tensorflow/addons/issues/118 issue step by step, so I try to make the PR a minimum change (most of the code is from https://github.com/tensorflow/custom-op, thank Yifei) , and it just works.

cc @Squadrick who might be interested.

WORKSPACE
```diff
@@ -1,5 +1,7 @@
 load("//tf_dependency:tf_configure.bzl", "tf_configure")
+load("//external/gpu:cuda_configure.bzl", "cuda_configure")
```
facaiy: @seanpmorgan Sean, perhaps we could move tf_dependency to external directory as well, any idea?

tensorflow_addons/custom_ops/image/cc/kernels/euclidean_distance_transform_op.cc
```diff
@@ -77,17 +77,18 @@ TF_CALL_double(REGISTER);
 
 #undef REGISTER
 
-#if GOOGLE_CUDA
+// TODO: fix compile issue of the gpu kernel.
+#if 0 && GOOGLE_CUDA
```
facaiy: @mels630 Matt, seems that we can migrate gpu kernel after the PR is merged, right?

configure.sh
```diff
@@ -53,3 +53,21 @@ write_action_env_to_bazelrc "TF_HEADER_DIR" ${TF_CFLAGS:2}
...
+write_to_bazelrc "build -c opt"
+write_action_env_to_bazelrc "TF_NEED_CUDA" ${TF_NEED_CUDA}
+
+# TODO(yifeif): do not hardcode path
+if [[ "$TF_NEED_CUDA" == "1" ]]; then
+    # TODO: use CUDA_HOME here?
+    write_action_env_to_bazelrc "CUDNN_INSTALL_PATH" "/usr/lib/x86_64-linux-gnu"
+    write_action_env_to_bazelrc "TF_CUDA_VERSION" "10.0"
+    write_action_env_to_bazelrc "TF_CUDNN_VERSION" "7"
+    write_action_env_to_bazelrc "CUDA_TOOLKIT_PATH" "/usr/local/cuda"
```
seanpmorgan: Could we make this an ENV variable with a default
`${CUDA_HOME:=/usr/local/cuda}`

configure.sh
```diff
@@ -53,3 +53,21 @@ write_action_env_to_bazelrc "TF_HEADER_DIR" ${TF_CFLAGS:2}
 write_action_env_to_bazelrc "TF_SHARED_LIBRARY_DIR" ${SHARED_LIBRARY_DIR}
 write_action_env_to_bazelrc "TF_SHARED_LIBRARY_NAME" ${SHARED_LIBRARY_NAME}
 write_action_env_to_bazelrc "TF_CXX11_ABI_FLAG" ${TF_CXX11_ABI_FLAG}
+
+write_to_bazelrc "build:cuda --define=using_cuda=true --define=using_cuda_nvcc=true"
+write_to_bazelrc "build:cuda --crosstool_top=@local_config_cuda//crosstool:toolchain"
+write_to_bazelrc "build --spawn_strategy=standalone"
+write_to_bazelrc "build --strategy=Genrule=standalone"
+write_to_bazelrc "build -c opt"
```
seanpmorgan: Probably can remove this as we use it in our CLI. That or we can remove it from our documentation.

WORKSPACE
seanpmorgan: I agree they should both live in a sub-directory. How would you feel about `build_deps`?

seanpmorgan:
Oh could we also update `CONTRIBUTING.md` to explain when to use `custom-op-gpu` container

tensorflow_addons/custom_ops/image/cc/kernels/euclidean_distance_transform_op.cc
mels630: no objection to fixing in a later PR. As long as we can avoid losing track of the issue!

WORKSPACE
seanpmorgan: Would you prefer we move tf_dependency into `build_deps` in a separate PR?

seanpmorgan:
So tests are failing OOM because we launch parallel jobs on a single card. Once #344 is merged we'll see that everything is working though 2 new tests fail when running on GPU:

https://github.com/tensorflow/addons/blob/master/tensorflow_addons/seq2seq/beam_search_ops_test.py#L73

And all of the `testSparseRepeatedIndices` in:
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/optimizers/weight_decay_optimizers_test.py

I'm okay with commenting these tests out and cutting new issues to address them.

facaiy:
Thank you, Sean. I have disabled all those failures temporally, and filed the corresponding issues to track them.

tensorflow_addons/custom_ops/image/cc/kernels/euclidean_distance_transform_op.cc
facaiy: sure, file https://github.com/tensorflow/addons/issues/349 issue

WORKSPACE
facaiy: totally agree

PR #297, opened 2019-06-17
adding .DS_store to gitignore

SSaishruthi:
Is it ok if we add `.DS_store` to .gitignore file for convenience?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.DS_Store
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,9 @@
-.DS_Store
+# Ignore all hidden files
+.*
+
+# Include travis.yml
+\!.travis.yml
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]


facaiy:
How about using `.*` to ignore all hidden files?

Squadrick:
@facaiy Yeah, sure. That seems like a better idea.

seanpmorgan:
> How about using `.*` to ignore all hidden files?

I previously removed that ignore because `.travis.yml` are you okay with more specific ignores?

SSaishruthi:
Is that change ok or I need to change something?

facaiy:
> > How about using `.*` to ignore all hidden files?
> 
> I previously removed that ignore because `.travis.yml` are you okay with more specific ignores?

Sean, I prefer to ignore all hidden files by default, and then we can use `git add --force .yourfile` to add whatever file we need. What do you think?

seanpmorgan:
> > > How about using `.*` to ignore all hidden files?
> > 
> > 
> > I previously removed that ignore because `.travis.yml` are you okay with more specific ignores?
> 
> Sean, I prefer to ignore all hidden files by default, and then we can use `git add --force .yourfile` to add whatever file we need. What do you think?

That works for me. Currently we have only 1 tracked hidden file anyway. @SSaishruthi can you please edit this PR to ignore all hidden files and then we can merge

Gokkulnath:
@seanpmorgan  Why dont we add a explicit include .travis.yml using ! prefix ?? That way you need not remember to force add it.  Check here for more info https://git-scm.com/docs/gitignore 
we just need to add \!.travis.yml to gitignore

seanpmorgan:
> @seanpmorgan Why dont we add a explicit include .travis.yml using ! prefix ??
> we just need to add !.travis.yml to gitignore
> That way you need not remember to force add it. Check here for more info https://git-scm.com/docs/gitignore

That sounds great. Thanks @Gokkulnath !

SSaishruthi:
@seanpmorgan I have made changes based on my understanding. Is that ok?

PR #299, opened 2019-06-17
Make gradient tests compatible with eager mode

WindQAQ:
1. TODO in distort_image_ops_test.py: test gradients in TF2.0 favor.
2. TODO in transform_ops_test.py: use `tf.test.compute_gradient` in TF2.0 version. ~~Note that the tolerance in TF1.x (previous code) is 1e-10, but this will cause < 1% mismatch in TF2.0 gradient checker. It will pass if I set the tolerance to 1e-4 (also tested 1e-5, but failed...).~~
3. There's no need to override `setUp()` method in order to set random seed.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/test_util.py#L1729

reference code:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/optimizer_v2/gradient_descent_test.py#L136

### DIFF
--- a/tensorflow_addons/image/transform_ops_test.py
+++ b/tensorflow_addons/image/transform_ops_test.py
@@ -77,26 +77,12 @@ class ImageOpsTest(tf.test.TestCase):
             output_shape=tf.constant([3, 5]))
         self.assertAllEqual([3, 5], result.shape)
 
-    def _test_grad(self, shape_to_test):
-        test_image_shape = shape_to_test
-        test_image = tf.random.normal(test_image_shape, dtype=tf.float32)
+    def _test_grad(self, input_shape, output_shape=None):
+        test_image = tf.random.normal(input_shape, dtype=tf.float32)
 
-        def transform_fn(x):
-            x.set_shape(test_image_shape)
-            transform = transform_ops.angles_to_projective_transforms(
-                np.pi / 2, 4, 4)
-            return transform_ops.transform(x, transform)
-
-        theoretical, numerical = tf.test.compute_gradient(
-            transform_fn, [test_image])
-
-        self.assertAllClose(theoretical[0], numerical[0], rtol=1e-4, atol=1e-4)
-
-    def _test_grad_different_shape(self, input_shape, output_shape):
-        test_image_shape = input_shape
-        test_image = tf.random.normal(test_image_shape, dtype=tf.float32)
-
-        if len(output_shape) == 2:
+        if output_shape is None:
+            resize_shape = None
+        elif len(output_shape) == 2:
             resize_shape = output_shape
         elif len(output_shape) == 3:
             resize_shape = output_shape[0:2]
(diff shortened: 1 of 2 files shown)


tensorflow_addons/image/transform_ops_test.py
```diff
@@ -72,59 +72,38 @@ def test_transform_static_output_shape(self):
...
-            test_image_tensor = tf.constant(test_image, shape=test_image_shape)
-            test_transform = transform_ops.angles_to_projective_transforms(
+        def transform_fn(x):
+            x.set_shape(input_shape)
+            transform = transform_ops.angles_to_projective_transforms(
                 np.pi / 2, 4, 4)
+            return transform_ops.transform(
+                images=x, transforms=transform, output_shape=resize_shape)
+
+        theoretical, numerical = tf.test.compute_gradient(
```
facaiy: Tzu-Wei, take the private op `max_error` a try, and see if it works?

https://github.com/tensorflow/tensorflow/blob/0ff7955a0c1a42e2767afb0a5cc202dfe4d6ff19/tensorflow/python/kernel_tests/relu_op_test.py#L119

WindQAQ: The reported error from `max_error` is 7.25388e-5... Still exceeds original tolerance 1e-10.

facaiy:
@suharshs Hi, Suharsh, do you know the loss of precision when computing gradients in V2 version?  or can you redirect as necessary?  thanks!
cc @karmel

WindQAQ:
@facaiy, thanks for redirecting the problem. I think I know what happened. V1 and V2 gradient checker indeed lead to the same result. Seems that the normal distribution sampled from TensorFlow and Numpy is very different, and makes the the difference between theoretical and numerical gradient larger. Apologized for not checking this earlier.  

https://colab.research.google.com/drive/1C9bD6-PNa8lnGJ_YYQ-E2PYvxKInijH_

facaiy:
Oh, I see. Tzu-Wei, can we use some constant arrays as input for gradient test? In some cases, random data might cause numeric instability. And because it might be not reproducible easily, we'd better to avoid random data if possible :-)

WindQAQ:
No problem :P See if this is better.
I notice the monthly meeting note states that we have to remove private APIs, so I do not use `max_error` from `gradient_checker_v2` in this case.

tensorflow_addons/image/transform_ops_test.py
```diff
@@ -72,59 +72,43 @@ def test_transform_static_output_shape(self):
...
-                transforms=test_transform,
-                output_shape=resize_shape)
-            left_err = tf.compat.v1.test.compute_gradient_error(
-                test_image_tensor,
-                test_image_shape,
-                output,
-                output_shape,
-                x_init_value=test_image)
-            self.assertLess(left_err, 1e-10)
+        self.assertAllClose(theoretical[0], numerical[0], rtol=1e-6, atol=1e-6)
```
facaiy: Can we remove rtol and atol here?

facaiy:
> No problem :P See if this is better.
> I notice the monthly meeting note states that we have to remove private APIs, so I do not use `max_error` from `gradient_checker_v2` in this case.

Looks great!

PR #304, opened 2019-06-17
Speed up median_filter2d

WindQAQ:
Native `tf.image.extract_patches` could extract multi-channeled image patches. Speed up a lot when `batch_size` and `channels` are large.

https://colab.research.google.com/drive/1UkNXxjTMHifKmHB2N5ojOHFojtXrOUIG

### DIFF
--- a/tensorflow_addons/image/filters.py
+++ b/tensorflow_addons/image/filters.py
@@ -174,34 +174,40 @@ def median_filter2d(image,
         if rank == 3:
             image = tf.expand_dims(image, axis=0)
 
+        image_shape = tf.shape(image)
+        batch_size = image_shape[0]
+        height = image_shape[1]
+        width = image_shape[2]
+        channels = image_shape[3]
+
         # Explicitly pad the image
         image = _pad(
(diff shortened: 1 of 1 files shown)


tensorflow_addons/image/filters.py
```diff
@@ -174,34 +174,40 @@ def median_filter2d(image,
...
+        patches = tf.reshape(
+            patches, shape=[batch_size, height, width, area, channels])
+
+        patches = tf.transpose(patches, [0, 1, 2, 4, 3])
+
+        # Note the returned median is casted back to the original type
+        # Take [5, 6, 7, 8] for example, the median is (6 + 7) / 2 = 3.5
+        # It turns out to be int(6.5) = 6 if the original type is int
+        top = tf.nn.top_k(patches, k=ceil).values
+        output = tf.cast(
```
facaiy: Seems replication computation for odd `area` here?

WindQAQ: Facai, see if this is better :-)

PR #306, opened 2019-06-18
Fix #289

### LINKED ISSUE #289: Beam search decoder _check_static_batch_beam_maybe check correct?
**System information**
- TensorFlow installed from (source or binary): binary
- TensorFlow version (use command below): 2.0.0=alpha
- TensorFlow Addons installed from (source, PyPi): PyPi
- TensorFlow Addons version: 0.3.1
- Python version and type (eg. Anaconda Python, Stock Python as in Mac, or homebrew installed Python etc): Google Colab Python
- Is GPU used? (yes/no): yes
- GPU model (if used): T4

The check in beam_search_decoder.py#L183 seems faulty. As a MWE, consider:
```
tfa.seq2seq.beam_search_decoder._check_static_batch_beam_maybe(tf.TensorShape([3, 5, None]), 3, 5)
```
which yields:
```
TensorArray reordering expects elements to be reshapable to (3, 5, None) which is incompatible with the current shape (3, 5, None).
```

Either this error message is misleading or the check is faulty. Sadly, since I do not quite understand what it is supposed to be checking, I cannot propose a solution (other than to remove the last condition).

sjmielke:
Fix #289 

Unclutter the big conditional and change one negation.

### DIFF
--- a/tensorflow_addons/seq2seq/beam_search_decoder.py
+++ b/tensorflow_addons/seq2seq/beam_search_decoder.py
@@ -181,19 +181,21 @@ def _check_static_batch_beam_maybe(shape, batch_size, beam_width):
     reshaped to [batch_size, beam_size, -1]."""
     reshaped_shape = tf.TensorShape([batch_size, beam_width, None])
     assert len(shape.dims) > 0
-    known_size = batch_size is not None and shape.dims[0].value is not None
-    case_1 = (shape[0] == batch_size * beam_width)
-    case_2_guard = shape.ndims >= 2 and shape.dims[1].value is not None
-    case_2 = case_2_guard and shape[0] == batch_size and shape[1] == beam_width
-    if known_size and not (case_1 or case_2):
-        tf.get_logger().warn(
-            "TensorArray reordering expects elements to be "
-            "reshapable to %s which is incompatible with the "
-            "current shape %s. Consider setting "
-            "reorder_tensor_arrays to False to disable TensorArray "
-            "reordering during the beam search." % (reshaped_shape, shape))
-        return False
-    return True
+    if batch_size is None or shape.dims[0].value is None:
+        return True  # not statically known => no check
+    if shape[0] == batch_size * beam_width:
+        return True  # flattened, matching
+    has_second_dim = shape.ndims >= 2 and shape.dims[1].value is not None
+    if has_second_dim and shape[0] == batch_size and shape[1] == beam_width:
+        return True  # non-flattened, matching
+    # Otherwise we could not find a match and warn:
+    tf.get_logger().warn(
+        "TensorArray reordering expects elements to be "
+        "reshapable to %s which is incompatible with the "
+        "current shape %s. Consider setting "
+        "reorder_tensor_arrays to False to disable TensorArray "
+        "reordering during the beam search." % (reshaped_shape, shape))
+    return False
 
 
 def _check_batch_beam(t, batch_size, beam_width):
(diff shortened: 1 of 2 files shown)


tensorflow_addons/seq2seq/beam_search_decoder.py
```diff
@@ -180,10 +180,12 @@ def _check_static_batch_beam_maybe(shape, batch_size, beam_width):
     """Raises an exception if dimensions are known statically and can not be
     reshaped to [batch_size, beam_size, -1]."""
     reshaped_shape = tf.TensorShape([batch_size, beam_width, None])
-    if (batch_size is not None and shape.dims[0].value is not None
-            and (shape[0] != batch_size * beam_width or
-                 (shape.ndims >= 2 and shape.dims[1].value is not None and
-                  (shape[0] != batch_size or shape[1] != beam_width)))):
+    assert len(shape.dims) > 0
+    known_size = batch_size is not None and shape.dims[0].value is not None
+    case_1 = (shape[0] == batch_size * beam_width)
```
facaiy: Hi, can we give it a meaningful name if possible?

facaiy:
@qlzh727 Qianli, can you take a look?

facaiy:
@sjmielke Sebastian, can you add a test case to catch the known bug?

tensorflow_addons/seq2seq/beam_search_decoder.py
```diff
@@ -180,10 +180,12 @@ def _check_static_batch_beam_maybe(shape, batch_size, beam_width):
...
+    case_2_guard = shape.ndims >= 2 and shape.dims[1].value is not None
```
qlzh727: At this point, the code is quite hard to understand now. I think the more readable way is to exit early, like:

```
if not batch_size or not beam_width or not shape[0]:
  return  # Shape not static, do not check.
if shape[0] == batch_size * beam_width:
  return  
if shape.ndims >= 2 and shape[0] == batch_size and (
        shape[1] == beam_width or shape[1] is None):
     return
tf.get_logger().warn(....)
```

tensorflow_addons/seq2seq/beam_search_decoder.py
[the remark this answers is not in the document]
sjmielke: Yeah, I was worried about that but since the code didn't catch that before and I wanted to keep the change minimal... I can add an explicit assertion (i.e. keep the failure) or just return False with the warning message here (and hope the rest of the code catches that / makes this never happen) -- I'd prefer the former, if that makes sense?

tensorflow_addons/seq2seq/beam_search_decoder.py
sjmielke: That is much prettier, yes. I guess in Python having many ifs vs. one big conditional doesn't matter anyway 🙃 I'll update this PR today or tomorrow. Thanks for the feedback!

sjmielke:
Added more explicit branching like suggested by @qlzh727 and tests for the method.

sjmielke:
Does this need anything else? :) (Oh, also, in case anyone knows: does tfa have a schedule to backport fixes like these into TF1.contrib or should I best create similar PRs there for that?)

seanpmorgan:
> Does this need anything else? :) (Oh, also, in case anyone knows: does tfa have a schedule to backport fixes like these into TF1.contrib or should I best create similar PRs there for that?)

We have no process for backporting fixes at the moment.. I'm not sure if changes to contrib will be accepted in TF core right now but seeing as seq2seq is popularly used they might merge it in prior to a TF2 release.

PR #309, opened 2019-06-20
Migrate npairs_loss

WindQAQ:
Addresses #26. Will proceed with multilabel npairs loss after this.

### DIFF
--- a/tensorflow_addons/losses/npairs.py
+++ b/tensorflow_addons/losses/npairs.py
@@ -31,7 +31,7 @@ def npairs_loss(y_true, y_pred):
     The loss takes each row of the pair-wise similarity matrix, `y_pred`,
     as logits and the remapped multi-class labels, `y_true`, as labels.
 
-    The similarity matrix `y_pred` between two embedding matrics `a` and `b`
+    The similarity matrix `y_pred` between two embedding matrices `a` and `b`
     with shape `[batch_size, hidden_size]` can be computed as follows:
 
     ```python
@@ -72,7 +72,7 @@ class NpairsLoss(tf.keras.losses.Loss):
     The loss takes each row of the pair-wise similarity matrix, `y_pred`,
     as logits and the remapped multi-class labels, `y_true`, as labels.
 
-    The similarity matrix `y_pred` between two embedding matrics `a` and `b`
+    The similarity matrix `y_pred` between two embedding matrices `a` and `b`
     with shape `[batch_size, hidden_size]` can be computed as follows:
 
     ```python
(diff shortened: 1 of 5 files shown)


tensorflow_addons/losses/npairs.py
```diff
@@ -0,0 +1,94 @@
...
+@tf.function
+def npairs_loss(y_true, y_pred):
+    """Computes the npairs loss between `y_true` and `y_pred`.
+
+    Npairs loss expects paired data where a pair is composed of samples from
+    the same labels and each pairs in the minibatch have different labels.
+    The loss takes each row of the pair-wise similarity matrix, `y_pred`,
+    as logits and the remapped multi-class labels, `y_true`, as labels.
+
+    The similarity matrix `y_pred` between two embedding matrics `a` and `b`
```
Squadrick: Typo: matrics -> matrices

tensorflow_addons/losses/npairs.py
```diff
@@ -0,0 +1,94 @@
...
+      y_pred: 2-D float `Tensor` with shape `[batch_size, batch_size]` of
+        similarity matrix between embedding matrices.
+
+    Returns:
+      npairs_loss: float scalar.
+    """
+    y_pred = tf.convert_to_tensor(y_pred)
+    y_true = tf.cast(y_true, y_pred.dtype)
+
+    y_true = tf.reshape(y_true, (tf.shape(y_true)[0], 1))
```
Squadrick: You could use `y_true = tf.expand_dims(y_true, 1)`. It's more succinct.

tensorflow_addons/losses/npairs.py
```diff
@@ -0,0 +1,94 @@
...
+@keras_utils.register_keras_custom_object
+class NpairsLoss(tf.keras.losses.Loss):
+    """Computes the npairs loss between `y_true` and `y_pred`.
+
+    Npairs loss expects paired data where a pair is composed of samples from
+    the same labels and each pairs in the minibatch have different labels.
+    The loss takes each row of the pair-wise similarity matrix, `y_pred`,
+    as logits and the remapped multi-class labels, `y_true`, as labels.
+
+    The similarity matrix `y_pred` between two embedding matrics `a` and `b`
```
Squadrick: Typo: matrics -> matrices.

tensorflow_addons/losses/npairs.py
WindQAQ: It will but while compiling keras model, `y_true` will have shape `[None, None]`. I'm afraid that `tf.expand_dims(y_true, 1)` will generate `[None, 1, None]` shape, and further break following computation?

Squadrick: Will `y_true = tf.expand_dims(y_true, -1)` work instead?

Squadrick: It'll reshape `[None, None]` -> [None, None, 1]`. If it doesn't make sense, ignore my comment.

WindQAQ: After a rough test, I could confirm that either way could successfully compile keras model. So as you suggested, I use `tf.expand_dims(y_true, -1)` for conciseness :P

PR #310, opened 2019-06-20
Emphasize PEP8 (non-TF spacing) compliance in README.md

mels630:


mels630:
split out from https://github.com/tensorflow/addons/pull/308

STYLE_GUIDE.md
```diff
@@ -43,3 +43,6 @@ foo._protected_member
 #### TensorFlow Conventions
 
 Follow the guidance in the [TensorFlow Style Guide - Conventions](https://www.tensorflow.org/community/contribute/code_style#tensorflow_conventions_and_special_uses).
+
+Please note that Addons follows these conventions but _not_ the entirety of the [TensorFlow Style Guide](https://www.tensorflow.org/community/contribute/code_style).
+In particular, as stated above, Python code should conform to [PEP8](https://www.python.org/dev/peps/pep-0008/) _without_ the Tensorflow spacing exception.
```
seanpmorgan: Thanks for  adding this. Could you modify this slightly to say that we follow "the conventions of the TensorFlow library, but choose to format our code using PEP8 guidelines" or something along those lines?

My first read of this was it was a bit back and forth on which guide to use. Though we format with PEP8 and use the conventions for building upon the library.

mels630: sure, makes sense. see if the updated commit (cea8962) reads more cleanly

PR #314, opened 2019-06-21
Add conditional random fields

### LINKED ISSUE #22: Implement Conditional Random Field
Per the [RFC](https://github.com/tensorflow/community/blob/393dfc9e03dd4150a839de6357a26e0259aa97d6/rfcs/20181214-move-to-addons.md), we need to move [CRF](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/crf) from contrib to addons:

This will involve moving the code and modifying test cases to run in TF2.

Squadrick:
~WIP till I can resolve some of the failing test cases.~
Closes #22 

Ref: [`tf.contrib.crf`](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/crf)


Squadrick:
The test cases were failing due to not masking the inputs to the RNN.

Squadrick:
Ping @seanpmorgan @facaiy

tensorflow_addons/text/crf_ops.py
```diff
@@ -0,0 +1,469 @@
...
+
+    masks = tf.sequence_mask(
+        sequence_lengths, maxlen=tf.shape(tag_indices)[1], dtype=tf.float32)
+    truncated_masks = tf.slice(masks, [0, 1], [-1, -1])
+    binary_scores = tf.reduce_sum(binary_scores * truncated_masks, 1)
+    return binary_scores
+
+
+class CrfForwardRnnCell(tf.keras.layers.Layer):
+    def __init__(self, transition_params, **kwargs):
```
seanpmorgan: Seems the doc strings were dropped for CrfForwardRNNCell?

tensorflow_addons/text/crf_ops.py
```diff
@@ -0,0 +1,469 @@
...
+                              flattened_transition_indices)
+
+    masks = tf.sequence_mask(
+        sequence_lengths, maxlen=tf.shape(tag_indices)[1], dtype=tf.float32)
+    truncated_masks = tf.slice(masks, [0, 1], [-1, -1])
+    binary_scores = tf.reduce_sum(binary_scores * truncated_masks, 1)
+    return binary_scores
+
+
+class CrfForwardRnnCell(tf.keras.layers.Layer):
```
seanpmorgan: Could you explain why we subclass keras Layer for these cells instead of keras AbstractRNNCell?

tensorflow_addons/text/BUILD
```diff
@@ -6,6 +6,7 @@ py_library(
     name = "text",
     srcs = ([
         "__init__.py",
+        "crf_ops.py",
```
seanpmorgan: I think we can just name this as `crf` and `crf_tests`. For reference `skip_gram_ops`, `distort_image_ops`, etc. all utilize custom_ops

tensorflow_addons/text/crf_ops.py
```diff
@@ -0,0 +1,469 @@
...
+        super(CrfDecodeBackwardRnnCell, self).__init__(**kwargs)
+        self._num_tags = num_tags
+
+        self.state_size = 1
+        self.output_size = 1
+
+    def build(self, input_shape):
+        super(CrfDecodeBackwardRnnCell, self).build(input_shape)
+
+    def call(self, inputs, state, training=None):
```
seanpmorgan: Unused parameter `training`

facaiy:
I don't know it. @qlzh727 Qianli, are you a good person to look at this?

cc @WindQAQ  @howl-anderson @NLP-ZY who might be interested.

tensorflow_addons/text/crf_ops.py
Squadrick: I was looking at how `SimpleRNNCell` was implemented. But yes, using `AbstractRNNCell` makes more sense.

Squadrick:
@seanpmorgan I'm making a colab notebook with usage right now, I'll open a separate PR once this merges.

howl-anderson:
Since TF 2.0 encourages users using Keras API, implement CRF functions in the form of Keras layer will be a great feature. keras-contrib already has a CRF layer (https://github.com/keras-team/keras-contrib/blob/master/keras_contrib/layers/crf.py) which is pretty widely used in Keras community. We'd better consistent with it.

howl-anderson:
I am working on implementing the CRF  Keras layer based on Squadrick's solution. When it's ready I will post a new PR.

Squadrick:
@howl-anderson Will you using any of the functions implemented here? Similar to how `tf.keras.layers` internally wraps `tf.nn` or will you doing the computation from scratch. 

If it's the latter, I'll close this PR and we can proceed with our approach. If it's the latter, I'll keep this open. 

P.S.: Implement your CRF layer under `tfa.layers`.

Squadrick:
IMO, it's best to include both. The raw functions under `tfa.text` for those who want low level control, and an end-to-end layer under `tfa.layers` for easy integration with the Keras API. What do you think? Standardising the naming scheme used in both will be an important aspect. 

What do y'all think? Any thoughts on this? @seanpmorgan @WindQAQ @facaiy

howl-anderson:
@Squadrick CRF layer will be based on your functions (internally wraps your functions). This is a friendly feature for Keras users. low-level functions (which compatible with tf.contrib) for CRF are still very useful for advanced users and it's very important to users who migrate they code from 1.x to 2.0 and don't want to change their code on a large scale.

tensorflow_addons/text/crf.py
```diff
@@ -0,0 +1,509 @@
...
+        # Sequence length is not allowed to be less than zero.
+        sequence_lengths_less_one = tf.maximum(
+            tf.constant(0, dtype=sequence_lengths.dtype), sequence_lengths - 1)
+
+        forward_layer = tf.keras.layers.RNN(
+            forward_cell, return_sequences=True, return_state=True)
+
+        mask = tf.sequence_mask(sequence_lengths_less_one,
+                                tf.shape(inputs)[1] - 1)
+        _, alphas = forward_layer(rest_of_input, first_input, mask=mask)
```
qlzh727: What's the real input an initial state here? probably better to have the initial_state kwarg to make it clear.

tensorflow_addons/text/crf.py
```diff
@@ -0,0 +1,509 @@
...
+    return binary_scores
+
+
+class CrfForwardRnnCell(tf.keras.layers.AbstractRNNCell):
+    """Computes the alpha values in a linear-chain CRF.
+
+    See http://www.cs.columbia.edu/~mcollins/fb.pdf for reference.
+    """
+
+    def __init__(self, transition_params, **kwargs):
```
qlzh727: Also note that for tensor value passed in from constructor, it won't be easily get serialized into json in get_config() and rebuild at from_config(). Currently the cell can't be saved/reloaded by keras framework. It is fine for now If you don't use keras model, but need to keep in mind.

tensorflow_addons/text/crf.py
```diff
@@ -0,0 +1,509 @@
...
+    """Computes backward decoding in a linear-chain CRF."""
+
+    def __init__(self, num_tags, **kwargs):
+        """Initialize the CrfDecodeBackwardRnnCell.
+
+        Args:
+          num_tags: An integer. The number of tags.
+        """
+        super(CrfDecodeBackwardRnnCell, self).__init__(**kwargs)
+        self._num_tags = num_tags
```
qlzh727: the num_tags is never used within this class.

tensorflow_addons/text/crf.py
```diff
@@ -0,0 +1,509 @@
...
+
+    @property
+    def state_size(self):
+        return 1
```
qlzh727: 1? shouldn't it be _num_tags?

tensorflow_addons/text/crf.py
```diff
@@ -0,0 +1,509 @@
...
+        backpointers = tf.reverse_sequence(
+            backpointers, sequence_length_less_one, seq_axis=1)
+
+        crf_bwd_cell = CrfDecodeBackwardRnnCell(num_tags)
+        initial_state = tf.cast(tf.argmax(last_score, axis=1), dtype=tf.int32)
+        initial_state = tf.expand_dims(initial_state, axis=-1)
+        crf_bwd_layer = tf.keras.layers.RNN(
+            crf_bwd_cell,
+            return_sequences=True,
+            return_state=True,
```
qlzh727: If the returned state is ignored, then just set return_state to False.

tensorflow_addons/text/crf.py
```diff
@@ -0,0 +1,509 @@
...
+        sequence_length_less_one = tf.maximum(
+            tf.constant(0, dtype=sequence_length.dtype), sequence_length - 1)
+
+        mask = tf.sequence_mask(sequence_length_less_one, tf.shape(inputs)[1])
+        crf_fwd_cell = CrfDecodeForwardRnnCell(transition_params)
+        crf_fwd_layer = tf.keras.layers.RNN(
+            crf_fwd_cell,
+            return_sequences=True,
+            return_state=True,
+            time_major=False)
```
qlzh727: The default value for time_major is False, you can skip it here and below.

howl-anderson:
> Haven't looked deep into the logic yet. Probably need to have some one compare the original implementation again this PR.

@qlzh727 I have some experiences with the original implementation of CRF in the tf.contrib and familiar with how CRF works, I can check/compare this PR with the original one.

tensorflow_addons/text/crf.py
Squadrick: State for `CrfDecodeBackwardRnnCell` is the tag index of the next step. It'll be `[batch_size, 1]`.

Squadrick:
@qlzh727 @howl-anderson Modified the CRF code to use `tf.scan` for decode backward and forward alpha calculation. `decode_forward` still uses an RNN cell, but I can rewrite it to use a `tf.while_loop`, but it ends up being far more complicated then just using an RNN cell. 

Also, `transition_params` will be not tracked by Keras RNN.

howl-anderson:
@Squadrick Thanks for your update

Squadrick:
@qlzh727 @howl-anderson Any feedback on this?

howl-anderson:
@howl-anderson Don't worry,  I am still working on it.

gjxcq:
you said the crf is in the tfa.text, but I not find it, are you sure that merge is successful??

tensorflow_addons/text/crf.py
```diff
@@ -0,0 +1,448 @@
...
+
+    masks = tf.sequence_mask(
+        sequence_lengths, maxlen=tf.shape(tag_indices)[1], dtype=tf.float32)
+    truncated_masks = tf.slice(masks, [0, 1], [-1, -1])
+    binary_scores = tf.reduce_sum(binary_scores * truncated_masks, 1)
+    return binary_scores
+
+
+@tf.function
+def crf_forward(inputs, state, transition_params, sequence_lengths):
```
howl-anderson: Adding some comments (just like the tf.contrib) about the meaning and shape info of arguments of this function will be a great improvement for other code readers or maintainers.

tensorflow_addons/text/crf.py
```diff
@@ -0,0 +1,448 @@
...
+        state = tf.expand_dims(state[0], 2)
+        transition_scores = state + self._transition_params
+        new_state = inputs + tf.reduce_max(transition_scores, [1])
+        backpointers = tf.argmax(transition_scores, 1)
+        backpointers = tf.cast(backpointers, dtype=tf.int32)
+        return backpointers, new_state
+
+
+@tf.function
+def crf_decode_forward(inputs, state, transition_params, sequence_lengths):
```
howl-anderson: Adding some comments (like the tf.contrib) about the meaning and shape info of arguments of this function will more awesome!

tensorflow_addons/text/crf.py
```diff
@@ -0,0 +1,448 @@
...
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import numpy as np
+import tensorflow as tf
+
+
+@tf.function
```
guillaumekln: Do we want to annotate these functions with `tf.function`?

`max_seq_len` usually varies a lot between batches which would cause many retraces. An input signature should at least be set, but even then there is a current TF issue which causes graphs to be rebuilt internally when the input shape changes, see https://github.com/tensorflow/tensorflow/issues/29075.

Squadrick: I've removed the `tf.function` wrappers, and added a TODO.

Squadrick:
@howl-anderson @guillaumekln @seanpmorgan Made the changes, this should be good to merge.

howl-anderson:
@Squadrick Nice job!

seanpmorgan:
Looks great @Squadrick Thanks! Going to leave this open just for a bit to test #387

seanpmorgan:
Assuming this is some edge case because the sanity checks are working on other PRs. Merging. Thanks @Squadrick !

Squadrick:
Haha, not sure what caused the sanity check to fail. Finally got this long running PR closed.

Thanks for the reviews, guys.

lingvisa:
@howl-anderson did you have this implementation in the tf-addons in TF 2.0? like this one: https://github.com/keras-team/keras-contrib/blob/master/keras_contrib/layers/crf.py

seanpmorgan:
Hi @lingvisa , we do have support for CRF, but as it stands today they are only TF functions:
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/text/crf.py

We have an open PR for wrapping these around a layer, please feel free to review and let us know if it meets your expectations. At a first glance I think the keras-contrib version is much simplier and probably benefits from not have CRF functions carried over from tf.contrib. I'll make a note in that PR to consider dropping CRF functions and adopting the keras-contrib version more or less as is

PR #316, opened 2019-06-25
Add MeanMetricWrapper to metrics_utils

Squadrick:
As per [this comment](https://github.com/tensorflow/tensorflow/issues/28601#issuecomment-505098700). Reference: [`tf.keras.metrics.MeanMetricWrapper`](https://github.com/tensorflow/tensorflow/blob/71241a6afd27a9f99039a7b373cb46553e8f701f/tensorflow/python/keras/metrics.py#L538).

I couldn't use `ragged_assert_compatible_and_get_flat_values` and `squeeze_or_expand_dimensions`, since they are not part of the public API either. But I don't think it should be an issue if we skip those checks. We just need to ensure the `fn` can handle the input `y_true` and `y_false`.

### DIFF
--- a/tensorflow_addons/metrics/BUILD
+++ b/tensorflow_addons/metrics/BUILD
@@ -9,6 +9,7 @@ py_library(
         "cohens_kappa.py",
         "f1_scores.py",
         "r_square.py",
+        "utils.py",
     ],
     srcs_version = "PY2AND3",
     deps = [
--- a/tensorflow_addons/metrics/utils.py
+++ b/tensorflow_addons/metrics/utils.py
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
-"Utilities for metrics"
+"""Utilities for metrics."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
 
 import tensorflow as tf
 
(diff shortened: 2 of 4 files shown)


facaiy:
Yeah, Dheeraj, I agree to skip those checks at first, can we leave an TODO there? 
And I have two more questions:
1. How about moving metrics_utils to metric module? And perhaps `wrapper.py` or `base.py` are good candidate name as well :-)
2. Why to leave `get_config` method out?

Squadrick:
@facaiy I'll add the TODO. 
I thought it would be a better fit in `utils`, but I'm fine with calling it `base.py` and moving it to `metrics`. 
Also, the `get_config` of `Mean` is overridden in this case, `MeanMetricWrapper` doesn't implement `get_config`. 

The way to implement a few metric would be to extend `MeanMetricWrapper` and define a `get_config` and calling `super` with `fn`. I could add a simple useless wrapper like this:
```
def get_config(self):
     return super(MeanMetricWrapper, self).get_config()
```

facaiy:
> I thought it would be a better fit in utils

Do we expect that other modules will use `MeanMetricWrapper` as well?

> MeanMetricWrapper doesn't implement get_config.

If I'm not wrong, `MeanMetricWrapper` implements the get_config method by itself:  https://github.com/tensorflow/tensorflow/blob/71241a6afd27a9f99039a7b373cb46553e8f701f/tensorflow/python/keras/metrics.py#L582

Squadrick:
@facaiy You're right, `get_config` is implemented. I missed that in a hurry last time around. My bad. 

I'll also move `metrics_utils.py` to `metrics/utils.py`

Squadrick:
Test failure unrelated. Fixed here: #318

PR #325, opened 2019-07-01
Run CI testing inside docker containers

seanpmorgan:
Follow up to the discussion in #294 

@facaiy Let me know your thoughts on this setup. After we're in agreement we can request Amit to approve and modify anything on their side.


seanpmorgan:
Just a note.. this would require:
* Nvidia-docker to be installed on their GCP image
* Requires a pass argument to run with python3

PR #330, opened 2019-07-04
Convert unknown rank image to 4D image

### LINKED ISSUE #243: Make image transforms compatible with TF Dataset

facaiy:
Fix #243

1. The name `*_4D_image` looks kind of confusing, any good idea?
2. I didn't check all image ops and see if they can work well with unknown rank image, perhaps we should file a new issue with check list.

### DIFF
--- a/tensorflow_addons/image/utils.py
+++ b/tensorflow_addons/image/utils.py
@@ -40,7 +40,7 @@ def to_4D_image(image):
         tf.debugging.assert_rank_in(
             image, [2, 3, 4], message='`image` must be 2/3/4D tensor')
     ]):
-        # ydpf: enable
+        # yapf: enable
         ndims = image.get_shape().ndims
         if ndims is None:
             return _dynamic_to_4D_image(image)
--- a/tensorflow_addons/image/utils.py
+++ b/tensorflow_addons/image/utils.py
@@ -1,0 +1,96 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Image util ops."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow as tf
+
+
+def get_ndims(image):
+    return image.get_shape().ndims or tf.rank(image)
+
+
+@tf.function
+def to_4D_image(image):
+    """Convert 2/3/4D image to 4D image.
+
+    Args:
+      image: 2/3/4D tensor.
+
+    Returns:
+      4D tensor with the same type.
+    """
+    tf.debugging.assert_rank_in(image, [2, 3, 4])
+    ndims = image.get_shape().ndims
+    if ndims is None:
+        return _dynamic_to_4D_image(image)
+    elif ndims == 2:
+        return image[None, :, :, None]
+    elif ndims == 3:
+        return image[None, :, :, :]
+    else:
+        return image
+
+
+def _dynamic_to_4D_image(image):
+    shape = tf.shape(image)
+    original_rank = tf.rank(image)
+    # 4D image => [N, H, W, C]
+    # 3D image => [1, H, W, C]
+    # 2D image => [1, H, W, 1]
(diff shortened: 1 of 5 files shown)


tensorflow_addons/image/utils.py
```diff
@@ -0,0 +1,98 @@
...
+    elif ndims == 2:
+        return image[None, :, :, None]
+    elif ndims == 3:
+        return image[None, :, :, :]
+    else:
+        return image
+
+
+def _dynamic_to_4D_image(image):
+    shape = tf.shape(image)
```
kyleabeauchamp: nit: might improve clarity to call this original_shape.

tensorflow_addons/image/utils.py
```diff
@@ -0,0 +1,98 @@
...
+    # yapf: enable
+    return tf.reshape(image, new_shape)
+
+
+@tf.function
+def from_4D_image(image, ndims):
+    """Convert back to an image with `ndims` rank.
+
+    Args:
+      image: 4D tensor.
```
kyleabeauchamp: From looking at the case when `ndims==3`, I think we're assuming that channels are placed last in the image dimensions.  If that's the case, maybe we should add `(N, H, W, C)` to the docstring for the image arg, so that it's made clear?

tensorflow_addons/image/utils.py
```diff
@@ -0,0 +1,98 @@
...
+def to_4D_image(image):
+    """Convert 2/3/4D image to 4D image.
+
+    Args:
+      image: 2/3/4D tensor.
+
+    Returns:
+      4D tensor with the same type.
+    """
+    tf.debugging.assert_rank_in(image, [2, 3, 4])
```
WindQAQ: I wonder if we should use `tf.control_dependencies` here as stated in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/check_ops.py#L1445-L1449
>   Returns:
    Op raising `InvalidArgumentError` unless rank of `x` is in `ranks`.
    If static checks determine `x` has matching rank, a `no_op` is returned.
    This can be used with `tf.control_dependencies` inside of `tf.function`s
    to block followup computation until the check has executed.

tensorflow_addons/image/utils.py
```diff
@@ -0,0 +1,98 @@
...
+    """Convert back to an image with `ndims` rank.
+
+    Args:
+      image: 4D tensor.
+      ndims: The original rank of the image.
+
+    Returns:
+      `ndims`-D tensor with the same type.
+    """
+    tf.debugging.assert_rank(image, 4)
```
WindQAQ: ditto

facaiy:
@kyleabeauchamp Good question, Kyle, I didn't notice that. I think `*_4D_image` methods work for both channel-last and channel-first format, so I revise the corresponding code comments. And for `rotate`, `transform` methods, they are designed only for channel-last format which has been written in the corresponding documents.

@WindQAQ It's wired that I find all tests pass without tf.control_dependencies. For safety, I have add the op as suggested :-)

All comments has been solved, can you take a look again, Kyle and Tzu-Wei?

tensorflow_addons/image/utils.py
facaiy: I see, how about renaming original_shape to new_shape?

tensorflow_addons/image/utils.py
facaiy: I think the code works for both HWC and CHW format, Kyle. I have rewritten the code comments below. What do you think?

WindQAQ:
@facaiy thank you! @kyleabeauchamp, Kyle, feel free to merge this after taking a look.

facaiy:
Forget to mention: `*_4D_image` and `get_ndims` are only visible in the image module by now, namely, they are not public endpoint of addons.

tensorflow_addons/image/utils.py
```diff
@@ -35,10 +35,12 @@ def to_4D_image(image):
...
       4D tensor with the same type.
     """
+    # yapf:disable
     with tf.control_dependencies([
-            tf.debugging.assert_rank_in(
-                image, [2, 3, 4], message='`image` must be 2/3/4D tensor')
+        tf.debugging.assert_rank_in(
+            image, [2, 3, 4], message='`image` must be 2/3/4D tensor')
     ]):
+        # ydpf: enable
```
kyleabeauchamp: Should this be `yapf`?

facaiy: Fixed, Kyle :-)

Issue #83, opened 2019-03-10
Remove numpy implementation in test case?


facaiy:
I find, in most of test cases of Addons / tf.contrib, we are used to write a duplicated numpy function to verify the correction of our implementations. It's straightforward, but the drawbacks are:
1. both tf and numpy implications could be incorrect;
2. numpy might have a bug, or will introduce a bug in the future;
3. difficult to maintain test cases, and might also too complex to understand it.

Hence I propose to remove all numpy implementations, and replace them by explicit numerical value (and  add necessary comments to describe how to figure it out, say intermediate result of calculation) in test cases.

```python
# No
def np_add(a, b):
    return a + b

self.assertEqual(np_add(1, 2), tf_add(1, 2)) 

# Yes
# Add some comments if necessary:
# c = a + b = 1 + 2 = 3
self.assertEqual(3, tf_add(1, 2))
```

cc @seanpmorgan @karmel

facaiy:
By the way, do we need to post those `discussion needed` issues to mailing list? Perhaps we could gain more attention, and collect more feedback there.

seanpmorgan:
Some initial thoughts...

* If possible I would prefer our unit tests not include another library
* [TF Keras testing](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/layers/normalization_test.py) does extensively use it, doesn't mean we have to but worth considering
* I do think the likelihood of a bug in such fundamental pieces of numpy is probably low
* Without having thoroughly checked all usages of np; I would be in favor of removing usage when there are TF equivalents.

I also think that emailing `dicussion needed` tags is a great idea. It could probably be made a bot if someone was willing.

Smokrow:
I think the chances of hitting such a bug are quite low. (Of course as long as you stay with the „standard“ function and don’t do something crazy) 

Implementing something wrong twice might happen. But it’s still better then implementing it once, since you have to express the algo in two different apis. 

Also the amount of Tests you can create without numpy might be quite limited. 

Of course the dependencie to numpy it self is bad, but in the usefulness of numpy is quite big. 


Greets :)

facaiy:
If an input X is given, then its output Y has been determinate. We should explicitly claim what we want, and what the result should be. 

[...]

PR #333, opened 2019-07-06
TEST: Add test cases for pairwise distance

seanpmorgan:


seanpmorgan:
> @seanpmorgan LGTM, Sean! Just wonder if we really need to have a numpy replicated implementation in our test. I do not oppose it but it also appears in `triplet_loss.py`.

Hmmm good point I didn't notice that.  Would you be opposed to importing the numpy implementation in triplet test?  I'll modify the PR if that works.

WindQAQ:
> Hmmm good point I didn't notice that. Would you be opposed to importing the numpy implementation in triplet test? I'll modify the PR if that works.

That would be a better workaround. It turns out to be the discussion of https://github.com/tensorflow/addons/issues/83#issuecomment-472778678.

seanpmorgan:
> > Hmmm good point I didn't notice that. Would you be opposed to importing the numpy implementation in triplet test? I'll modify the PR if that works.
> 
> That would be a better workaround. It turns out to be the discussion of [#83 (comment)](https://github.com/tensorflow/addons/issues/83#issuecomment-472778678).

Agree I've altered the PR. Thanks for the review.

Issue #332, opened 2019-07-05
Check which image ops are compatible with tf.data.Dataset


facaiy:
Follow up https://github.com/tensorflow/addons/issues/243.

Here is a check list:
+ [x] dense_image_warp
+ [ ] sparse_image_warp		
+ [x] interpolate_bilinear	
+ [x] euclidean_distance_transform	
+ [x] adjust_hsv_in_yiq	
+ [x] random_hsv_in_yiq	
+ [x] mean_filter2d	
+ [x] median_filter2d	
+ [x] angles_to_projective_transforms	
+ [x] matrices_to_flat_transforms	
+ [x] rotate	
+ [x] transform

Like what we did in https://github.com/tensorflow/addons/pull/330, feel free to open a pull request for each op:
1. Add test for unknown rank image tensor.
2. Fix it if possible. Otherwise, please notice users that the op is not compatible with unknown rank tensor (shape=None) in its document.

satishjasthi:
Hello,
Can anyone help me how to test whether a tfa op is compatible with tf.Dataset.
I would like to check for 
- dense_image_warp
- interpolate_bilinear
My question might sound silly, I'm a newbie please help me.

seanpmorgan:
Hi @satishjasthi welcome to addons! So this colab notebook shows what the issue was (has since been fixed for that op):
https://colab.research.google.com/drive/1ZDhnGrorvSf04wzS-0utztZ6Nnzusvl7

Basically, if an image op can't be used in a tf.dataset mapping it's likely because we're not handling an unknown shape correctly. tf.dataset runs as a subgraph so the dynamic shape is not known until the graph is ran. 

A good test to see if it's compatible can be see in previous pull requests that adressed this:
https://github.com/tensorflow/addons/pull/343/files#diff-7137321db62bba0af8193f6f6213fd46R243

Yan added an image util which can be used to infer shape:
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/utils.py#L29

Let us know if you run into any issues, and thanks for the help!

mels630:
regarding the last two ops:

* angles_to_projective_transforms maps a scalar (or batch of N scalars) to an output of dimension (N, 8) representing the equivalent projective transform. It additionally requires image_width and image_height (as scalars or tensors of same dimension as the input angles) to compute the transform.
[...]

PR #328, opened 2019-07-02
add run_docker.sh script

facaiy:
Based on #325 . Related with #294.

@seanpmorgan @av8ramit Hi, does it look complicated?

```bash
bash tools/run_docker.sh -c 'make code-format' 
# Github checkers
bash tools/run_docker.sh -c 'make sanity-check'  # Ubuntu Sanity Check
bash tools/run_docker.sh -s linux -d cpu -p py2 -c 'make unit-test'  # Ubuntu Python2
bash tools/run_docker.sh -s linux -d cpu -p py3 -c 'make unit-test'  # Ubuntu Python3
bash tools/run_docker.sh -s linux -d gpu -p py2 -c 'make gpu-unit-test'  # Ubuntu GPU Python2
bash tools/run_docker.sh -s linux -d gpu -p py3 -c 'make gpu-unit-test'  # Ubuntu GPU Python3
```

PR #338, opened 2019-07-08
Add unknown shape support to distance_transform_ops

Squadrick:
Update `euclidean_dist_transform` as per #332

### DIFF
--- a/tensorflow_addons/image/distance_transform_test.py
+++ b/tensorflow_addons/image/distance_transform_test.py
@@ -21,7 +21,7 @@ from __future__ import print_function
 import numpy as np
 import tensorflow as tf
 
-from tensorflow_addons.image import distance_transform as distance_tranform_ops
+from tensorflow_addons.image import distance_transform as distance_transform_ops
 from tensorflow_addons.utils import test_utils
 
 
@@ -45,7 +45,7 @@ class DistanceOpsTest(tf.test.TestCase):
         image = tf.constant(image, dtype=tf.uint8)
 
         for output_dtype in [tf.float16, tf.float32, tf.float64]:
-            output = distance_tranform_ops.euclidean_dist_transform(
+            output = distance_transform_ops.euclidean_dist_transform(
                 image, dtype=output_dtype)
             output_flat = tf.reshape(output, [-1])
 
@@ -73,7 +73,7 @@ class DistanceOpsTest(tf.test.TestCase):
         # yapf: enable
         images = tf.constant([image] * batch_size, dtype=tf.uint8)
         for output_dtype in [tf.float16, tf.float32, tf.float64]:
-            output = distance_tranform_ops.euclidean_dist_transform(
+            output = distance_transform_ops.euclidean_dist_transform(
                 images, dtype=output_dtype)
             output_flat = tf.reshape(output, [-1])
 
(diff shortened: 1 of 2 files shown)


JasonMts:
I am currently getting an error importing tensorflow addons because it cannot find 'euclidean_dist_transform'. Is it related to this issue?

kyleabeauchamp:
That does not sound related, IMHO

seanpmorgan:
> I am currently getting an error importing tensorflow addons because it cannot find 'euclidean_dist_transform'. Is it related to this issue?

Probably not related but just to answer are you importing as `tfa.image.euclidean_dist_transform`? If so and it's still not working please file an issue and we'll work it out

tensorflow_addons/image/distance_transform_test.py
```diff
@@ -97,38 +97,37 @@ def test_image_with_invalid_dtype(self):
             # pylint: disable=bad-continuation
             with self.assertRaisesRegex(
                     TypeError, "`dtype` must be float16, float32 or float64"):
-                _ = distance_tranform_ops.euclidean_dist_transform(
+                _ = distance_transform_ops.euclidean_dist_transform(
                     image, dtype=output_dtype)
 
     def test_image_with_invalid_shape(self):
-        for invalid_shape in ([1], [2, 1], [2, 4, 4, 4, 1]):
```
facaiy: Why to remove the test case for 1D and 5D images?

tensorflow_addons/image/distance_transform_test.py
```diff
@@ -97,38 +97,37 @@ def test_image_with_invalid_dtype(self):
...
 
     def test_all_ones(self):
         image = tf.ones([10, 10, 1], tf.uint8)
-        output = distance_tranform_ops.euclidean_dist_transform(image)
+        output = distance_transform_ops.euclidean_dist_transform(image)
         expected_output = np.full([10, 10, 1], tf.float32.max)
         self.assertAllClose(output, expected_output)
 
+    def test_unknown_shape(self):
+        fn = distance_transform_ops.euclidean_dist_transform.\
```
facaiy: How about giving an alias for distance_transform_ops? In most cases, we'd better to avoid backlash because it's fragile as shown in the discussion https://www.python.org/dev/peps/pep-3125/

tensorflow_addons/image/distance_transform_test.py
Squadrick: This [test](https://github.com/tensorflow/addons/blob/d16b4077de06ed0636a93fbe88adc4ed7faec096/tensorflow_addons/image/utils_test.py#L45) in `image/utils_test.py` takes care of that, since I'm calling `to_4D_image` in `euclidean_dist_tranform`.

tensorflow_addons/image/distance_transform_test.py
Squadrick: I'm renaming `distance_transform_ops` -> `dist_ops`.

Squadrick:
Test case fails with:
```
/bin/bash: tools/run_docker.sh: No such file or directory
```
I'm guessing this has something to do with #328

facaiy:
> Test case fails with:
> 
> ```
> /bin/bash: tools/run_docker.sh: No such file or directory
> ```
> 
> I'm guessing this has something to do with #328

Yes, let's wait #328 to finish.

tensorflow_addons/image/distance_transform_test.py
facaiy: Oh, I see. We won't get any warning if the behavior of `to_4D_image` changes, however.

Squadrick: As long as the behaviour of `to_4D_image` is to convert the image to a 4D Tensor, euclidean distance transform should function. The checks for other aspects like single channel, etc. are still in place. I can add back the test cases if you think it's a good idea.

facaiy: Both are fine to me. OK, let's remove it, and we can add the test case back later.

PR #339, opened 2019-07-08
CLN: Statically copy normalize_tuple

seanpmorgan:


tensorflow_addons/utils/keras_utils.py
```diff
@@ -19,10 +19,50 @@
...
-
 
 def register_keras_custom_object(cls):
     tf.keras.utils.get_custom_objects()[cls.__name__] = cls
     return cls
+
+
+def normalize_tuple(value, n, name):
+    """Transforms a single integer or iterable of integers into an integer
+    tuple. Ported from tensorflow.python.keras.util.
```
facaiy: How about splitting the paragraph to two single lines? Because summary line should be one line, as required in https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

tensorflow_addons/utils/keras_utils.py
```diff
@@ -19,10 +19,50 @@
...
+        for single_value in value_tuple:
+            try:
+                int(single_value)
+            except (ValueError, TypeError):
+                raise ValueError('The `' + name +
+                                 '` argument must be a tuple of ' + str(n) +
+                                 ' integers. Received: ' + str(value) + ' '
+                                 'including element ' + str(single_value) +
+                                 ' of type' + ' ' + str(type(single_value)))
+        return value_tuple
```
kyleabeauchamp: So I noticed that the indentation of the `return` statement seems to have drifted one unit to the right.  It doesn't matter semantically but I suppose the intent is a bit clearer with the new approach---because the return statement lives in the conditional?

kyleabeauchamp:
LGTM once you get your docstring fixes merged.

tensorflow_addons/utils/keras_utils.py
seanpmorgan: Interesting... it differs from keras and tensorflow:
https://github.com/keras-team/keras/blob/master/keras/utils/conv_utils.py
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/utils/conv_utils.py

Quite hard to notice on the TF version (2 spaces...), but as you said it behaves the same.

tensorflow_addons/utils/keras_utils.py
seanpmorgan: Done. Thanks!

tensorflow_addons/utils/keras_utils.py
facaiy: keras Version looks weird, I think both coding styles below are acceptable:

```python
if condition:
    return value
return default_value

# or
if condition_a:
    return value_a
else:
    return value_b
```

facaiy:
@seanpmorgan Sorry, I forget to mention that we'd better to add unit tests for normalize_tuple if time allows :-)

seanpmorgan:
> @seanpmorgan Sorry, I forget to mention that we'd better to add unit tests for normalize_tuple if time allows :-)

Good catch... added.

PR #346, opened 2019-07-12
CLN: distort_image_ops_test

Squadrick:


tensorflow_addons/image/distort_image_ops_test.py
```diff
@@ -310,21 +302,14 @@ def _benchmark_adjust_saturation_in_yiq(self, device, cpu_count):
...
-        tag = "%s" % (cpu_count) if cpu_count is not None else "_all"
-        print(
-            "benchmarkAdjustSaturationInYiq_299_299_3_cpu%s step_time: %.2f us"
-            % (tag, step_time * 1e6))
-        self.report_benchmark(
-            name="benchmarkAdjustSaturationInYiq_299_299_3_cpu%s" % (tag),
-            iters=benchmark_rounds,
-            wall_time=step_time)
+                benchmark_values = self.run_op_benchmark(
+                    sess,
```
Squadrick: Any update on the status of [`tf.test.Benchmark`](https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/test/Benchmark) in TF2.x? As long as we use it, we'll need to use `tf.Session`.

seanpmorgan: Hmmm looks like there is a `tf.compat.v2.test.Benchmark` but it still utilizes a Session.. not sure who to follow up with that but you can put it in the monthly agenda if you'd like and we can ask there

Squadrick: Yup, I'll do that.

PR #354, opened 2019-07-15
Improve API for resetting AttentionMechanism memory

guillaumekln:
This simplifies the usage of the layer outside of Keras where a typical usage would be to create the layer once and then setup the memory dynamically, e.g.:

```python
attention = tfa.seq2seq.LuongAttention(units)
# ... later in the code ...
attention.setup_memory(memory, memory_sequence_length)
```

### DIFF
--- a/tensorflow_addons/seq2seq/attention_wrapper.py
+++ b/tensorflow_addons/seq2seq/attention_wrapper.py
@@ -116,6 +116,7 @@ class _BaseAttentionMechanism(AttentionMechanism, tf.keras.layers.Layer):
         if not callable(probability_fn):
             raise TypeError("probability_fn must be callable, saw type: %s" %
                             type(probability_fn).__name__)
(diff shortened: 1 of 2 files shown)


qlzh727:
By changing the setup_memory to be public, this could imply to user that it can be called multiple times, which should cause issue when using the overall keras model. On the other hand, it should work if this is only used for building up graph, and train with customized logic. 

Can we add some warning to user about this side effect if the memory is reset?

guillaumekln:
I'm not very familiar with the whole Keras workflow. Could you provide the warning string that I should add in this case?

qlzh727:
Sorry for the late reply. I guess the Attention itself is already not very aligned with keras framework (eg, require tensor as inputs to layer from call() instead of init argument). If user want just use keras to build the network and not use model.fit/eval/predict, then this change probably will not affect them.

PR #362, opened 2019-07-19
BLD: Pin bazel version for macos

seanpmorgan:
Resolve `load() statements must be called before any other statement. First non-load() statement appears at ` Error.

Seems we could alter the cuda_configure script to be compatible with 0.25, but for now I think we should align our bazel versions anyway.


feihugis:
@seanpmorgan Just in case you need it: #363 is submitted to update `cuda_configure.bzl` to be compatible with bazel 0.25.

PR #363, opened 2019-07-20
Make cuda_configure.bzl compatible with bazel

feihugis:
This PR tries to fix the error `load() statements must be called before any other statement` when building the project on MacOS. It moves `load()` to the top of `cuda_configure.bzl`.


kyleabeauchamp:
Seems reasonable to me, but I'm not particular informed on bazel stuff so I'll wait for another reviewer.

feihugis:
> Seems reasonable to me, but I'm not particular informed on bazel stuff so I'll wait for another reviewer.

@kyleabeauchamp, thanks for the quick review!

PR #369, opened 2019-07-25
Add `training` parameter in seq2seq decoders

### LINKED ISSUE #364: [Feature Request] Support setting `training=False` argument in `BeamSearchDecoder`
**System information**
- TensorFlow version (you are using): 2.0-beta
- TensorFlow Addons version: 0.40
- Is it in the tf.contrib (if so, where): No
- Are you willing to contribute it (yes/no): No
- Are you willing to maintain it going forward? (yes/no): No

Current `BeamSearchDecoder` does not support setting `training=False` argument during the computation. This limits many scenarios in which model behaves differently between training/inference stage, e.g., dropout.

An extra argument that supports setting the `training` argument should be added to this [line](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/seq2seq/beam_search_decoder.py#L547)

Everyone who wants the RNN Cell to behave differently between training and inference (e.g., dropout).
**Any Other info.**

Squadrick:
Fixes #364 

Also fixes a few typos using `tf.contrib`.

### DIFF
--- a/tensorflow_addons/seq2seq/basic_decoder.py
+++ b/tensorflow_addons/seq2seq/basic_decoder.py
@@ -111,18 +111,19 @@ class BasicDecoder(decoder.BaseDecoder):
             tf.nest.map_structure(lambda _: dtype, self._rnn_output_size()),
             self.sampler.sample_ids_dtype)
 
-    def step(self, time, inputs, state):
+    def step(self, time, inputs, state, training=None):
         """Perform a decoding step.
 
         Args:
           time: scalar `int32` tensor.
           inputs: A (structure of) input tensors.
           state: A (structure of) state tensors and TensorArrays.
+          training: Python boolean.
 
         Returns:
           `(outputs, next_state, next_inputs, finished)`.
         """
-        cell_outputs, cell_state = self.cell(inputs, state)
+        cell_outputs, cell_state = self.cell(inputs, state, training=training)
         if self.output_layer is not None:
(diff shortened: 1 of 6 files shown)


tensorflow_addons/seq2seq/basic_decoder.py
```diff
@@ -111,18 +111,19 @@ def output_dtype(self):
             tf.nest.map_structure(lambda _: dtype, self._rnn_output_size()),
             self.sampler.sample_ids_dtype)
 
-    def step(self, time, inputs, state):
+    def step(self, time, inputs, state, training=None):
```
feihugis: Would `training=False` be better?

Squadrick: That's what I initially figured, but keras seems to set `training=None` like [here](https://github.com/tensorflow/tensorflow/blob/545135634295853ad44ef9082a1782f2111db4dd/tensorflow/python/keras/layers/recurrent.py#L676).

feihugis: Yeah, you are right. Find another explanation [here](https://github.com/tensorflow/tensorflow/blob/545135634295853ad44ef9082a1782f2111db4dd/tensorflow/python/keras/engine/training.py#L2658-L2660):
> training: Boolean or None. Only relevant in symbolic mode. Specifies
        whether to build the model's graph in inference mode (False), training
        mode (True), or using the Keras learning phase (None).

PR #377, opened 2019-07-30
Add Conditional Random Field (CRF) layer

howl-anderson:
related to: #314 #22
depend on: #314 

# What's include
* TensorFlow Keras layer for CRF
* Loss function for CRF

# How to use it
The document is coming soon.

# Progress
[x] Layer class for CRF
[x] Test cases

# Limitation
* Currently, only support right padding mask (due to the underline CRF functions), CRF layer will detect and report to user if mask contains left padding.

# Roadmap (next version)
* Include `constraints`: an optional list of allowed transitions (from_tag_id, to_tag_id). Learned from [AllenNLP](https://github.com/allenai/allennlp)


foxik:
Any progress here?

tensorflow_addons/layers/crf.py
```diff
@@ -0,0 +1,334 @@
...
+                                                  constraint=self.boundary_constraint)
+
+        # or directly call self.built = True
+        super(CRF, self).build(input_shape)
+
+    def call(self, inputs, mask=None, **kwargs):
+        # mask: Tensor(shape=(?, ?), dtype=bool) or None
+
+        if mask is not None:
+            assert K.ndim(mask) == 2, 'Input mask to CRF must have dim 2 if not None'
```
Squadrick: Check using `if` and raise an Error.

Squadrick:
@howl-anderson Can update on the remaining TODOs?

howl-anderson:
@foxik @Squadrick Sorry about that, I quite busy recently. I will fix it up as soon as possible.

foxik:
Sorry to be pushy, but is there any progress?

howl-anderson:
@foxik Hi foxik, sorry about this, in the next week, I will have a holiday (about 3 days), I will use it for this project. I hope it can finish this PR.

OttoseiAza:
Does CRF loss work with tensorflow 2.0 already? I get the error even when I just running crf_loss_test.py        

crf, idx = y_pred._keras_history[:2]
    AttributeError: 'Tensor' object has no attribute '_keras_history'

howl-anderson:
@OttoseiAza Hi, this PR still WIP, it passed CI test but still not ready for release. When it's ready I will let you know.

howl-anderson:
@foxik @OttoseiAza This PR is ready to review, feel free to try it.

tensorflow_addons/layers/crf.py
```diff
@@ -0,0 +1,491 @@
...
+        if self.use_boundary:
+            self.potentials = self.add_boundary_energy(
+                self.potentials, mask, self.left_boundary, self.right_boundary)
+
+        self.sequence_length = self._get_sequence_length(inputs, mask)
+
+        decoded_sequence, _ = self.get_viterbi_decoding(
+            self.potentials, self.sequence_length)
+
+        return decoded_sequence
```
OttoseiAza: Is it possible to return tf.one_hot here to make it look like logits so that it could be compatible with tf.keras.metrics?

tensorflow_addons/layers/crf.py
```diff
@@ -0,0 +1,491 @@
...
+    def _get_sequence_length(self, input_, mask):
+        """
+        Currently underline CRF fucntion (provided by tensorflow_addons.text.crf)
+        do not support bi-direction masking (left padding / right padding),
+        it support right padding by tell it the sequence length.
+
+        this function is compute the sequence length from input and mask.
+        """
+        if mask is not None:
+            int_mask = tf.keras.backend.cast(mask, tf.int8)
```
OttoseiAza: can't int8 overflow in tf.keras.backend.sum later when calculating sequence length?

tensorflow_addons/layers/README.md
```diff
@@ -8,6 +8,7 @@
 | poincare |  |  |
 | sparsemax | @AndreasMadsen | amwwebdk+github@gmail.com |
 | wrappers | @seanpmorgan | seanmorgan@outlook.com |
+| crf | @howl-anderson | u1mail2me@gmail.com |
```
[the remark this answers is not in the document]
howl-anderson: Thanks, already fixed in a new commit.

tensorflow_addons/layers/README.md
```diff
@@ -18,6 +19,7 @@
 | poincare | PoincareNormalize | https://arxiv.org/abs/1705.08039    |
 | sparsemax| Sparsemax | https://arxiv.org/abs/1602.02068 |
 | wrappers | WeightNormalization | https://arxiv.org/abs/1602.07868 |
+| crf | CRF | https://repository.upenn.edu/cgi/viewcontent.cgi?article=1162&context=cis_papers |
```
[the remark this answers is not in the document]
howl-anderson: Thanks, already fixed in a new commit.

tensorflow_addons/layers/crf.py
howl-anderson: As far as I know,  metric functions (for example tf.keras.metrics.Accuracy) can work on the current output format. Are there some metrics that one_hot can do but the current output format can't?

tensorflow_addons/layers/crf.py
howl-anderson: What do you mean? Using if in here may raise an error?

howl-anderson:
@Squadrick This PR is ready, please take a look.

tensorflow_addons/layers/crf.py
OttoseiAza: What I'm trying to do here is make an example with CRF and SparseCategoricalCrossentropy look the same. So the example metric is SparseCategoricalAccuracy, but in the real life it may be some custom metrics like Precision/Recall on named entities in NER

howl-anderson: Thank you for your explain, I know what you mean now. Since this is the first release of CRF layer, I am trying to keep it as simple as possible so it can release quickly. What you want is the layer has a friendly output for builtin metrics and custom metric which is an important feature for me too. I am planning to write some custom metrics around the NER when the CRF layer is ready.  But I am not sure one_hot format is good for everyone. Maybe we can try using the CRF after it is released, if we still have trouble with the output format, then we can have a detailed plan how to make it better in the next release of CRF layer.

tensorflow_addons/layers/crf.py
howl-anderson: Thank you! Already fixed.

tensorflow_addons/layers/crf.py
```diff
@@ -0,0 +1,485 @@
...
+
+        return tf.keras.backend.cast(output, self.chain_kernel.dtype)
+
+    def __call__(self, inputs, *args, **kwargs):
+        outputs = super(CRF, self).__call__(inputs, *args, **kwargs)
+
+        # A hack that add _keras_history to EagerTensor, make it more like normal Tensor
+        for tensor in tf.nest.flatten(outputs):
+            if not hasattr(tensor, '_keras_history'):
+                tensor._keras_history = (self, 0, 0)
```
OttoseiAza: It's probably not a top priority right now, but do you have ideas/plans for the future on a better way to do it? It works alright now in most cases, but I get an error that _keras_history not found in some cases when input is unknown at the compile time and build is delayed to fit time.

howl-anderson: Can you tell me the case that you get the error, so I can try to resolve it? The reason I do a little hack here because, in eager execution mode, there is (as far as I know)  no way to find the layer who produced the output tensor. But CRF loss function depends on use tensor info to find the CRF layer instance. I have to add _keras_history to the tensor to make the eager tensor behave like the normal tensor generated under graph mode. The better way to do this maybe need TensorFlow core to add more information in eager tensor, or provide another way to find the layer who produced the tensor. I will try to submit an issue or PR for this.

OttoseiAza: One way to reproduce it is subclassing:
    class CRFModel(tf.keras.Model):
            def __init__(self):
                super(CRFModel, self).__init__()
                self.layer =  CRF(5)
            def call(self, inputs):
                return self.layer(inputs)

But I believe it's also possible to get the same error with custom first layer in sequential (if there is no input layer)

howl-anderson: Currently, using the CRF layer in a subclassing model need a little hack or patch, I already submit a commit contains a test case to show how to hack. But the graceful way to resolve this need writes a patch to TensorFlow core.

seanpmorgan:
Hi All, thank you for all your work on this. Seeing as [keras-contrib is migrating to tensorflow-addons ](https://github.com/keras-team/keras-contrib/issues/519) I'm thinking we should compare this implementation with the CRF Layer that exists there. 

We've had discussions about removing CRF from tfa.text and this may be a way to accomplish that and a CRF layer in one pass. It would also help the keras-contirb user-base to adopt tf-addons.

Thoughts? @howl-anderson @facaiy @tensorflow/sig-addons-maintainers

howl-anderson:
This implementation has a similar user interface with the keras-contrib one by design. The biggest difference is this one using CRF function from tfa.text while the keras-contrib one using builtin CRF function. Since CRF may remove from tfa.text in the near future, I would like to implement CRF function within this CRF layer.

gjxcq:
I want to use the keras CRF layer,but the default package in pypi not exist it,how can I use it or how much time it will be contained in the pypi package

howl-anderson:
@gjxcq Yes, you can when this feature is released, you can use CRF layer from tf_addons in tf.keras. Since  CRF functions will removed from tf_addons, I will update this PR with builtin (provided by the CRF layer it self) CRF funtions instead of the one from tf_adons.text.

tensorflow_addons/layers/crf.py
toxicmender: I think he means you should raise an Error in else case, i.e; When `mask is None`

toxicmender: I think He means you should raise a ValueError (or another appropriate Error, like maybe AttributeError) when `mask is None`, i.e; 
```py
if mask is not None:
    assert K.ndim(mask) == 2, 'Input mask to CRF must have dim 2 if not None'
else:
    raise ValueError(A tensor for variable %s of type %s is required'
                         ' for layer %s' % (name, dtype.base_dtype, self.name))
```
name = mask
dtype.base_dtype = type(mask)
self.name = Arbitrary Name of the layer object (if any)

Something along theses lines, I suppose

tensorflow_addons/layers/crf.py
```diff
@@ -0,0 +1,474 @@
...
+    def _compute_mask_left_boundary(mask):
+        """input mask: 0011100, output left_boundary: 0010000."""
+        # shift mask to right by 1: 0011100 => 0001110
+        offset = 1
+        right_shifted_mask = tf.keras.backend.concatenate(
+            [tf.keras.backend.zeros_like(mask[:, :offset]), mask[:, :-offset]],
+            axis=1)
+
+        # 0011100 > 0001110 => 0010000
+        left_boundary = tf.keras.backend.greater(mask, right_shifted_mask)
```
w-cheng: This throws an error if any previous layers set mask_zero=True.

Here is the code to reproduce the error:

from tensorflow.keras.layers import Embedding
from tensorflow.keras.models import Sequential
from tensorflow_addons.layers import CRF

model = Sequential()
model.add(Embedding(3001, 300, mask_zero=True))
crf = CRF(10, name='crf_layer')
model.add(crf)

Here is the relevant part of the error:
 TypeError: Value passed to parameter 'x' has DataType bool not in list of allowed values: float32, float64, int32, uint8, int16, int8, int64, bfloat16, uint16, float16, uint32, uint64

Cast both masks to int32 seem to solve the problem:
left_boundary = tf.keras.backend.greater(tf.dtypes.cast(mask, tf.int32), tf.dtypes.cast(right_shifted_mask, tf.int32))

tensorflow_addons/layers/crf.py
```diff
@@ -0,0 +1,474 @@
...
+            alt_mask = tf.ones(raw_input_shape)
+
+            sequence_length = self.mask_to_sequence_length(alt_mask)
+
+        return sequence_length
+
+    def mask_to_sequence_length(self, mask):
+        """compute sequence length from mask."""
+        sequence_length = tf.keras.backend.cast(
+            tf.keras.backend.sum(mask, 1), tf.int64)
```
w-cheng: This throws the same type error if mask_zero=True

TypeError: Value passed to parameter 'x' has DataType bool not in list of allowed values: float32, float64, int32, uint8, int16, int8, int64, bfloat16, uint16, float16, uint32, uint64

Cast mask to int32 seems to work:
tf.keras.backend.sum(tf.dtypes.cast(mask, tf.int32), 1), tf.int64)

lkytal:
Hi, I also tried the codes and find out it does not work with `tf.distribute.MirroredStrategy`, while those codes works well on a single device. An example of error messages are:

`WARNING:tensorflow:Gradients do not exist for variables ['conv1d_27/kernel:0', 'instance_normalization_29/gamma:0', ............... ] when minimizing the loss.
INFO:tensorflow:Error reported to Coordinator: Tried to create variable conv1d_27/kernel/m with mismatching name on device 1
  File "/home/liukaiy/anaconda3/envs/dl/lib/python3.6/site-packages/tensorflow_core/python/distribute/shared_variable_creator.py", line 84, in reuse_variable
    v = shared_variable_store[canonical_name][variable_index]
KeyError: 'conv1d_27/kernel/m'`

howl-anderson:
@w-cheng @lkytal Thank you all for your report, I start to fix it from today. When it's done, I will let you know.

howl-anderson:
Update: some APIs changed in tf.addon since the last commit I made, it will take me some time to fix it and test it.

tensorflow_addons/layers/crf.py
```diff
@@ -0,0 +1,474 @@
...
+    ```python
+        from tensorflow_addons.layers import CRF
+
+        model = Sequential()
+        model.add(Embedding(3001, 300, mask_zero=True)
+
+        crf = CRF(10, name='crf_layer')
+        model.add(crf)
+
+        model.compile('adam', loss={'crf_layer': crf.loss})
```
SilverLiningSL: maybe I'm wrong, but what will the crf.loss get? The traceback says 'CRF' object has no attribute 'loss'.

hobofan: I built a model with this yesterday, and the documentation is indeed wrong ATM.

`crf.loss` would need to be replaced with `ConditionalRandomFieldLoss()` from the `losses` module to get it working

SilverLiningSL: > I built a model with this yesterday, and the documentation is indeed wrong ATM.
> 
> `crf.loss` would need to be replaced with `ConditionalRandomFieldLoss()` from the `losses` module to get it working

Thanks for your reply. So what we should do is to replace crf.loss with ConditionalRandomFieldLoss(),and the y of model.fit should be a [batch_size,sequence_length] array, is that right? The loss of my model stays around 16-20, which is abnormal, and in contrast with the result I got from keras+keras-contrib.

hobofan: Yeah, that sounds about right.

I think it's very hard to compare this to the CRF from keras-contrib, as they seem to be quite different.

SilverLiningSL: > Yeah, that sounds about right.
> 
> I think it's very hard to compare this to the CRF from keras-contrib, as they seem to be quite different.

Thank you!

tpanza:
Hello @howl-anderson , @Squadrick , @SilverLiningSL , @seanpmorgan  is there any news regarding this?

After reading through #22 , #314 , it looks like there are low-level CRF functions available in `tensorflow_addons.text`, but _still_ no CRF **layer**.

If we are using `tf.keras`, [we cannot intermix that with the "pure" keras](https://stackoverflow.com/a/51831434/2491761). 

This limitation presumably extends to the `keras_contrib` modules as well, since I am getting 
`AttributeError: 'Node' object has no attribute 'output_masks'` when trying to use the CRF layer from `keras_contrib.layers` with other layers from `tf.keras.layers`.  

So we cannot use the CRF layer from `keras_contrib.layers`. (And with TF 2.0, there is no `tf.contrib.crf`.)

May I ask what is the plan and intention going forward? Is keras-team/keras-contrib#519 going to supersede this?

WarrenGreen:
@Squadrick Looks like your changes have been addressed and progress is 100%. Are you ready to approve?

Squadrick:
@WarrenGreen Thanks for the reminder, I'll take a look over the weekend.

howl-anderson:
@tpanza Sorry for the delay.  At the end of a year, usually, we are busier than another period. In the recent week, I am working on CRF stuff, but the first step I am working on migrating CRF Keras layer to TF 1.15 in the project [tf_crf_layer](https://github.com/howl-anderson/tf_crf_layer). It works with TF 1.14 but broken with TF 1.15. Since this is the code, we used in the production environment. I need to fix this first. The good news is it nearly finished. @Squadrick Please give me one more week, and I will fix the docs stuff, and add more tests.

tensorflow_addons/layers/__init__.py
```diff
@@ -25,4 +25,5 @@
 from tensorflow_addons.layers.optical_flow import CorrelationCost
 from tensorflow_addons.layers.poincare import PoincareNormalize
 from tensorflow_addons.layers.sparsemax import Sparsemax
-from tensorflow_addons.layers.wrappers import WeightNormalization
\ No newline at end of file
+from tensorflow_addons.layers.wrappers import WeightNormalization
+from tensorflow_addons.layers.crf import CRF
```
Squadrick: Order this alphabetically.

tensorflow_addons/layers/crf.py
```diff
@@ -0,0 +1,478 @@
...
+            )
+            self.right_boundary = self.add_weight(
+                shape=(self.units,),
+                name="right_boundary",
+                initializer=self.boundary_initializer,
+                regularizer=self.boundary_regularizer,
+                constraint=self.boundary_constraint,
+            )
+
+        # or directly call self.built = True
```
Squadrick: You can remove this comment.

tensorflow_addons/layers/crf.py
```diff
@@ -0,0 +1,478 @@
...
+    def call(self, inputs, mask=None, **kwargs):
+        # mask: Tensor(shape=(batch_size, sequence_length), dtype=bool) or None
+
+        if mask is not None:
+            if tf.keras.backend.ndim(mask) != 2:
+                raise ValueError(
+                    "Input mask to CRF must have dim 2 if not None")
+
+        # left padding of mask is not supported, due the underline CRF function
+        # detect it and report it to user
```
Squadrick: Create an issue for this. We can add support to it soon.

tensorflow_addons/layers/crf.py
```diff
@@ -0,0 +1,478 @@
...
+
+            sequence_length = self.mask_to_sequence_length(alt_mask)
+
+        return sequence_length
+
+    def mask_to_sequence_length(self, mask):
+        """compute sequence length from mask."""
+        sequence_length = tf.keras.backend.cast(
+            tf.keras.backend.sum(tf.keras.backend.cast(mask, tf.int8), 1),
+            tf.int64)
```
Squadrick: Use `tf.cast` instead of `tf.keras.backend.cast`

tensorflow_addons/layers/crf.py
```diff
@@ -0,0 +1,478 @@
...
+        # TODO(howl-anderson): for below code
+        # Original code in keras_contrib:
+        # end_mask = K.cast(
+        #   K.greater(self.shift_left(mask), mask),
+        #   K.floatx()
+        # )
+        # May have a bug, it's better confirmed
+        # by the original keras_contrib maintainer
+        # Luiz Felix (github: lzfelix),
+        # mailed him already and waiting for reply.
```
Squadrick: Any update on this?

cc: @lzfelix

tensorflow_addons/layers/crf.py
```diff
@@ -0,0 +1,478 @@
...
+        left_boundary = tf.keras.backend.greater(
+            tf.dtypes.cast(mask, tf.int32),
+            tf.dtypes.cast(right_shifted_mask, tf.int32))
+        # left_boundary = tf.keras.backend.greater(mask, right_shifted_mask)
+
+        return left_boundary
+
+    def add_boundary_energy(self, potentials, mask, start, end):
+        def expend_scalar_to_3d(x):
+            # expend tensor from shape (x, ) to (1, 1, x)
```
Squadrick: Typo? `expend` -> `expand`?

tensorflow_addons/layers/crf.py
```diff
@@ -0,0 +1,478 @@
...
+            return tf.keras.backend.expand_dims(
+                tf.keras.backend.expand_dims(x, 0), 0)
```
Squadrick: Instead of two `tf.expand_dims`, you can use a single `tf.reshape`.

tensorflow_addons/layers/crf.py
```diff
@@ -0,0 +1,478 @@
...
+            mask = tf.keras.backend.cast(self.mask, tf.keras.backend.floatx())
+            return (tf.keras.backend.sum(judge * mask) /
+                    tf.keras.backend.sum(mask))
+
+    def _dense_layer(self, input_):
+        if self.use_kernel:
+            output = self.activation(
+                tf.keras.backend.dot(input_, self.kernel) + self.bias)
+        else:
+            output = input_
```
Squadrick: Any chance of replacing this with a `tf.keras.layers.Dense`?

tensorflow_addons/layers/crf.py
```diff
@@ -0,0 +1,478 @@
...
+        for tensor in tf.nest.flatten(outputs):
+            if not hasattr(tensor, '_keras_history'):
+                tensor._keras_history = (self, 0, 0)
+
+        return outputs
+
+    @property
+    def _compute_dtype(self):
+        # fixed output dtype from underline CRF functions
+        return tf.int32
```
Squadrick: I don't see this used anywhere?

Squadrick: Unless it's overriding `tf.keras.layers.Layer`

tensorflow_addons/losses/BUILD
```diff
@@ -17,6 +18,7 @@ py_library(
     ],
     deps = [
         "//tensorflow_addons/activations",
+        "//tensorflow_addons/layers",
```
Squadrick: You'll need to add `//tensorflow_addons/text/` as a dep. Better yet, create a separate target only for `text.crf` in `text/BUILD`, and add that as a dependency.

tensorflow_addons/losses/crf.py
```diff
@@ -0,0 +1,66 @@
...
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow as tf
+
+from tensorflow_addons.layers.crf import CRF
+
+
+@tf.keras.utils.register_keras_serializable(package="Addons")
+class ConditionalRandomFieldLoss(object):
```
Squadrick: Why not extend `tf.keras.losses.Loss`, and implement `call`?

tensorflow_addons/losses/crf_test.py
```diff
@@ -0,0 +1,335 @@
...
+            "adam",
+            loss=crf.ConditionalRandomFieldLoss(),
+            metrics=[tf.keras.metrics.Accuracy()])
+
+        model.fit(self.logits, self.tags, epochs=10, batch_size=1)
+
+    def test_dump_and_load(self):
+        self.skipTest("don't work now")
+
+        MODEL_PERSISTENCE_PATH = './test_saving_crf_model.h5'
```
Squadrick: Use [`tf.TestCase.get_temp_dir()`](https://www.tensorflow.org/api_docs/python/tf/test/TestCase?version=nightly#get_temp_dir) instead.

tensorflow_addons/layers/__init__.py
howl-anderson: fixed

tensorflow_addons/layers/crf.py
howl-anderson: fixed

tensorflow_addons/layers/crf.py
howl-anderson: fixed

tensorflow_addons/layers/crf.py
howl-anderson: fixed

tensorflow_addons/layers/crf.py
howl-anderson: fixed

tensorflow_addons/losses/BUILD
howl-anderson: I am not sure what I do is the thing exactly what you want, but I am do updated codes to fix the dependencies issue.

tensorflow_addons/losses/crf_test.py
howl-anderson: fixed

tensorflow_addons/layers/crf.py
howl-anderson: fixed

tensorflow_addons/layers/crf.py
howl-anderson: Yes, it overriding the base layer's method. CRF layer's output dtype is int32, but default dtype of layer is tf.keras.backend.floatx()

tensorflow_addons/layers/crf.py
lzfelix: Hello. My apologies in the delay, I was vacationing for a few days.

When this code was hosted under `keras-contrib` I fixed some issues regarding model saving/loading and at that time people asked me to be the maintainer of this layer, however I didn't write the code for it (and particularly find it really hard to follow).

Indeed, the CRF layer seems to have some bugs, for instance, loss becomes negative if the model is trained for too many iterations (notice that mathematically the CRF loss can't be negative), and this behaviour disappears when a masking layer is used.

Some users suggested potential bugs in the `keras-contrib` repo. I'll find all of them and update this comment, but it seems that this is one of them.

howl-anderson: Hi lzfelix, nice to have your reply. Since the bug with negative loss likely exists in this implementation too, I will try to find out why and fix it. I hope you can be the reviewer for those bugs (including this boundary one), that will be very helpful!

lzfelix: Sure. I'll brush up my math on the CRF and help you with that!

DachuanZhao:
Hi , @howl-anderson . When do you plan to finish this feature ? Is there anything I can do to help you speed it up ?  ヽ(✿ﾟ▽ﾟ)ノ

howl-anderson:
@DachuanZhao Thank you for your kindness. Test this layer in your context and then report it is a good help to me.

DachuanZhao:
> @DachuanZhao Thank you for your kindness. Test this layer in your context and then report it is a good help to me.

 I'm glad to use it . But could you please give me some example code ? (｡･∀･)ﾉﾞ

howl-anderson:
@DachuanZhao There are some code examples in the test cases of the CRF layer.

Squadrick:
@howl-anderson Update? I can pick up this PR if you're busy with other work. This is a highly requested feature, that already been in the works for a few months now. It's best we get it done ASAP.

howl-anderson:
@Squadrick Hi, thank you for your concern. The easy part of this PR is already done. The hard part (loss function etc.) is also implemented in a way that is not considered graceful but works. I am trying to provide a more graceful implementation. The hard part related to Keras core in TF 2.0 (graph mode / eager mode). I read the Keras core before, but I need some time to refresh my knowledge about the detail (code of Keras core may change over time). The good news is a Chinese traditional holiday (The Spring Festival) is coming (within about one week), I will get enough time for this PR.  I will give a detailed answer to the feasible implementation choices we can have and implement the best one.

tensorflow_addons/layers/crf.py
spasis: Any updates on this issue? I have still the same error when using mask_zero=True in my Embedding layer...

drdozer:
I'm very interested in this functionality. Im hoping to use it to make input steam editors, where the output edited stream is a near perfect copy of the input, but with tokens potentially replaced as well as insertions of new tokens and deletions that skip over some input. I think this crf layer I'd what I need.

design_docs/crf.md
```diff
@@ -0,0 +1,60 @@
...
+model.compile('adam', loss=crf_loss)
+
+model.fit(x, y)
+```
+
+#### pros ####
+the standard way to use loss
+
+#### cons ####
+in the eager mode, there need a complicated patch to make this solution works.
```
AakashKumarNain: Can you provide a minimal example for this?

howl-anderson: Sure, this is the way currently I implement the CRF layer (code already included in this PR).  I will add more information about how it works.

design_docs/crf.md
```diff
@@ -0,0 +1,60 @@
...
+
+model.fit(x, y)
+```
+
+#### pros ####
+easy to implement and no more need patch
+
+#### cons ####
+
+This solution has a shortage that this model can not be save and load from disk anymore.
```
AakashKumarNain: When loading the model, all we need to do is to build the `CRF` layer again and then load it in the model as a custom object which should be fine IMO

howl-anderson: In TF 2.x, there are two export/load formats one is HDF5 (.h5), another one is SavedModel which is the default value. In the current code of loading saved model (`tensorflow_core.python.keras.saving.save.load_model`), it doesn't use the arg`custom_objects`, so it doesn't work as a general solution.

When using the HDF5 format, the CRF layer is implemented as a class of layer, when Keras restore such an object from a .h5 file, it will look up the class not the instance from `custome_objects`  (`tensorflow_core.python.keras.utils.generic_utils.deserialize_keras_object`), so directly passing a CRF layer instance through `custome_objects` doesn't work.

The above is my personal understanding of the code. It is not guaranteed to be completely correct. If there is an error, please correct me.

design_docs/crf.md
howl-anderson: I have updated the documentation to explain where the trick is.

tensorflow_addons/losses/crf.py
howl-anderson: I add a doc (`design_docs/crf.md`) to explain why loss function cannot be implemented as a subclass of `tf.keras.losses.Loss`.  If there is an error, please correct me.

design_docs/crf.md
AakashKumarNain: Either of us is missing something here. Loading custom objects should work as per my understanding but I would be happy to look at an example, like a colab if you have one ready

howl-anderson: For now, I don't have an example for this.  But for PoC, I will add a Jupyter notebook example for this later.

AakashKumarNain: Sure. Thank you

howl-anderson: PoC was uploaded as `design_docs/PoC_of_crf_layer.ipynb`.

howl-anderson:
@googlebot I consent.

design_docs/crf.md
AakashKumarNain: Thanks. I will look into it

AakashKumarNain: @howl-anderson I can save and load the model with custom layer. Check it out:
https://colab.research.google.com/drive/1k12THfKb5gPa2RsWip_LPZ6zo2Vn8Ufm

howl-anderson: @AakashKumarNain Yes, you can use `custome_objects` to load a custom layer/loss/..., but what I meaning is that you can not use it load an instance of a layer/loss/... into a model. In solution 2, users need to create a loss instance from `crf_layer.get_loss()`, and load the CRF layer instance and loss instance into the model.

tensorflow_addons/layers/crf.py
howl-anderson: @w-cheng @spasis Sorry for the late reply, this issue fixed in the latest code.

tensorflow_addons/layers/crf.py
howl-anderson: @w-cheng Sorry for the late reply, this issue fixed in the latest code.

howl-anderson:
@lkytal Can you provide a mini code to reproduce the issue with `tf.distribute.MirroredStrategy`, thank you!

tensorflow_addons/layers/crf.py
howl-anderson: PR #1089 is created for this.

tensorflow_addons/layers/crf.py
howl-anderson: The bug which simillar to keras-team/keras-contrib#508 was already fixed in this PR.  This CRF layer uses `tensorflow_addons.text.crf` to compute the loss which is mainly ported from `tf.contrb.crf`, currently, as far as I know, there is no report on GitHub that `tf.contrb.crf` has a similar negative loss issue same with keras-team/keras-contrib#253, so far so good.

design_docs/crf.md
AakashKumarNain: @howl-anderson If I understand correctly, you are saying that when loading model, if you generate a new instance of the layer with diff param values and then pass it to the custom_object, it won't work correctly. Here are my views on the same:

1. When you are saving a model, that model has a state. The state defines the architecture and the weights values used in the model definition.
2. When loading the model, you expect the model to load the same values that you saved. If you are changing anything at the time of loading, then you are basically changing the model itself which shouldn't be the way to dot it.
3. If you want something, like a layer, to have a state different than the one it was saved with, then you should compile the model again after making changes to the states that you want. I have updated this in the notebook. 

cc: @Squadrick @WindQAQ @seanpmorgan

howl-anderson: > If I understand correctly, you are saying that when loading model, if you generate a new instance of the layer with diff param values and then pass it to the custom_object, it won't work correctly

No,  you misunderstand what I want to express. I use that case to demo the PoC which is used to prove that user cannot pass a layer instance (no matter with diff parameters or not) into a model. that PoC is used to respond to your comment:
> When loading the model, all we need to do is to build the CRF layer again and then load it in the model as a custom object which should be fine IMO

-----

Anyway, I think I find a way to load model with the `get_keras_loss` compatibility, although I don't think it is as elegance/simple as solution 1. I will update the doc later.

AakashKumarNain: `I use that case to demo the PoC which is used to prove that user cannot pass a layer instance (no matter with diff parameters or not) into a model.`

I got it. I am looking for a specific use case where you want to pass an `instance` while loading the saved model. And more specifically, what would be the advantage of doing it that way?

howl-anderson: I don't mean I want to pass an instance while loading a model from disk, the reason I say it, it is because of your comment:

> When loading the model, all we need to do is to build the CRF layer again and then load it in the model as a custom object which should be fine IMO

From what I understand, I think you are trying to create a CRF layer instance and pass it through `custom_objects`.

so I say:

> When using the HDF5 format, the CRF layer is implemented as a class of layer, when Keras restore such an object from a .h5 file, it will look up the class not the instance from custome_objects (tensorflow_core.python.keras.utils.generic_utils.deserialize_keras_object), so directly passing a CRF layer instance through custome_objects doesn't work.

then you want a colab for this:

> Either of us is missing something here. Loading custom objects should work as per my understanding but I would be happy to look at an example, like a colab if you have one ready

This is why I add the custom_objects part for my PoC code.

AakashKumarNain: @howl-anderson thanks for the detailed info. Let other maintainers review it, then I will take a look into it again

gabrieldemarmiesse:
@googlebot I consent

gabrieldemarmiesse:
@howl-anderson I've merged master into your branch to update it and fixed any formatting/conflicts it might have. If you need to do some more modifications, please do `git pull` beforehand.

howl-anderson:
@gabrieldemarmiesse Thank you!

howl-anderson:
@lkytal I am not sure if it is related but I found an issue report on `tf.distribute.MirroredStrategy` with custom loss: tensorflow/tensorflow#35093

HStuart18:
Is there any update on a guide/document for this?

Thanks

howl-anderson:
Hi @HStuart18, this CRF implement is already stable to me, but since it uses a patch to make it works, but tensorflow/tensorflow#37119 gives another implement suggestion, I am trying to apply it.

HStuart18:
Hi @howl-anderson,
Thanks for this. I am sure the implementation is stable, but I’m looking for a document on how to add a CRF later to a Keras model and compile with CRF loss. Are there any such examples? I was unable to find any in the repo. Thanks.

howl-anderson:
@googlebot I consent

HStuart18:
I should mention that I am after a layer implementation snippet for TF2, and unfortunately, `tf_crf_layer` is not configured to work with TF2 it seems.

howl-anderson:
@HStuart18 Here is a simple demo for using CRF layer:
```python
import numpy as np
import tensorflow as tf

model = tf.keras.models.Sequential(
    [
        tf.keras.layers.Input(3),
        tf.keras.layers.Embedding(1000, 100, mask_zero=True),
        tfa.layers.CRF(10),
    ]
)


model.summary()

model.compile("adam", loss=tfa.losses.crf_loss)

x = np.array([[1, 2, 3]])
y = np.array([[3, 4, 5]])

model.fit(x, y, epochs=10)
```

and,

PyPI package `tf_crf_layer` is designed only used in TF 1.15.x

HStuart18:
I get the error:

`module 'tensorflow_addons.layers' has no attribute 'CRF'`

I can't seem to find such a file in this repo either.

howl-anderson:
@HStuart18 This feature is not released yet, so it not included in the tensorflow_addons package now. If you want use CRF feature, consider git clone from my [source branch](https://github.com/howl-anderson/addons/tree/feature/crf_layers)(note: it on branch `feature/crf_layers`)

HStuart18:
When using `keras_contrib.layers.CRF`, this is the model summary that is produced in my case:

```
Model: "model1"
_________________________________________________________________
Layer (type)                 Output Shape              Param #   
=================================================================   
_________________________________________________________________
embedding_1 (Embedding)      (None, 140, 150)          4525950   
_________________________________________________________________
bidirectional_1 (Bidirection (None, 140, 300)          361200    
_________________________________________________________________
lstm_2 (LSTM)                (None, 140, 300)          721200    
_________________________________________________________________
time_distributed_1 (TimeDist (None, 140, 18)           5418      
_________________________________________________________________
crf_1 (CRF)                  (None, 140, 18)           702       
=================================================================
Total params: 5,614,470
Trainable params: 5,614,470
Non-trainable params: 0
_________________________________________________________________
```
However, using the CRF in tf addons results in this summary:

```
Model: "model1"
_________________________________________________________________
Layer (type)                 Output Shape              Param #
=================================================================
embedding (Embedding)        (None, 80, 40)            18480
_________________________________________________________________
bidirectional (Bidirectional (None, 80, 80)            25920
_________________________________________________________________
lstm_1 (LSTM)                (None, 80, 80)            51520
_________________________________________________________________
time_distributed (TimeDistri (None, 80, 10)            810
_________________________________________________________________
crf (CRF)                    (None, 80)                230
=================================================================
Total params: 96,960
Trainable params: 96,960
Non-trainable params: 0
_________________________________________________________________
```
Note the dimensionality of the output shape of the crf layer. Ignore the differences in the exact shape parameter values. Why does the tf addons implementation result in one less dimension than the  `keras_contrib.layers.CRF` layer? How can I replicate the `keras_contrib.layers.CRF` layer?

Thanks, Harry

howl-anderson:
@HStuart18 `keras_contrib.layers.CRF` output softmax encoded tag sequence such as `[[0, 1, 0], [[0, 0, 1]]`, while `tfa.layers.CRF` output tag sequence `[1, 2]`.

follow code can help convert between two formats:
```python
import tensorflow as tf

data_2d = tf.constant([[0,1,0,0],[1,0,0,0],[0,0,0,1]])

data_1d = tf.argmax(data, axis=-1)  # data_1d will be [1, 0, 3]

data_2d_round_trip = tf.one_hot(data_1d, depth=4, dtype=tf.int32  # data_2d_round_trip will be [[0,1,0,0],[1,0,0,0],[0,0,0,1]]
```

design_docs/crf.md
```diff
@@ -0,0 +1,153 @@
...
+
+So it has same issue that solution 1.
+
+### Solution 3: implement loss as a callable class
+
+#### pros
+Nothing breaks. `mark` property is still a meaningful tensor which standard as a mark.
+
+#### cons
+this solution need understanding how keras process a loss function, which is not documented and not recommend way in TF 2.x.
```
gabrieldemarmiesse: I'm not sure that's true anymore: https://www.tensorflow.org/api_docs/python/tf/keras/losses/Loss

I don't fully understand the complexity of CRF but I'll try to make this pull request progress.

gabrieldemarmiesse:
To implement a layer that acts as a loss function, I recommend using the method `add_loss()`. Could it help you? https://www.tensorflow.org/guide/keras/custom_layers_and_models#layers_recursively_collect_losses_created_during_the_forward_pass 

Maybe we wouldn't need to `_keras_history` hack. 

Also the expected outputs would need to be passed as input to the model but I don't think it's as hacky as using `_keras_history`.

bodak:
I added a PR onto @howl-anderson 's branch (https://github.com/howl-anderson/addons/pull/1) a while ago with an implementation that:
1. Allows for sequence lengths > 128 https://github.com/bodak/addons/commit/684bde57fb8c3dbb6e4a9bfc33041ed140f5805e
2. Does not use the `_keras_history` hack. https://github.com/bodak/addons/commit/289ba791c7b36892d26add8484f5e0ae043beaf9

The issue that remains with this API approach (i.e. passing in the CRF layer to the loss fn at init) is serialization and deserialization. I am not familiar enough with `.get_config` / `.from_config` to make this work (or even if this is possible).

Tests are updated in https://github.com/bodak/addons/commit/033cd24f78dc9f27115a2aeb6a9abd5fc3663ee8

Example Functional API
```
model.add(tf.keras.layers.Input(shape=(3, 5)))
model.add(crf)
model.compile(
            "adam",
            loss=ConditionalRandomFieldLoss(crf_layer=crf),
            metrics=[tf.keras.metrics.Accuracy()])```

gabrieldemarmiesse:
I believe that with `add_loss()` we woudn't have to use `_keras_history` nor we would have issues with serialization and deserialization.

howl-anderson:
@gabrieldemarmiesse I am working on using the `add_loss` way to implement the CRF layer. When it's done, I will update this PR.

gabrieldemarmiesse:
Thanks a lot @howl-anderson I hope it helped.

albertnanda:
Any update here?

howl-anderson:
@albertnanda I encountered some problems during the implementation and are trying to solve them.

gabrieldemarmiesse:
I moved some code around in this PR, and I made #1363 which has less hacks. @howl-anderson , @Squadrick, please take a look when you can.

@howl-anderson , I've put you as author of the first commit.

bodak:
> I moved some code around in this PR, and I made #1363 which has less hacks. @howl-anderson , @Squadrick, please take a look when you can.
> 
> @howl-anderson , I've put you as author of the first commit.

While `_keras_history` approach in this PR is unfortunate (but necessary to access the layer at loss), it has no impact on the end API user. 

The `add_loss()` approach unless I missed something won't work due to not having access to `y_true` inside the layer.

@gabrieldemarmiesse  Your approach seems to have introduced hacks that lead to multiple impacts on the API user: two models to track and fake targets to input. From an end user perspective, I would be hesitant to leave this to be handled externally.

gabrieldemarmiesse:
@bodak we can't use private APIs in code used by users. 

If at the next release, this attribute disapears, we get an attribute error and we have to redo all the work of this pull request to find another pattern that works for CRF.

Or worst case scenario, the behavior changes silently (for example tensorflow decide to enforce that we can't use tensors or variables not declared in the signature in the loss), and all users get wrong numerical results/horrible error messages and we spend a lot of time understanding what is going on. To in the end fall back to #1363 because it's the only way that is garanteed to last since we don't use private APIs. 

I agree with you that it leads to more confusing code for end users, but they'll have more flexibility and stability. For example, in the implementation proposed in   #1363, the user is not forced to have the CRF layer be the last layer of the model.

If users don't like the pattern of having two models or passing the labels as inputs, they're free to use only one model and do a custom training loop. Making a custom training loop is getting easier and easier nowadays.

mhartig:
Hi, huge thanks to you for making this effort and building a CRF layer for tensorflow 2. For my application I require the probabilities or confidences for the class predictions in order to conduct a majority vote. As far as I understand, if I use the variable `self.potentials` that is used in the function `get_viterbi_decoding`  as additional output, this will yield the probabilities that come from the dense layer, so they are not necessarily the "true" probabilities for the CRF class prediction. Is there a simple way to get the predicted class together with the probability?

howl-anderson:
Hi @mhartig, output the probabilities or confidences for the prediction is a good feature. After the CRF layer is merged, I can implement it.

mhartig:
I have tried to modify the output of the CRF layer myself, unfortunately I'm not very familiar with the underlying structure of tf.keras and the context in which the class methods are called. The output shape seems to be restricted in more than one place, but it is very hard to see exactly where the expected output shape needs to be changed in order to return for example a tensor of shape (batch_size, sequence length, number_of_classes + 1) and dtype float32 instead of a tensor of shape (batch_size, sequence_length) and dtype int32.
If it doesn't take too much time for you, it would be awesome if you could share a list of places where the output settings need to be modified. I could do the manipulations within the layer by myself.

howl-anderson:
@mhartig Since this CRF layer implement is not merged, so I am not planning to implement this feature now. But if I get enough free time, I willing to help you with this.

NLP-ZY:
Any update？

howl-anderson:
@NLP-ZY waiting for tensorflow/tensorflow#37818

NLP-ZY:
It's difficult to get response from  tensorflow developers. Many tensorflow 2.0 issues haven't  get response, maybe I should consider pytorch.

soumayan:
any example of how to implement this crf layer?

howl-anderson:
@soumayan Currently, it's waiting for other PR. Doc will be added to the repo when it's ready.

soumayan:
@howl-anderson when I'm installing tensorflow_addons package from your github repository using

!pip install git+https://github.com/howl-anderson/addons.git

I'm getting error--
ERROR: Failed building wheel for tensorflow-addons
Failed to build tensorflow-addons
ERROR: Could not build wheels for tensorflow-addons which use PEP 517 and cannot be installed directly

howl-anderson:
@soumayan You should download it and built it to a wheel package. https://github.com/tensorflow/addons has docs for how to build from source.

zhoushaoxiang:
excuse me，how to get accuray while training？I can‘t find crf_accuracy.py

howl-anderson:
@zhoushaoxiang I removed it for more focus on CRF itself. I am planning to add metric functions (just like the ACC function ) after the CRF implement merged.

gabrieldemarmiesse:
@howl-anderson , I'll close this pull request in favor of #1733 which has an API as clean as this one, is flexible and can be serialized, the whole package without using any private api.

Issue #279, opened 2019-06-11
Move python3 CI testing to python3.6


seanpmorgan:
Recently there has not been any [py34 tf2-nightlies published](https://pypi.org/project/tf-nightly-2.0-preview/#files). I've asked SIG-Build if this is a permanent change seeing as py34 is EOL, but regardless I think it's wise to move our CI testing to python3.6

Can this be easily done? @karmel @av8ramit

seanpmorgan:
As a discussion topic: should we stop supporting python3.4 now that it's EOL?

kyleabeauchamp:
I agree with deprecating 34.

av8ramit:
If the official agreement is to do this with Python3.6 we can look into this. I'll let @karmel make the final decision.

seanpmorgan:
**Update:**
It has been confirmed as a bug and tf2-nightlies for py34 have started being published again. They did mention that tf-core will indeed remove py34 support in the future -- so I think addons should just align with tensorflow core in doing this. For now this issue can be closed but we'll need to re-open this in the future it seems.

seanpmorgan:
Re-opening this issue. We need to adjust our docker scripts to install py36 and test with that as py34 is no longer being published

PR #380, opened 2019-07-30
Add fallback flag to seq2seq tests

karmel:
This test uses a Keras anti-pattern in where the layer optionally accepts the output of another layer in the constructor. We are in the process of refactoring the v2 training loop to make it work with tf.function (faster! better! stronger!), but this code will need to be updated as a result. For now, we are falling back to the v1 training loop with an experimental flag; @qlzh727 will look for a longer-term solution when he gets through some pressing release issues.

(The flag won't be in the nightlies until tomorrow, but setting up this PR early, and will wait until tests are green to submit.)


seanpmorgan:
Just FYI py3 failures will continue until we get https://github.com/tensorflow/addons/issues/279 completed (py34 is no longer being published).

karmel:
@seanpmorgan , would you prefer we wait for the py3 builds? This should not be affected by py version.

PR #382, opened 2019-07-31
TEST: Upgrade python3 env to py36

### LINKED ISSUE #279: Move python3 CI testing to python3.6

seanpmorgan:
Closes #279.

Py34 pip packages for tf-nightly have not been produced in ~2 weeks. When the new build env arrives we'll be re-vamping our testing scripts anyway.

### DIFF
--- a/tools/ci_testing/addons_cpu.sh
+++ b/tools/ci_testing/addons_cpu.sh
@@ -31,6 +31,7 @@ if [[ ${PLATFORM} == "darwin" ]]; then
     N_JOBS=$(sysctl -n hw.ncpu)
 else
     N_JOBS=$(grep -c ^processor /proc/cpuinfo)
+    tools/ci_testing/install_py36.sh # Patch to test on py36
 fi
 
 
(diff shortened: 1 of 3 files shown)


seanpmorgan:
Hmmm so I just realized that we tell people to run these shell scripts from our makefile in the CONTRIBUTING doc. Those scripts shouldn't be installing anything if thats the case, but I don't think we should restrict what our ci_test scripts do.

IMO we should have separate scripts for CI and Local. @facaiy Thoughts?

seanpmorgan:
Alternatively we could append this install script in the docker build:
https://github.com/tensorflow/addons/blob/master/tools/run_docker.sh#L91

facaiy:
https://github.com/tensorflow/addons/pull/384 looks better. Thank Sean

PR #384, opened 2019-07-31
FIX: Better implementation of py36 testing

seanpmorgan:
@Squadrick @tensorflow/sig-addons-maintainers  this should work better. Apologies.

Installing py36 is only safe to do knowing the user is running in a docker container.

### DIFF
--- a/tools/ci_testing/install_py36.sh
+++ b/tools/ci_testing/install_py36.sh
@@ -25,6 +25,7 @@ apt-get -y -qq update && apt-get -y -qq install python3.6
 
 python3.6 get-pip.py -q
 python3.6 -m pip --version
+rm get-pip.py
 
 ln -sfn /usr/bin/python3.6 /usr/bin/python3
 pip3 install scipy  # Pre-installed in custom-op
--- a/tools/run_docker.sh
+++ b/tools/run_docker.sh
@@ -88,7 +88,7 @@ if [[ -z "${COMMAND}" ]]; then
     exit 1
 fi
 
-DOCKER_CMD="ln -sf ${PYTHON_LIB} /usr/bin/python && ${COMMAND}"
+DOCKER_CMD="tools/ci_testing/install_py36.sh && ln -sf ${PYTHON_LIB} /usr/bin/python && ${COMMAND}"
 echo "Docker image: ${DOCKER_IMAGE}"
 echo "Docker command: ${DOCKER_CMD}"
 docker run ${DOCKER_OPTS}                   \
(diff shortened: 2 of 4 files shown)


Squadrick:
We can get this merged as soon as the test suite passes.

tools/run_docker.sh
```diff
@@ -88,7 +88,7 @@ if [[ -z "${COMMAND}" ]]; then
     exit 1
 fi
 
-DOCKER_CMD="ln -sf ${PYTHON_LIB} /usr/bin/python && ${COMMAND}"
+DOCKER_CMD="tools/ci_testing/install_py36.sh && ln -sf ${PYTHON_LIB} /usr/bin/python && ${COMMAND}"
```
facaiy: How about moving the installation to 79 line ?

```bash
case ${PYTHON} in
    py2) xxx
    py3) install_py36.sh and set environment variable
```

tools/ci_testing/install_py36.sh
```diff
@@ -25,6 +25,7 @@ apt-get -y -qq update && apt-get -y -qq install python3.6
 
 python3.6 get-pip.py -q
 python3.6 -m pip --version
+rm get-pip.py
 
 ln -sfn /usr/bin/python3.6 /usr/bin/python3
 pip3 install scipy  # Pre-installed in custom-op
```
facaiy: Perhaps we could add scipy dependency in build_deps/requirements.txt and requirements_gpu.txt?

tools/run_docker.sh
facaiy: I misunderstand it.

tools/ci_testing/install_py36.sh
seanpmorgan: I'm open to it, but it's only required in one test script. Custom-op has it pre-installed (I assume because it's a dependency of TF -- though that may not be true for recent versions)

tools/run_docker.sh
seanpmorgan: Hmmm I believe that  would install py36 on host? We need it to run as a command in the docker container

seanpmorgan: Oh CMD is  `ln -sf ${PYTHON_LIB}` 
so if PYTHON_LIB = "/usr/bin/python3 && install_py36.sh" it would work

tools/ci_testing/install_py36.sh
facaiy: Thank you for clarification, Sean, I understand.

PR #389, opened 2019-08-02
BLD: Update environment to ubuntu16.04

### LINKED ISSUE #75: Package Addons for python3.7

seanpmorgan:
Preparing to close #75 and #119 now that the docker containers are available. At the moment the packaged whls are failing manylinux2010 compliance but it could be because we link to the tf-nightly which is not yet built in this env. The switch for nightlies on core should be within a couple weeks.

Also need to setup the cuda crosstool (though we haven't finalized our gpu pip package):
https://github.com/tensorflow/custom-op/commit/b74be358b804ee563a011cb255642ec17c6bbc1a

Another note: this will drop py34 support as tf-core has done the same.


tensorflow_addons/seq2seq/attention_wrapper_test.py
```diff
@@ -466,7 +466,8 @@ def _testWithMaybeMultiAttention(self,
                     expected_final_alignment_history,
                     final_alignment_history_info)
 
-    @parameterized.parameters([np.float32, np.float64])
+    # TODO: #407 Float64 test is failing
```
seanpmorgan: Related to #407

build_deps/requirements_gpu.txt
```diff
@@ -1 +1 @@
-tf-nightly-gpu-2.0-preview==2.0.0.dev20190731
+tf-nightly-gpu-2.0-preview>=2.0.0.dev20190802
```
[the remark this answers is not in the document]
seanpmorgan: So this is the date when tensorflow began publishing manylinux2010 whls. The greater or equal to will need to stay so I added a comment to the files. Thanks!

seanpmorgan:
@tensorflow/sig-addons-maintainers  Fixed the issues... ready for review.

build_pip_pkg.sh
```diff
@@ -44,6 +44,7 @@ function main() {
   cp ${PIP_FILE_PREFIX}setup.py "${TMPDIR}"
   cp ${PIP_FILE_PREFIX}MANIFEST.in "${TMPDIR}"
   cp ${PIP_FILE_PREFIX}LICENSE "${TMPDIR}"
+  touch ${TMPDIR}/stub.cc
```
WindQAQ: Sean, I don't find where we use this file. Could you point out the place or it is just for testing?

setup.py
```diff
@@ -73,6 +74,7 @@ def has_ext_modules(self):
     author='Google Inc.',
     author_email='opensource@google.com',
     packages=find_packages(),
+    ext_modules=[Extension('_foo', ['stub.cc'])],
```
WindQAQ: ditto

build_pip_pkg.sh
seanpmorgan: Yeah good find... without that sort-of hack the whl fails manylinux2010 compliance because:
```
RuntimeError: Invalid binary wheel, found the following shared library/libraries in purelib folder: ... The wheel has to be platlib compliant in order to be repaired by auditwheel.
```

This is the same patch as is done on TF-Core:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/devtoolset/platlib.patch

PR #392, opened 2019-08-02
Remove unnecessary tensor_shape module import

guillaumekln:
The function `as_shape` only creates a TensorShape if not already one.

### DIFF
--- a/tensorflow_addons/seq2seq/beam_search_decoder.py
+++ b/tensorflow_addons/seq2seq/beam_search_decoder.py
@@ -224,13 +224,13 @@ def _check_batch_beam(t, batch_size, beam_width):
     return tf.Assert(condition, [error_message])
 
 
-def _as_shape(shape):
+def _as_shape(value):
     """Converts the argument to a TensorShape if not already one."""
-    if not isinstance(shape, tf.TensorShape):
-        if isinstance(shape, tf.Tensor):
-            shape = tf.get_static_value(shape)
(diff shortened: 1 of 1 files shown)


tensorflow_addons/seq2seq/beam_search_decoder.py
```diff
@@ -225,6 +224,15 @@ def _check_batch_beam(t, batch_size, beam_width):
     return tf.Assert(condition, [error_message])
 
 
+def _as_shape(shape):
```
qlzh727: the input is not a strict shape, it can be a tensor as well, probably rename it to not be shape, which is bit confusing.

guillaumekln: Maybe something like `dims_or_shape`?

qlzh727: shouldn't it be shape_or_tensor?

guillaumekln: It could also be a Python list of dimensions.

qlzh727: Hum, since it is polymorphic, how about just call it input?

guillaumekln: Ok, I went with `value` as `input` clashes with a Python builtin.

PR #393, opened 2019-08-04
Sync custom-op

WindQAQ:
Prefix of cuda include path is changed in core Tensorflow so we need to sync build files in custom-op.
Need this PR for [beam_search_ops_gpu.cc](https://github.com/tensorflow/addons/blob/d57ffc2296a599542833da6c1c1482799ce811ad/tensorflow_addons/custom_ops/seq2seq/cc/kernels/beam_search_ops_gpu.cu.cc#L21) and [adjust_hsv_in_yiq_op_gpu.cu.cc](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/custom_ops/image/cc/kernels/adjust_hsv_in_yiq_op_gpu.cu.cc#L21). We could keep working on GPU kernel building after this :-)

Related links:
https://github.com/tensorflow/custom-op/commit/b267cc546365f30963ef2f6b690ff0a4e7391913#diff-0c845043da8b3200ae16b7a3e1781cc2
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/util/gpu_kernel_helper.h#L22

### DIFF
--- a/build_deps/gpu/cuda/BUILD.tpl
+++ b/build_deps/gpu/cuda/BUILD.tpl
@@ -1,3 +1,5 @@
+load(":build_defs.bzl", "cuda_header_library")
+
 licenses(["restricted"])  # MPL2, portions GPL v3, LGPL v3, BSD-like
 
 package(default_visibility = ["//visibility:public"])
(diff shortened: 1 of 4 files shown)


seanpmorgan:
Thanks Tzu-Wei! In this PR could you pin the requirement packages to 20190731? The manylinux2010 packages are causing failing builds for us (will work on in #389) The py2 test is passing because the docker's installed pip is probably too old to install manylinux2010 packages and so it installs 20190731

WindQAQ:
@seanpmorgan Thanks for pointing out that, Sean :P

PR #394, opened 2019-08-04
[image/distort_image_ops] build GPU kernel

WindQAQ:
Build GPU kernel


tensorflow_addons/custom_ops/image/cc/kernels/adjust_hsv_in_yiq_op_gpu.cu.cc
```diff
@@ -16,12 +16,21 @@ limitations under the License.
...
+namespace {
+
+template <typename T>
+inline se::DeviceMemory<T> AsDeviceMemory(const T* cuda_memory, uint64 size) {
+  se::DeviceMemoryBase wrapped(const_cast<T*>(cuda_memory), size * sizeof(T));
+  se::DeviceMemory<T> typed(wrapped);
+  return typed;
+}
+}  // namespace
+
```
WindQAQ: These are copied from https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/gpu_utils.h#L37 because `tensorflow/core/kernels/*` are not included in custom-op container. Please correct me if I'm wrong.

Issue #12, opened 2019-01-14
[Bug] skip_gram_sample assert rank is 1


seanpmorgan:
While converting the tests for skip_gram_ops, I noticed a test that was failing in TF2 eager execution, but I'm not sure why.

This [test](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/text/python/skip_gram_ops_test.py#L244):
```
with self.assertRaises(ValueError):
    invalid_tensor = constant_op.constant([[b"the"], [b"quick"], [b"brown"]])
    text.skip_gram_sample(invalid_tensor)
```

Should raise an error because of [this check in the op definition](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/text/cc/ops/skip_gram_ops.cc#L37). When decorating with `@test_util.run_deprecated_v1` the test correctly passes (error is raised), but otherwise it fails to raise an Error. 

Is this related to inspecting the rank of EagerTensors? Or is something else at play?

karmel:
@jaingaurav - Is this something you have seen? Can you take a look? (Note that this is for the Addons repo, which is migrating functionality currently in contrib.)

jaingaurav:
`@test_util.run_deprecated_v1` will cause the test to run in graph mode hence the check fires during graph construction. In Eager, you should see a [slightly different assertion](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/text/kernels/skip_gram_kernels.cc#L50) pop up due to a runtime check. This results in an `InvalidArgumentError` exception. The test need to be modified to handle both types of assertions. One approach would be to use something like `self.assertRaisesRegexp(Exception, <some string that matches both exceptions>)`

seanpmorgan:
@jaingaurav Thank you very much for looking into this. So if I remove [this comment and the assertion context](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/text/python/skip_gram_ops_test.py#L271)  then I still see the test as passing and no Error at all being raised.

Seems like I would need to add a new OP_REQUIRES in the Compute method that verifies tensor rank?

jaingaurav:
@seanpmorgan: Yes, that would be advisable to ensure better safety in eager mode.

facaiy:
@jaingaurav So, will shape inference be supported for eager mode later? Otherwise, duplicate codes exist in both graph constructing and kernel computing.

PR #395, opened 2019-08-05
[image/distort_image_ops] fix graph mode tests

### LINKED ISSUE #373: distort_image_ops_test.py fails on tf-nightly-2.0-gpu
The CI Build for PY2-GPU failed because of `distort_image_ops_test.py`.
https://source.cloud.google.com/results/invocations/7ed10b8d-6cba-47d9-b764-d823365528f4/log

Ideally, the manager should catch the exception raised in CC file
https://github.com/tensorflow/addons/blob/436580c5427af1b0666faf4955cd493f5ccace72/tensorflow_addons/image/distort_image_ops_test.py#L110-L112

https://github.com/tensorflow/addons/blob/436580c5427af1b0666faf4955cd493f5ccace72/tensorflow_addons/custom_ops/image/cc/kernels/adjust_hsv_in_yiq_op.cc#L70-L73

WindQAQ:
Closes #373. 
This is probably related to https://github.com/tensorflow/addons/issues/12 and https://github.com/tensorflow/tensorflow/issues/24945 (in the totally opposite side). I think the best solution now is to do shape inference in SetShapeFn while registering OP, which seems to successfully raise the same exception (ValueError) and the same error message in both graph and eager modes.

### DIFF
--- a/tensorflow_addons/image/distort_image_ops_test.py
+++ b/tensorflow_addons/image/distort_image_ops_test.py
@@ -23,11 +23,10 @@ import numpy as np
 import tensorflow as tf
 from tensorflow_addons.image import distort_image_ops
 
-# from tensorflow_addons.utils import test_utils
+from tensorflow_addons.utils import test_utils
 
 
-# TODO: #373 Get this to run in graph mode as well
-# @test_utils.run_all_in_graph_and_eager_modes
+@test_utils.run_all_in_graph_and_eager_modes
 class AdjustHueInYiqTest(tf.test.TestCase):
     def _adjust_hue_in_yiq_np(self, x_np, delta_h):
         """Rotate hue in YIQ space.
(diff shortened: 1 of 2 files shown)


WindQAQ:
The sanity check also failed due to timeout...

tensorflow_addons/image/distort_image_ops_test.py
```diff
@@ -101,15 +100,18 @@ def test_adjust_random_hue_in_yiq(self):
...
+        with self.assertRaisesRegex(ValueError, msg):
             self.evaluate(self._adjust_hue_in_yiq_tf(x_np, delta_h))
+
+    def test_invalid_channels(self):
+        msg = "Dimension must be 3 but is 4"
         x_np = np.random.rand(4, 2, 4) * 255.
         delta_h = np.random.rand() * 2.0 - 1.0
-        with self.assertRaisesOpError("input must have 3 channels "
-                                      "but instead has 4 channels"):
+        with self.assertRaisesRegex(ValueError, msg):
```
feihugis: I did not figure out why the check failure [here](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/custom_ops/image/cc/kernels/adjust_hsv_in_yiq_op.cc#L72) is not raised in the graph mode in the GPU version of tensorflow-2.0-preview, but it is raised in the CPU version.

WindQAQ: I also have no idea about it... I'd suppose it should either fail or pass regardless of the device.

Issue #396, opened 2019-08-06
FBetaScore is missing sample_weight parameter in update_state function


PhilipMay:
**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow): see https://gist.github.com/PhilipMay/8897cd3980b4367f22cee7a0f8cb2ee5#file-poc_f1_bug-ipynb
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): macOS Mojave
- TensorFlow installed from (source or binary): binary (pypi with pip)
- TensorFlow version (use command below): tensorflow  2.0.0b1
- TensorFlow Addons installed from (source, PyPi): pypi
- TensorFlow Addons version: tfa-nightly 0.5.0.dev20190804
- Python version and type (eg. Anaconda Python, Stock Python as in Mac, or homebrew installed Python etc): conda python 3.6.9
- Bazel version (if compiling from source): -
- GCC/Compiler version (if compiling from source): -
- Is GPU used? (yes/no): no
- GPU model (if used): -

**Describe the bug**
Using `F1Score`from metrics with tf.keras does not work. See exception here:
https://gist.github.com/PhilipMay/8897cd3980b4367f22cee7a0f8cb2ee5#file-poc_f1_bug-ipynb

**Describe the expected behavior**
Should calculate the f1 metric

**Code to reproduce the issue**
see: https://gist.github.com/PhilipMay/8897cd3980b4367f22cee7a0f8cb2ee5#file-poc_f1_bug-ipynb

**Solution**
FBetaScore is missing sample_weight parameter in update_state function:
https://github.com/tensorflow/addons/blob/dab18121fa4253cf05320d41962bfbf1803e1f1b/tensorflow_addons/metrics/f_scores.py#L178

This would be a workaround for the problem:
`def update_state(self, y_true, y_pred, sample_weight=None):`

Although this way sample_weight is not considered in metric calculation.

Please note that all `update_state`functions from `tf.keras.metrics` do contain the `sample_weight` parameter: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/metrics.py

If you want me to I can prepare a PR.

Same problem should be here:
- https://github.com/tensorflow/addons/blob/master/tensorflow_addons/metrics/multilabel_confusion_matrix.py#L100
- https://github.com/tensorflow/addons/blob/master/tensorflow_addons/metrics/r_square.py#L54

**Other info / logs**
```
$ pip list
Package              Version             
[...]

PR #402, opened 2019-08-08
[WIP] Adding sample weight to Multi-label Confusion matrix

SSaishruthi:
This PR related to #396


SSaishruthi:
@Squadrick 
Adding the same comment here as well just for reference.

I have created this PR to add sample weight parameter to F1. Right now, I did not modify the code to take sample weight into account. Just added the param so that it can be used when model training as per the issue. 

I am referring https://github.com/tensorflow/tensorflow/blob/7dcd936389a16028843c822315a8e3a0f7591653/tensorflow/python/keras/metrics.py#L113 for adding broadcasting weights. 

Would like to know your comments on further steps and if the above approach is ok to try.

Thanks in advance.

Squadrick:
Could you add a `TODO` so that we can get to that later? Also, having a test to check whether each `Metric` works with the Keras API would be very beneficial.

SSaishruthi:
Thanks will update the PR

Squadrick:
@SSaishruthi Yup. We need to check everything with the Keras API.

SSaishruthi:
@Squadrick I have the updated code ready but do you think the confusion matrix will be used as an evaluation metric when training (fit)? 
Usually, they are either used in the validation or test step.

SSaishruthi:
@Squadrick I have updated the code. But did not add keras test as I wanted to confirm my above question

SSaishruthi:
@Squadrick @seanpmorgan any update on this?

SSaishruthi:
@seanpmorgan Any update on my above question?

Squadrick:
@SSaishruthi Sorry about the delay. You're right, a confusion matrix is usually used during validation, and much less during training. However, if it's not too much extra work, we'd like to support both scenarios in case someone wants to use it during training.

gabrieldemarmiesse:
I'll close this pull request as it has been inactive for a while now. Feel free to comment if you want me to reopen the pull request.

PR #406, opened 2019-08-10
[image/sparse_image_warp] make tests compatible with graph and eager modes

WindQAQ:
Make tests compatible with graph and eager modes.

### DIFF
--- a/tensorflow_addons/image/interpolate_spline_test.py
+++ b/tensorflow_addons/image/interpolate_spline_test.py
@@ -21,7 +21,6 @@ import numpy as np
 from scipy import interpolate as sc_interpolate
 
 import tensorflow as tf
-import tensorflow.compat.v1 as tf1  # TODO: locate placeholder
 from tensorflow_addons.utils import test_utils
 from tensorflow_addons.image import interpolate_spline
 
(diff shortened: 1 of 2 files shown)


tensorflow_addons/image/interpolate_spline_test.py
```diff
@@ -237,92 +213,81 @@ def test_nd_linear_interpolation(self):
...
+        # Get concrete functions such that the batch size, number of train points,
+        # and number of query points are not known at graph construction time.
+        fn = tf.function(interpolate_spline).get_concrete_function(
+            tf.TensorSpec(
+                shape=[None, None, feature_dim], dtype=train_points.dtype),
+            tf.TensorSpec(
+                shape=[None, None, value_dim], dtype=train_values.dtype),
+            tf.TensorSpec(
+                shape=[None, None, feature_dim], dtype=query_points.dtype),
+            order, reg_weight)
```
WindQAQ: As interpolate_spline and sparse_image_warp are internal requests from Google Team, I am not sure if we should directly decorate them with `tf.function`. So I only decorate it when tests need it.

seanpmorgan: I think this is fine. @yhliang2018 may be able to comment if it's okay to decorate at all times.

PR #420, opened 2019-08-15
Gelu activation

AakashKumarNain:
Implement [GELU](https://arxiv.org/abs/1606.08415) activation heavily used in BERT models

### DIFF
--- a/tensorflow_addons/activations/__init__.py
+++ b/tensorflow_addons/activations/__init__.py
@@ -19,3 +19,4 @@ from __future__ import division
 from __future__ import print_function
 
 from tensorflow_addons.activations.sparsemax import sparsemax
+from tensorflow_addons.activations.gelu import gelu
--- a/tensorflow_addons/activations/gelu.py
+++ b/tensorflow_addons/activations/gelu.py
@@ -1,0 +1,43 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Implementing GELU activation."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import math
+import tensorflow as tf
+from tensorflow_addons.utils import keras_utils
+
+
+@tf.function
+@keras_utils.register_keras_custom_object
+def gelu(x):
+    """Gaussian Error Linear Unit.
+
+    A smoother version of ReLU generally used 
+    in the BERT or BERT architecture based models.
+    Original paper: https://arxiv.org/abs/1606.08415
+    
+    Args:
+        x: float Tensor to perform activation.
+    Returns:
+        `x` with the GELU activation applied.
+    """
+
+    x = tf.convert_to_tensor(x, dtype=tf.float32)
+    return 0.5 * x * (1 + tf.tanh(tf.sqrt(2 / math.pi) * \
+             (x + 0.044715 * tf.pow(x, 3))))
--- a/tensorflow_addons/activations/BUILD
+++ b/tensorflow_addons/activations/BUILD
@@ -26,3 +26,16 @@ py_test(
         ":activations",
     ],
 )
+
+py_test(
+    name = "gelu_test",
+    size = "medium",
+    srcs = [
+        "gelu_test.py",
+    ],
+    main = "gelu_test.py",
+    srcs_version = "PY2AND3",
+    deps = [
+        ":activations",
+    ],
+)
--- a/tensorflow_addons/activations/gelu_test.py
+++ b/tensorflow_addons/activations/gelu_test.py
@@ -1,0 +1,48 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Tests for GELU activation."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import numpy as np
+import tensorflow as tf
+from tensorflow_addons.activations import gelu
+from tensorflow_addons.utils import test_utils
+
+
+@test_utils.run_all_in_graph_and_eager_modes
+class GELUTest(tf.test.TestCase):
+    def random_test(self):
+        x = tf.constant([0.5, 1.2, -0.3], dtype=tf.float32)
+        val = tf.constant([0.345714, 1.0617027, -0.11462909], 
+                            dtype=tf.float32)
+        act = gelu(x)
+        self.assertAllClose(val, self.evaluate(act), atol=1e-5)
+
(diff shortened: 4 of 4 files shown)


tensorflow_addons/activations/gelu.py
```diff
@@ -0,0 +1,43 @@
...
+    in the BERT or BERT architecture based models.
+    Original paper: https://arxiv.org/abs/1606.08415
+    
+    Args:
+        x: float Tensor to perform activation.
+    Returns:
+        `x` with the GELU activation applied.
+    """
+
+    x = tf.convert_to_tensor(x, dtype=tf.float32)
```
WindQAQ: Do we really need to cast to float32? I mean maybe float16 and float64 are friendly to the following computation.

tensorflow_addons/activations/__init__.py
```diff
@@ -19,3 +19,4 @@
 from __future__ import print_function
 
 from tensorflow_addons.activations.sparsemax import sparsemax
+from tensorflow_addons.activations.gelu import gelu
```
WindQAQ: Please arrange them in alphabetical order :-)

AakashKumarNain: Ahh...thanks for pointing it out

tensorflow_addons/activations/gelu.py
AakashKumarNain: Most of the operations in TF happens in `float32`. `float64` is rarely used and for `fp16`, I need to check all the changes I have to make. If I am right, then all other floating operations have to be in `fp16` then.

AakashKumarNain:
> Thanks! I do think this is heavily used in NLP task now, but safely request for approval from Sean @seanpmorgan :-)
> 
> Also, I suppose this can also be a `tf.keras.layers.Layer` subclass like what relu does in core TF. If so, would you mind adding it into `tfa.layers` later (maybe another PR is much more suitable).

Yeah I can implement this as a layer as well, thought it might take a while.

AakashKumarNain:
@WindQAQ Can you please explain why this code throws this error:
```python
#@parameterized.parameters([np.float32, np.float64])
@test_utils.run_all_in_graph_and_eager_modes
class GELUTest(tf.test.TestCase):
    def random_test(self, dtype):
        x = np.array([0.5, 1.2, -0.3]).astype(dtype)
        val = np.array([0.345714, 1.0617027, -0.11462909]).astype(dtype)
     
        test_utils.layer_test(
            GeLU,
            kwargs={'dtype': dtype},
            input_data=x,
            expected_output=val)
```

```
ValueError: in converted code:

    <ipython-input-3-74907d021250>:23 call  *
        x = K.cast_to_floatx(inputs)
    /usr/local/lib/python3.6/dist-packages/tensorflow/python/keras/backend.py:174 cast_to_floatx
        return np.asarray(x, dtype=floatx())
    /usr/local/lib/python3.6/dist-packages/numpy/core/numeric.py:538 asarray
        return array(a, dtype, copy=False, order=order)

    ValueError: setting an array element with a sequence.
```

On the other hand, this works fine though:
```
K.cast_to_floatx(np.array([0.5, 1.2, -0.3]).astype(np.float32))
```

WindQAQ:
> @WindQAQ Can you please explain why this code throws this error:
> 
> ```python
> #@parameterized.parameters([np.float32, np.float64])
> @test_utils.run_all_in_graph_and_eager_modes
> class GELUTest(tf.test.TestCase):
>     def random_test(self, dtype):
>         x = np.array([0.5, 1.2, -0.3]).astype(dtype)
>         val = np.array([0.345714, 1.0617027, -0.11462909]).astype(dtype)
>      
>         test_utils.layer_test(
>             GeLU,
>             kwargs={'dtype': dtype},
>             input_data=x,
>             expected_output=val)
> ```
> 
> ```
> ValueError: in converted code:
> 
>     <ipython-input-3-74907d021250>:23 call  *
>         x = K.cast_to_floatx(inputs)
>     /usr/local/lib/python3.6/dist-packages/tensorflow/python/keras/backend.py:174 cast_to_floatx
>         return np.asarray(x, dtype=floatx())
>     /usr/local/lib/python3.6/dist-packages/numpy/core/numeric.py:538 asarray
>         return array(a, dtype, copy=False, order=order)
> 
>     ValueError: setting an array element with a sequence.
> ```
> 
> On the other hand, this works fine though:
> 
> ```
> K.cast_to_floatx(np.array([0.5, 1.2, -0.3]).astype(np.float32))
> ```

I cannot locate where the errors occur. Would you mind committing the updates so that I can get the full codes directly? Thanks.

AakashKumarNain:
I have changed this to a layer. Should I close this PR and open another one or should I make changes in this branch only and once finalized, then move to layer with a new PR?

WindQAQ:
I am fine with both of them. Just choose whatever you like :P

AakashKumarNain:
Okay. Instead of a new PR, let us fix this here only. Would be much easier for all of us. 
Here is the colab with full source code
https://colab.research.google.com/drive/1jPT9jrh1hROhM4Zic9OrUpKE_qiMSsO6

WindQAQ:
It's due to casting. `K.cast_to_floatx` takes numpy array as inputs.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/backend.py#L150-L175

So the modified notebook with float16, float32 and float64 supports is in the following.
https://colab.research.google.com/drive/1011MVfDbqa8zZ4rSJ5l7XP07NYF9rPq7

AakashKumarNain:
Yeap, my bad! Thanks for looking into this. I guess I will close this PR and will open a new one which would have GeLU as a layer similar to ReLU. This is almost done. Let me know what you think

AakashKumarNain:
@WindQAQ please check #424

tensorflow_addons/activations/gelu.py
```diff
@@ -0,0 +1,45 @@
...
+
+import math
+import tensorflow as tf
+from tensorflow.keras import backend as K
+from tensorflow_addons.utils import keras_utils
+
+
+@tf.function
+@keras_utils.register_keras_custom_object
+def GeLU(x, dtype='float32'):
```
WindQAQ: Could we follow the naming convention in core TF: say `activations.gelu` and `layers.GeLU`?

tensorflow_addons/activations/gelu.py
```diff
@@ -0,0 +1,45 @@
...
+    Args:
+        x: float Tensor to perform activation.
+    Returns:
+        `x` with the GELU activation applied.
+    """
+
+    x  = K.cast(x, dtype=dtype)
+    pi = K.cast(math.pi, dtype=dtype)
+    return 0.5 * x * (1 + tf.tanh(tf.sqrt(2 / pi) * \
+             (x + 0.044715 * tf.pow(x, 3))))
```
WindQAQ: Also, seems that `tf.keras.activations.*` do not have `dtype` argument.  I think for gelu, we could explicitly cast `pi` and `0.044715` to `x.dtype`:

So IMO, the better impl and API would be
```python
def gelu(x):
    pi = K.cast(math.pi, x.dtype)
    coeff = K.cast(0.044715, x.dtype)
    return 0.5 * x * (1 + tf.tanh(tf.sqrt(2 / pi) * (x + coeff * tf.pow(x, 3))))
```
What's your thought on this @AakashKumarNain :-)

AakashKumarNain:
@WindQAQ I was going to make those changes, that's why didn't say anything haha. Anyways, I have found a better way to deal with all these nuances. We will keep things as done in the core API. Just give me some time. Also, I didn't update the README because there was one thing that is still left to do. As you said that for `layer`, we should run a model test as done in the core API. There is one problem with that, I am trying to figure out to sort all these. Hopefully in the next commit

AakashKumarNain:
@WindQAQ I made the necessary changes in both PRs. On my end, every unit test passed. Can you please check now? Also, I will update the `README` as soon as everything runs fine on your end too

WindQAQ:
Great! Let's move on :-)

AakashKumarNain:
@WindQAQ did you run the tests?

seanpmorgan:
Good to close this PR given the quicker ops in #427 ? Keras layer PR + the fused op seems to me to be the right solution

AakashKumarNain:
@seanpmorgan will close this today after making changes in the file. Thanks

AakashKumarNain:
@seanpmorgan @WindQAQ I made the changes in code format. On my box, everything ran fine. It is weird that if I run `sanity-check` outside of the docker, then it doesn't complain about code format and passes all the test. Can you please run the checks once again now? Thank you

AakashKumarNain:
@WindQAQ the last commit should run fine now. I pulled the repo on another workstation after this commit and this time `code-formatter` passed. Formatting the code took 3x than coding the actual thing :face_with_thermometer: :man_facepalming:

PR #424, opened 2019-08-16
GeLU activation as a layer

AakashKumarNain:
I closed the PR that I opened it yesterday which proposed `gelu` as an activation function. This one is a much better implementation as the activation function here is represented as a keras layer


tensorflow_addons/layers/gelu_test.py
```diff
@@ -0,0 +1,58 @@
...
+from tensorflow.keras import backend as K
+from tensorflow_addons.layers.gelu import GeLU
+from tensorflow_addons.utils import test_utils
+from absl.testing import parameterized
+
+
+@parameterized.parameters([np.float16, np.float32, np.float64])
+@test_utils.run_all_in_graph_and_eager_modes
+class GELUTest(tf.test.TestCase):
+    def random_test(self, dtype):
```
WindQAQ: Seems that most of tests in addons use naming convention of `class TestGeLU`  and `def test_random`.  Also, there might not be any random part in the test. Should we rename it? Maybe `test_gelu` is enough.

tensorflow_addons/layers/gelu_test.py
```diff
@@ -0,0 +1,58 @@
...
+                        -4.9960107e-04], 
+                        dtype=dtype).reshape(3,1)
+     
+        test_utils.layer_test(
+            GeLU,
+            kwargs={'dtype': dtype},
+            input_data=x,
+            expected_output=val) 
+
+    def random_test_with_numpy(self, dtype):
```
WindQAQ: I think this one is can be combined with above test method?

tensorflow_addons/layers/gelu_test.py
AakashKumarNain: Yeah seems fair. I will make the necessary changes.

tensorflow_addons/layers/gelu_test.py
AakashKumarNain: Okay. Should we rather include a test with a dummy model, like in the link that you referenced?

AakashKumarNain:
@WindQAQ no worries. I have a very simple query. When I run `sanity-check` and `unit-test`, is it possible to make it run for a particular set of files only? For example, I just want it to test for `GeLU`, is there any simple way to do it? 
PS: Sorry noob question but I had to ask

WindQAQ:
unit-test can be done by running
```
bazel test -c opt -k \
--test_timeout 300,450,1200,3600 \
--test_output=all \
//tensorflow_addons/layers:gelu_test
```
(please type use to auto-complete the rule, say `//tensorflow_addons/xxxxxx`)

for sanity check, I think you refer to linter check right? If so, `pylint --rcfile=tools/ci_build/pylintrc --output-format=parseable [YOUR FILES]` would be useful.

AakashKumarNain:
I  don't have `bazel` installed locally. Can you please the equivalent `docker` command or something that I need to change in the files present in `tools` directory? Thanks

WindQAQ:
In my case, I always attach to the container to run tests. So I will first lanuch a container
```sh
docker run --rm -it -v ${PWD}:/addons -w /addons tensorflow/tensorflow:custom-op-ubuntu16 /bin/bash
pip install --upgrade pip # Pip must be upgraded to install manylinux2010 pkg
./configure.sh  # Links project with TensorFlow dependency
```
And run the commands I paste above.

mostafaelhoushi:
I was about to create a pull request for Gelu too, but rather as an op implemented as C++:
https://github.com/mostafaelhoushi/tensorflow-addons

Here are the main files I have added:
[gelu.cc](https://github.com/mostafaelhoushi/tensorflow-addons/blob/master/tensorflow_addons/custom_ops/text/cc/kernels/gelu.cc)
[gelu_op.cc](https://github.com/mostafaelhoushi/tensorflow-addons/blob/master/tensorflow_addons/custom_ops/text/cc/ops/gelu_op.cc)
[gelu.py](https://github.com/mostafaelhoushi/tensorflow-addons/blob/master/tensorflow_addons/text/gelu.py)
[gelu_test.py](https://github.com/mostafaelhoushi/tensorflow-addons/blob/master/tensorflow_addons/text/gelu_test.py)

I will wait for this pull request to be merged so that I can see how to rebase on top of it.

seanpmorgan:
> I was about to create a pull request for Gelu too, but rather as an op implemented as C++:
> https://github.com/mostafaelhoushi/tensorflow-addons
> 
> Here are the main files I have added:
> [gelu.cc](https://github.com/mostafaelhoushi/tensorflow-addons/blob/master/tensorflow_addons/custom_ops/text/cc/kernels/gelu.cc)
> [gelu_op.cc](https://github.com/mostafaelhoushi/tensorflow-addons/blob/master/tensorflow_addons/custom_ops/text/cc/ops/gelu_op.cc)
> [gelu.py](https://github.com/mostafaelhoushi/tensorflow-addons/blob/master/tensorflow_addons/text/gelu.py)
> [gelu_test.py](https://github.com/mostafaelhoushi/tensorflow-addons/blob/master/tensorflow_addons/text/gelu_test.py)
> 
> I will wait for this pull request to be merged so that I can see how to rebase on top of it.

@mostafaelhoushi looks like there is a PR #427 which implements the C++ op. Would you mind doing a review?

@AakashKumarNain Thanks very much! This looks great. Waiting for the activation to be merged first then we can review this and merge

AakashKumarNain:
@WindQAQ @seanpmorgan since #427 has been merged, should we close #420 and update this layer implementation to use #427 ?

tensorflow_addons/layers/gelu.py
```diff
@@ -0,0 +1,56 @@
...
+        super(GeLU, self).__init__(**kwargs)
+        self.supports_masking = True
+
+    def call(self, inputs):
+        return gelu(inputs, dtype=inputs.dtype)
+
+    def get_config(self):
+        config = {}
+        base_config = super(GeLU, self).get_config()
+        return dict(list(base_config.items()) + list(config.items()))
```
facaiy: Seems `config` is just an empty dict?

AakashKumarNain: That was intentional. We weren't sure which implementation we were going to call. Since #427 is merged now, I will update this whole thing in the evening

AakashKumarNain:
@facaiy @WindQAQ I have updated the code. All test cases passed on my end, so I am positive that this commit should be enough to close this PR

AakashKumarNain:
@seanpmorgan all checks passed! Wooo!! :tada:

tensorflow_addons/layers/gelu_test.py
```diff
@@ -0,0 +1,56 @@
...
+    def test_random(self, dtype):
+        x = np.array([[0.5, 1.2, -0.3]]).astype(dtype)
+        val = np.array([[0.345714, 1.0617027, -0.11462909]]).astype(dtype)
+        test_utils.layer_test(
+            GeLU, kwargs={'dtype': dtype}, input_data=x, expected_output=val)
+
+
+@keras_parameterized.run_all_keras_modes
+@keras_parameterized.run_with_all_model_types
+class TestGeLUModel(keras_parameterized.TestCase):
```
facaiy: Do we need this case? I'm afraid that `layer_test` above has tested both Sequential and Functional model.

facaiy:
Great, Aakash! @WindQAQ Tzu-Wei, could you take a look again?

AakashKumarNain:
> LGTM. Probably need only one test case (`TestGeLU`) with two test methods, say `test_gelu` and `test_layer_as_activation`. See https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/layers/advanced_activations_test.py for more details. Thanks!

@WindQAQ  @facaiy I wanted to put them in a single Test case but I found out that `keras_parameterized.run_all_keras_modes`, `keras_parameterized.run_with_all_model_types`, and `test_utils.run_all_in_graph_and_eager_modes` don't play along well. In short, `keras_parameterized` doesn't allow any other decorator. But as @facaiy pointed out that `layer_test` has already tested it in both modes, I think it is safe to remove it. What do you say?

WindQAQ:
@AakashKumarNain  Yep, then just follow Facai's suggestions :P

AakashKumarNain:
@facaiy @WindQAQ I made the necessary changes. I think we are good to go

AakashKumarNain:
Thank you @WindQAQ @facaiy @seanpmorgan @Squadrick

PR #427, opened 2019-08-19
[activations] fused gelu kernel

WindQAQ:
Fused GeLU kernel (forward and backward). On my local machine, achieve 3x and 9x speed up on forward and backward direction respectively compared with python implementation. Also support original impl stated in https://arxiv.org/pdf/1606.08415.pdf.

References:
https://on-demand.gputechconf.com/ai-conference-2019/T1-3_Minseok%20Lee_Adding%20custom%20CUDA%20C++%20Operations%20in%20Tensorflow%20for%20boosting%20BERT%20Inference.pdf
https://github.com/pytorch/pytorch/blob/master/caffe2/operators/gelu_op.cu


AakashKumarNain:
This looks good to me. We can use this with the layer implementation

WindQAQ:
@AakashKumarNain Thanks! Do you mind if I also put your name and email on README? It would be great if you could also participate in maintaining gelu activation function :-)

AakashKumarNain:
@WindQAQ No problem at all. I would love to! Thank you

tensorflow_addons/custom_ops/activations/cc/kernels/gelu_op.h
```diff
@@ -0,0 +1,145 @@
...
+  //           outputs (using either one yields the same result here).
+  // approximate: whether to enable approximation.
+  // backprops: gradients to backpropagate to the Gelu inputs.
+  void operator()(const Device& d, typename TTypes<T>::ConstTensor gradients,
+                  typename TTypes<T>::ConstTensor features, bool approximate,
+                  typename TTypes<T>::Tensor backprops) {
+    if (approximate) {
+      const T kAlpha = static_cast<T>(M_2_SQRTPI * M_SQRT1_2);
+      const T kBeta = kAlpha * static_cast<T>(0.044715) * static_cast<T>(3);
+      const auto y =
```
mostafaelhoushi: I see here that the output `y` is recalculated. It was already calculated in the forward pass.

How about having the output of the forward pass, to be passed as an additional input to the backward pass in order to save time recalculating? i.e., have `typename TTypes<T>::ConstTensor activations` passed to the function, and replace this line with:
```
const auto y = activations
```

tensorflow_addons/custom_ops/activations/cc/ops/gelu_op.cc
```diff
@@ -0,0 +1,37 @@
...
+REGISTER_OP("Gelu")
+    .Input("features: T")
+    .Output("activations: T")
+    .Attr("T: {half, float, double}")
+    .Attr("approximate: bool = true")
+    .SetShapeFn(shape_inference::UnchangedShape);
+
+REGISTER_OP("GeluGrad")
+    .Input("gradients: T")
+    .Input("features: T")
```
mostafaelhoushi: Referring to my previous comment, we can add:
```
.Input("activations: T")
```
so that we don't recalculate the output when calculating the gradient.

tensorflow_addons/activations/gelu_test.py
```diff
@@ -0,0 +1,102 @@
...
+    else:
+        return 0.5 * x * (
+            1.0 + tf.math.erf(x / tf.cast(tf.sqrt(2.0), x.dtype)))
+
+
+@test_utils.run_all_in_graph_and_eager_modes
+class TestGelu(tf.test.TestCase, parameterized.TestCase):
+    @parameterized.named_parameters(("float16", np.float16),
+                                    ("float32", np.float32),
+                                    ("float64", np.float64))
```
mostafaelhoushi: What about testing integer and quantization types?
For integer types, I believe that Gelu will simply behave as Relu.

WindQAQ: I do not really get your point. Like `int32`, do you mean we first cast it to `float` and do computations in `float` and finally cast back to `int32`? If so, it's weird why users don't cast `int32` to `float` first?

Actually, most of activation ops in core TF can support only floating points input. ReLU/ReLU6 is an exception because `cwiseMax`/`cwiseMin` can run in non-floating dtype.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/nn_ops.cc#L1053-L1144

tensorflow_addons/custom_ops/activations/cc/kernels/gelu_op.h
WindQAQ: I wonder if this causes excessive memory occupation. Such as BERT model, `gelu` is called multiple times in forward pass with multiple huge tensors of `[batch_size, sequence_length, depth]`. If we do the caching mechanism as you stated, the operation will store such large tensor in *each* forward node. I'm afraid it'll harm the model throughput because of OOM.

Or if you have already had some benchmarking results?

tensorflow_addons/activations/gelu_test.py
WindQAQ: BTW, after a rough computing with google calculator, I found there are some gap between ReLU and GeLU with int type. When input=2, approximate version shows the result of 1.95459769409 and non-approximate version shows the one of 1.9544997361. Get deeper into the definition of GeLU:
```python
gelu(x) = x * P(X <= x) = x * normcdf(x)
``` 
When `x=2`, gelu(2) = 2 * normcdf(2) = 2 * 0.9772 != 2.

[Approximate](https://www.google.com/search?ei=GgBdXbSjPIbW0gSaxaPoAQ&q=0.5+*+2.0+*+%281+%2B+tanh%28sqrt%282+%2F+pi%29+*+%282.0+%2B+0.044715+*+2.0%5E3%29%29%29&oq=0.5+*+2.0+*+%281+%2B+tanh%28sqrt%282+%2F+pi%29+*+%282.0+%2B+0.044715+*+2.0%5E3%29%29%29&gs_l=psy-ab.3...1226.1638..2012...0.0..0.63.118.2......0....1..gws-wiz.GZzLvi8SKuQ&ved=0ahUKEwi0mO-WxZPkAhUGq5QKHZriCB0Q4dUDCAo&uact=5)
[non-approximate](https://www.google.com/search?q=0.5+*+2.0+*+(1.0+%2B+erf(2.0+%2F+sqrt(2.0)))&oq=0.5+*+2.0+*+(1.0+%2B+erf(2.0+%2F+sqrt(2.0)))&aqs=chrome..69i57j6.2966j0j7&sourceid=chrome&ie=UTF-8)

AakashKumarNain: @mostafaelhoushi `For integer types, I believe that Gelu will simply behave as Relu.` I don't think that this is true. Can you elaborate a bit on this?

mostafaelhoushi: Thanks @WindQAQ and @AakashKumarNain for your feedback.
I meant if both the input and _output_ are constrained to be integer, then Gelu will behave as Relu. e.g., 

for the example that @WindQAQ mentioned: `gelu(2) ~= 2*0.9772 - 1.9554`
but if the activations are constrained to be input then we need to round the output to the nearest integer... so `round(gelu(2)) = 2 = relu(2)`

However, @WindQAQ mentioned an important point that  "most of activation ops in core TF (and PyTorch) can support only floating points input. ReLU/ReLU6". Hence, I think you may safely ignore this suggestion.

mostafaelhoushi: Thanks @WindQAQ and @AakashKumarNain for your feedback.
I meant if both the input and _output_ are constrained to be integer, then Gelu will behave as Relu. e.g., 

for the example that @WindQAQ mentioned: `gelu(2) ~= 2*0.9772 - 1.9554`
but if the activations are constrained to be input then we need to round the output to the nearest integer... so `round(gelu(2)) = 2 = relu(2)`

However, @WindQAQ mentioned an important point that  "most of activation ops in core TF (and PyTorch) can support only floating points input. ReLU/ReLU6". Hence, I think you may safely ignore this suggestion.

WindQAQ: okay! Thanks again for the review :-)

tensorflow_addons/custom_ops/activations/cc/kernels/gelu_op.h
WindQAQ: Hello, seems that
`y = tanh(sqrt(2 / pi) * (x + 0.044715 * x^3)))`, and
`gelu(x) = 0.5 * x * (1 + tanh(sqrt(2 / pi) * (x + 0.044715 * x^3)))`

Though we can compute `y` via `y = gelu(x) / (0.5 * x) - 1`, this will indeed lose some precision during division. So I tend to preserve the current implementation. What's your thoughts on this @mostafaelhoushi and @AakashKumarNain .

AakashKumarNain: The current implementation looks good to me.

AakashKumarNain:
@WindQAQ I think we should inject this in #424 now. What do you say?

WindQAQ:
@tensorflow/sig-addons-maintainers mind approving this one? Thanks!

tensorflow_addons/custom_ops/activations/cc/kernels/gelu_op.h
WindQAQ: So I think further changes could be done in the future. Let's move on to Layer's subclass! Thanks for you guys' feedback again.

PR #434, opened 2019-08-20
Revert back to manylinux1 package which breaks spec

seanpmorgan:
This patch will make it so we publish spec breaking (but at least working) manylinux1 packages.

The pinned installs match the previous auditwheel: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/install/install_auditwheel.sh

I tested the new pip package against a TF manylinux2010 package and it works.

We can revert this after https://github.com/tensorflow/tensorflow/issues/31807 is solved


seanpmorgan:
Closing as Yong Tang may have a fix.

PR #435, opened 2019-08-21
Patch manylinux2010 auditwheel repair so libtensorflow is not copied

### LINKED ISSUE #119: Make whls manylinux2010 compatible

seanpmorgan:
Copying @yongtang 's patch https://github.com/tensorflow/io/commit/02dcf4a . Pure wizardry :smile: 

Closes #119

### DIFF
--- a/tools/ci_build/builds/wheel_verify.sh
+++ b/tools/ci_build/builds/wheel_verify.sh
@@ -19,9 +19,8 @@ set -e
 if [[ $(uname) == "Darwin" ]]; then
     CMD="delocate-wheel -w wheelhouse"
 else
-    LD_PATH="$(cat .bazelrc | grep TF_SHARED_LIBRARY_DIR | sed 's/"//g' | awk -F'=' '{print $2}')"
-    export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$LD_PATH
-    CMD="auditwheel repair --plat manylinux2010_x86_64"
+    pip3 install auditwheel==2.0.0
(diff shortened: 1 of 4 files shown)


tools/ci_build/builds/wheel_verify.sh
```diff
@@ -19,9 +19,8 @@ set -e
 if [[ $(uname) == "Darwin" ]]; then
     CMD="delocate-wheel -w wheelhouse"
 else
-    LD_PATH="$(cat .bazelrc | grep TF_SHARED_LIBRARY_DIR | sed 's/"//g' | awk -F'=' '{print $2}')"
-    export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$LD_PATH
-    CMD="auditwheel repair --plat manylinux2010_x86_64"
+    pip3 install auditwheel==2.0.0
```
facaiy: Can we use pip here?

seanpmorgan: So auditwheel==1.5.0 (too old for manylinux2010) is by default installed on py36. I've added a new commit to specifically update the py36 version.

Also the patch will only get called once now instead of each time auditwheel is ran.

PR #440, opened 2019-08-23
Add mAP Callback

fsx950223:


Squadrick:
Can this be implemented as a metric instead of a callback?

fsx950223:
I try it

PR #458, opened 2019-08-29
bugfix: force WeightNormalization to execute in order

facaiy:
It seems that autograph is not smart enough, so we have to control execute order by ourselves.

### DIFF
--- a/tensorflow_addons/layers/wrappers.py
+++ b/tensorflow_addons/layers/wrappers.py
@@ -93,8 +93,9 @@ class WeightNormalization(tf.keras.layers.Wrapper):
 
         if self.data_init:
             # Used for data initialization in self._data_dep_init.
-            self._naked_clone_layer = tf.keras.layers.deserialize(
-                tf.keras.layers.serialize(self.layer))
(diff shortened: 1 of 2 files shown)


facaiy:
Following #453

facaiy:
The new implementation still performs slightly better than the regression model (but worser than the precious implementation). I don't know what's wrong

https://colab.research.google.com/drive/1yXMehUQdFQ0mFtgJjYaj6FBCp8fr4Dt8#scrollTo=r66OUUrK_l_S

facaiy:
I run it for five times, sometimes better, sometimes worser...

https://colab.research.google.com/drive/1yXMehUQdFQ0mFtgJjYaj6FBCp8fr4Dt8#scrollTo=UZd6V90eR4Gm

seanpmorgan:
@facaiy I think it was for the wrong reason that it always outperformed using the previous  version (at least through one epoch). Could you try to increase the number of epochs a significant amount? Also mind making the colab sharable?

facaiy:
sharable link: https://colab.research.google.com/drive/1yXMehUQdFQ0mFtgJjYaj6FBCp8fr4Dt8

PR #459, opened 2019-08-29
Build correlation cost GPU kernel

### LINKED ISSUE #447: Correlation cost layer is not placed on gpu
**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow): yes
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Ubuntu 18.04.3
- TensorFlow installed from (source or binary): PyPi
- TensorFlow version (use command below): tf-nightly-gpu-2.0-preview 2.0.0.dev20190824   
- TensorFlow Addons installed from (source, PyPi): PyPi (also tested with source)
- TensorFlow Addons version: tfa-nightly 0.5.0.dev20190824 (/master)
- Python version and type (eg. Anaconda Python, Stock Python as in Mac, or homebrew installed Python etc): system
- Bazel version (if compiling from source): (0.25.1)
- GCC/Compiler version (if compiling from source): (gcc/7.4.0)
- Is GPU used? (yes/no): yes
- GPU model (if used): Nvidia GTX 1080Ti

**Describe the bug**
The correlation cost layer gets executed on CPU instead of on GPU.

**Describe the expected behavior**
The correlation layer should be run on the GPU.

```
import tensorflow as tf
import numpy as np
import time

from tensorflow.python.eager import context
context.context().log_device_placement = True


im = np.random.rand(1, 500, 500, 3)

model = tf.keras.Sequential([
    tfa.layers.CorrelationCost(
        kernel_size=1, max_displacement=4,
        stride_1=1, stride_2=1,
        pad=4, data_format="channels_last")])

start = time.time()
corr_volume = model([im, im])
print(f"computing correlation took {str(time.time() - start)} sec")
```

Output:
```
...
input_a: (_Arg): /job:localhost/replica:0/task:0/device:GPU:0
input_b: (_Arg): /job:localhost/replica:0/task:0/device:GPU:0
correlation_cost/CorrelationCost: (CorrelationCost): /job:localhost/replica:0/task:0/device:CPU:0
correlation_cost/transpose: (Transpose): /job:localhost/replica:0/task:0/device:GPU:0
Identity: (Identity): /job:localhost/replica:0/task:0/device:GPU:0
identity_RetVal: (_Retval): /job:localhost/replica:0/task:0/device:GPU:0
correlation_cost/transpose/perm: (Const): /job:localhost/replica:0/task:0/device:GPU:0
...
computing correlation took 14.041364908218384 sec
```

**Other info / logs**
Replacing the correlation cost with a Conv2D shows, that the GPU is correctly utilized for other layers.

WindQAQ:
- Fix #447 
- Build GPU kernel for correlation cost
- Rename cub prefix from gpu to third_party, which is the same with [core TF](https://github.com/tensorflow/tensorflow/blob/master/third_party/cub.BUILD).

image

Will omit the following errors, but still pass unittests. Probably due to autograph? Related to https://github.com/tensorflow/tensorflow/issues/32018.
```
2019-08-29 08:01:48.033866: E tensorflow/compiler/jit/compilability_check_util.cc:346] unsupported op node : {name:'correlation_cost_1/correlation_cost/CorrelationCost' id:5 op device:{/job:localhost/replica:0/task:0/device:GPU:0} def:{{{node correlation_cost_1/correlation_cost/CorrelationCost}} = CorrelationCost[T=DT_FLOAT, data_format="NHWC", kernel_size=1, max_displacement=2, pad=4, stride_1=1, stride_2=2, _device="/job:localhost/replica:0/task:0/device:GPU:0"](input_3, input_4)}}
```

### DIFF
--- a/build_deps/gpu/cub.BUILD
+++ b/build_deps/gpu/cub.BUILD
@@ -18,7 +18,7 @@ filegroup(
 cc_library(
     name = "cub",
     hdrs = if_cuda([":cub_header_files"]),
-    include_prefix = "gpu",
+    include_prefix = "third_party",
     deps = [
         "@local_config_cuda//cuda:cuda_headers",
     ],
--- a/build_deps/gpu/cub.BUILD
+++ b/build_deps/gpu/cub.BUILD
@@ -18,7 +18,7 @@ filegroup(
 cc_library(
     name = "cub",
     hdrs = if_cuda([":cub_header_files"]),
-    include_prefix = "third_party",
+    include_prefix = "gpu",
     deps = [
         "@local_config_cuda//cuda:cuda_headers",
     ],
--- a/tensorflow_addons/custom_ops/layers/cc/kernels/correlation_cost_op.cc
+++ b/tensorflow_addons/custom_ops/layers/cc/kernels/correlation_cost_op.cc
@@ -331,7 +331,7 @@ TF_CALL_float(REGISTER_CORRELATIONCOST_OP_CPU);
 #undef REGISTER_CORRELATIONCOST_OP_CPU
 
 // Register the GPU kernels.
-#if GOOGLE_CUDA
+#ifdef GOOGLE_CUDA
 
 #define REGISTER_CORRELATIONCOST_OP_GPU(T)                                   \
   REGISTER_KERNEL_BUILDER(                                                   \
--- a/tensorflow_addons/custom_ops/layers/cc/kernels/correlation_cost_op.cc
+++ b/tensorflow_addons/custom_ops/layers/cc/kernels/correlation_cost_op.cc
@@ -331,7 +331,7 @@ TF_CALL_float(REGISTER_CORRELATIONCOST_OP_CPU);
 #undef REGISTER_CORRELATIONCOST_OP_CPU
 
 // Register the GPU kernels.
-#ifdef GOOGLE_CUDA
+#if GOOGLE_CUDA
 
 #define REGISTER_CORRELATIONCOST_OP_GPU(T)                                   \
   REGISTER_KERNEL_BUILDER(                                                   \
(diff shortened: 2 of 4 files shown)


tensorflow_addons/custom_ops/layers/cc/kernels/correlation_cost_op.cc
```diff
@@ -331,7 +331,7 @@ TF_CALL_float(REGISTER_CORRELATIONCOST_OP_CPU);
 #undef REGISTER_CORRELATIONCOST_OP_CPU
 
 // Register the GPU kernels.
-#ifdef GOOGLE_CUDA
+#if GOOGLE_CUDA
```
facaiy: Seems `#ifdef` looks better here?

build_deps/gpu/cub.BUILD
```diff
@@ -18,7 +18,7 @@ filegroup(
 cc_library(
     name = "cub",
     hdrs = if_cuda([":cub_header_files"]),
-    include_prefix = "gpu",
+    include_prefix = "third_party",
```
facaiy: Because the file is placed at `tensorflow/third_party/cub.BUILD`, I guess that's why its prefix is `third_party`?

WindQAQ: Ohoh, got it. I thought TF wants to unify header prefix previously. Will revert this real quick :-)

seanpmorgan:
Thanks good catch!

tensorflow_addons/custom_ops/layers/cc/kernels/correlation_cost_op.cc
facaiy: @WindQAQ Tzu-Wei, my fault, we should use #if bacause `if_cuda(["-DGOOGLE_CUDA=1"])` is passed

PR #463, opened 2019-08-30
BLD: increase minor version after 0.5.0 release

facaiy:
we need to increase version for tfa-nightly


facaiy:
Thank you, Kyle :-)

PR #466, opened 2019-08-30
[losses] migrate npairs multilabel loss

### LINKED ISSUE #26: Implement Metric Losses

WindQAQ:
Closes #26. Finally the last one :-)

- Migrate `npairs_loss_multilabel` from contrib
- Rename `npairs_loss_multilabel` to `npairs_multilabel_loss`, which makes sense to addons' naming style
- Modify input spec from
    - `labels` (list of sparse tensor)
    - `embedding_anchor`
    - `embedding_posititve`

    to
    - `y_true` (either tensor or sparse tensor)
    - `y_pred` (similarity matrix, which is the multiplication of anchor and positive)

### DIFF
--- a/tensorflow_addons/losses/README.md
+++ b/tensorflow_addons/losses/README.md
@@ -17,6 +17,7 @@
 | focal_loss | SigmoidFocalCrossEntropy | https://arxiv.org/abs/1708.02002  |
 | lifted | LiftedStructLoss | https://arxiv.org/abs/1511.06452       |
 | npairs | NpairsLoss | http://www.nec-labs.com/uploads/images/Department-Images/MediaAnalytics/papers/nips16_npairmetriclearning.pdf |
+| npairs | NpairsMultilabelLoss | http://www.nec-labs.com/uploads/images/Department-Images/MediaAnalytics/papers/nips16_npairmetriclearning.pdf |
 | sparsemax_loss | SparsemaxLoss |  https://arxiv.org/abs/1602.02068 |
 | triplet | TripletSemiHardLoss | https://arxiv.org/abs/1503.03832       |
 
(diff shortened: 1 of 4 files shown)


facaiy:
@Squadrick Hi, Dheeraj. I think the PR looks good. Do you have time to double check it? Thanks :-)

WindQAQ:
> Looks good, Tzu-wei!
> 
> By the way, can we provide a helper method which calculates `similarity_matrix`? It might be user-friendly for tf 1.x users. What do you think?

Seems good, but it will be a one line function like `tf.matmul(a, b)`. Previously, we inherit from `LossFunctionWrapper`, and that's why we have input signature `y_true` and `y_pred` only, resulting in computation of similarity matrix outside the loss function. But now, we inherit from `Loss`, I wonder if we could use signature `label`, `embedding_anchor` and `embedding_positive`, which makes much sense to me in the field of metric learning. What do you feel about this?

facaiy:
@WindQAQ Oh, you're right, forget it

Squadrick:
Maybe add a test to check that the new loss is compatible with the Keras Model API? It would be great if we add it under `utils` and every loss has to use the check. What do y'all think?

facaiy:
Like `layer_test` or https://github.com/tensorflow/tensorflow/blob/0197a2d8a3070af763cb67227835ee63df095e6d/tensorflow/python/keras/losses_test.py#L68 ? I have no objection. How about filing an issue and finishing it in next PR?

Issue #454, opened 2019-08-28
Use package manager to pin TF dependency


seanpmorgan:
Per https://github.com/tensorflow/community/pull/135 we should be using pip/conda to specify which Addons releases work with which TF versions. 

This is blocked until the `tensorflow` and `tensorflow-gpu` consolidation scheduled for 2.1 release

guillaumekln:
What is the reason to wait for the package consolidation? It seems possible to depend on `tensorflow-gpu` today.

seanpmorgan:
Good question.. one reason would be that TF does not package `tensorflow-gpu` for MacOS:
https://pypi.org/project/tensorflow-gpu/2.0.0rc0/#files

guillaumekln:
Ah, it makes sense. A possible workaround could be to express the [platform dependencies](https://www.python.org/dev/peps/pep-0508/#environment-markers) in `setup.py`, e.g.:

```python
install_requires=[
    "tensorflow-gpu;platform_system!='Darwin'",
    "tensorflow;platform_system=='Darwin'",
]
```

Not sure it is something we want though...

kyleabeauchamp:
I recall always having issues when using logic in the setup script.  Might be possible to do something with the new PEP518 approaches (e.g., https://www.python.org/dev/peps/pep-0518/).  Not 100% sure.

guillaumekln:
TensorFlow Text is using a less restrictive TensorFlow version requirement:

> `'tensorflow>=2.0.0b1, <2.1'`

https://github.com/tensorflow/text/blob/v1.0.0-beta2/oss_scripts/pip_package/setup.py#L76

Should we do something similar?

seanpmorgan:
> TensorFlow Text is using a less restrictive TensorFlow version requirement:
> 
> > `'tensorflow>=2.0.0b1, <2.1'`
> 
> https://github.com/tensorflow/text/blob/v1.0.0-beta2/oss_scripts/pip_package/setup.py#L76
> 
> Should we do something similar?

In general I think we should support a greater-than versioning if we know that there were no breaks between releases (Might have to adapt our travis tests).

For this specific case I believe 2.0.0b1 will break with TFA 0.6 as Op Namespacing wasn't implemented at that point (~90% sure of that, but not positive)

facaiy:
[...]

PR #474, opened 2019-09-03
BLD: add tf 2.0 dependency for addons

facaiy:
Related with #454

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,6 +40,16 @@ from setuptools import Extension
 
 DOCLINES = __doc__.split('\n')
 
+TFA_NIGHTLY = 'tfa-nightly'
+TFA_RELEASE = 'tensorflow-addons'
+
+if '--nightly' in sys.argv:
+    project_name = TFA_NIGHTLY
+    nightly_idx = sys.argv.index('--nightly')
+    sys.argv.pop(nightly_idx)
+else:
+    project_name = TFA_RELEASE
+
 # Version
 version = {}
 base_dir = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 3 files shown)


seanpmorgan:
Thanks @facaiy this is a nice cleanup. Only thing I want to be positive is that `tensorflow-gpu` does not crash on a CPU only box... I believe that is true but haven't tested. Has anyone confirmed this?

We should probably cherry pick and publish 0.5.1 with this as well.

setup.py
```diff
@@ -39,24 +40,44 @@
...
-    project_name = 'tfa-nightly'
-    nightly_idx = sys.argv.index('--nightly')
-    sys.argv.pop(nightly_idx)
-    version['__version__'] += datetime.strftime(datetime.today(), "%Y%m%d")
-else:
-    project_name = 'tensorflow-addons'
+if project_name == TFA_RELEASE:
+    # TODO: remove if-else condition when tf supports package consolidation.
+    if platform.system() == 'Linux':
+        REQUIRED_PACKAGES.append('tensorflow-gpu >= 2.0.0-rc0')
```
guillaumekln: Is the version requirement `>=` correct? For example if a user froze `tensorflow_addons` to a version built against TensorFlow 2.0 and then update to TensorFlow 2.1, will it work correctly?

guillaumekln:
> Only thing I want to be positive is that tensorflow-gpu does not crash on a CPU only box... I believe that is true but haven't tested. Has anyone confirmed this?

Yes, it works since TensorFlow 1.14.

setup.py
seanpmorgan: Even though backwards compatible API guarantees should be there, there may new features we're using so I think this is a good catch.

Should be `==`

facaiy: Thanks, done :-)

facaiy:
> We should probably cherry pick and publish 0.5.1 with this as well.

Totally agree, will do later :-)

PR #478, opened 2019-09-04
Improve issue templates

seanpmorgan:
Improvements based on the information we actually require. Happy to make some changes to this if there are other thoughts.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-performance-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-performance-report.md
@@ -5,23 +5,17 @@ about: Use this template for reporting a bug or a performance issue.
 ---
 
 **System information**
-- Have I written custom code (as opposed to using a stock example script provided in TensorFlow):
 - OS Platform and Distribution (e.g., Linux Ubuntu 16.04):
 - TensorFlow installed from (source or binary):
-- TensorFlow version (use command below):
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug-performance-report.md
```diff
@@ -5,24 +5,18 @@ about: Use this template for reporting a bug or a performance issue.
 ---
 
 **System information**
-- Have I written custom code (as opposed to using a stock example script provided in TensorFlow):
 - OS Platform and Distribution (e.g., Linux Ubuntu 16.04):
 - TensorFlow installed from (source or binary):
-- TensorFlow version (use command below):
-- TensorFlow Addons installed from (source, PyPi):
+- TensorFlow Addons installed from (source or binary):
```
facaiy: move one line up? We might want to put `TensorFlow installed from` and `TensorFlow version` together.

seanpmorgan:
> LGTM. The python type ("e.g., Anaconda") might be useful in some cases but it's not critical for all reports.

Agree, if we find it to be a common requirement we can add it back

.github/ISSUE_TEMPLATE/bug-performance-report.md
seanpmorgan: How do you feel about the change?

PR #479, opened 2019-09-04
CLN: reorganize ci_testing

seanpmorgan:


facaiy:
It seems that both check_futures_test.py and file_name_test.py only are invoked by ci_sanity.sh. How about putting those files in the `ci_build`? Maybe we can create a subdirectory for them if necessary, say `ci_build/check`, `ci_build/verify` or something else.

And we can delete `_test` suffix which looks kind of misleading, what do you think, Sean?

seanpmorgan:
> It seems that both check_futures_test.py and file_name_test.py only are invoked by ci_sanity.sh. How about putting those files in the `ci_build`? Maybe we can create a subdirectory for them if necessary, say `ci_build/check`, `ci_build/verify` or something else.
> 
> And we can delete `_test` suffix which looks kind of misleading, what do you think, Sean?

I like that. Updated.

tools/ci_build/ci_sanity.sh
```diff
@@ -227,13 +227,13 @@ do_bazel_nobuild() {
...
-    python check_futures_test.py
+    cd "$ROOT_DIR/tools/ci_build/verify"
+    python check_futures.py
 }
 
 do_check_file_name_test() {
-    cd "$ROOT_DIR/tools/test"
-    python file_name_test.py
+    cd "$ROOT_DIR/tools/ci_build/verify"
+    python file_name.py
```
facaiy: nit: check_file_name.py ?

WindQAQ:
Got  the error `AttributeError: 'module' object has no attribute 'keras' ` for CI tests. Probably an issue from upstream?

PR #480, opened 2019-09-04
BLD: add tf 2.0 dependency for addons (#474)

facaiy:
cherry-pick for 0.5.1

* BLD: add tf dependency for addons
* DOC: remove note in readme
* BLD: support tfa-nightly
* BLD: pin to 2.0.0-rc0


facaiy:
> Still need to update the minor version I believe

Yes, will do in the next PR

PR #499, opened 2019-09-11
Adding threshold parameter to f1

### LINKED ISSUE #490: Bug in F1 and FBeta implementation and test
**Describe the bug**
The tests for FBeta and F1 both use a softmax function with just one output. 
- https://github.com/tensorflow/addons/blob/76249547c481ed47bd6f2f8092a99cd64d37fd2f/tensorflow_addons/metrics/fbeta_test.py#L122
- https://github.com/tensorflow/addons/blob/76249547c481ed47bd6f2f8092a99cd64d37fd2f/tensorflow_addons/metrics/f1_test.py#L116

The effect is that the output (prediction) is always 1. The right activation function for a binary calssification (with just one output is sigmoid). But this is not the only problem. This bug in the test hides an even worse other bug.

The other bug is that F1 and FBeta both can not handle values other then 1 (1.0) and 0 (0.0) as the predicted result. But the predictions of a binary classification (with simoid) and multi class classification with softmax is always somewhere between 1 and 0 (one hot encoded for multi class). The result of this bug is that this implementation of F1 and FBeta always return 0.0 when the predicted results are not exactly 0 (0.0) or 1 (1.0) - which is not realistic.

Also all other tests of F1 and FBeta have values of 0 or 1 as the predicted results. This does not reflect reality.
- https://github.com/tensorflow/addons/blob/76249547c481ed47bd6f2f8092a99cd64d37fd2f/tensorflow_addons/metrics/fbeta_test.py#L91
- https://github.com/tensorflow/addons/blob/76249547c481ed47bd6f2f8092a99cd64d37fd2f/tensorflow_addons/metrics/f1_test.py#L88

Just change softmax to sigmoid in the tests and change `verbose=1` and see the results.

SSaishruthi:
This PR closes #490 

Changes:

1. Threshold parameter has been added to both F1 and F-Beta score.
2. Using the threshold parameter output is changed to 0 and 1.

### DIFF
--- a/tensorflow_addons/metrics/f1_test.py
+++ b/tensorflow_addons/metrics/f1_test.py
@@ -28,7 +28,7 @@ import numpy as np
 @test_utils.run_all_in_graph_and_eager_modes
 class F1ScoreTest(tf.test.TestCase):
     def test_config(self):
-        f1_obj = F1Score(num_classes=3, average=None)
+        f1_obj = F1Score(num_classes=3, threshold=0.75, average=None)
         self.assertEqual(f1_obj.name, 'f1_score')
         self.assertEqual(f1_obj.dtype, tf.float32)
         self.assertEqual(f1_obj.num_classes, 3)
@@ -42,7 +42,7 @@ class F1ScoreTest(tf.test.TestCase):
 
     def initialize_vars(self, average):
         # initialize variables
-        f1_obj = F1Score(num_classes=3, average=average)
+        f1_obj = F1Score(num_classes=3, threshold=0.75, average=average)
 
         self.evaluate(tf.compat.v1.variables_initializer(f1_obj.variables))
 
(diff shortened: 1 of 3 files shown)


PhilipMay:
@SSaishruthi please see my comment here: https://github.com/tensorflow/addons/issues/490#issuecomment-530288447

SSaishruthi:
Left a comment in the issue.

SSaishruthi:
@PhilipMay 

I am taking binary accuracy as a reference here: https://github.com/tensorflow/tensorflow/blob/2ff39d00faf8f7e433ddcae0aa278f6e573b0c55/tensorflow/python/keras/metrics.py#L630

In order for this to work you need to provide the threshold value as 0.49 above. I can probably fix the threshold default to be 0.5. I think below 0.5, it may not be considered as a good prediction.


I have this tested with scikit learn as well.https://colab.research.google.com/drive/1qSq0SsYkPqjdKUgM1RM4kKM67X75ocFj

Squadrick:
I'm closing this issue, I've created a new PR #502 that has `threshold` as well as some refactoring to f_scores and f_tests.

PR #502, opened 2019-09-11
CLN: Refactor f_scores and f_test

### LINKED ISSUE #490: Bug in F1 and FBeta implementation and test

Squadrick:
* Add `threshold` param to f-scores
* Tests now compare with sklearn
* Add sklearn to requirements

Fixes #490


SSaishruthi:
Hi @Squadrick 
I think I have provided the same threshold solution. Any reason for not accepting that?

Squadrick:
It's extended to include a case when `threshold=None` that it defaults to using the max. The reason for closing your PR was more to do with the fact that this includes a pretty major refactor as well as the `threshold` functionality.

SSaishruthi:
Sounds good. 

Reason I asked is I had the same `threshold` functionality (except `None`), test cases and was waiting for review to make sure if we are ok with the fix before adding further tests/changes. Was confused when I saw this one.

Thanks for the modifications.

Squadrick:
Hey, sorry about that, my bad. Should've checked before opening this PR.

build_deps/requirements.txt
```diff
@@ -1,2 +1,3 @@
 # TensorFlow greater than this date is manylinux2010 compliant
 tf-nightly-2.0-preview>=2.0.0.dev20190802
+sklearn
```
seanpmorgan: At first this was concerning, but the requirements files are only used in our builds/tests and are not used as requirements in `setup.py` so I'm fine with this.

seanpmorgan:
Also ping @PhilipMay for review

tensorflow_addons/metrics/f_scores.py
```diff
@@ -12,220 +12,154 @@
...
+        self.false_negatives = _zero_wt_init('false_negatives')
+        self.weights_intermediate = _zero_wt_init('weights_intermediate')
+
     # TODO: Add sample_weight support, currently it is
     # ignored during calculations.
     def update_state(self, y_true, y_pred, sample_weight=None):
+        y_pred = tf.cast(y_pred, tf.float32)
+
+        if self.threshold is None:
+            threshold = tf.reduce_max(y_pred, axis=-1, keepdims=True)
```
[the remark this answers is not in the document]
Squadrick: `tf.math.reduce_max` is an alias for `tf.reduce_max`, they both do the exact same thing.

Squadrick:
@PhilipMay I test for `F-1` score implicitly by testing for `F-Beta` with `beta=1.0`. But binary classification with `num_classes=1` is still misleading. Like @SSaishruthi mentioned, `tf.shape` can be used to infer `num_classes`/`label_length` from `y_true` and `y_pred`, but initialization of the weights (for tracking false positives, etc.) in happen `update_states` instead.

Squadrick:
I've added a very simple `F1Test` to check that it the same as `FBetaScore(beta=1.0)`, and since we test `FBeta` pretty extensively, imo, it should be fine.

SSaishruthi:
@Squadrick I will start with sample weight addition after this PR gets merged

PhilipMay:
> 
> 
> I've added a very simple `F1Test` to check that it the same as `FBetaScore(beta=1.0)`, and since we test `FBeta` pretty extensively, imo, it should be fine.

Yes. Thanks. That's what I mean. A small "smoke test".

tensorflow_addons/metrics/f_scores.py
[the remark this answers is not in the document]
PhilipMay: Ok.

tensorflow_addons/metrics/f_scores.py
```diff
@@ -237,111 +171,69 @@ def get_config(self):
...
+
+        weighted: Metrics are computed for each class
+            and returns the mean weighted by the
+            number of true instances in each class.
     """
 
-    def __init__(self, num_classes, average, name='f1_score',
+    def __init__(self,
+                 num_classes,
+                 average,
```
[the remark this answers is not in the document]
PhilipMay: Ok now.

PhilipMay:
LGTM - the bug seems to be fixed now.

A different thought: On tf.keras.metrics there are already many implemented basic metrics. Confusion matrix, precision, recall and so on. Wouldn't it be a good idear to build on them?

SSaishruthi:
Both works in the same way. We change calculation according to the type. This seems to the better way after investigation.

Squadrick:
> Both works in the same way. We change calculation according to the type. This seems to the better way after investigation.

I don't quite understand. What are the alternatives that work the same way?

Squadrick:
> @Squadrick I will start with sample weight addition after this PR gets merged

Like I mentioned above, I think a better approach would be to use `MeanMetricWrapper` and let it handle the `sample_weight`. I currently don't see an easy way of adding sample weight functionality to F-scores.

facaiy:
Hi, thank everyone:-) I haven't did a full review yet

1. For implimentation:
> @PhilipMay  A different thought: On tf.keras.metrics there are already many implemented basic metrics. Confusion matrix, precision, recall and so on.
> @Squadrick  a better approach would be to use MeanMetricWrapper and let it handle the sample_weight. 

+1, I'm wondering if we can refer to [AUC](
https://github.com/tensorflow/tensorflow/blob/848227c8a0d354d79b801b3e0e1b342d0c0cd65b/tensorflow/python/keras/metrics.py#L1588) metric.

2. 
> Tests now compare with sklearn
> Add sklearn to requirements

I prefer to keep dependency minimal, and F1Score is not so complex that we cannot calculate it easily. Maybe we can learn test cases in https://github.com/tensorflow/addons/pull/466/

3. 
> In case of "single-label categorial classification" 
> actuals = np.array([[0], [1], [1], [1]])

If I'm not wrong, tf.keras use one-hot encoding for labels `y_true` and logits for `y_pred` by default. We can clarify the requirement in the document (like AUC metric). If we really care about it, please refer to subclass solution for accuracy: Accuracy, BinaryAccuracy, CategoricalAccuracy, SparseCategoricalAccuracy, etc. 

What do you think, Philip, Saishruthi, Dheeraj? Thank all for your contribution

tensorflow_addons/metrics/f_test.py
```diff
@@ -0,0 +1,128 @@
...
+
+        actuals = np.random.randint(2, size=(10, 4))
+        preds = np.random.uniform(size=(10, 4))
+
+        self.evaluate(fbeta.update_state(actuals, preds))
+        self.evaluate(f1.update_state(actuals, preds))
+        self.assertAllClose(
+            self.evaluate(fbeta.result()), self.evaluate(f1.result()))
+
+    def test_keras_model(self):
```
facaiy: It seems that _get_model is designed to test model.evaluate: https://github.com/tensorflow/tensorflow/blob/e832dcc31c3c68b86c9c4ab5ebbb78cfdd28dcee/tensorflow/python/keras/metrics_test.py#L1977 Can we keep consistency with tf-core/keras?

tensorflow_addons/metrics/f_test.py
```diff
@@ -0,0 +1,128 @@
...
+        self._test_fbeta_score(actuals, preds, None)
+
+    def test_keras_model(self):
+        fbeta = FBetaScore(5, 'micro', 1.0)
+        utils.test_keras_model(fbeta, 5)
+
+
+@test_utils.run_all_in_graph_and_eager_modes
+class F1ScoreTest(tf.test.TestCase):
+    def test_eq(self):
```
facaiy: How about checking result (instead of implementation) correct directly? In most cases, we might want to avoid using random inputs for reproducibility.

As we override `get_config` method, perhaps we should add a test case for it.

facaiy:
@WindQAQ Tzu-Wei, could you take a look if time allows? Thank you 😄

facaiy:
By the way, @seanpmorgan, as F1 Score is migrating to tf/core in https://github.com/tensorflow/tensorflow/pull/31818, not sure if we need to deprecate the metric.

PhilipMay:
> If I'm not wrong, tf.keras use one-hot encoding for labels `y_true` and logits for `y_pred` by default. 

When using Keras we use 1 or 0 to encode the classes for binary calssification and one-hot for more then one class. Binary case has one output with simoid activation and multi class has softmax.

PhilipMay:
My suggestion is to add a test for `get_config` and then accept this PR since it fixes the bug. Then we can open up a new issue to look for a more elegant way to implement F1.

PhilipMay:
> Generally looks good to me. But I'm not sure if we should use sklearn for testing. IMO, hard-coding the numeric results might be a better choice.

I agree- @Squadrick can you please hardcode the sklearn part so we can complete the PR and fix thw bug?

SSaishruthi:
Is this PR ready to merge?

PhilipMay:
@SSaishruthi the bug is fixed but @WindQAQ suggested an improvement to the tests. Unfortunately @Squadrick seems to be busy and does not do the change.

If you want you can accept this PR. I will improve the test in a 2nd PR afterwards.

Squadrick:
Sorry, guys. Got a little busy with some other work, I'll make the changes soon.

tensorflow_addons/metrics/f_scores.py
```diff
@@ -12,220 +12,156 @@
...
+        if self.threshold is None:
+            threshold = tf.reduce_max(y_pred, axis=-1, keepdims=True)
+            # make sure [0, 0, 0] doesn't become [1, 1, 1]
+            # Use (x - 0 > eps) to check for fp zero equality
+            y_pred = tf.logical_and(y_pred >= threshold, y_pred - 0 > 1e-12)
+        else:
+            y_pred = y_pred > self.threshold
+
         y_true = tf.cast(y_true, tf.int32)
         y_pred = tf.cast(y_pred, tf.int32)
```
PhilipMay: This line is redundant to line number 124 where the same is executed.

Squadrick:
I'd prefer using `sklearn` as the ground truth rather than hard-coding the values for two reasons:
1. Adopting `sklearn` across tests for `tfa.metrics` will guarantee that any param style we borrow from `sklearn` will behave as expected. Example: Type of `average` in F-scores.
2. Specifically, in this case, rewriting to include hard-coded values makes the code much longer with a lot of duplication, since we need to test different type of `average`s.

I'm open to hard-coding the results if you think that's the better approach. What do you all think?

PhilipMay:
@Squadrick for me both ways are good. Both have pro and cons.

facaiy:
I'm afraid that sklearn is too heavy, what do you think @seanpmorgan @WindQAQ ?

PhilipMay:
@facaiy "too heavy" sounds very abstract for me. Can you explain what you think is the concrete disadvantage? Download needs too much time, installing docker image for testing needs too much time? What is it that makes you think "too heavy"?

facaiy:
@PhilipMay Hi, Philip, it's easy to add a new dependency while difficult (sometimes impossible) to remove one, that's why I suggest to act conservatively. Moreover, sklearn is a quite complicated python wheel which has many dependencies on its own.

PhilipMay:
Would it be a solution to split into test and install dependencies? See here: https://stackoverflow.com/questions/15422527/best-practices-how-do-you-list-required-dependencies-in-your-setup-py

facaiy:
Sorry for the delay, Philip. I'm referring to test dependencies when I use 'dependency' above. Anyway, I'm not against the sklearn proposal if you insist :-) What do you think, Dheeraj @Squadrick ?

WindQAQ:
> I'm afraid that sklearn is too heavy, what do you think @seanpmorgan @WindQAQ ?

Agree +1. As a plugin/addons package, it would be great if we could make the wheel lightweight. So in this case, if we could do unittests even without `sklearn`, I would think this dependency is not a must.

PhilipMay:
Ok. So let’s do this without Sklearn. For me finishing this PR has priority anyway.

seanpmorgan:
> Ok. So let’s do this without Sklearn. For me finishing this PR has priority anyway.

Agree we've had a similar discussion before... both options have pros and cons, though we have precedent throughout the repo of using pre-calucated values.

Squadrick:
Sorry about the delay, hardcoded the tests. 

@PhilipMay @SSaishruthi @seanpmorgan

PhilipMay:
@Squadrick thanks for finalizing this. :-)

Issue #375, opened 2019-07-27
seq2seq.SequenceLoss is probably incompatible with tf.keras.Model.compile


kazemnejad:
**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow): yes
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Google Colab
- TensorFlow installed from (source or binary): binary
- TensorFlow version (use command below): 2.0.0.b1
- TensorFlow Addons installed from (source, PyPi): PyPi
- TensorFlow Addons version: 0.4.0
- Python version and type (eg. Anaconda Python, Stock Python as in Mac, or homebrew installed Python etc): Stock Python
- Bazel version (if compiling from source): - 
- GCC/Compiler version (if compiling from source): - 
- Is the GPU used? (yes/no): Yes

**Describe the bug**

I think there are two incompatibilities when we use `tfa.seq2seq.SequenceLoss` and `tf.keras.Model` at the same time. In the following example, I replicated a common scenario in building a Seq2seq model. i. e.  The input is a 2D int32 tensor containing word ids, and the output/logit is a 3D tensor which represents distributions over our vocabulary in each time step.
```python
import tensorflow as tf

model = tf.keras.Sequential([
    tf.keras.layers.Embedding(vocab_size, 50, mask_zero=True),
    tf.keras.layers.Dense(vocab_size, activation='softmax')
])
model.summary()
```
Now, Let's compile the model and use `tfa.seq2seq.SequenceLoss()` as its loss function:
```python
import tensorflow_addons as tfa
loss_obj = tfa.seq2seq.SequenceLoss()
model.compile(optimizer='adam', loss=loss_obj)
```
```
/usr/local/lib/python3.6/dist-packages/tensorflow/python/framework/tensor_util.py in make_tensor_proto(values, dtype, shape, verify_shape, allow_broadcast)
    453   else:
    454     if values is None:
--> 455       raise ValueError("None values not supported.")
    456     # if dtype is provided, forces numpy array to be the type
    457     # provided if possible.

ValueError: None values not supported.
```
[...]

PR #503, opened 2019-09-11
SequenceLoss keras compatibility

kazemnejad:
Fixing the incompatibility issue between the SequenceLoss and Keras built-in training loops by adding support for dense targets. For more information please refer to #375


tensorflow_addons/seq2seq/loss.py
```diff
@@ -168,6 +182,11 @@ def __init__(self,
         self.sum_over_batch = sum_over_batch
         self.softmax_loss_function = softmax_loss_function
 
+        # Delete the reduction attribute to inform Keras that it
+        # should call this class by the __call__(...) method.
+        if 'reduction' in dir(self):
```
guillaumekln: `hasattr` seems more appropriate here?

kazemnejad: Ops, I forget that. Thanks for the comment. it's now fixed in the next commit.

guillaumekln:
@qlzh727 Can you take a quick look when you have time?

Issue #422, opened 2019-08-16
Implement RAdam optimizer


seanpmorgan:
**System information**
- Are you willing to contribute it (yes/no): Yes, but would like to see if someone else will
- Are you willing to maintain it going forward? (yes/no): Yes, but would like to see if someone else will

**Describe the feature and the current behavior/state.**
New paper describing RAdam looks like a drop in replacement for Adam optimizer with better results.

https://arxiv.org/abs/1908.03265v1
https://github.com/LiyuanLucasLiu/RAdam

yassineAlouini:
That would be awesome indeed.
For the record, here is a Keras implementation (not official): https://github.com/CyberZHG/keras-radam.

rafiqhasan:
This is going to be a great addon ! For any one who is looking out for another detailed comparison can read through - https://medium.com/@lessw/new-state-of-the-art-ai-optimizer-rectified-adam-radam-5d854730807b

talipini:
I vote for this as well. Looking for this!

SSaishruthi:
Great addition. 
Can I try implementation of this?

seanpmorgan:
@SSaishruthi Sure! I know @sayoojbk has also shared interest in helping with this so if you could make a WIP PR as soon as you get it started that'd be great so we could have a few eyes on this and push it through.

physicso:
Looking forward to that!

WindQAQ:
Seems that there is an unofficial implementation for TF/Keras.
https://github.com/CyberZHG/keras-radam

physicso:
I have found one for TF:

physicso:
I have found one for TF: https://github.com/taki0112/RAdam-Tensorflow

SSaishruthi:
Thanks for the links. I have marked all the links and planning to kick start with the implementation in 2 days. Will keep posted.

luminoso:
@SSaishruthi it looks like RAdam has already an improvement called 'RAdam+lookahead'. One possible implementation of lookahead:
https://github.com/bojone/keras_lookahead

It's called 'Ranger' (the combination of RAdam + lookahead):
Small article about it: https://medium.com/@lessw/new-deep-learning-optimizer-ranger-synergistic-combination-of-radam-lookahead-for-the-best-of-2dc83f79a48d

Paper: https://arxiv.org/abs/1907.08610v1
[...]

PR #506, opened 2019-09-13
Add RAdam and LookAhead optimizers

CyberZHG:
Related to #422:

* The name of the optimizer is `RectifiedAdam` in case of conflicting with other abbreviations. (The name appears in the first line of the second page in the paper)
* Add `Lookahead` optimizer wrapper.
* Unit tests for RAdam are compared with the results of the [official implementation](https://github.com/LiyuanLucasLiu/RAdam).
* Unit tests for Ranger (RAdam + Lookahead) are compared with the results of the [implementation](https://github.com/lessw2020/Ranger-Deep-Learning-Optimizer) from the proposer.

### DIFF
--- a/tensorflow_addons/optimizers/lookahead.py
+++ b/tensorflow_addons/optimizers/lookahead.py
@@ -59,12 +59,12 @@ class Lookahead(tf.keras.optimizers.Optimizer):
                 The ratio for updating the slow weights.
             name: Optional name for the operations created when applying
                 gradients. Defaults to "RectifiedAdam".
-            **kwargs: keyword arguments. Allowed to be {`clipnorm`, `clipvalue`,
-                `lr`, `decay`}. `clipnorm` is clip gradients by norm;
-                `clipvalue` is clip gradients by value, `decay` is included for
-                backward compatibility to allow time inverse decay of learning
-                rate. `lr` is included for backward compatibility, recommended
-                to use `learning_rate` instead.
+            **kwargs: keyword arguments. Allowed to be {`clipnorm`,
+                `clipvalue`, `lr`, `decay`}. `clipnorm` is clip gradients
+                by norm; `clipvalue` is clip gradients by value, `decay` is
+                included for backward compatibility to allow time inverse
+                decay of learning rate. `lr` is included for backward
+                compatibility, recommended to use `learning_rate` instead.
         """
         super(Lookahead, self).__init__(name, **kwargs)
 
@@ -114,16 +114,20 @@ class Lookahead(tf.keras.optimizers.Optimizer):
         step_back = slow_var + slow_step_size * (var - slow_var)
         sync_cond = tf.equal(local_step % sync_period, 0)
         with tf.control_dependencies([step_back]):
-            slow_update = slow_var.assign(tf.where(
-                sync_cond,
-                step_back,
-                slow_var,
-            ), use_locking=self._use_locking)
-            var_update = var.assign(tf.where(
-                sync_cond,
-                step_back,
-                var,
-            ), use_locking=self._use_locking)
+            slow_update = slow_var.assign(
+                tf.where(
+                    sync_cond,
+                    step_back,
+                    slow_var,
+                ),
+                use_locking=self._use_locking)
+            var_update = var.assign(
+                tf.where(
+                    sync_cond,
+                    step_back,
+                    var,
+                ),
+                use_locking=self._use_locking)
         return tf.group(slow_update, var_update)
 
     @property
--- a/tensorflow_addons/optimizers/lookahead.py
+++ b/tensorflow_addons/optimizers/lookahead.py
@@ -88,7 +88,7 @@ class Lookahead(tf.keras.optimizers.Optimizer):
     def apply_gradients(self, grads_and_vars, name=None):
         var_list = [v for (_, v) in grads_and_vars]
 
-        with tf.keras.backend.name_scope(self._scope_ctx):
+        with tf.keras.backend.name_scope(self._name):
             with ops.init_scope():
                 _ = self.iterations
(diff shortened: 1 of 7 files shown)


Squadrick:
I think a more useful use of `LookAhead` would be to make a generic wrapper that can take a `tf.keras.optimizers.Optimizer` and apply lookahead. (Similar to [`tfa.optimizers.MovingAverage`](https://github.com/tensorflow/addons/blob/4022d3ac392a34e540c0f856fdec591a9eadc7d3/tensorflow_addons/optimizers/moving_average.py#L44)). 

So the final API for using `Ranger` would look like this:
```python
radam = tfa.optimizers.RectifiedAdam(lr=1e-3)
ranger = tfa.optimizers.LookAhead(radam, step=4, ratio=0.5)
```
This will allow for other optimizers to be used with `LookAhead`.

CyberZHG:
I've added a standalone Lookahead wrapper.

tensorflow_addons/optimizers/rectified_adam.py
```diff
@@ -0,0 +1,308 @@
...
+# limitations under the License.
+# ==============================================================================
+"""Rectified Adam (RAdam) optimizer."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow as tf
+from tensorflow.python import ops
+from tensorflow.python.ops import (math_ops, state_ops,
```
Squadrick: Look at my comments regarding private/public API in `lookahead.py`

CyberZHG:
It seems that the `mod` operation is not working properly on GPU. I've changed the implementation to use `floordiv` in `Lookahead`. I don't know if there will be any numerical stability issue with the modification.

tensorflow_addons/optimizers/lookahead.py
```diff
@@ -0,0 +1,171 @@
...
+    def __init__(self,
+                 optimizer,
+                 sync_period=6,
+                 slow_step_size=0.5,
+                 name="Lookahead",
+                 **kwargs):
+        r"""Wrap optimizer with the lookahead mechanism.
+
+        Args:
+            optimizer: A Tensor or a floating point value.
```
WindQAQ: Probably wrong docstring here.

seanpmorgan:
> It seems that the `mod` operation is not working properly on GPU. I've changed the implementation to use `floordiv` in `Lookahead`. I don't know if there will be any numerical stability issue with the modification.

I'm not sure why TF2 isn't using a soft device placement for this. There is no GPU implementation for the overloaded `__mod__`:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/cwise_op_gpu_mod.cu.cc#L22

But I would have assumed it'd fall back to CPU with dynamic kernels implemented. cc @karmel @martinwicke

tensorflow_addons/optimizers/lookahead.py
```diff
@@ -0,0 +1,171 @@
...
+
+        Args:
+            optimizer: The original optimizer that will be used to compute
+                and apply the gradients.
+            sync_period: An integer. The synchronization period of lookahead.
+                Enable lookahead mechanism by setting it with a positive value.
+            slow_step_size: A floating point value.
+                The ratio for updating the slow weights.
+            name: Optional name for the operations created when applying
+                gradients. Defaults to "RectifiedAdam".
```
Squadrick: Rename `"RectifiedAdam"` -> `"Lookahead"`

tensorflow_addons/optimizers/lookahead.py
```diff
@@ -108,11 +108,13 @@ def _init_op(self, var):
     def _look_ahead_op(self, var):
         var_dtype = var.dtype.base_dtype
         slow_var = self.get_slot(var, 'slow')
-        local_step = tf.cast(self.iterations + 1, var_dtype)
-        sync_period = self._get_hyper('sync_period', local_step.dtype)
+        local_step = tf.cast(self.iterations + 1, tf.dtypes.int64)
+        sync_period = self._get_hyper('sync_period', tf.dtypes.int64)
         slow_step_size = self._get_hyper('slow_step_size', var_dtype)
         step_back = slow_var + slow_step_size * (var - slow_var)
-        sync_cond = tf.equal(local_step % sync_period, 0)
```
facaiy: > I would have assumed it'd fall back to CPU with dynamic kernels implemented.

I believe so. Have you ever tried `tf.math.floormod`?

CyberZHG: I've tried this one and it's not working.

CyberZHG: w/blob/d30ba8443710e94d6a1ab1c71a2c61c437e24790/tensorflow/core/kernels/cwise_op_gpu_floor_mod.cu.cc#L22

facaiy: Do you use tf.device to force all ops place on GPU? Can you give a minimal reproducing example? Thanks

CyberZHG: The outputs are the same as the results of this [CI](https://source.cloud.google.com/results/invocations/d11ec6bf-ede3-40cf-9c10-3112fe54ba12/targets). It fails the [fit_simple_linear_model](https://github.com/CyberZHG/addons/blob/b4cea8189724f88663637daa1df82d035d97cba4/tensorflow_addons/optimizers/lookahead_test.py#L110-L128) test.

facaiy: Thanks for the information, really useful! Could you file an issue for it and we might resolve it later?

Squadrick:
LGTM. @CyberZHG Thank you so much for this contribution, I'm sure the community will find the optimizers very useful.

PR #510, opened 2019-09-15
FIX: MovingAverage from_config, get_config

Squadrick:
* `get_config` correctly serializes the wrapped optimizer using `tf.keras.optimizers.serialize()`

* `from_config` to deserialize the config correctly, including the wrapped optimizer

* Wrapper optimizer can be specified using keras optimizer strings like: adam, rmsprop, adagrad, etc.

* Add tests for above new functionality

* Minor refactoring in `moving_average.py`

### DIFF
--- a/tensorflow_addons/optimizers/moving_average.py
+++ b/tensorflow_addons/optimizers/moving_average.py
@@ -50,6 +50,9 @@ class MovingAverage(tf.keras.optimizers.Optimizer):
 
         super(MovingAverage, self).__init__(name, **kwargs)
 
+        if isinstance(optimizer, str):
+            optimizer = tf.keras.optimizers.get(optimizer)
+
         if not isinstance(optimizer, tf.keras.optimizers.Optimizer):
             raise TypeError(
                 "optimizer is not an object of tf.keras.optimizers.Optimizer")
@@ -60,24 +63,23 @@ class MovingAverage(tf.keras.optimizers.Optimizer):
         if not isinstance(sequential_update, bool):
             raise TypeError("sequential_update must be of bool type")
 
-        self._optimizer = optimizer
-
         with tf.name_scope(name):
             self._ema = tf.train.ExponentialMovingAverage(
                 average_decay, num_updates=num_updates)
 
+        self._optimizer = optimizer
         self._set_hyper("average_decay", average_decay)
         self._num_updates = num_updates
         self._sequential_update = sequential_update
-        self._init = True
+        self._initialized = False
 
     def apply_gradients(self, grads_and_vars, name=None):
         var_list = [v for (_, v) in grads_and_vars]
 
-        if tf.executing_eagerly() and self._init:
+        if tf.executing_eagerly() and not self._initialized:
             # this to ensure that var_list is registered initially
             self._ema.apply(var_list)
-            self._init = False
+            self._initialized = True
 
         train_op = self._optimizer.apply_gradients(grads_and_vars, name=name)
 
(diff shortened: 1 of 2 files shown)


tensorflow_addons/optimizers/moving_average.py
```diff
@@ -50,6 +50,9 @@ def __init__(self,
 
         super(MovingAverage, self).__init__(name, **kwargs)
 
+        if isinstance(optimizer, str):
```
WindQAQ: Should we really need to add this condition? Maybe we could let [tf.keras.optimizer.get](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/optimizers.py#L820-L850) check this implicitly.

tensorflow_addons/optimizers/moving_average_test.py
```diff
@@ -107,25 +107,30 @@ def test_model_weights_update(self):
...
+        _ = MovingAverage('adam')
+
     def test_config(self):
         sgd_opt = tf.keras.optimizers.SGD(
             lr=2.0, nesterov=True, momentum=0.3, decay=0.1)
         opt = MovingAverage(
             sgd_opt,
             average_decay=0.5,
-            num_updates=100,
+            num_updates=None,
```
WindQAQ: Any reason why we change this from 100 to `None` for testing?

tensorflow_addons/optimizers/moving_average.py
Squadrick: If the user passes an invalid arg to `optimizer`, the error message is less obvious:
```
ValueError('Could not interpret optimizer identifier:', identifier)
```
So I decided to do a manual check with more obvious error messages. What do you think?

tensorflow_addons/optimizers/moving_average_test.py
Squadrick: `num_updates` isn't registered as a hyperparameter using `self._set_hyper()` specifically, because `num_updates` could take the value of `None` which was not compatible with `self._set_hyper()` which only supports numeric values IIRC. Wanted to make sure that passing a `None` to the `config` wouldn't create an issue.

tensorflow_addons/optimizers/moving_average.py
```diff
@@ -47,9 +47,26 @@ def __init__(self,
...
+                to compute and apply gradients.
+            average_decay: float. Decay to use to maintain the moving averages
+                of trained variables. See tf.train.ExponentialMovingAverage
+                for details.
+            num_updates: Optional count of the number of updates applied to
+                variables. See tf.train.ExponentialMovingAverage for details.
+            sequential_update: Bool. If False, will compute the moving average
+                at the same time as the model is updated, potentially doing
+                benign data races. If True, will update the moving average
+                after gradient updates.
```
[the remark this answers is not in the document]
WindQAQ: Also for tf.* and type, I think we could add `backticks` for better documentation rendering

tensorflow_addons/optimizers/moving_average_test.py
```diff
@@ -107,25 +107,30 @@ def test_model_weights_update(self):
         self.evaluate(mean_update)
         self.assertAllClose(model.variables[0].read_value(), [[0.9]])
 
+    def test_optimizer_string(self):
+        _ = MovingAverage('adam')
```
facaiy: @Squadrick  Hi, Dheeraj, do we forget to check the result in the case?

saberkun:
How does this optimizer affect checkpointing?
If we store the model, will it take a moving average value?
In TF1, both original value and average value are stored and we need to apply EMA when we load the model for inference. How does it work here?

tensorflow_addons/optimizers/moving_average_test.py
Squadrick: @facaiy No, just needed to ensure that creating the optimizer with a string doesn't throw any errors.

Squadrick:
@saberkun It's doesn't save or checkpoint the moving average of the weights. The user must call `opt.assign_average_vars(model.variables)` before saving the model.

When the model is loaded, it only stores a single set of weights, either the original or the moving average based on whether the weights were assigned before saving to disk. 

This won't work well with using `model.fit`, but it should be fine if you're using a custom training loop.

I'll create an issue to save the moving average weights when saving the `MovingAverage` optimizer. This will be closer to TF1.x behavior.

Issue #226, opened 2019-05-01
Link API docs with with tensorflow.org


seanpmorgan:
Given the discussion in this comment: https://github.com/tensorflow/addons/pull/140#discussion_r276059512

And the API docs generated for release 0.3:
https://github.com/tensorflow/addons/tree/r0.3/docs/api_docs/python

I believe we're almost ready to link our docs with the tensorflow.org website? [Looks like we need some landing page files and a few configs?](https://github.com/tensorflow/probability/tree/master/tensorflow_probability/g3doc)

Is there information somewhere about what all needs to be done? @lamberta @Markdaoust

lamberta:
Hi Sean,
Glad the API gen is working! I have a WIP internal config that will import the API docs to publish on tensorflow.org, but we still need to figure out how to surface it. Probably under https://www.tensorflow.org/api_docs/python/tf/addons but unclear how the versoning will work. The root docs reflect stable (currently r1.13) but TFA (0.3) isn't tied to this.
Eventually we want to surface all the reference APIs in one place so we'll need to figure this out ... but probably will have to wait until after Google I/O next week.

These files are all you need, for now. We don't have plans to create a tensorflow.org/addons site, but if you get some docs/notebooks together, we can link here from https://www.tensorflow.org/resources/libraries-extensions

seanpmorgan:
Thanks for the clarification Billy! I agree the APIs might want to wait for TF2 seeing as we're not compatible with the current TF release.

We'll look to add some narative docs on our master branch to make the directory more friendly, but in the mean time could we add Addons to that extensions page? Or should that wait for TF2 release as well?

lamberta:
SG. I filed b/131839445 to add TF Addons to the Resources page.

shashvatshahi1998:
@seanpmorgan can I contribute in it ?

seanpmorgan:
Hi @lamberta with 2.0 RC0 being out we were wondering what the status of this issue is?

https://github.com/tensorflow/addons/tree/r0.5/docs/api_docs/python is pinned to the release candidate, but will linking to tensorflow.org need to wait for the full release?

Also gentle ping on b/131839445

MarkDaoust:
@adammichaelwood,

While you're looking at tensorflow/agents can you look at tensorflow/addons? It's exactly the same process.

Two quick questions for stake-holders:

1. Is there any reason not to make this a real project-subsite? `tensorflow.org/addons`?

[...]

PR #518, opened 2019-09-18
Add standard flags to build_docs.py

MarkDaoust:
Hi, 

To generate docs for tensorflow.org we need to standardize the `FLAGS` so this script works with our pipeline without throwing errors about unrecognized flags.

This change doesn't affect the output of the script (except path prefixes is .yaml files) or the current usage described in README.md.

For #226

### DIFF
--- a/tools/docs/build_docs.py
+++ b/tools/docs/build_docs.py
@@ -52,12 +52,13 @@ flags.DEFINE_string(
     default=None,
     help='The name of the corresponding branch on github.')
 
-flags.DEFINE_string("output_dir", "/tmp/addons_api",
-                    "Where to output the docs")
-
 CODE_PREFIX_TEMPLATE = "https://github.com/tensorflow/addons/tree/{git_branch}/tensorflow_addons"
 flags.DEFINE_string("code_url_prefix", None,
                     "The url prefix for links to the code.")
+flags.mark_flags_as_mutual_exclusive(['code_url_prefix', 'git_branch'])
+
+flags.DEFINE_string("output_dir", "/tmp/addons_api",
+                    "Where to output the docs")
 
 flags.DEFINE_bool("search_hints", True,
                   "Include metadata search hints in the generated files")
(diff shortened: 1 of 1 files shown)


tools/docs/build_docs.py
```diff
@@ -49,31 +49,46 @@
...
 
+flags.DEFINE_string("output_dir", "/tmp/addons_api",
+                    "Where to output the docs")
+
+CODE_PREFIX_TEMPLATE = "https://github.com/tensorflow/addons/tree/{git_branch}/tensorflow_addons"
 flags.DEFINE_string(
-    'output_dir',
-    default='docs/api_docs/python/',
-    help='Where to write the resulting docs to.')
+    "code_url_prefix", None,
```
facaiy: How about putting `git_branch` and `code_url_prefix` together, and mark them as mutual exclusion？

tools/docs/build_docs.py
```diff
@@ -49,31 +49,46 @@
...
 def main(argv):
     if argv[1:]:
         raise ValueError('Unrecognized arguments: {}'.format(argv[1:]))
 
-    code_url_prefix = ('https://github.com/tensorflow/addons/tree/'
-                       '{git_branch}/tensorflow_addons'.format(
-                           git_branch=FLAGS.git_branch))
-
+    
+    if FLAGS.git_branch:
```
facaiy: nit: perhaps we could check code_url_prefix first?

facaiy:
Thanks for the nice PR, Mark. Could you fix Sanity Check? 😄

seanpmorgan:
> Thanks for the nice PR, Mark. Could you fix Sanity Check?

`bash tools/run_docker.sh -c 'make code-format'`  for reference

tools/docs/build_docs.py
MarkDaoust: Sure. Fixed.

tools/docs/build_docs.py
MarkDaoust: Yes.

MarkDaoust:
> bash tools/run_docker.sh -c 'make code-format' for reference

Thanks. Done.

PR #520, opened 2019-09-18
Implement alternatives to some TensorFlow private APIs

guillaumekln:
Most of these functions are small and can be replaced by a simple implementation. However, they included some additional logic about static shapes but it does not seem to be needed here (at least the tests are not covering such cases).

@qlzh727 Let me know if there is a strong reason to rely on these internal functions.


guillaumekln:
Sorry, this is not ready yet.

PR #523, opened 2019-09-18
typo fix

georgesterpu:


georgesterpu:
I have a question on this topic: are we supposed to explicitly call the `call` method of the `BeamSearchDecoder` class (e.g. `mydecoder.call(embedding=...)`) or the `__call__` one (e.g. `mydecoder(embedding=...`) ? I receive the following error in the latter case when naming the `embedding` argument (`embeddning` as it is now):

`TypeError: __call__() missing 1 required positional argument: 'inputs'`

code snippet here: https://pastebin.com/Lf17f4km

Thanks

qlzh727:
> I have a question on this topic: are we supposed to explicitly call the `call` method of the `BeamSearchDecoder` class (e.g. `mydecoder.call(embedding=...)`) or the `__call__` one (e.g. `mydecoder(embedding=...`) ? I receive the following error in the latter case when naming the `embedding` argument (`embeddning` as it is now):
> 
> `TypeError: __call__() missing 1 required positional argument: 'inputs'`
> 
> code snippet here: https://pastebin.com/Lf17f4km
> 
> Thanks

You should just invoke decoder by calling it, and not invoke the call() method directly. There are common logic in __call__() method, which we don't want user to skip. Also, you don't need to call with the keyword "embedding", just pass the value in as positional arg is fine.

georgesterpu:
@qlzh727 I understand. For `BeamSearchDecoder`, the first argument `embedding` becomes the `inputs` of `__call__`, while the other ones (`start_tokens`, `end_token`, 'initial_state`, `training`) are stored in `kwargs`. Is this because its call method does not take a proper input like most layers do ? Is `embedding` supposed to represent the sequence of embedded decoder inputs (of shape [batch_size, seq_len, embedding_size]) or the lookup table (of shape [vocab_size, embedding_size]) used to retrieve the embedding of an integer id in the inputs ?

PR #536, opened 2019-09-22
Fix dropout bug in BeamSearchDecoder / LSTMCell

georgesterpu:
The implementation of `call` in the `LSTMCell` class makes the assumption that the dropout mask can be cached across consecutive calls.
When using a `BasicDecoder` and a `BeamSearchDecoder` sharing the same cell with non-zero dropout, an error is triggered in `LSTMCell.call`, since the two decoders have different input shapes (the batch size gets multiplied by the beam width).

I think that the problem can also be addressed in `DropoutRNNCellMixin` by additionally checking in `get_dropout_mask_for_cell` if the input shape is different from the cached dropout mask shape.

### DIFF
--- a/tensorflow_addons/seq2seq/beam_search_decoder.py
+++ b/tensorflow_addons/seq2seq/beam_search_decoder.py
@@ -717,6 +717,13 @@ class BeamSearchDecoder(BeamSearchDecoderMixin, decoder.BaseDecoder):
         self._start_tokens = tf.tile(
             tf.expand_dims(self._start_tokens, 1), [1, self._beam_width])
         self._start_inputs = self._embedding_fn(self._start_tokens)
+        
+        # We need to prevent the RNN cell from caching the dropout masks
+        # since they will have different shapes for a BasicDecoder
+        # and a BeamSearchDecoder sharing the same cell
+        if self._cell._cell.dropout > 0.0:
+            self._cell._cell.reset_dropout_mask()
+            self._cell._cell.reset_recurrent_dropout_mask()
 
         self._finished = tf.one_hot(
             tf.zeros([self._batch_size], dtype=tf.int32),


qlzh727:
Humm, if the input shape is different between two calls, then it will also cause issue since the weights has already been initialized in one shape. Does it give you any error when you didn't set the dropout?

The reason we need to cache the dropout mask is more for correctness, for the "variational dropout".

georgesterpu:
I don't get an error with the default dropout of 0.0.
The error occurs on [this line](https://github.com/tensorflow/tensorflow/blob/r2.0/tensorflow/python/keras/layers/recurrent.py#L2258) when calling the `BeamSearchDecoder` instantiated with the same cell as the `BasicDecoder` previously used in training. The first dimension of `inputs` is beam_width times greater that the original batch size, i.e. the first dimension of the cached `dp_mask`.

On my toy example, the error is:

> tensorflow.python.framework.errors_impl.InvalidArgumentError: Incompatible shapes: [640,227] vs. [64,227]

I also suspect that the BasicDecoder may require a similar cell dropout mask reset when alternating between training and validation, but I first need to implement a full training and evaluation loop these days to be sure.

qlzh727:
I see. Seems that the weights doesn't rely on the batch dim for the shape, it will proceed without issue.

I don't think the current fix is correct, since it will forcefully reset the mask and generate a new one, which basically introduce some noise to the model. I think the correct fix should probably in the tf.keras, where the input shape should be checked against the cached dropout mask.

Wrt to training/validation, the dropout is only working in the training stage, and disabled for validation.

georgesterpu:
Since I am not so familiar with that part of the codebase, I am not confident about the correctness of the solution. Browsing through the code, I could find that `tf.keras.layers.LSTM` does a `self.reset_dropout_mask()` in its call method, as instructed by the docstring of `reset_dropout_mask`, but `LSTMCell` doesn't. The seq2seq decoders take `LSTMCell` objects, possibly wrapped with `AttentionWrapper`, however none of these classes call the dropout mask reset method.

How do you recommend to turn off dropout in evaluation, when the LSTMCell is instantiated only once in the layer's constructor with a non-zero dropout parameter ?

georgesterpu:
@qlzh727
I am not entirely sure about this, but I suspect there is a more important bug / side effect associated with dropout mask caching. This is related to my question in the previous message.

Despite having `training=False` flag, `get_dropout_mask_for_cell` is able to ignore it as long as `self._dropout_mask` has not been set to `None`, e.g. after a training iteration. Thus the evaluation dropout mask is the same as in training, rather than a Tensor of ones. This applies to the `BeamSearchDecoder`; the `LSTM` layer in the encoder appears to correctly reset the dropout mask between training and evaluation sessions.

With the proposed PR, the BeamSearchDecoder appears to have the correct behaviour in evaluation.
Perhaps both `BasicDecoder` and `BeamSearchDecoder` should reset the dropout masks right before `dynamic_decode` ?

qlzh727:
@georgesterpu.

1. training vs eval. If you check the body of get_dropout_mask_for_cell, the generated mask is based on the boolean tensor "training", which is a dynamic value and be set differently between training and eval. So I am not so sure about "Despite having training=False flag" statement.

2. reset_dropout_mask() is called at layer level and not on cell level. This is expected because of "variational dropout". If you check the details about "variational dropout", it means the same mask should be used for the time sequence input, rather than generate different one for every timestep. 

3. The fix I propose is at tf.keras.layer.LSTM/GRU, in the body of get_dropout_mask_for_cell(). It should check the input shape and test whether it matches to the existing cached value, and generate new one if it doesn't match.

Hope this answer all your questions.

georgesterpu:
Thanks a lot for the explanations, @qlzh727 
Unfortunately, they did not clear up my problems.

On 1, if we call `get_dropout_mask_for_cell` with `training=False` but with a `self._dropout_mask` different from None, `dp_mask` takes the reference to that mask. If the object stored there was created in a different call with no mask set and with `training=True`, then `dp_mask` will contain incorrect data.

On 2, `BeamSearchDecoder` never calls `reset_dropout_mask`. Is this intentional ?

On 3, I suppose that will work. In my first message I was thinking of this as a more defensive approach, however my opinion was that the application logic should be controlled at the cell level, but rather at the layer (`BeamSearchDecoder`) one. With this fix, we still have the training flag problem in 1; for example, if we use a BasicDecoder for both training and evaluation (i.e. not changing the input shape), the dropout mask used in evaluation will be the same as in training. This is why I thought that it may be good for both `BasicDecoder` and `BeamSearchDecoder` to reset the dropout masks at the top of their `call` methods.

qlzh727:
Think about this more, I think you are correct for the point 2, that the caller of the cell (Decoder in this case), should invoke reset_dropout_mask() and reset_recurrent_dropout_mask() if the cell has those methods, especially when the cell is reused.

Those methods are invoked at call() for keras.layers.LSTM/GRU already, and we should port them here as well, not only the BeamSearchDecoder, but all the Decoder. Maybe put them in the initialize() method.

georgesterpu:
@qlzh727 
If the want to reset the mask in `initialize`, then the current PR could stay, just need to do the same for `BasicDecoder`. However, the object stored in self.cell is not necessarily an instance of `LSTMCell` or `DropoutRNNCellMixin` to have a `reset_dropout_mask` method implemented. For example, it could be an `AttentionWrapper`, which inherits from `AbstractRNNCell`, and the LSTMCell object is stored in `self.cell._cell` in this case. Do you recommend to customise the decoders only for these two cases depending on the instance type (`self.cell` and `self.cell._cell`) ?

It appears to me that anyone extending the functionalities of RNN cells should be aware of the dropout masking mechanism in order to manually call `reset..`. Maybe this is too error-prone ?

I would like to point out one more corner case where the batch size changes silently. The batching methods of TensorFlow datasets have the default behaviour to output smaller batches. This should happen once or at least once, since it is fairly common to group data into buckets. Calling the dropout reset method in `initialize` appears to address this issue. However, it seems this is just slightly suboptimal, because when I log the dropout mask events, I get a sequence of "generate, reset, generate, cache, cache, cache, cache ..., cache" pe batch.

tensorflow_addons/seq2seq/beam_search_decoder.py
```diff
@@ -717,6 +717,13 @@ def initialize(self, embedding, start_tokens, end_token, initial_state):
         self._start_tokens = tf.tile(
             tf.expand_dims(self._start_tokens, 1), [1, self._beam_width])
         self._start_inputs = self._embedding_fn(self._start_tokens)
+        
+        # We need to prevent the RNN cell from caching the dropout masks
+        # since they will have different shapes for a BasicDecoder
+        # and a BeamSearchDecoder sharing the same cell
+        if self._cell._cell.dropout > 0.0:
```
qlzh727: note that the _cell might not have dropout if it is a customize cell. Please do some type or method existence check.

qlzh727:
@georgesterpu 

For any layer that will include a RNN cell within it, it will need to handle the reset of dropout mask. This is not a common use case (most of time people will write new cell, but not layer). We should document this in the public API for clarity.

In this case, the classes that handles cell directly are AttentionWrapper and Decoder, and they should check the cell's class/method and reset the mask.

For the changing of batch size, it is very common to have final partial batch, which does have a smaller batch size compare to the normal batch size. This is correctly handled in the code already by the keras training loop (eg the batch size is None and will be populated when the real data is feed). However, when the same tensor is used in two places with different batch size within the same graph, it will raise error because of the shape conflict. 

Overall, I think resetting the cached masking in the call() of AttentionWrapper and Decoder is the proper fix here. I have also done a similar fix recently in https://github.com/tensorflow/tensorflow/commit/2eb6dc0f2e7f5455d368c59c35458709eef03a55. You can take the similar approach in this PR as well.

georgesterpu:
@qlzh727 can we calculate how large is the performance penalty for adding all the `_maybe_reset_cell_dropout_mask` calls in that commit ? I wonder if it offsets the benefit of caching the dropout mask.

qlzh727:
Sorry for the late reply. The caching is for correctness (variational dropout), rather than performance.

georgesterpu:
Hello again @qlzh727 
Sorry as well for the delay. I had to fix a few other things in order to make my system work as well as on TF1.x. Now I am currently comparing the difference in accuracy and performance between relying on the LSTM/LSTMCell dropout and writing my own dropout wrapper.

The issue that I have with resetting the dropout mask in the decoder's `initialize` method by type checking is that custom cell wrappers won't work, since we need to know the structure of the class members (and where exactly is the instance of `DropoutRNNCellMixin` referenced). Or perhaps I got the wrong idea from the commit you referenced above.

qlzh727:
I think in the case of wrapped cell, eg attention in this case, the attention wrapper should also have the method of _maybe_reset_cell_dropout_mask() for the cell included, and then the decoder will just access that method if needed. Having decoder to traverse all the nested cell/wrapper is not ideal.

georgesterpu:
@qlzh727 
My workaround for the dropout mask issue in Decoders is to turn off the feature entirely for LSTMCell objects, and instead wrap them with the existing rnn cell dropout wrapper. The wrapper's implementation looks similar to the one from TF 1.x, which has worked well for me in the last couple of years. One drawback is the incompatibility of the wrapper with the `LSTM` class used in the Encoder (I assume that `RNN` + list of `LSTMCell` alternative is slower), but at least the code does not return the errors seen in the two decoders.

Since I do not like anymore the design of my initial PR or an isinstance upgrade of it for the maintainability reasons stated above, I am closing the PR. I agree that, at least for my current use case, writing a `maybe_reset...(cell)` method for the `seq2seq.BaseDecoder` class would work.

PR #539, opened 2019-09-23
Link to cudart_static and remove cuda_libs

### LINKED ISSUE #532: Package import fails when using tensorflow-gpu and CUDA libraries are missing
**System information**
- OS Platform and Distribution: Ubuntu 16.04
- TensorFlow version and how it was installed: 2.0.0rc1 (binary)
- TensorFlow-Addons version and how it was installed: 0.5.1 (binary)
- Python version: 3.5.2
- Is GPU used? No

**Describe the bug**

When using `tensorflow-gpu`, importing the `tensorflow_addons` package fails if CUDA libraries are missing while importing `tensorflow` itself works without issue.

On a non-GPU system:

```text
$ pip install tensorflow-addons==0.5.1
$ python
>>> import tensorflow_addons as tfa
```

**Other info / logs**

> `2019-09-20 18:04:16.387041: W tensorflow/stream_executor/platform/default/dso_loader.cc:55] Could not load dynamic library 'libcudart.so.10.0'; dlerror: libcudart.so.10.0: cannot open shared object file: No such file or directory`
> `Segmentation fault`

guillaumekln:
This is what is done in the TensorFlow codebase for custom ops:

https://github.com/tensorflow/tensorflow/blob/r2.0/tensorflow/tensorflow.bzl#L1771-L1772

After building a pip package with this change, I no longer get the error reported in #532.

Fixes #532.

### DIFF
--- a/tensorflow_addons/custom_ops/activations/BUILD
+++ b/tensorflow_addons/custom_ops/activations/BUILD
@@ -21,8 +21,8 @@ cc_library(
         "@local_config_tf//:libtensorflow_framework",
         "@local_config_tf//:tf_header_lib",
     ] + if_cuda_is_configured([
-        "@local_config_cuda//cuda:cuda_libs",
         "@local_config_cuda//cuda:cuda_headers",
+        "@local_config_cuda//cuda:cudart_static",
     ]),
     alwayslink = 1,
 )
@@ -43,8 +43,8 @@ cc_library(
         "@local_config_tf//:libtensorflow_framework",
         "@local_config_tf//:tf_header_lib",
     ] + if_cuda_is_configured([
-        "@local_config_cuda//cuda:cuda_libs",
         "@local_config_cuda//cuda:cuda_headers",
+        "@local_config_cuda//cuda:cudart_static",
     ]),
     alwayslink = 1,
 )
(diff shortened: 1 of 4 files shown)


qlzh727:
Adding @gunan for confirmation.

seanpmorgan:
Thanks @guillaumekln! Tested the packaged version for GPU and CPU machines and LGTM. I don't see an issue with this static linking as we only support a single version of CUDA, but we might want to wait for @yifeif or @gunan to confirm before we cherry-pick this.

gunan:
I do not think we statically link CUDA into anything in TF. I believe the tensorflow.bzl snippet to be a stale piece of code. I think @yifeif is looking into how you can build the custom ops using the same strategy as TF, dlopen-dlsym then call.

It is true that this would be a workaround but this can greatly increase your binary sizes.
We also should check how the issues around the license for redistribution of statically linked CUDA libraries work.

guillaumekln:
> I do not think we statically link CUDA into anything in TF. I believe the tensorflow.bzl snippet to be a stale piece of code.

It seems `tf_custom_op_library` was still used for many contrib kernels.

seanpmorgan:
Just for information I'm seeing a very modest increase from in binary size to 1.6MB. The licensing issues I'm unclear on though.

guillaumekln:
Actually, `nvcc` will by default link statically to the CUDA runtime, see:

https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#file-and-path-specifications-cudart

I did not find where this option is changed in the context of Addons.

guillaumekln:
Should we move forward on this? We can always  improve it in a later version if there is a better way to do it. In the meantime, it can unblock a few users.

facaiy:
> We can always improve it in a later version if there is a better way to do it. In the meantime, it can unblock a few users.

+1. We can make it better later.

facaiy:
Linked to #541

facaiy:
@guillaumekln Hi, Guillaume. Could you backport the fix to r0.5 branch? Thanks 😄

PR #545, opened 2019-09-26
correction, typeError

autoih:


autoih:
Thanks, @WindQAQ. Yes, I changed `cohens_kappa.py` back. And, I think these two are TypeError instead of ValueError because we are checking the type of `inspect.isclass(f)` in `test_utils.py`;  in `keras_utils.py`, it goes to `except TypeError` clause and raise `TypeError`  when try clause raises an exception. May I have your opinions? 
See https://docs.python.org/3/library/exceptions.html#TypeError :
 > Passing arguments of the wrong type (e.g. passing a list when an int is expected) should result in a TypeError, but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a ValueError.

facaiy:
@autoih Thank you, prcvih! Make sense for me, would you mind adding some unit tests for the change?

autoih:
Thanks @facaiy. I'd love to do that, but I'm not sure where to start. Could you please give me some reference?

facaiy:
@autoih Sure, please take a look at https://github.com/tensorflow/addons/blob/c8372e2b6499502908f1fa5551f6daca1097792e/tensorflow_addons/utils/keras_utils_test.py

autoih:
When I ran sanity check locally, there is no error. May I have your suggestions, @WindQAQ ?

seanpmorgan:
> When I ran sanity check locally, there is no error. May I have your suggestions, @WindQAQ ?

Looks like just a formatting issue. Could you please run 
`bash tools/run_docker.sh -c 'make code-format'`

If that's already been done then could you double check that the commit has been made and pushed to the branch. Thanks!

tensorflow_addons/utils/keras_utils_test.py
```diff
@@ -30,11 +30,13 @@ def test_normalize_tuple(self):
                          keras_utils.normalize_tuple((2, 1, 2),
                                                      n=3,
                                                      name='strides'))
+        # with self.assertRaises(TypeError):
```
facaiy: Why to comment this two lines?

tensorflow_addons/image/filters_test.py
```diff
@@ -94,10 +94,15 @@ def test_invalid_filter_shape(self):
...
 
-        for filter_shape in [(3, 3, 3), (3, None, 3), None]:
+        for filter_shape in [(3, 3, 3), (3, None, 3)]:
             with self.subTest(filter_shape=filter_shape):
                 with self.assertRaisesRegexp(ValueError, msg):
                     mean_filter2d(image, filter_shape=filter_shape)
+        
+        for filter_shape in [None]:  
+            with self.subTest(filter_shape=filter_shape):
+                with self.assertRaisesRegexp(TypeError, msg):
```
facaiy: Does it make sense? cc  @WindQAQ

tensorflow_addons/utils/keras_utils_test.py
autoih: I think it may be a duplicate. How do you think?

tensorflow_addons/image/filters_test.py
autoih: It looks weird, but I have to separate `None` and other tests. May I have your suggestions?

WindQAQ: Probably get rid of the first for loop?

WindQAQ: I think that's why core TF raises `ValueError` though

tensorflow_addons/image/filters_test.py
```diff
@@ -94,10 +94,15 @@ def test_invalid_filter_shape(self):
         msg = ("The `filter_shape` argument must be a tuple of 2 integers.")
         image = tf.ones(shape=(1, 28, 28, 1))
 
-        for filter_shape in [(3, 3, 3), (3, None, 3), None]:
+        for filter_shape in [(3, 3, 3), (3, None, 3)]:
             with self.subTest(filter_shape=filter_shape):
                 with self.assertRaisesRegexp(ValueError, msg):
                     mean_filter2d(image, filter_shape=filter_shape)
+        
+        if filter_shape is None:
```
WindQAQ:
```suggestion
        with self.subTest(filter_shape=None):
            with self.assertRaisesRegexp(TypeError, msg):
```

tensorflow_addons/image/filters_test.py
```diff
@@ -214,11 +219,16 @@ def test_invalid_filter_shape(self):
         msg = ("The `filter_shape` argument must be a tuple of 2 integers.")
         image = tf.ones(shape=(1, 28, 28, 1))
 
-        for filter_shape in [(3, 3, 3), (3, None, 3), None]:
+        for filter_shape in [(3, 3, 3), (3, None, 3)]:
             with self.subTest(filter_shape=filter_shape):
                 with self.assertRaisesRegexp(ValueError, msg):
                     median_filter2d(image, filter_shape=filter_shape)
 
+        if filter_shape is None:
```
WindQAQ: ditto

tensorflow_addons/utils/keras_utils_test.py
WindQAQ: Please remove it if it's duplicated.

tensorflow_addons/image/filters_test.py
autoih: I am not sure if it makes sense with the way I modified.

tensorflow_addons/utils/keras_utils.py
```diff
@@ -48,11 +48,16 @@ def normalize_tuple(value, n, name):
         return (value,) * n
     else:
         try:
+            try:
+                obj_value = iter(value)
+            except TypeError as te:
+                print (value, ' is not iterable')
+
```
WindQAQ: Why we need this?

autoih: When filter_shape is None(not [None]), value_tuple = tuple(value) can cause the error `NoneType' object is not iterable. I think it looks not good. I would like to ask your suggestion of this issue.

WindQAQ: Screenshot from 2019-10-04 17-38-33
But it still raises `TypeError`, and will be caught by the following `except: TypeError`.

autoih: Thanks @WindQAQ.

autoih:
Thanks @WindQAQ. I've checked locally (`bash tools/run_docker.sh -c 'make code-format'`), and have all pass, not sure if it's okay now.

tensorflow_addons/image/filters_test.py
```diff
@@ -94,11 +94,16 @@ def test_invalid_filter_shape(self):
...
 
-        for filter_shape in [(3, 3, 3), (3, None, 3), None]:
+        for filter_shape in [(3, 3, 3), (3, None, 3)]:
             with self.subTest(filter_shape=filter_shape):
                 with self.assertRaisesRegexp(ValueError, msg):
                     mean_filter2d(image, filter_shape=filter_shape)
 
+        filter_shape = None
+        with self.subTest(filter_shape=filter_shape):
+            with self.assertRaisesRegexp(TypeError, msg):
```
facaiy: seems kind of inconsistency, Tzu-Wei @WindQAQ

WindQAQ: I suppose that's why keras raise `ValueError` originally. Wonder if we should choose better engineering or standard.

PR #546, opened 2019-09-27
Add support for masked input in TrainingSampler

### LINKED ISSUE #534: BasicDecoder fails with masked input tensor
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Linux Ubuntu 18.04
- TensorFlow version and how it was installed (source or binary): binary(2.0.0-dev20190920)
- TensorFlow-Addons version and how it was installed (source or binary): binary(0.6.0-dev)
- Python version: 3.6
- Is GPU used? (yes/no): yes

**Describe the bug**

When the `BasicDecoder` receives a tensor equipped with masking meta-data as the input, it will fail due to the unhandled mask argument in `TrainingSampler.initialize(..)`. This is a common scenario in which the input tensor is computed using the embedding layer.  

```python
units = 32
vocab_size = 1000
embedding = tf.keras.layers.Embedding(
    input_dim=vocab_size, output_dim=units, mask_zero=True)
attention_mechanism = attention_wrapper.LuongAttention(units)
cell = tf.keras.layers.LSTMCell(units)
cell = attention_wrapper.AttentionWrapper(cell, attention_mechanism)
output_layer = tf.keras.layers.Dense(vocab_size)
sampler = sampler_py.TrainingSampler()
decoder = basic_decoder.BasicDecoder(
    cell, sampler, output_layer=output_layer)

# setup the attention mechanism's memory with a random tensor
fake_mem = tf.random.normal((2, 3, units))
fake_mem_mask = tf.ones((2,3), dtype=tf.bool)
attention_mechanism(fake_mem, mask=fake_mem_mask, setup_memory=True)

word_ids = tf.random.uniform((2, 5), 0, vocab_size, dtype=tf.int64) \
            * tf.constant([[1, 1, 1, 0, 0], [1, 1, 1, 1, 1]], dtype=tf.int64)
word_embeds = embedding(word_ids)
mask = embedding.compute_mask(word_ids)

initial_state = cell.get_initial_state(batch_size=2, dtype=tf.float32)
outputs = decoder(
    inputs=word_embeds,
    initial_state=initial_state,
    sequence_length=tf.math.reduce_sum(tf.cast(mask, tf.int32), axis=1),
)
```
It will raise the following error:
```
  File "tensorflow_addons/seq2seq/basic_decoder.py", line 72, in initialize
    return self.sampler.initialize(inputs, **kwargs) + (initial_state,)
TypeError: initialize() got an unexpected keyword argument 'mask'
```

**Other info / logs**
This issue happens because we capture the mask argument on the `BaseDecoder.call(..., **kwargs)` and propagate it all the way through `TrainingSample.initialize(...)` except that the `initialize` method doesn't accept the mask argument. 
One can temporarily fix this error by adding `delattr(word_embeds, '_keras_mask')` just before calling the decoder instance.

*I encountered this issue when i was working on #335*

Ideas to fix this issue:
- Fix the `BasicDecoder` to only pass the appropriate argument when it calls the sampler. or even fix the `TrainingSampler` to capture all other arguments using the `kwargs`
- Add the appropriate support for the mask argument in the `TrainingSampler` in additon to `sequence_length` (similar to `AttentionMechanism`)

I think there is no harm in the latter option as it provides a more Keras-friendly way to work with the framework.

kazemnejad:
By applying this patch, `TrainingSampler` will be able to support both `mask` and `sequnce_length` arguments. But `sequnce_length` will have higher priority. So if both are provided, the `mask` argument will be ignored. I had to add this priority because Keras implicitly passes the mask information to layers upon the invocation (even in model sub-classing). Without this priority, the user will face an error by using the following snippet (due to the automatically passed mask argument):
```python
some_masked_input = ...
my_decoder(some_masked_input, initial_state=..., sequence_length=sequence_length)
```
To overcome this issue, the user had to manually delete the `_keras_mask` attribute from the `some_masked_input` variable.( e.g. `delattr(some_masked_input, '_keras_mask')`) to be able to apply the `sequence_length` parameter only.

It Fixes (#534)  by adding support for masked inputs. Also, this will bring better Keras integration. 

Fixes #534


kazemnejad:
@seanpmorgan Could you please force run the CI tests on this PR?

tensorflow_addons/seq2seq/sampler.py
```diff
@@ -219,14 +223,29 @@ def initialize(self, inputs, sequence_length=None):
...
+            if mask.get_shape().ndims != 2:
+                raise ValueError(
+                    "Expected mask to a 2D tensor, but received shape: %s" %
+                    mask)
+            if not mask.dtype.is_bool:
+                raise ValueError(
+                    "Expected mask to be a boolean tensor, but received "
+                    "dtype: %s" % repr(mask.dtype))
+
+            self.sequence_length = tf.math.reduce_sum(
```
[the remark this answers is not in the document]
kazemnejad: Thanks for the link. 
Does checking the padding type of `mask` change the result sequeunce_length? I think both right and left padded will result in the same sequence length.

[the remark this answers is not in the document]
guillaumekln: The seq2seq code assumes that sequences are right padded. For example, consider this statement in `TrainingSampler`:

```python
finished = (next_time >= self.sequence_length)
```

It is only correct if sequences are right padded.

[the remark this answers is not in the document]
kazemnejad: Uh, I see. I added that in the next commit. 
By the way, should we reference the snippet you mentioned earlier in our code (As the original function is not a public API)?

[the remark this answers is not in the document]
guillaumekln: I don't think that is required. The function is quite simple.

tensorflow_addons/seq2seq/sampler.py
```diff
@@ -219,14 +223,34 @@ def initialize(self, inputs, sequence_length=None):
...
+                    mask)
+            if not mask.dtype.is_bool:
+                raise ValueError(
+                    "Expected mask to be a boolean tensor, but received "
+                    "dtype: %s" % repr(mask.dtype))
+
+            axis = 1 if not self.time_major else 0
+            with tf.control_dependencies([
+                    _check_sequence_is_right_padded(  # pylint: disable=bad-continuation
+                        mask, self.time_major)
```
qlzh727: nit, I feel this can be fit in the previous line.

tensorflow_addons/seq2seq/basic_decoder_test.py
```diff
@@ -122,6 +122,132 @@ def testStepWithTrainingHelperOutputLayer(self, use_output_layer):
...
+            self.assertEqual((batch_size, cell_depth),
+                             first_state[0].get_shape())
+            self.assertEqual((batch_size, cell_depth),
+                             first_state[1].get_shape())
+            self.assertEqual((batch_size, cell_depth),
+                             step_state[0].get_shape())
+            self.assertEqual((batch_size, cell_depth),
+                             step_state[1].get_shape())
+
+            self.assertEqual(len(output_layer.variables), 1)
```
qlzh727: You can use self.assertLen()

tensorflow_addons/seq2seq/basic_decoder_test.py
```diff
@@ -122,6 +122,132 @@ def testStepWithTrainingHelperOutputLayer(self, use_output_layer):
...
+
+            self.evaluate(tf.compat.v1.global_variables_initializer())
```
qlzh727: Assume we use the tf 2.0 as dependency, I guess this is no longer necessary?

tensorflow_addons/seq2seq/sampler.py
```diff
@@ -201,12 +202,15 @@ def sample_ids_shape(self):
...
-    def initialize(self, inputs, sequence_length=None):
+    def initialize(self, inputs, sequence_length=None, mask=None):
         """Initialize the TrainSampler.
 
         Args:
           inputs: A (structure of) input tensors.
-          sequence_length: An int32 vector tensor.
+          sequence_length: An int32 vector tensor. Note that the
+            sequence_length has higher priority than mask, so mask
+            will be ignored if both of them are provided.
```
qlzh727: I feel we should throw error if both of them are provided, rather than silently ignore the mask.

kazemnejad: I agree with you. But as I mentioned earlier, the problem with throwing an error when both are provided is that Keras implicitly passes the `mask` attribute in the Layer's `.call(..)` method. 

So even if the user does not specify the mask attribute and wants to use the `sequence_length`-type invocation it could face our error considering both `sequence_length` and `mask`(automatically) are provided. Please consider the following example:
```python
# masked_input will have the ._keras_mask attribute internally, 
# thus the Keras framework will pass this attribute 
# as a mask argument in the upcoming layers.
masked_input = ...
decoder = BasicDecoder(.., TrainingSampler(), ...)
decoder(masked_input, initial_state=..., sequence_length=sequence_length)
```
*Note that the decoder calls the `TrainingSampler.initialize(...)` internally.*

As you can see, the user doesn't specify the mask argument, but we'll receive it the `TrainingSampler.initialize(...)` independent of the user intention. 
For the user to overcome this issue, s/he will need to remove the `_keras_mask` attr. from the input tensor. 
```python
# a working example
masked_input = ...
delattr(masked_input, '_keras_mask')
decoder = BasicDecoder(.., TrainingSampler(), ...)
decoder(masked_input, initial_state=..., sequence_length=sequence_length)
```

So considering this annoyance for the end user, should we still throw an error if both of them are provided?

qlzh727: If I remember it correctly, the mask value is only provided by keras if user either have a masking layer in the model, or use embedding layer with mask_zero = True. If you don't have those specified, I think keras framework will just inject none as mask when invoking the call().

kazemnejad: Yes, I think that is the case too. However, the decoder's input is commonly produced by feeding the word ids through an embedding layer (with `mask_zero=True` attribute). Thus the user will probably face the problem mentioned above.

Despite that, I think there will be some cases that Keras unintentionally passed a not-`None` mask attribute to our function.

tensorflow_addons/seq2seq/basic_decoder_test.py
kazemnejad: Yes, you're right, I fixed it in the new commits.  However, I used the previous tests as a base for this one, so I guess we need to update them too (in a separate PR though)

tensorflow_addons/seq2seq/sampler.py
qlzh727: If the inputs are from embedding layer with mask_zero=True, then the mask should carry the same information as sequence_length, and user don't need to specify seqence_length at all.

I don't think keras will unintentionally pass a no-none mask to layer. Do you have any example?

kazemnejad: Yes, in that user doesn't need to pass the  sequence_length (which the most frequent case). 
Actually, I was worried about some use case that the user wants to implement a customized behavior in which the seqeucne_length and mask aren't the same (i don't have an example in my mind though). But it'd definitely a rare case. So I will change the `initialize` method to throw an error if both are provided.

kazemnejad:
@qlzh727 Did you have a chance to look at the new commits?

tensorflow_addons/seq2seq/sampler.py
qlzh727: Let's update this, since both of them can't be provided at same time.

tensorflow_addons/seq2seq/sampler.py
```diff
@@ -219,14 +223,36 @@ def initialize(self, inputs, sequence_length=None):
...
+                    "dtype: %s" % repr(mask.dtype))
+
+            axis = 1 if not self.time_major else 0
+            with tf.control_dependencies(
+                [_check_sequence_is_right_padded(mask, self.time_major)]  # pylint: disable=bad-continuation
+            ):
+                self.sequence_length = tf.math.reduce_sum(
+                    tf.cast(mask, tf.int32), axis=axis, name="sequence_length")
+        else:
+            raise ValueError("At least, one of the sequence_length or mask "
```
qlzh727: I don't think this is needed.

kazemnejad: So In the case that none of them is provided, we should assume that the sequence_lengh is a `[max_length]*batch_size` vector?

qlzh727: Correct. If neither of them are provided, then it means all the inputs are in full length (we should process all of them).

kazemnejad: Thanks.

kazemnejad: Thanks.
One last question, do you have any suggestions on how to infer batch_size from a structured input in a clean way?

I think, First, we need to flatten it (using `tf.nest.flatten`) and then use the first tensor as a reference to infer the batch_size

tensorflow_addons/seq2seq/sampler.py
```diff
@@ -218,21 +220,45 @@ def initialize(self, inputs, sequence_length=None):
...
+                    "dtype: %s" % repr(mask.dtype))
+
+            axis = 1 if not self.time_major else 0
+            with tf.control_dependencies(
+                [_check_sequence_is_right_padded(mask, self.time_major)]  # pylint: disable=bad-continuation
+            ):
+                self.sequence_length = tf.math.reduce_sum(
+                    tf.cast(mask, tf.int32), axis=axis, name="sequence_length")
+        else:
+            max_seq_len = tf.shape(tf.nest.flatten(inputs)[0])[0]
```
qlzh727: Please add a note here since the inputs has been converted to time_major. I was bit confused about tf.shape()[0] until I saw the self.time_major.

tensorflow_addons/seq2seq/sampler.py
qlzh727: Yea, tf.nest.flatten() and get the first element should be fine.

kazemnejad:
@qlzh727 @guillaumekln Thank you very much for your feedbacks.

guillaumekln:
Thank you for these improvements. Looking forward to the seq2seq tutorial!

PR #551, opened 2019-09-30
unify activations and tests

WindQAQ:
- Remove `name` arg for sparsemax
- Test theoretical gradients only
- Test outputs against numerical values instead of reference function
- Move serialization test to standalone file to reduce duplicated codes
- CC trivial fix

### DIFF
--- a/tensorflow_addons/activations/gelu_test.py
+++ b/tensorflow_addons/activations/gelu_test.py
@@ -27,50 +27,27 @@ from tensorflow_addons.activations import gelu
 from tensorflow_addons.utils import test_utils
 
 
-def _ref_gelu(x, approximate=True):
-    x = tf.convert_to_tensor(x)
-    if approximate:
-        pi = tf.cast(math.pi, x.dtype)
-        coeff = tf.cast(0.044715, x.dtype)
-        return 0.5 * x * (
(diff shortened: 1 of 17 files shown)


tensorflow_addons/activations/gelu_test.py
```diff
@@ -19,58 +19,33 @@
...
+            [-0.04550028, -0.15865526, 0.0, 0.8413447, 1.9544997], dtype=dtype)
+        self.assertAllCloseAccordingToType(gelu(x, False), expected_result)
 
     @parameterized.named_parameters(("float32", np.float32),
                                     ("float64", np.float64))
     def test_theoretical_gradients(self, dtype):
         # Only test theoretical gradients for float32 and float64
         # because of the instability of float16 while computing jacobian
-        x = tf.constant([1.0, 2.0, 3.0], dtype=dtype)
+        x = tf.constant([-2.0, -1.0, 0.0, 1.0, 2.0], dtype=dtype)
```
seanpmorgan: Should it be in the README that we test gradients across positive/zero/negative or should we leave that to be specific to activation function

WindQAQ: Probably want to test some special values like 0 or some edge cases, so I think it's better to leave it as activation-specific test.

PR #483, opened 2019-09-06
MatthewsCorrelationCoefficient 

autoih:
This is still in progress about (https://github.com/tensorflow/addons/issues/457). I'm trying to use colab to compare&test to make sure the correctness.

PR #557, opened 2019-10-01
avoid substraction ops to maintain precision

autoih:
This PR avoids to use `get_shape` from https://github.com/tensorflow/addons/pull/483 ; it also helps keep precision.


SSaishruthi:
@autoih 
Can you please add these two lines before the calculation?
```
y_true_negative = tf.math.not_equal(y_true,1)
y_pred_negative = tf.math.not_equal(y_pred,1)
```

I am testing with the working notebook I have for the metric and we need to add this calculation

autoih:
Hi @seanpmorgan, I did all test via [this](https://github.com/tensorflow/addons/blob/master/CONTRIBUTING.md) and pass locally. Hopefully, there is not other errors.

seanpmorgan:
Thanks @autoih this is a better implementation. Would you mind formatting with `bash tools/run_docker.sh -c 'make code-format'` and then we can merge this.

autoih:
Thanks @seanpmorgan.

Issue #437, opened 2019-08-22
please add more activation functions


bionicles:
@jvishnuvardhan @yongtang @seanpmorgan follow-up on the tf issue
**System information**
- TensorFlow version (you are using): 2b1
- TensorFlow Addons version: pip
- Is it in the tf.contrib (if so, where): idk
- Are you willing to contribute it (yes/no): yes
- Are you willing to maintain it going forward? (yes/no): yes

**Describe the feature and the current behavior/state.** 
activations are high-yield because they dramatically influence performance for little code

**Will this change the current api? How?**
just adds more activations

**Who will benefit with this feature?**
people doing hyperparameter search can benefit especially 

**Any Other info.**
here is an updated python file with some activations (converted the if/elif stuff into a lookup table at the bottom)
```
from tensorflow_addons.activations import sparsemax
import tensorflow as tf

K = tf.keras

B, L = K.backend, K.layers

RRELU_MIN, RRELU_MAX = 0.123, 0.314
HARD_MIN, HARD_MAX = -1., 1.
SOFT_ARGMAX_BETA = 1e10
FN = 'lrelu'


def swish(x):
    """
    Searching for Activation Functions
    https://arxiv.org/abs/1710.05941
    """
    return (B.sigmoid(x) * x)


def soft_argmax(x, beta=SOFT_ARGMAX_BETA):
    """
    https://stackoverflow.com/questions/46926809/getting-around-tf-argmax-which-is-not-differentiable
    https://lucehe.github.io/differentiable-argmax/
    """
    x_range = tf.range(x.shape.as_list()[-1], dtype=x.dtype)
    return tf.math.reduce_sum(
        tf.nn.softmax(x * beta) * x_range, axis=-1)


def gaussian(x):
    return B.exp(-B.pow(x, 2))


def hard_tanh(x, min=HARD_MIN, max=HARD_MAX):
    if x > max:
        return max
    elif x < min:
        return min
    else:
        return x


def hard_lisht(x, min=HARD_MIN, max=HARD_MAX):
    if x < min or x > max:
        return max
    else:
        return tf.math.abs(x)


def lisht(x):
    """
    LiSHT: Non-Parametric Linearly Scaled Hyperbolic Tangent
    https://github.com/swalpa/LiSHT
    """
    return (B.tanh(x) * x)


def rrelu(x, min=RRELU_MIN, max=RRELU_MAX):
    return x if x >= 0 else tf.random.uniform(min, max) * x


[...]

PR #569, opened 2019-10-05
add mish kernel

WindQAQ:
- Address #437 - mish kernel [Mish: A Self Regularized Non-Monotonic Neural Activation Function](https://arxiv.org/abs/1908.08681)


WindQAQ:
@digantamisra98 Hello Diganta, would you mind taking a look at this. Also, I would like to drop your name on the maintainer list if you agree. Thanks!

digantamisra98:
Hey @WindQAQ. Going through it. As of now, everything looks fine. The arXiv identifier in your first comment is incorrect though, just to point out. 
Yeah, sure, feel free to add me on the maintainer list. 
Thank you!

WindQAQ:
> Hey @WindQAQ. Going through it. As of now, everything looks fine. The arXiv identifier in your first comment is incorrect though, just to point out.
> Yeah, sure, feel free to add me on the maintainer list.
> Thank you!

Thanks for the review! Already updated the maintainer list. Please feel free to ping me if the mail address or the user id is wrong :smiley:

digantamisra98:
@WindQAQ The ID and email are correct. I would also encourage you to take a look at this repository (based on CUDA) - https://github.com/thomasbrandon/mish-cuda. 
It's much faster. Though I'm not sure, it is the right approach (Since it still doesn't work pretty well on double precision)
Also what is the wait time on the review process?
Thanks

WindQAQ:
> @WindQAQ The ID and email are correct. I would also encourage you to take a look at this repository (based on CUDA) - https://github.com/thomasbrandon/mish-cuda.
> It's much faster. Though I'm not sure, it is the right approach (Since it still doesn't work pretty well on double precision)

@digantamisra98 Eigen's GPUDevice is also based on CUDA (and ROCm). My previous experiment on gelu activations shows that Eigen's parallelism has the same performance with customized CUDA kernel in TensorFlow.

> Also what is the wait time on the review process?

Maybe next week. Thanks!

WindQAQ:
My performance test on gelu: TFA (Eigen's parallelism) vs PyTorch.
https://colab.research.google.com/drive/1pQFOp5sEp3Gw0LGHlZ6dPsNGfZPWXcgc

Will do a thorough test on TFA versus PyTorch on activation after all of these get merged :P

digantamisra98:
Thanks for that notebook link. Now I have better clarity on the same. Keep me posted on the progress of the TFA/Torch tests. Thanks!

thomasbrandon:
As author of the PyTorch kernel linked by @digantamisra98 I'll just expand a little.
I tried a similar gradient calculation to the one you have (though with less optimised reuse of intermediates). I found this to be very unstable and training an actual networks generally resulted in `nan` output killing the whole network (though tensorflow/eigen might have some special handling for this, at least not killing the whole network). This mostly affected float16 calculations but it would also occur with float32 sometimes.
Switching to a symbolically derived gradient calculation resulted in much better stability and also improved performance (likely less pronounced in the more optimised version here which reuses more of the expensive `exp` calculations).
I just use a fairly naive backward that re-calculates the entire forward then derives the gradient. I have tested a method that stored intermediates in the forward to avoid this but found that the extra memory accesses for this actually reduced performance on CUDA float32 (I didn't test but would assume this would also apply to float16). Though I was only testing synthetically (with CUDA events), there may well be less impact on overall network performance.
Storing intermediates may be more useful for CUDA float64 and CPU where the extra calculations are likely slower than the extra memory accesses, so probably the ideal method would be to selectively use intermediates depending on device/type which I will likely try at some point.

WindQAQ:
> As author of the PyTorch kernel linked by @digantamisra98 I'll just expand a little.
> I tried a similar gradient calculation to the one you have (though with less optimised reuse of intermediates). I found this to be very unstable and training an actual networks generally resulted in `nan` output killing the whole network (though tensorflow/eigen might have some special handling for this, at least not killing the whole network). This mostly affected float16 calculations but it would also occur with float32 sometimes.
> Switching to a symbolically derived gradient calculation resulted in much better stability and also improved performance (likely less pronounced in the more optimised version here which reuses more of the expensive `exp` calculations).
> I just use a fairly naive backward that re-calculates the entire forward then derives the gradient. I have tested a method that stored intermediates in the forward to avoid this but found that the extra memory accesses for this actually reduced performance on CUDA float32 (I didn't test but would assume this would also apply to float16). Though I was only testing synthetically (with CUDA events), there may well be less impact on overall network performance.
> Storing intermediates may be more useful for CUDA float64 and CPU where the extra calculations are likely slower than the extra memory accesses, so probably the ideal method would be to selectively use intermediates depending on device/type which I will likely try at some point.

Storing intermediate values is kind of trade-off, but as you noted, I will investigate it on the combination of device/type. For the stability, I am aware the softplus operation in the forward pass is very likely to underflow/overflow. Will see how you deal with the backward gradients too! Thank you so much for the suggestion!

thomasbrandon:
Yeah, the intermediate stuff was mainly as I gather that's what @digantamisra98 was referring to  with "not sure, it is the right approach (Since it still doesn't work pretty well on double precision)".
Was mainly noting what seem to be [stable forward/backwards](https://github.com/thomasbrandon/mish-cuda/blob/master/csrc/mish.h). For forward I used a trick from PyTorch's softplus which is `inp < 20 ? : log1p(exp(inp)) : inp`. I gather the threshold of 20 is because `exp(inp)` overflows a float16 at around `inp>=20`.

WindQAQ:
Hi @digantamisra98 @thomasbrandon, I have already dealt with precision problem. The softplus part is mainly copied from core TF and backward part is from @thomasbrandon implementation. Would you mind reviewing the PR again? Thank you so much for the time.

digantamisra98:
@WindQAQ will check tomorrow and let you know. Thanks!

digantamisra98:
@WindQAQ I did a quick check and was going through the log of the failed Ubuntu GPU build and couldn't interpret it correctly. Could you give it a bit of clarity?

WindQAQ:
> @WindQAQ I did a quick check and was going through the log of the failed Ubuntu GPU build and couldn't interpret it correctly. Could you give it a bit of clarity?

It seems to be the upstream's issue.

seanpmorgan:
Hi @WindQAQ when time allows, mind refactoring to use the `custom_op_library` from #581

seanpmorgan:
> @WindQAQ I did a quick check and was going through the log of the failed Ubuntu GPU build and couldn't interpret it correctly. Could you give it a bit of clarity?

@digantamisra98 would you mind taking a look again. The fail was fixed in upstream tensorflow

digantamisra98:
@seanpmorgan yeah checked, all good. Thank you for notifying.

seanpmorgan:
@WindQAQ  mind resolving conflicts when you get a chance?

WindQAQ:
To whom it may concern about the speed, I benchmark some activations in tensorflow/tensorflow-addons against pytorch's implementation. All run in eager mode without `tf.function`, `tf.xla`, `torch.jit`, `torch.jit.trace` and so on.

The GPU part is mostly the same speed (even slightly faster). CPU part depends on what the activation is. (but `hardshrink` is quite slow on CPU, I will investigate it when time allows.)

Because it's hard to build a toolchain on Colab, @thomasbrandon would you mind installing `tfa-nightly` to run some tests on Mish if you still have concerns on it? Bug report would be really appreciated. Thank you all!

https://colab.research.google.com/drive/1zKuef-upkN_4jFnBRoHLk06xmtIDRemi

digantamisra98:
@seanpmorgan is there any scope of moving Mish to TF and Keras since now it is included in PyTorch 1.9. [Reference](https://pytorch.org/docs/stable/generated/torch.nn.Mish.html?highlight=mish)

bhack:
@digantamisra98 We don't handle anymore on our side migrations in the TF ecosystem as you can see in https://github.com/tensorflow/community/pull/241#issuecomment-689947089.

Also as the [`ecosystem review" draft process](https://github.com/tensorflow/addons/blob/master/CONTRIBUTING.md#requirements-for-new-contributions-to-the-repository) is still not multi-lateral in the TF ecosystem if you are going to create a PR to introduce `MISH` again in keras like:

https://github.com/keras-team/keras/pull/13440
https://github.com/keras-team/keras/pull/13439

we don't know how the deprecation it will be triggered/handled. The process isn't currently documented and my best bet is to collaboratively expand and improve the `ecosystem review` process in the whole ecosystem.

/cc @seanpmorgan @yarri-oss @theadactyl @qlzh727

PR #572, opened 2019-10-05
correct link

autoih:
correct missing link.


autoih:
Thanks @WindQAQ, I've updated the links.

PR #598, opened 2019-10-16
CLN: decorate CRF functions

seanpmorgan:
https://github.com/tensorflow/tensorflow/issues/29075 was fixed


guillaumekln:
Should an input signature be set? CRF are usually applied on variable length inputs so this would generate a lot of retrace if users are using it the functions directly.

seanpmorgan:
Closing until we decide how we're going to handle CRF

PR #600, opened 2019-10-16
DOC: Standardize doc information

seanpmorgan:
Seeing as these doc strings appear on the landing page for our API docs. Probably best that they're a bit more standardized:
https://www.tensorflow.org/addons/api_docs/python/tfa

### DIFF
--- a/tensorflow_addons/text/__init__.py
+++ b/tensorflow_addons/text/__init__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
-"""Text-processing ops."""
+"""Additional text-processing ops."""
 from __future__ import absolute_import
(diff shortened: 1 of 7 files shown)


tensorflow_addons/text/__init__.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
-"""Text-processing ops."""
+"""Additional text-processing ops."""
```
WindQAQ: My minor concern is that crf in `text` is more like a probabilistic model rather than a text-processing operation.

seanpmorgan: Agree. I'm thinking CRF will should be moved to Layers. Individual CRF functions can still be called but they'll probably be marked private.

PR #601, opened 2019-10-16
Standardize lambda with other frameworks

seanpmorgan:
Per this discussion:
https://github.com/tensorflow/addons/pull/570#discussion_r332281330

Matching the default lambda values with other frameworks:
https://github.com/pytorch/pytorch/blob/446a79b95992dee7b987d0f35364fbb4ed1372db/torch/csrc/api/include/torch/nn/options/activation.h#L166

https://github.com/PaddlePaddle/Paddle/blob/8fb569e5b913958619f07243decebc24e5a6aa48/paddle/fluid/operators/activation_op.cc#L393

https://github.com/pytorch/pytorch/blob/446a79b95992dee7b987d0f35364fbb4ed1372db/torch/csrc/api/include/torch/nn/options/activation.h#L33

### DIFF
--- a/tensorflow_addons/activations/hardshrink.py
+++ b/tensorflow_addons/activations/hardshrink.py
@@ -27,7 +27,7 @@ _activation_ops_so = tf.load_op_library(
 
 @keras_utils.register_keras_custom_object
 @tf.function
-def hardshrink(x, lower=-1.0, upper=1.0):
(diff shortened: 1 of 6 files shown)


seanpmorgan:
Marking as WIP. Forgot the test values will need to be updated

PR #604, opened 2019-10-17
Minor correction

autoih:
Exception correction.

### DIFF
--- a/tensorflow_addons/image/transform_ops.py
+++ b/tensorflow_addons/image/transform_ops.py
@@ -96,7 +96,7 @@ def transform(images,
         else:
             transforms = transform_or_transforms
             raise ValueError(
-                "Transforms should have rank 1 or 2, but got rank %d" 
+                "transforms should have rank 1 or 2, but got rank %d" 
                 % len(transforms.get_shape()))
 
         output = _image_ops_so.addons_image_projective_transform_v2(
@@ -270,7 +270,7 @@ def _image_projective_transform_grad(op, grad):
     else:
         transforms = transform_or_transforms
         raise ValueError(
-            "Transforms should have rank 1 or 2, but got rank %d" 
+            "transforms should have rank 1 or 2, but got rank %d" 
             % len(transforms.get_shape()))
 
     # Invert transformations
--- a/tensorflow_addons/image/transform_ops.py
+++ b/tensorflow_addons/image/transform_ops.py
@@ -90,13 +90,14 @@ def transform(images,
             transforms = transform_or_transforms[None]
         elif transform_or_transforms.get_shape().ndims is None:
             raise ValueError(
-                "transform_or_transforms rank must be statically known")
+                "transforms rank must be statically known")
         elif len(transform_or_transforms.get_shape()) == 2:
             transforms = transform_or_transforms
         else:
+            transforms = transform_or_transforms
             raise ValueError(
                 "Transforms should have rank 1 or 2, but got rank %d" 
-                % len(transform_or_transforms.get_shape()))
+                % len(transforms.get_shape()))
 
         output = _image_ops_so.addons_image_projective_transform_v2(
             images,
@@ -225,7 +226,7 @@ def angles_to_projective_transforms(angles,
         elif len(angle_or_angles.get_shape()) == 1:
             angles = angle_or_angles
         else:
-            raise ValueError("Angles should have rank 0 or 1.")
+            raise ValueError("angles should have rank 0 or 1.")
         # yapf: disable
         x_offset = ((image_width - 1) -
                     (tf.math.cos(angles) * (image_width - 1) -
@@ -267,9 +268,10 @@ def _image_projective_transform_grad(op, grad):
     elif len(transform_or_transforms.get_shape()) == 2:
         transforms = transform_or_transforms
     else:
+        transforms = transform_or_transforms
         raise ValueError(
             "Transforms should have rank 1 or 2, but got rank %d" 
-            % len(transform_or_transforms.get_shape()))
+            % len(transforms.get_shape()))
 
     # Invert transformations
     transforms = flat_transforms_to_matrices(transforms=transforms)
--- a/tensorflow_addons/image/transform_ops.py
+++ b/tensorflow_addons/image/transform_ops.py
@@ -249,6 +249,7 @@ def angles_to_projective_transforms(angles,
 @tf.RegisterGradient("Addons>ImageProjectiveTransformV2")
 def _image_projective_transform_grad(op, grad):
     """Computes the gradient for ImageProjectiveTransform."""
+    
     images = op.inputs[0]
     transforms = op.inputs[1]
     interpolation = op.get_attr("interpolation")
--- a/tensorflow_addons/image/transform_ops.py
+++ b/tensorflow_addons/image/transform_ops.py
@@ -94,7 +94,7 @@ def transform(images,
         elif len(transform_or_transforms.get_shape()) == 2:
             transforms = transform_or_transforms
         else:
-            raise TypeError("Transforms should have rank 1 or 2.")
+            raise ValueError("Transforms should have rank 1 or 2.")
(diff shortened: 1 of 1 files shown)


facaiy:
The exception message seems not consistency for unknown shape and wrong shape, what do you think, Tzu-Wei @WindQAQ 

image

WindQAQ:
> The exception message seems not consistency for unknown shape and wrong shape, what do you think, Tzu-Wei @WindQAQ
> 
> image

It should be same. I'd go for `transforms should have rank 1 or 2, but got rank xxx`.

autoih:
Thanks @facaiy and @WindQAQ. In this case, it may be better to use valueError. What do you think?

WindQAQ:
> Thanks @facaiy and @WindQAQ. In this case, it may be better to use valueError. What do you think?

Agree. Would you mind also changing the message? Thanks!

autoih:
> Agree. Would you mind also changing the message? Thanks!

Yes, the message is the following, `raise ValueError(
                "Transforms should have rank 1 or 2, but got rank %d" 
                % len(transform_or_transforms.get_shape()))`

tensorflow_addons/image/transform_ops.py
```diff
@@ -89,12 +89,14 @@ def transform(images,
         if len(transform_or_transforms.get_shape()) == 1:
             transforms = transform_or_transforms[None]
         elif transform_or_transforms.get_shape().ndims is None:
-            raise TypeError(
+            raise ValueError(
                 "transform_or_transforms rank must be statically known")
```
WindQAQ: maybe make the variable name same with input argument
```suggestion
            raise ValueError(
                "transforms rank must be statically known")
```

tensorflow_addons/image/transform_ops.py
```diff
@@ -89,12 +89,14 @@ def transform(images,
...
         elif len(transform_or_transforms.get_shape()) == 2:
             transforms = transform_or_transforms
         else:
-            raise TypeError("Transforms should have rank 1 or 2.")
+            raise ValueError(
+                "Transforms should have rank 1 or 2, but got rank %d" 
+                % len(transform_or_transforms.get_shape()))
```
WindQAQ: ditto
```suggestion
            raise ValueError(
                "transforms should have rank 1 or 2, but got rank %d" 
                % len(transform_or_transforms.get_shape()))
```

tensorflow_addons/image/transform_ops.py
```diff
@@ -223,7 +225,7 @@ def angles_to_projective_transforms(angles,
         elif len(angle_or_angles.get_shape()) == 1:
             angles = angle_or_angles
         else:
-            raise TypeError("Angles should have rank 0 or 1.")
+            raise ValueError("Angles should have rank 0 or 1.")
```
WindQAQ: ditto
```suggestion
            raise ValueError("angles should have rank 0 or 1.")
```

tensorflow_addons/image/transform_ops.py
```diff
@@ -258,13 +261,15 @@ def _image_projective_transform_grad(op, grad):
...
+        raise ValueError("Invalid dtype %s." % image_or_images.dtype)
     if len(transform_or_transforms.get_shape()) == 1:
         transforms = transform_or_transforms[None]
     elif len(transform_or_transforms.get_shape()) == 2:
         transforms = transform_or_transforms
     else:
-        raise TypeError("Transforms should have rank 1 or 2.")
+        raise ValueError(
+            "Transforms should have rank 1 or 2, but got rank %d" 
+            % len(transform_or_transforms.get_shape()))
```
WindQAQ: ditto
```suggestion
        raise ValueError(
            "transforms should have rank 1 or 2, but got rank %d" 
            % len(transform_or_transforms.get_shape()))
```

tensorflow_addons/image/transform_ops.py
autoih: Do we prefer to use lower case for `angles`?

tensorflow_addons/image/transform_ops.py
```diff
@@ -89,12 +89,15 @@ def transform(images,
...
-                "transform_or_transforms rank must be statically known")
+            raise ValueError(
+                "transforms rank must be statically known")
         elif len(transform_or_transforms.get_shape()) == 2:
             transforms = transform_or_transforms
         else:
-            raise TypeError("Transforms should have rank 1 or 2.")
+            transforms = transform_or_transforms
+            raise ValueError(
+                "Transforms should have rank 1 or 2, but got rank %d"
```
WindQAQ: Transforms -> transforms to make it consistent with input argument.

tensorflow_addons/image/transform_ops.py
```diff
@@ -258,13 +262,16 @@ def _image_projective_transform_grad(op, grad):
...
+        raise ValueError("Invalid dtype %s." % image_or_images.dtype)
     if len(transform_or_transforms.get_shape()) == 1:
         transforms = transform_or_transforms[None]
     elif len(transform_or_transforms.get_shape()) == 2:
         transforms = transform_or_transforms
     else:
-        raise TypeError("Transforms should have rank 1 or 2.")
+        transforms = transform_or_transforms
+        raise ValueError(
+            "Transforms should have rank 1 or 2, but got rank %d"
```
WindQAQ: Transforms -> transforms

autoih:
Thanks @WindQAQ, I've tested locally, I didn't see any format issue. Can you help me point it out?

WindQAQ:
Please follow https://source.cloud.google.com/results/invocations/ed1e5142-4ca5-476c-ad55-d5a85044600e/targets/tensorflow_addons%2Fubuntu%2Fcpu%2Fsanity%2Fpresubmit/log

autoih:
Thanks, @WindQAQ. Here is the partial error message, `1 failed; 2 passed. Use make code-format command to format codes automatically Format checks FAILED`, but I may need your help to find the exact incorrect place. 

The following snapshot is the result after running `bash tools/run_docker.sh -c 'make code-format'`
Screen Shot 2019-10-23 at 9 52 40 AM

WindQAQ:
After formatting codes, make sure you run something like `git diff` to check if any file is changed. Anyway, try to apply this patch on your branch. 
```
diff --git tensorflow_addons/image/transform_ops.py tensorflow_addons/image/transform_ops.py                                                 
index 4a4167f..3d13935 100644
--- tensorflow_addons/image/transform_ops.py                                                                                           [0/60]
+++ tensorflow_addons/image/transform_ops.py
@@ -89,15 +89,14 @@ def transform(images,
         if len(transform_or_transforms.get_shape()) == 1:
             transforms = transform_or_transforms[None]
         elif transform_or_transforms.get_shape().ndims is None:
-            raise ValueError(
-                "transforms rank must be statically known")
+            raise ValueError("transforms rank must be statically known")
         elif len(transform_or_transforms.get_shape()) == 2:
             transforms = transform_or_transforms
         else:
             transforms = transform_or_transforms
             raise ValueError(
-                "transforms should have rank 1 or 2, but got rank %d"
-                % len(transforms.get_shape()))
+                "transforms should have rank 1 or 2, but got rank %d" % len(
+                    transforms.get_shape()))

         output = _image_ops_so.addons_image_projective_transform_v2(
             images,
@@ -268,9 +267,8 @@ def _image_projective_transform_grad(op, grad):
         transforms = transform_or_transforms
     else:
         transforms = transform_or_transforms
-        raise ValueError(
-            "transforms should have rank 1 or 2, but got rank %d"
-            % len(transforms.get_shape()))
+        raise ValueError("transforms should have rank 1 or 2, but got rank %d"
+                         % len(transforms.get_shape()))

     # Invert transformations
     transforms = flat_transforms_to_matrices(transforms=transforms)
```

autoih:
Thanks @WindQAQ for the information. Yes, sometimes locally running sanity check doesn't fully have the error messages.

Issue #607, opened 2019-10-19
Add constrained decoding to CRF


napsternxg:
**Describe the feature and the current behavior/state.**
We can reuse code from the `crf_multitag_sequence_score` to allow constrained decoding. 

**Relevant information**
- Are you willing to contribute it (yes/no): yes
- Are you willing to maintain it going forward? (yes/no): yes
- Is there a relevant academic paper? (if so, where): we only need to reuse the functions in `tfa.text.crf` to get this functionality
- Is there already an implementation in another framework? (if so, where): yes easy to extend `tfa.text.crf`
- Was it part of tf.contrib? (if so, where): 

**Which API type would this fall under (layer, metric, optimizer, etc.)**
This will be under `tfa.text.crf`

**Who will benefit with this feature?**
People in NLP community, constrained decoding is important to deal with padding issues as well to denote start and end of a sequence. 

**Any other info.**

PR #608, opened 2019-10-19
Refactor crf code to support constrained decoding

napsternxg:
Fixes tensorflow/addons#607    

    - We can have a common function for creating `filtered_inputs` used in
    `crf_multitag_sequence_score`
    - This function can be reused to modify the input to `crf_decode` to
    support constrained decoding.


napsternxg:
@googlebot I signed it!

napsternxg:
@googlebot I fixed it.

PR #611, opened 2019-10-21
Fix GroupNormalization

DawyD:
* Fix a bug, where the GroupNormalization layer was normalizing over the second axis instead of the selected axis.
* Update tests (which seem to be irrelevant anyway)

### DIFF
--- a/tensorflow_addons/layers/normalizations.py
+++ b/tensorflow_addons/layers/normalizations.py
@@ -161,7 +161,7 @@ class GroupNormalization(tf.keras.layers.Layer):
 
         group_shape = [tensor_input_shape[i] for i in range(len(input_shape))]
         group_shape[self.axis] = input_shape[self.axis] // self.groups
-        group_shape.insert(1, self.groups)
+        group_shape.insert(self.axis, self.groups)
         group_shape = tf.stack(group_shape)
         reshaped_inputs = tf.reshape(inputs, group_shape)
         return reshaped_inputs, group_shape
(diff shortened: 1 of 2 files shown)


seanpmorgan:
Thanks @DawyD!

@Smokrow could you take a look as code owner? We might need add additional test cases

Smokrow:
As far as I see it the reduction is working as intended.  
The axis that is introduced is the group axis and should always be on position 1.

DawyD:
The problem lies in the reshaping. `tf.reshape` reads and writes the elements in C-like index order, with the last axis index changing fastest, back to the first axis index changing slowest.

You insert a new axis at index 1 that has size `nr_groups` and then you reshape assuming that the selected axis will be split into the inserted axis. In your code, this works only if the selected axis is the axis with index 2. If you select, for example, the axis with index 4, the axis that is split into the inserted one is the axis with index 2 and not 4.

Smokrow:
Hey there. I completely missed this one. Thank you. 
@seanpmorgan the fix looks good.

Smokrow:
@seanpmorgan This will actually change the behavior of the layer since the GroupAxis will be moved to a different position. I am not sure if this can cause trouble down the line. Maybe we should fix that. What do you think?

seanpmorgan:
> @seanpmorgan This will actually change the behavior of the layer since the GroupAxis will be moved to a different position. I am not sure if this can cause trouble down the line. Maybe we should fix that. What do you think?

Since the group axis only exists while [reshaping in the layer call](https://github.com/tensorflow/addons/blob/91e951507c88e087215791551a6cc9126593e2ea/tensorflow_addons/layers/normalizations.py#L118-L124) this shouldn't produce any side-effects as far as I can tell.

PR #622, opened 2019-10-24
FIX: pass numpy arrays to keras model

seanpmorgan:
Nightly breaks when passing list to predict. No reason this test should have been doing that.


seanpmorgan:
Apologies for the spam, was trying to do this on a machine I couldn't test locally. Since https://github.com/tensorflow/tensorflow/commit/268229dfedd5ee95ddb8028625a3b27ae5361a50 the fallback to v1 keras predict without compile has been removed. 

The eager version had a few issues solved in this PR.

PR #635, opened 2019-10-28
Bugfix: correct input shape handling for CorrelationCost Keras layer

mels630:
* Fixes in CorrelationCost::compute_output_shape: integer division and channel indexing
* Cleanup in CorrelationCostTest for duplicated code


mels630:
Note that the Keras tests in optical_flow_test.py were segfaulting when I ran outside the Docker container in a Mac/CPU environment. The computed shapes were definitely wrong in Keras (that is, different than the shape of outputs created by the _forward_simple method), not sure why the segfault doesn't occur in the standard run/test environments.

PR #637, opened 2019-10-29
Add ctags about symbols

autoih:
This PR is adding ctags to generate a text file about symbols used in addons.

### DIFF
--- a/tfa.ctags
+++ b/tfa.ctags
@@ -1,0 +1,1 @@
+--exclude=bazel\*


WindQAQ:
This should be a user-specific plugin, so I am going to close this PR. I'd suggest you could load it in your vscode workspace config or .vimrc etc.

autoih:
Thanks @WindQAQ, but I think it will be good to have it as well, [here](https://github.com/pytorch/pytorch/blob/master/.ctags.d/pytorch.ctags) is from PyTorch.

WindQAQ:
Yes, but it's still not a prerequisite of a GitHub repository or a Python package. I mean it's a feature that the repository could still work without it, and users could easily setup it on their own workspace without any extra effort. Thank you.

PR #695, opened 2019-11-12
Test in eager model

autoih:
To be consistent with all tests in optimizers, layers, etc; replacing `run_in_graph_and_eager_modes` (only supports test methods) by `run_all_in_graph_and_eager_modes`.

### DIFF
--- a/tensorflow_addons/layers/normalizations_test.py
+++ b/tensorflow_addons/layers/normalizations_test.py
@@ -52,7 +52,9 @@ class NormalizationTest(tf.test.TestCase):
             reshaped_inputs, group_shape = group_layer._reshape_into_groups(
                 inputs, (10, 10, 10), tensor_input_shape)
             for i in range(len(expected_shape)):
-                tf.assertEqual(tf.cast(group_shape[i],tf.int32), tf.cast(expected_shape[i], tf.int32))
+                tf.assertEqual(
+                    tf.cast(group_shape[i], tf.int32),
+                    tf.cast(expected_shape[i], tf.int32))
 
         input_shape = (10, 10, 10)
         expected_shape = [10, 5, 10, 2]
--- a/tensorflow_addons/layers/normalizations_test.py
+++ b/tensorflow_addons/layers/normalizations_test.py
@@ -24,6 +24,7 @@ from tensorflow_addons.layers.normalizations import InstanceNormalization
 from tensorflow_addons.utils import test_utils
 
 
+@test_utils.run_all_in_graph_and_eager_modes
 class NormalizationTest(tf.test.TestCase):
 
     # ------------Tests to ensure proper inheritance. If these suceed you can
@@ -50,7 +51,8 @@ class NormalizationTest(tf.test.TestCase):
             reshaped_inputs, group_shape = group_layer._reshape_into_groups(
                 inputs, (10, 10, 10), tensor_input_shape)
             for i in range(len(expected_shape)):
-                self.assertEqual(int(group_shape[i]), expected_shape[i])
+                self.assertEqual(
+                    self.evaluate(group_shape[i]), expected_shape[i])
 
         input_shape = (10, 10, 10)
         expected_shape = [10, 5, 10, 2]
@@ -99,13 +101,13 @@ class NormalizationTest(tf.test.TestCase):
             axis=axis, groups=groups, center=center, scale=scale)
         model = tf.keras.models.Sequential()
         model.add(layer)
-        outputs = model.predict(inputs)
+        outputs = model.predict(inputs, steps=1)
         self.assertFalse(np.isnan(outputs).any())
 
         # Create shapes
         if groups is -1:
             groups = input_shape[axis]
-        np_inputs = inputs.numpy()
+        np_inputs = self.evaluate(inputs)
         reshaped_dims = list(np_inputs.shape)
         reshaped_dims[axis] = reshaped_dims[axis] // groups
         reshaped_dims.insert(1, groups)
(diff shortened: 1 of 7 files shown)


tensorflow_addons/layers/normalizations_test.py
```diff
@@ -184,7 +187,7 @@ def test_axis_error(self):
         with self.assertRaises(ValueError):
             GroupNormalization(axis=0)
 
-    @test_utils.run_in_graph_and_eager_modes
+    # @test_utils.run_in_graph_and_eager_modes
```
seanpmorgan: Hmmm why these commented out decorators below the decorated Class?

seanpmorgan: ditto for the below ~4 other cases of these comments

tensorflow_addons/layers/normalizations_test.py
```diff
@@ -24,6 +24,8 @@
 from tensorflow_addons.utils import test_utils
 
 
+@test_utils.run_all_in_graph_and_eager_modes
+@tf.function
```
seanpmorgan: What is the logic on decorating a TestCase class with `tf.function`? Does it make all eager methods run as a `tf.function`? I wasn't aware of this being used this way?

tensorflow_addons/layers/normalizations_test.py
autoih: Thanks @seanpmorgan. Updated.

tensorflow_addons/layers/normalizations_test.py
autoih: Thanks @seanpmorgan. My original thought with `tf.function` is for [test here](https://github.com/tensorflow/addons/blob/1374b46f2d12e993fed903485b7cf8cd3e9b897e/tensorflow_addons/layers/normalizations_test.py#L56). If there is a better way to do, I'd like to have your suggestions. Thanks again.

Squadrick: `tf.function` works only on python functions, it doesn't work on classes. [Code here](https://github.com/tensorflow/tensorflow/blob/befd063e093b306084642872c1c7e95317e0fee4/tensorflow/python/eager/def_function.py#L971).

Further, we'd never want to wrap any test function in `tf.function` since that'll force it to execute in a non-eager (graph) way whereas we want to test both eager and graph execution. Wrapping in one of the decorators in`test_utils` should be sufficient.

tensorflow_addons/layers/normalizations_test.py
```diff
@@ -50,7 +52,9 @@ def run_reshape_test(axis, group, input_shape, expected_shape):
             reshaped_inputs, group_shape = group_layer._reshape_into_groups(
                 inputs, (10, 10, 10), tensor_input_shape)
             for i in range(len(expected_shape)):
-                self.assertEqual(int(group_shape[i]), expected_shape[i])
+                tf.assertEqual(
```
Squadrick: Any reason for switching from `self.assertEqual` to `tf.assertEqual`?

tensorflow_addons/optimizers/weight_decay_optimizers_test.py
```diff
@@ -27,6 +27,7 @@
 WEIGHT_DECAY = 0.01
 
 
+@test_utils.run_all_in_graph_and_eager_modes
```
Squadrick: Don't think this is necessary. The expected use case would be to extend `OptimizerTestBase`, and wrap the child class with `test_utils.run_all_in_graph_and_eagar_modes`. 

What do you think @seanpmorgan?

tensorflow_addons/metrics/r_square_test.py
```diff
@@ -20,8 +20,11 @@
 
 import tensorflow as tf
 from tensorflow_addons.metrics import RSquare
+from tensorflow_addons.utils import test_utils
 
 
+@test_utils.run_all_in_graph_and_eager_modes
+@tf.function
```
Squadrick: Remove the `tf.function`.

tensorflow_addons/layers/normalizations_test.py
autoih: Thanks @Squadrick. In this case, using `@test_utils.run_in_graph_and_eager_modes` in functions might be more appropriate for `test_utils`.  What do you think?

tensorflow_addons/layers/normalizations_test.py
autoih: Thanks @Squadrick.

tensorflow_addons/metrics/r_square_test.py
autoih: Thanks @Squadrick.

tensorflow_addons/layers/normalizations_test.py
Squadrick: No no, you can wrap the class in `test_utils.run_all_in_graph_and_eager_modes`. Just delete `tf.function`.

tensorflow_addons/optimizers/weight_decay_optimizers_test.py
autoih: Thanks @Squadrick. I updated it at child class.

tensorflow_addons/layers/normalizations_test.py
autoih: Thanks @Squadrick. Yea, in this case, I may need to change the test a bit, since errors will be raised by the [line](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/layers/normalizations_test.py#L53)

autoih: Thanks @Squadrick. I've remove `tf.function` and fix the errors cause by running `graph` mode. Can you help review when you are convenient?

autoih:
Thanks @Squadrick. The new commit uses `run_all_in_graph_and_eager_modes` for all test. Can you help review when you are available?

autoih:
Thanks @Squadrick and @seanpmorgan for the suggestions.

PR #696, opened 2019-11-13
Fix WeightNormalization wrapped layer has use_biase=False

tzachar:
When using WeightNormalization wrapper layer on a layer which set use_bias=False we get an exception. 

Make sure to check that a bias exists before trying to assign to it.


tzachar:
@googlebot I signed it!

PR #700, opened 2019-11-15
Fix keras model compile for lifted_struct_loss

### LINKED ISSUE #699: Lifted Struct Loss Error
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Google Colab (Linux)
- TensorFlow version and how it was installed (source or binary): binary 2.0
- TensorFlow-Addons version and how it was installed (source or binary): binary 0.6.0
- Python version: 3
- Is GPU used? (yes/no): Yes

**Describe the bug**
I get an assert shape error when I try to initialize the LiftedStructLoss obj

`assert lshape.shape == 1 AssertionError` is the error I get. However, it can easily be resolved by removing the assert line. I have one more question, why does it say that the vector should not be l2 normalized? while in the case of semihardtriplet loss it says that the vector should be l2 normalized
`import tensorflow_addons as tfa`
`tfa.losses.LiftedStructLoss()`

WindQAQ:
Closes #699.

### DIFF
--- a/tensorflow_addons/losses/lifted_test.py
+++ b/tensorflow_addons/losses/lifted_test.py
@@ -64,7 +64,9 @@ class LiftedStructLossTest(tf.test.TestCase):
         labels = np.random.randint(
             0, num_classes, size=(num_data)).astype(np.float32)
         # Reshape labels to compute adjacency matrix.
+        # pylint: disable=E1136
         labels_reshaped = np.reshape(labels, (labels.shape[0], 1))
+        # pylint: enable=E1136
 
         # Compute the loss in NP
         # pylint: disable=E1111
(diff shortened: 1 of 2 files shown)


tensorflow_addons/losses/lifted_test.py
```diff
@@ -67,7 +67,9 @@ def testLiftedStruct(self):
         labels_reshaped = np.reshape(labels, (labels.shape[0], 1))
 
         # Compute the loss in NP
+        # pylint: disable=E1111
         adjacency = np.equal(labels_reshaped, labels_reshaped.T)
+        # pylint: enable=E1111
```
WindQAQ: This seems to be the false alarm for pylint in python2 so I disable it partly.

Squadrick:
@WindQAQ Thanks for the PR

PR #658, opened 2019-11-02
Revert pinned nightly

seanpmorgan:
Checking https://github.com/tensorflow/tensorflow/commit/f6a40276a1eaa472b25e87a6371310bc80e3fb56 fix

PR #701, opened 2019-11-15
Pin tf-nightly in setup.py

jkr26:
Currently, the following sequence of steps fails:

* Create and activate fresh virtualenv
* install tfa-nightly
* import tensorflow_addons

This raises an op not found error.

I believe this change should fix at least this behavior.


seanpmorgan:
Thanks @jkr26 and sorry for not responding sooner. We had the nightly pinned to an older version for a lot longer than we ever expect to. Going forward the next time we pinn we should include this type of fix so that it's not broken like this. 

https://github.com/tensorflow/addons/pull/658 just went in though so all should be good now. Thanks again!

Issue #663, opened 2019-11-04
Missing symbol for Abseil ParseTime Op


seanpmorgan:
**Describe the bug**
Since the 20101030 nightly, our library fails when opening the `_parse_time_op.so` because of an undefined symbol error. Typically this occurs when there is an ABI incompatibility between core TensorFlow and our custom-ops. 

```
tensorflow.python.framework.errors_impl.NotFoundError: .../custom_ops/text/_parse_time_op.so: undefined symbol: _ZN4absl9ParseTimeERKSsS1_PNS_4TimeEPSs
```

This particular [custom-op utilizes absl](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/custom_ops/text/cc/kernels/parse_time_kernel.cc#L75) and the mangled symbol name does not match what our compiled operation thinks it should be. I'm not really sure how this would happen since the rest of the TensorFlow library was compiled with a compatible gcc version.

CC'ing some people who may have insight into how this can happen:
@yifeif @r4nt @perfinion @gunan 

I suppose we could make absl a dependency in our TFA build and compile it ourselves but this seems to be going down a bad path.

Failing Log:
https://source.cloud.google.com/results/invocations/a71d103a-ecac-421c-bc53-ef06648fdc9f/targets/tensorflow_addons%2Fubuntu%2Fgpu%2Fpy3%2Fpresubmit/log

r4nt:
@scentini for whether this might be due to us not exporting all symbols from absl any more

scentini:
Yep, seems like it is related to [f58c9f8](https://github.com/tensorflow/tensorflow/commit/f58c9f835c32e02895bdd629c7660567ae011001), which changes what we export, see https://github.com/bazelbuild/bazel/issues/7362.

The absl symbols probably came from [here](https://github.com/tensorflow/addons/blob/c3aba0818b6e74ac3a4341d00242c464584c8b50/tensorflow_addons/tensorflow_addons.bzl#L13), and due to [f58c9f8](https://github.com/tensorflow/tensorflow/commit/f58c9f835c32e02895bdd629c7660567ae011001) we just don't export the symbol anymore.
I think the right way forward is to make [_parse_time_op.so](https://github.com/tensorflow/addons/blob/c3aba0818b6e74ac3a4341d00242c464584c8b50/tensorflow_addons/custom_ops/text/BUILD#L16) depend on absl.

seanpmorgan:
[...]

PR #710, opened 2019-11-21
Don't build parse_time till TF r2.1

Squadrick:
As discussed, we won't build `parse_time` till #663 is completely resolved in r2.1


Squadrick:
cc: @seanpmorgan @WindQAQ

Issue #720, opened 2019-11-26
CohenKappa reset_states distributed training setup errors.


n3011:
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Google Colab
- TensorFlow version and how it was installed (source or binary): 2.0.0
- TensorFlow-Addons version and how it was installed (source or binary): 0.6.0
- Python version: 3.6
- Is GPU used? (yes/no): yes

**Describe the bug**
When using cohen kappa metric within the distribution strategy scope, `reset_states` throws the following errors:

     UFuncTypeError                            Traceback (most recent call last)
     <ipython-input-20-56dc62930e1d> in <module>()
     38     train_accuracy.reset_states()
     39     test_accuracy.reset_states()
      ---> 40     test_kappa.reset_states()

     2 frames
      /tensorflow-2.0.0/python3.6/tensorflow_core/python/distribute/values.py in assign(self, *args, 
       **kwargs)
     1218         tensor = args[0]
     1219         if self._aggregation == vs.VariableAggregation.SUM: 
         -> 1220           tensor *= 1. / len(self.devices)
     1221         return control_flow_ops.group(tuple(
      1222             _assign_on_device(v.device, v, tensor) for v in self._values))

     UFuncTypeError: Cannot cast ufunc 'multiply' output from dtype('float64') to dtype('int32') with 
          casting rule 'same_kind'

**Code to reproduce the issue**
[Colab notebook](https://colab.research.google.com/drive/13MfjnWWsg6cKTINeKtBe5SKBhdeCHhr0)
attached.

Squadrick:
cc: @AakashKumarNain

n3011:
I have fixed it temporarily by changing the variables and weight matrix dtype to `tf.float32`. I think float should be correct dtype for weight matrix by default.

Squadrick:
@n3011 Could you please make a PR with the fix?

n3011:
@Squadrick Sure will do.

PR #721, opened 2019-11-27
fix distributed training error and nan result bugs

n3011:
It fixes the following two bugs related to the CohenKappa metric:
  - dtype issue when used within distributed strategy scope #720 
  - NaN result issue after state reset.

### DIFF
--- a/tensorflow_addons/metrics/cohens_kappa.py
+++ b/tensorflow_addons/metrics/cohens_kappa.py
@@ -167,9 +167,10 @@ class CohenKappa(Metric):
         # 6. Calculate Kappa score
         numerator = tf.reduce_sum(conf_mtx * weight_mtx)
         denominator = tf.reduce_sum(out_prod * weight_mtx)
-        return tf.cond(tf.math.is_nan(denominator),
-                       true_fn=lambda: 0.0,
-                       false_fn=lambda: 1 - (numerator / denominator))
+        return tf.cond(
+            tf.math.is_nan(denominator),
+            true_fn=lambda: 0.0,
+            false_fn=lambda: 1 - (numerator / denominator))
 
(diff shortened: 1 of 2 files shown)


tensorflow_addons/metrics/cohens_kappa.py
```diff
@@ -167,8 +167,9 @@ def result(self):
         # 6. Calculate Kappa score
         numerator = tf.reduce_sum(conf_mtx * weight_mtx)
         denominator = tf.reduce_sum(out_prod * weight_mtx)
-        kp = 1 - (numerator / denominator)
-        return kp
+        return tf.cond(tf.math.is_nan(denominator),
+                       true_fn=lambda: 0.0,
+                       false_fn=lambda: 1 - (numerator / denominator))
```
WindQAQ: Though I think the fix is correct, but a better approach should be change the division of [these two lines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/metrics/cohens_kappa.py#L161-L162) into `tf.math.divide_no_nan`. And modify `kp` to `kp = tf.math.divide_no_nan(denominator - numerator, denominator)`. How do you think about this @n3011?

n3011: @WindQAQ there are two problems with `tf.math.divide_no_nan` approach:
   - it will produce `nan` result only, as `nan/nan` not defined.
   - it slightly slower than `tf.cond` based aproach.

WindQAQ: Got it. After a rough test, your approach is actually faster :-) Thanks for the report.

n3011:
@WindQAQ thanks, pushed the formatted changes.

PR #726, opened 2019-11-30
Swish Activation Function

PhilipMay:
Pull request for swish activation function.
See paper "Searching for Activation Functions": https://arxiv.org/abs/1710.05941

The swish activation function is being used in EfficientNet for example:

- https://github.com/keras-team/keras-applications/blob/master/keras_applications/efficientnet.py#L96
- https://arxiv.org/abs/1905.11946

### DIFF
--- a/tensorflow_addons/activations/BUILD
+++ b/tensorflow_addons/activations/BUILD
@@ -13,6 +13,7 @@ py_library(
         "rrelu.py",
         "softshrink.py",
         "sparsemax.py",
+        "swish.py",
         "tanhshrink.py",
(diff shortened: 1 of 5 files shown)


PhilipMay:
Well... yes that’s right. Although I wonder why they put an activation function to tf.nn
I think I should close this PR again.

seanpmorgan:
Thanks for the PR @PhilipMay! I'm very much not a fan of the `tf.nn` grab-bag but regardless we'll probably not want to implement this in TFA. Sorry for the confusion.

PR #734, opened 2019-12-02
Minor correction docstring

autoih:


autoih:
Hi @facaiy, can you please check the minor corrections when you are available? Thanks.

tensorflow_addons/losses/focal_loss.py
```diff
@@ -41,10 +41,10 @@ class SigmoidFocalCrossEntropy(tf.keras.losses.Loss):
     fl = tfa.losses.SigmoidFocalCrossEntropy()
     loss = fl(
       [[0.97], [0.91], [0.03]],
-      [[1], [1], [0])
+      [[1.0], [1.0], [0.0]])
     print('Loss: ', loss.numpy())  # Loss: [[0.03045921]
                                             [0.09431068]
-                                            [0.31471074]
+                                            [0.31471074]]
```
seanpmorgan: Syntax fix looks good. I'm getting a different result though.. would you mind changing it?
https://colab.research.google.com/drive/1jQ9rkZs9501AP8ADhzwuv_3pQaRzW9I2

FWIW the old syntax would have worked since there is a [dtype cast in this loss](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/losses/focal_loss.py#L123). Still I think it's best to pass in matching types.

autoih: Thanks @seanpmorgan. I updated the results.

PR #736, opened 2019-12-03
Disable seq2seq test that uses deprecated fallback

omalleyt12:
The seq2seq layers need to be made to work w/o `experimentatl_run_tf_function=False`


seanpmorgan:
Thanks for pointing this out ahead of breakage @omalleyt12 . Is there a timeline for when `experimentatl_run_tf_function` will not be supported or any additional information on this? Is it https://github.com/tensorflow/community/pull/184 ?

cc @qlzh727 @guillaumekln

guillaumekln:
Let's get this in before it starts causing issue. I will create the related issue.

PR #738, opened 2019-12-04
fix bug in crf.CrfDecodeForwardRnnCell

tracy-talent:
When I read your code implementation of crf, I found a little bug in class CrfDecodeForwardRnnCell, check it!


tracy-talent:
Get it! Thank you, I suggest updating the comment, it's easy to mislead.

Squadrick:
Do you want to update it in this PR?

tracy-talent:
> Do you want to update it in this PR?
Yes, I do. I suggest
```
state: A tuple containing a [batch_size, num_tags] matrix containing the 
        previous step's score values.
```
is better.

Squadrick:
@tracy-talent Yup, it's what we should go with it.

PR #742, opened 2019-12-06
BLD: fix release

seanpmorgan:
Fix issues found in today's nightly build


seanpmorgan:
> Thanks. This is probably my mistake.

Nope my fault, would be hard to have known about the env variable in the container.

PR #745, opened 2019-12-06
Implements YogiOptimizer

manzilz:
Implement Yogi optimizer in paper [Adaptive Methods for Nonconvex Optimization](https://papers.nips.cc/paper/8186-adaptive-methods-for-nonconvex-optimization.pdf) as a tf.keras.optimizers.Optimizer

### DIFF
--- a/tensorflow_addons/optimizers/BUILD
+++ b/tensorflow_addons/optimizers/BUILD
@@ -15,6 +15,7 @@ py_library(
         "rectified_adam.py",
         "stochastic_weight_averaging.py",
         "weight_decay_optimizers.py",
+        "yogi.py",
     ],
     deps = [
         "//tensorflow_addons/utils",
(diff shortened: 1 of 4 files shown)


jkr26:
+1 to this PR!!

manzilz:
@WindQAQ Finished code formatting. PTAL

manzilz:
Fixed a small error.

avinava-o:
+1 for this PR!

manzilz:
Any update @WindQAQ @facaiy @Squadrick ?

relaxation82:
It seems that the Yogi optimiser was released as part of tensorflow-addons-0.7.0, but I am unable to find any documentation about it. In the case I was just blind, could someone provide a link to a page with the documentation, or in case it is actually missing for the documentation, could you provide a short example describing how it can be used?

seanpmorgan:
> It seems that the Yogi optimiser was released as part of tensorflow-addons-0.7.0, but I am unable to find any documentation about it. In the case I was just blind, could someone provide a link to a page with the documentation, or in case it is actually missing for the documentation, could you provide a short example describing how it can be used?

@relaxation82 thank you very much for this callout. Seems the import got missed and thus the API docs were not generated.

I've submitted a PR to fix this (#875) and we will likely do a patch release in the next couple of weeks. In the mean time it can still be used on 0.7.0 with:
```
from tensorflow_addons.optimizers.yogi import Yogi
```

The documentation will be whats present in the doc strings of the module:
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/optimizers/yogi.py

Issue #732, opened 2019-12-02
Limit GPU test memory so we can run in parallel


seanpmorgan:
Because TF allocates an entire GPU's memory when running, we have to limit our GPU tests to run a single test at a time. See the [ci script](https://github.com/tensorflow/addons/blob/08115506e7bdf9c19857540d662d7bfba66f448a/tools/ci_testing/addons_gpu.sh#L28) and [documentation](https://github.com/tensorflow/addons/blob/master/CONTRIBUTING.md#locally-testing-gpu).

I believe there are a few ways to work around this so that we can improve our test time. Some possible ways:
*  TF Core runs parallel gpu tests with this script: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/gpu_build/parallel_gpu_execute.sh
* Looks like there is some new experimental ways to allow GPU growth: https://www.tensorflow.org/api_docs/python/tf/config/experimental/set_memory_growth

PR #752, opened 2019-12-10
run gpu tests in parallel

WindQAQ:
Addresses #732.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,19 +79,12 @@ or run manually:
 docker run --rm -it -v ${PWD}:/addons -w /addons gcr.io/tensorflow-testing/nosla-ubuntu16.04-manylinux2010 /bin/bash
 ./configure.sh  # Links project with TensorFlow dependency
 
-bazel test -c opt -k \
---test_timeout 300,450,1200,3600 \
---test_output=all \
-//tensorflow_addons/...
+bash tools/ci_testing/addons_cpu.sh
 ```
 
 #### Locally Testing GPU
 
 ```bash
-export TF_GPU_COUNT=4 # Specify number of GPUs available
-export TF_TESTS_PER_GPU=8 # Specify number of tests per GPU
-export TF_PER_DEVICE_MEMORY_LIMIT_MB=1024 # Limit the memory used per test
-
 bash tools/run_docker.sh -d gpu -c 'make gpu-unit-test'
 ```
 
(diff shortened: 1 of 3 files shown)


WindQAQ:
@tensorflow/sig-addons-maintainers GPU CI tests on cloud now runs in 10 minutes! I suppose the hardware spec on cloud is the same with the one of core TF, that is, 4 gpus in total.

Before:
https://source.cloud.google.com/results/invocations/278c5a6c-5e20-4130-a880-905ffc8d3264/targets
After:
https://source.cloud.google.com/results/invocations/e221f26f-1707-49d8-87c2-d7ce71a605ba/targets

tools/ci_testing/addons_gpu.sh
```diff
@@ -39,12 +39,18 @@ ls -alh $PYTHON_BIN_PATH
 # Use default configuration here.
 yes 'y' | ./configure.sh
 
+wget https://raw.githubusercontent.com/tensorflow/tensorflow/master/tensorflow/tools/ci_build/gpu_build/parallel_gpu_execute.sh
```
seanpmorgan: I think it's a bit more controlled if we copy this script into our own repo? WDYT?

WindQAQ: Sure, I also want to copy it into the repo :-)

WindQAQ:
- Copy parallel_gpu_count.sh into repo.
- Updated instructions to specify GPU spec.

CONTRIBUTING.md
```diff
@@ -105,6 +106,14 @@ bazel test -c opt -k \
...
 ```
 
+Note that the commands run with 4 GPUs with 8 tests per GPU by default.
+Please use the following variables to controll the hardware specifications:
+```
+export TF_GPU_COUNT=4 # Specify number of GPUs available
+export TF_TESTS_PER_GPU=8 # Specify number of tests per GPU
+export TF_PER_DEVICE_MEMORY_LIMIT_MB=1024 # Limit the memory used per test
+```
+
```
seanpmorgan: Can you move this up so that it goes with the `run_docker.sh` way of running GPU tests. For the manual way that's right above this -- these shouldn't be required.

WindQAQ: Fixed

WindQAQ: Hmm, it might be wrong if we move it up because the env vars outside the container might not take effect inside the container.

WindQAQ: it turns out we have to do something like
```
bash tools/run_docker.sh -d gpu -c 'TF_GPU_COUNT=1 && make gpu-unit-test'
```
What to you think, Sean?

CONTRIBUTING.md
```diff
@@ -94,15 +92,12 @@ or run manually:
...
---test_timeout 300,450,1200,3600 \
---crosstool_top=//build_deps/toolchains/gcc7_manylinux2010-nvcc-cuda10.1:toolchain \
---test_output=all \
---jobs=1 \
-//tensorflow_addons/...
+export TF_GPU_COUNT=4 # Specify number of GPUs available
+export TF_TESTS_PER_GPU=8 # Specify number of tests per GPU
+export TF_PER_DEVICE_MEMORY_LIMIT_MB=1024 # Limit the memory used per test
+
+bash tools/ci_testing/addons_gpu.sh
```
seanpmorgan: Thinking a bit more on this, I think it's unlikely that developers will be using multi-gpus for testing. It's even less useful because you only need to run one target for most changes. 

How do you feel about making the `addons_gpu.sh` default to single GPU if some `kokoro` flag is not set. Then we can just remove this export information (or document it in the script) so that it's less confusing for those who just want to test a single GPU target change.

seanpmorgan: So I tested an if these ENV variables are not set (but only one GPU exists) then the test still runs successfully with `addons_gpu.sh`.

seanpmorgan:
@tensorflow/sig-addons-maintainers This is ready for review. Tzu-Wei is on vacation so I've made updates as this is a very useful PR we would like to merge. 

I'll override the CLA once it's been approved.

PR #758, opened 2019-12-12
Removing int check in moving average

### LINKED ISSUE #754: Allow passing in a tensor to tfa.optimizers.MovingAverage num_updates
Currently, `tfa.optimizers.MovingAverage` has an assert that requires `num_updates` to be of type int, [shown here](https://github.com/tensorflow/addons/blob/604a70de563f8797984c9c3f002aff70bef6c90b/tensorflow_addons/optimizers/moving_average.py#L81). This prevents me from passing in an integer tensor that changes with the global step, which is officially supported by `tf.train.ExponentialMovingAverage`.

Can this assert be updated to handle this use case?

SSaishruthi:
This PR closes #754


SSaishruthi:
@seanpmorgan can you please take a look at it
@Hyperparticle FYI

PR #759, opened 2019-12-12
Small update on how to run individual tests

PhilipMay:
According to @seanpmorgan s comment on gitter I improved the descrition on how to run individual tests.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -90,7 +90,7 @@ bazel test -c opt -k \
 //tensorflow_addons/<package>:<py_test_name>
 ```
 
-`<package>` can be `metrics` or an other package name for example.
+`<package>` can any package name like `metrics` for example.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -96,15 +106,22 @@ or run manually:
...
+
+Run individual tests:
 
+```bash
 bazel test -c opt -k \
 --test_timeout 300,450,1200,3600 \
 --crosstool_top=//build_deps/toolchains/gcc7_manylinux2010-nvcc-cuda10.1:toolchain \
 --test_output=all \
 --jobs=1 \
-//tensorflow_addons/...
```
Squadrick: Let's not delete the command for running all the bazel tests. We can have both, i.e., `//tensorflow_addons/...` and the newer `tensorflow_addons/<package>:<py_test_name>`

PhilipMay: Yes - good point. I fixed that.

PR #760, opened 2019-12-12
Create new AveragedOptimizerWrapper optimizer

### LINKED ISSUE #741: Doubt on using tfa.optimizers.MovingAverage
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Colab
- TensorFlow version and how it was installed (source or binary): 2.0.0
- TensorFlow-Addons version and how it was installed (source or binary): 0.6.0
- Python version: 3.6.0
- Is GPU used? (yes/no): N/A

**Describe the bug**
Would you mind clarify how one should use MovingAverage? Possibly, this is not a bug and it's just that I don't know to use it. 

This is the way I use it: `tfa.optimizers.MovingAverage(tf.keras.optimizers.SGD(0.001))`. When running training, it seems no gradient update occurred: loss is at the same value.

You can find my code here: https://colab.research.google.com/drive/1Q_Jgkf69dUXhfXli4IXANGjUYdZz2xeU#scrollTo=GEZmiZv56Cen.

Squadrick:
`SWA` and `MovingAverage` extend `AverageWrapper` and implement `average_op(var, average_var)`.

Added a new test to close #741


Squadrick:
cc: @shreyashpatodia

tensorflow_addons/optimizers/moving_average.py
```diff
@@ -69,109 +71,30 @@ def __init__(self,
...
-    def lr(self):
-        return self.learning_rate
-
-    @lr.setter
-    def lr(self, lr):
-        self.learning_rate = lr
+    def _create_slots(self, var_list):
+        self._optimizer._create_slots(var_list=var_list)  # pylint: disable=protected-access
+        for var in var_list:
+            self.add_slot(var, 'average', var.read_value())
```
Squadrick: Is using `var.read_value()` the right way to initialize a slot value?

Squadrick: Because this requires us to run a variables initializer twice, once when the model is defined and once after `model.compile()`

Squadrick:
Ready for revieew.

tensorflow_addons/optimizers/average_wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow as tf
+
+
+@tf.keras.utils.register_keras_serializable(package='Addons')
+class AverageWrapper(tf.keras.optimizers.Optimizer):
```
shreyashpatodia: Can we make this more explicit and call it something like `AveragedOptimizerWrapper`?

tensorflow_addons/optimizers/moving_average.py
```diff
@@ -18,10 +18,13 @@
 from __future__ import print_function
 
 import tensorflow as tf
+from tensorflow.python.training.moving_averages import assign_moving_average
```
Squadrick: `assign_moving_average` is currently not exposed via a public API. The implementation is simple enough to include directly into `tfa`, but the upstream `tf` implementation has support for averaging under a distribution strategy which would be desirable to have.

Squadrick: Created an issue to make `assign_moving_average` public: https://github.com/tensorflow/tensorflow/issues/35118

Squadrick:
dc12fdf0837bca5f82c425a74f41f1e4d3052cb7 is a workaround for #816

Squadrick:
@shreyashpatodia I've added a function called `fit_bn` in f283033e754d36d748e6ea2dbbff9ece1e1a33cf, it uses a constant momentum rate compared to using variable momentum [like in the original implementation.](https://github.com/timgaripov/swa/blob/4a2ddfdb2692eda91f2ac41533b62027976c605b/utils.py#L128)

tensorflow_addons/optimizers/utils.py
```diff
@@ -0,0 +1,60 @@
...
+                layer.moving_mean.assign(tf.zeros_like(layer.moving_mean)),
+                layer.moving_variance.assign(
+                    tf.ones_like(layer.moving_variance))
+            ])
+
+    assign_op = tf.group(assign_ops)
+
+    _trainable = model.trainable
+    _metrics = model._metrics
+    model.trainable = False
```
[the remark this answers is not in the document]
Squadrick: BatchNormalization works a little different. `moving_mean` and `moving_variance` aren't trainable variables. So when we do `trainable=False`, they can still be changed during fit. 

I printed the variables before and after `fit` to make sure they are being updated.

[Docs](https://www.tensorflow.org/api_docs/python/tf/keras/layers/BatchNormalization?version=nightly)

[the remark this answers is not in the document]
shreyashpatodia: I don't actually mean the moving_mean and moving_average terms. Sorry for not making my comment clear.

Batch norm does have two trainable parameters which are the `gamma` and the `beta` terms which might need to change since the weights have changed and the moving_mean and moving_variance preceding them is changing as well.

The `gamma` and `beta` terms are canonically used to scale and shift the batch normalization input after the use of the moving_mean and moving_variance.

[Link](https://github.com/keras-team/keras/blob/7a39b6c62d43c25472b2c2476bd2a8983ae4f682/keras/layers/normalization.py#L174) to Keras' call to the backend for a reference for `beta` and `gamma`. Tensorflow [doc](https://www.tensorflow.org/api_docs/python/tf/nn/batch_normalization?version=nightly) for batch normalization.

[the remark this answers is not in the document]
Squadrick: `beta` and `gamma` should not be changed as per [the implementation](https://github.com/timgaripov/swa/blob/4a2ddfdb2692eda91f2ac41533b62027976c605b/utils.py) of the paper's authors. They never do a backprop on the network (which is when `beta` and `gamma` are updated). 

> Batch norm does have two trainable parameters, the gamma and the beta terms which might need to change since the weights have changed and the moving_mean and moving_variance preceding them are changing as well.

The distribution of the activations (before scaling using `gamma` and `beta`) should (_may?_) be the same between using the original weights and averaged weights as long as `moving_mean` and `moving_variance` are updated. At least this is the assumption made when fine-tuning batch norm on new data, and I think the logic carries over to this case as well. 

Also, as soon as `beta` and `gamma` are changed, then all the following layers need to be updated as well. That boils down to another epoch of full network training. 

Thanks for taking the time to do this. Let me know if what I've understood is not correct.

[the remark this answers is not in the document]
shreyashpatodia: Yep. It seems like it `beta` and `gamma` don't change in the author's original implementation and yeah I guess if you think of this as fine-tuning batch norm then this would make sense.

I felt like there is a large enough change in `beta` and `gamma` to allow them to be learnable (without resetting) but I have no evidence to back it up.

Everything looks good otherwise. I'll resolve this comment.

tensorflow_addons/optimizers/moving_average.py
seanpmorgan: I wasn't able to find a better alternative. `.value()` may cause issues for distributed strategies, and didn't see an initializer from existing variable elsewhere.

PR #761, opened 2019-12-13
Temporarily remove stale ppa until image can be updated

seanpmorgan:
Current CI and nightly build fail to run `apt-get update` since this ppa is now stale. Some context:
https://launchpad.net/~jonathonf

Tested the build scripts as well and it should fix our nightly.

### DIFF
--- a/tools/ci_build/builds/release_linux.sh
+++ b/tools/ci_build/builds/release_linux.sh
@@ -15,6 +15,10 @@
 # ==============================================================================
 set -e -x
 
+# Remove the now private ppa. This can be removed after the docker image removes the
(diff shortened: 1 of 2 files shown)


seanpmorgan:
cc @r4nt: Do you happen to know the owner of:
*  `gcr.io/tensorflow-testing/nosla-cuda10.1-cudnn7-ubuntu16.04-manylinux2010` 
*  `gcr.io/tensorflow-testing/nosla-ubuntu16.04-manylinux2010 `

Understand that the image says no SLA, but we might want to update the image since I know addons and io are using this.

PR #765, opened 2019-12-13
add filter response normalization

AakashKumarNain:


AakashKumarNain:
@seanpmorgan @WindQAQ @facaiy I have added the latest contribution for normalization from the Google Brain team. Although the implementation is fully correct IMO but before I go on and write unit-tests for it, can you please take a look at the implementation and let me know if it looks okay. Thank you

tensorflow_addons/layers/normalizations.py
```diff
@@ -316,3 +316,225 @@ def __init__(self, **kwargs):
...
+        return dict(list(base_config.items()) + list(config.items()))
+
+    def call(self, inputs):
+        # Compute the mean norm of activations per channel.
+        nu2 = tf.reduce_mean(tf.square(inputs), axis=self.axis, keepdims=True)
+        # Perform FRN.
+        inputs = inputs * tf.math.rsqrt(nu2 + tf.math.abs(self.epsilon))
+        # Return after applying the Offset-ReLU non-linearity.
+        return tf.maximum(self.gamma * inputs + self.beta,
+                          self.tau + self.eps_var)
```
Squadrick: In the paper, there's only a single use of `epsilon`. Which can either be a constant value of `1e-6` **or** can be a learnable param with the following form: `eps = 1e-6 + abs(eps_l)` where `eps_l` is the param with initial value of `1e-4`. 

You've used two different epsilons: `self.epsilon` (which is constant) and a `self.eps_var` which is a param with zero init. Secondly, the use of `eps` in the calculation of `FRN` is okay. But in the paper, there's no use of `eps` in the TLU whereas you've done:
```
z = max(y, tau + eps)
```
instead of 
```
z = max(y, tau)
```

AakashKumarNain: In most of the cases, the default value of `epsilon=1e-6` would work but not for the architectures where 1x1 conv is involved. In that case, the authors add another learnable parameter `eps_l`, which should be non-negative, so that the optimal epsilon value can be found out. 

Regarding `TLU`, yes that's a typo. Actually the  `eps_var` had to be added to the `epsilon` in the above line. I will correct it. Thanks for looking into it

AakashKumarNain: Another thing we can do is keep only `epsilon` as the trainable parameter with a constraint that it should only be positive and initialized with `1e-6`. Let me know what you think.

tensorflow_addons/layers/normalizations.py
```diff
@@ -316,3 +316,225 @@ def __init__(self, **kwargs):
...
+        base_config = super(FilterResponseNormalization, self).get_config()
+        return dict(list(base_config.items()) + list(config.items()))
+
+    def call(self, inputs):
+        # Compute the mean norm of activations per channel.
+        nu2 = tf.reduce_mean(tf.square(inputs), axis=self.axis, keepdims=True)
+        # Perform FRN.
+        inputs = inputs * tf.math.rsqrt(nu2 + tf.math.abs(self.epsilon))
+        # Return after applying the Offset-ReLU non-linearity.
+        return tf.maximum(self.gamma * inputs + self.beta,
```
saurabhme: It would be better to make the activation configurable so that users can experiment with other combinations. While the paper does demonstrate that FRN+TLU combination works well in a variety of settings, in future we may find that a different combination works even better. Making the activation configurable will future proof the layer.

As a result I would suggest making TLU a separate activation layer similar to ReLU, PReLU etc. This way it can be swapped in/out. For example, I found it useful to replace **all** activations in a network with TLU instead of just those that followed normalization (e.g. in ResnetV1 backbone model used in object detection in paper)..

tensorflow_addons/layers/normalizations.py
saurabhme: As suggested, there shouldn't be an eps in TLU.

In line with my earlier comment, I would suggest making TLU its own layer. Further, I would suggest that the implementation also include a configuration option to produce Affine-TLU (ref. paper). While Affine-TLU was shown to be similar/worse than TLU for image classification, it may still be useful in as yet unexplored applications.

saurabhme: No, that is not a good idea and it did not work well. For applications where it needs to be learnable, the parameterization of 1e-6 + abs(eps_l) worked well (where eps_l was initialized to 1e-4 or larger). Initializing it to small value tended to blow up the learning.

AakashKumarNain: Yeah, It's a typo actually. I will remove it

tensorflow_addons/layers/normalizations.py
AakashKumarNain: That actually makes sense. Okay, I will implement FRN and TLU as separate layers then

tensorflow_addons/layers/normalizations.py
AakashKumarNain: So, we need the constant value of `epsilon=1e-6` as well as the learnable `eps_l` initialized to 1e-4.

saurabhme: Right, the constant value lower bounds the value of effective epsilon, while the higher initialization prevent blow up at the start of the training.

AakashKumarNain: @saurabhme `gamma` and `beta` would be part of `FRN` and not TLU. For TLU, only `tau` is required. Rest everything is part of FRN. Just trying to clear it before I modify this PR. Thanks

saurabhme: Yes. That is correct.

AakashKumarNain: Thanks for the clarification

AakashKumarNain: @saurabhme Sorry to  bug you again but if we implement TLU as a separate layer, how is it diffferent from PReLU? To me, it seems the same until unless I am missing something

saurabhme: PReLU computes
`max(x, ax)`

While TLU computes 
`max(x, t)`

While Affine-TLU computes
`max(x, ax+t)`

Where a, t are learnable parameters. The difference is in the form of function that threshold takes. In PReLU it is a scaling of x, in TLU it is a bias while in Affine-TLU it is a affine function of x.

AakashKumarNain: Cool. Thanks

AakashKumarNain:
@seanpmorgan @WindQAQ @Squadrick @saurabhme I have opened a PR for TLU. Please check #857

Squadrick:
@AakashKumarNain Any reason to put this under `layers` as opposed to `activations`? Also, could you update the `README.md`?

gabrieldemarmiesse:
@googlebot I consent

gabrieldemarmiesse:
@AakashKumarNain I've merged master into your branch to update it and fixed any formatting/conflicts it might have. If you need to do some more modifications, please do `git pull` beforehand.

AakashKumarNain:
@gabrieldemarmiesse this will come with a lot of changes. Will do in the coming weeks. It has been long overdue

AakashKumarNain:
@gabrieldemarmiesse @seanpmorgan the tests work fine in eager mode but are failing in graph mode:
```python
inputs = np.random.rand(28, 28, 1).astype(np.float32)
frn = FilterResponseNormalization(
            beta_initializer="zeros", gamma_initializer="ones"
        )
frn.build((28, 28, 1))
observed = frn(inputs)
expected = self.calculate_frn(inputs, beta=0, gamma=1)
self.assertAllClose(expected, observed) # this is where the test fails
```

Squadrick:
@AakashKumarNain I know the problem. You aren't initializing the variables, you need to add: `        self.evaluate(tf.compat.v1.global_variables_initializer())` after `frn.build(...)`. Also, break `test_random_input()` into multiple smaller function for each instance params of the frn layer. 

Here's a stripped-down version of `test_random_inputs()` that runs without any issues:

```python
    def test_random_inputs(self):
        inputs = np.random.rand(28, 28, 1).astype(np.float32)
        frn = FilterResponseNormalization(
            beta_initializer="zeros", gamma_initializer="ones"
        )
        frn.build((28, 28, 1))
        self.evaluate(tf.compat.v1.global_variables_initializer())  # <-- missing
        observed = frn(inputs)
        expected = self.calculate_frn(inputs, beta=0, gamma=1)
        self.assertAllClose(expected, observed[0])
```

The output log of the successful run;
```
================================================================================
Target //tensorflow_addons/layers:normalizations_test up-to-date:
  bazel-bin/tensorflow_addons/layers/normalizations_test
INFO: Elapsed time: 33.660s, Critical Path: 33.31s
INFO: 2 processes: 2 local.
INFO: Build completed successfully, 2 total actions
//tensorflow_addons/layers:normalizations_test                           PASSED in 33.3s

INFO: Build completed successfully, 2 total actions
```

Also, prefer not to use randomized inputs as inputs for better repro. If you want to use random input, at least use a fixed seed for `np` and `tf`.

AakashKumarNain:
@Squadrick yeah 🤦‍♂ . (With `session` gone, I always forget that part now). And thanks for pointing out the `seed` part. Will update it. Thanks a lot

AakashKumarNain:
@seanpmorgan @Squadrick @gabrieldemarmiesse I think we can merge this now. If you think, any changes are required, please let me know. Thanks

tensorflow_addons/layers/normalizations_test.py
```diff
@@ -21,6 +21,7 @@
 
 from tensorflow_addons.layers.normalizations import GroupNormalization
 from tensorflow_addons.layers.normalizations import InstanceNormalization
+from tensorflow_addons.layers.normalizations import FilterResponseNormalization
```
Squadrick: Order this alphabetically.

tensorflow_addons/layers/normalizations_test.py
```diff
@@ -331,5 +332,73 @@ def test_groupnorm_convnet_no_center_no_scale(self):
         )
 
 
+@test_utils.run_all_in_graph_and_eager_modes
+class FilterResponseNormalizationTest(tf.test.TestCase):
```
Squadrick: Test for serialization.

tensorflow_addons/layers/__init__.py
```diff
@@ -18,6 +18,7 @@
 from tensorflow_addons.layers.maxout import Maxout
 from tensorflow_addons.layers.normalizations import GroupNormalization
 from tensorflow_addons.layers.normalizations import InstanceNormalization
+from tensorflow_addons.layers.normalizations import FilterResponseNormalization
```
Squadrick: Order alphabetically.

tensorflow_addons/layers/normalizations.py
```diff
@@ -321,3 +321,207 @@ def __init__(self, **kwargs):
...
+        gamma_constraint: Optional constraint for the gamma weight.
+        learned_epsilon: (bool) Whether to add another learnable
+        epsilon parameter or not.
+        name: Optional name for the layer
+
+    Input shape
+        Arbitrary. Use the keyword argument `input_shape`
+        (tuple of integers, does not include the samples axis)
+        when using this layer as the first layer in a model. This layer
+        wokrs on a 4-D tensor where the tensor should have the shape [N X H X W X C]
```
Squadrick: Typo: wokrs -> works

tensorflow_addons/layers/normalizations.py
```diff
@@ -321,3 +321,207 @@ def __init__(self, **kwargs):
...
+        gamma_initializer: Initializer for the gamma weight.
+        beta_regularizer: Optional regularizer for the beta weight.
+        gamma_regularizer: Optional regularizer for the gamma weight.
+        beta_constraint: Optional constraint for the beta weight.
+        gamma_constraint: Optional constraint for the gamma weight.
+        learned_epsilon: (bool) Whether to add another learnable
+        epsilon parameter or not.
+        name: Optional name for the layer
+
+    Input shape
```
Squadrick: Can't it be extended to 5D?

tensorflow_addons/layers/normalizations.py
```diff
@@ -321,3 +321,207 @@ def __init__(self, **kwargs):
...
+            raise ValueError(
+                "You are trying to normalize your batch axis. You may want to "
+                "use tf.layer.batch_normalization instead"
+            )
+        if self.axis == -1:
+            raise ValueError(
+                "You are trying to normalize your channel axis. You may want to "
+                "use GroupNormalization layer instead from tf_addons"
+            )
+        if self.axis != [1, 2]:
```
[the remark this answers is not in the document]
Squadrick: I see you've hardcoded it to `[1, 2]` in `calculate_frn()` in the test file.

tensorflow_addons/layers/normalizations_test.py
```diff
@@ -331,5 +332,73 @@ def test_groupnorm_convnet_no_center_no_scale(self):
         )
 
 
+@test_utils.run_all_in_graph_and_eager_modes
```
gabrieldemarmiesse: We're trying to get rid of the `run_all_in_graph_and_eager_mode` which is a private decorator. You can look at #1328 for the instructions on how to update. It's fairly easy and you can ping me if you have issues.

AakashKumarNain: Will do that. Thanks

tensorflow_addons/layers/normalizations.py
AakashKumarNain: Theoretically it can be  but in paper they have proved it for 2D CNNs, so I think we can put 3D CNNs for hold as of now.

tensorflow_addons/layers/normalizations.py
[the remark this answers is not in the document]
AakashKumarNain: So, in the [paper](https://arxiv.org/pdf/1911.09737.pdf), the concept was proved only for 2D CNNs. We don't know if it works in other scenarios like 3D CNNs or for RNNs. But I am not saying that it couldn't work. I replicated the method in the paper. Whether we want to make it flexible to work for any scenario (assuming user understands when and how to use it), or not, is totally on us. Lemme know what you think

tensorflow_addons/layers/normalizations.py
```diff
@@ -321,3 +321,207 @@ def __init__(self, **kwargs):
...
+
+        if self.use_eps_learned:
+            self.eps_learned_initializer = tf.keras.initializers.Constant(1e-4)
+        else:
+            self.eps_learned_initializer = None
+
+        if isinstance(axis, list):
+            self.axis = axis[:]
+        elif isinstance(axis, int):
+            self.axis = axis
```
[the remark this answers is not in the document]
AakashKumarNain: Nice catch.

tensorflow_addons/layers/normalizations.py
```diff
@@ -321,3 +321,207 @@ def __init__(self, **kwargs):
...
+        else:
+            raise TypeError("axis must be int or list, type given: %s" % type(axis))
```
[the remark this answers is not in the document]
AakashKumarNain: Donr

tensorflow_addons/layers/normalizations.py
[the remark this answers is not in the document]
Squadrick: Sounds good. Can you add a TODO? We can get to it later. But in that case, parameterizing `axis` isn't needed.

tensorflow_addons/layers/normalizations.py
```diff
@@ -321,3 +321,207 @@ def __init__(self, **kwargs):
...
+        epsilon = tf.math.abs(self.epsilon)
+        if self.use_eps_learned:
+            epsilon += self.eps_learned
+        nu2 = tf.reduce_mean(tf.square(inputs), axis=self.axis, keepdims=True)
+        inputs *= tf.math.rsqrt(nu2 + epsilon)
+        return self.gamma * inputs + self.beta
+
+    def _create_input_spec(self, input_shape):
+        ndims = len(tf.TensorShape(input_shape))
+        for idx, x in enumerate(self.axis):
```
Squadrick: `self.axis` can be an `int`. Doing an `enumerate` of `int` will lead to an error:
```python
TypeError: 'int' object is not iterable
```

Squadrick: This issue can be fixed with the `isinstance` check in `__init__`.

tensorflow_addons/layers/normalizations.py
```diff
@@ -321,3 +321,207 @@ def __init__(self, **kwargs):
...
+                "use GroupNormalization layer instead from tf_addons"
+            )
+        if self.axis != [1, 2]:
+            raise ValueError(
+                "FilterResponseNormalization operates on per-channel basis."
+                "Axis values should be spatial dimensions"
+            )
+
+    def _check_if_input_shape_is_none(self, input_shape):
+        dim1, dim2 = input_shape[self.axis[0]], input_shape[self.axis[1]]
```
Squadrick: This assumes that `len(self.axis) == 2`, but that isn't asserted anywhere. Also, what about when `axis` is an int?

tensorflow_addons/layers/normalizations.py
```diff
@@ -321,3 +321,207 @@ def __init__(self, **kwargs):
...
+        shape = [1, 1, 1, dim]
+        # Initialize beta with shape (1, 1, 1, C)
+        self.beta = self.add_weight(
+            shape=shape,
+            name="beta",
+            dtype=self.dtype,
+            initializer=self.beta_initializer,
+            regularizer=self.beta_regularizer,
+            constraint=self.beta_constraint,
+        )
```
Squadrick: The 4D tensor shape of `gamma` and `beta` will only work for batched images. It won't work for output from FC. 

We have two ways forward:
1. Do an assert on `input_shape` to ensure it is 4D, and add a TODO to add support for more shapes in the future.
2. Modify the code to support various sizes. 

I'm in favour of 1 to reduce the scope of this PR.

tensorflow_addons/layers/normalizations.py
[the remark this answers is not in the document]
AakashKumarNain: Yeah in that case, we won't need the `axis` parameter. Will do that. Thanks

tensorflow_addons/layers/normalizations.py
AakashKumarNain: Will resolve this

tensorflow_addons/layers/normalizations.py
AakashKumarNain: Yes. Thanks

tensorflow_addons/layers/normalizations.py
AakashKumarNain: I am also in favor of 1 for now

tensorflow_addons/layers/normalizations.py
Squadrick: Also, can you explain how this would work for fully connected layers? 

The output of an FC layer has shape `[B, N]` (where B is the batch shape, and N is the number of output neurons). Doing an operation (`tf.reduce_mean` in this case) over the axes `[1, 2]` will lead to an error.

tensorflow_addons/layers/normalizations.py
AakashKumarNain: We are already checking axis values for integer values which would raise `VlaueError`, hence this check wouldn't be required

tensorflow_addons/layers/normalizations.py
AakashKumarNain: Same case. When checking for axis values, we are making sure that it's a list for now

tensorflow_addons/layers/normalizations_test.py
AakashKumarNain: Done

AakashKumarNain:
@gabrieldemarmiesse @Squadrick I have made all the changes. Also, I am using only `pytest` for all tests now (much better). Let me know if any other changes are required

tensorflow_addons/layers/normalizations.py
```diff
@@ -321,3 +321,214 @@ def __init__(self, **kwargs):
...
+        super().__init__(name=name, **kwargs)
+        self.epsilon = tf.cast(epsilon, dtype=self.dtype)
+        self.beta_initializer = tf.keras.initializers.get(beta_initializer)
+        self.gamma_initializer = tf.keras.initializers.get(gamma_initializer)
+        self.beta_regularizer = tf.keras.regularizers.get(beta_regularizer)
+        self.gamma_regularizer = tf.keras.regularizers.get(gamma_regularizer)
+        self.beta_constraint = tf.keras.constraints.get(beta_constraint)
+        self.gamma_constraint = tf.keras.constraints.get(gamma_constraint)
+        self.use_eps_learned = learned_epsilon
+        self.supports_masking = True
```
gabrieldemarmiesse: If this layer supports masking, maybe we should have a simple test for that?

tensorflow_addons/layers/normalizations.py
```diff
@@ -321,3 +321,214 @@ def __init__(self, **kwargs):
...
+                """Only 4-D tensors (CNNs) are supported
+        as of now."""
+            )
+        self._check_if_input_shape_is_none(input_shape)
+        self._create_input_spec(input_shape)
+        self._add_gamma_weight(input_shape)
+        self._add_beta_weight(input_shape)
+
+        if self.use_eps_learned:
+            self._add_eps_learned_weight()
```
gabrieldemarmiesse: This can be moved to the __init__ because we don't depend on the input shape.

AakashKumarNain: Cool. Will do that.

tensorflow_addons/layers/normalizations.py
AakashKumarNain: Ok

AakashKumarNain: Right now the layer supports only 4D(CNNs) only, so testing masking isn't meaningful right now but in the future we are planning to add support for more shapes. Once that is done, we can add test for it then

gabrieldemarmiesse: Then I propose that we set `self.supports_masking = False` until we supports shapes where it's meaningful to have a mask. What do you think?

tensorflow_addons/layers/normalizations.py
AakashKumarNain: Done. Thanks

tensorflow_addons/layers/normalizations.py
AakashKumarNain: I can do that for sure.

tensorflow_addons/layers/normalizations.py
```diff
@@ -321,3 +321,208 @@ def __init__(self, **kwargs):
...
+        self._check_if_input_shape_is_none(input_shape)
+        self._create_input_spec(input_shape)
+        self._add_gamma_weight(input_shape)
+        self._add_beta_weight(input_shape)
+        super().build(input_shape)
+
+    def call(self, inputs):
+        epsilon = tf.math.abs(self.epsilon)
+        if self.use_eps_learned:
+            epsilon += self.eps_learned
```
Squadrick: This is incorrect. The paper mentions that `self.eps_learned` should be `abs`. We can add a check in `__init__` to ensure that `epsilon` passed by the user is also positive. 
```suggestion
        epsilon = self.epsilon
        if self.use_eps_learned:
            epsilon += tf.math.abs(self.eps_learned)
```

tensorflow_addons/layers/normalizations.py
```diff
@@ -321,3 +321,208 @@ def __init__(self, **kwargs):
...
+        nu2 = tf.reduce_mean(tf.square(inputs), axis=self.axis, keepdims=True)
+        inputs *= tf.math.rsqrt(nu2 + epsilon)
```
Squadrick: IIRC, tensors are passed by ref. So `inputs` will be changed. Also, it's easier to reason about `call` if it's purely functional. 

So avoid using `*=`, create a new tensor instead.

tensorflow_addons/layers/normalizations.py
AakashKumarNain: There is a `constraint` associated with this parameter that makes sure that it is always non-negative. Isn't that enough?

tensorflow_addons/layers/normalizations_test.py
```diff
@@ -331,5 +332,100 @@ def test_groupnorm_convnet_no_center_no_scale(self):
...
+@pytest.mark.parametrize("dtype", [np.float16, np.float32, np.float64])
+def test_serialization(dtype):
+    frn = FilterResponseNormalization(
+        beta_initializer="ones", gamma_initializer="ones", dtype=dtype
+    )
+    serialized_frn = tf.keras.layers.serialize(frn)
+    new_layer = tf.keras.layers.deserialize(serialized_frn)
+    assert frn.get_config() == new_layer.get_config()
+
+
```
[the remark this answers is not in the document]
AakashKumarNain: Yeah I can do that

tensorflow_addons/layers/normalizations.py
```diff
@@ -321,3 +321,208 @@ def __init__(self, **kwargs):
...
+            if x < 0 or x >= ndims:
+                raise ValueError("Invalid axis: %d" % x)
+
+        if len(self.axis) != len(set(self.axis)):
+            raise ValueError("Duplicate axis: %s" % self.axis)
+
+        axis_to_dim = {x: input_shape[x] for x in self.axis}
+        self.input_spec = tf.keras.layers.InputSpec(ndim=ndims, axes=axis_to_dim)
+
+    def _check_axis(self, axis):
```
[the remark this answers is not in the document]
AakashKumarNain: `axis` would be int when we add support for `FC` layer, that's why I didn't remove these checks but if you want, I can remove them

tensorflow_addons/layers/normalizations.py
AakashKumarNain: Cool

tensorflow_addons/layers/normalizations.py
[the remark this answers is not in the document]
Squadrick: Remove them now, we can add it later when we add support for FC.

tensorflow_addons/layers/normalizations.py
Squadrick: My bad didn't see the `non_neg` constraint. You can ignore the `abs` around `self.eps_learned`. 

Weirdly enough, `non_neg` constraint doesn't use `abs`, it does [this](https://github.com/tensorflow/tensorflow/blob/5943793102a96fb47e5bc007576e1f7f41600bb6/tensorflow/python/keras/constraints.py#L94) instead. No idea why.

AakashKumarNain: Actually that is correct. `non-neg` mathematically means a lower bound of zero. There should be an `abs` for epsilon instead of this constraint.

tensorflow_addons/layers/normalizations_test.py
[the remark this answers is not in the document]
AakashKumarNain: Done

Squadrick:
Created #1441 for adding more tensor shape support to this layer.

Squadrick:
Unblocks #1441

AakashKumarNain:
Thanks @Squadrick . Will work on #1441 now.

Squadrick:
@AakashKumarNain, I think someone else has started working on it already. You can check the comments in #1441.

PR #776, opened 2019-12-17
as_numpy_dtype is a property, not a method

brianwa84:
This makes it more visually-explicit that as_numpy_dtype(0) returns a scalar value, not a dtype.

### DIFF
--- a/tensorflow_addons/image/transform_ops_test.py
+++ b/tensorflow_addons/image/transform_ops_test.py
@@ -121,7 +121,7 @@ class ImageOpsTest(tf.test.TestCase):
         for dtype in _DTYPES:
             image = tf.constant([[1, 2], [3, 4]], dtype=dtype)
             self.assertAllEqual(
-                np.array([[4, 4], [4, 4]]).astype(dtype.as_numpy_dtype(0)),
+                np.array([[4, 4], [4, 4]]).astype(dtype.as_numpy_dtype),
                 transform_ops.transform(image, [1] * 8))
 
     def test_transform_eager(self):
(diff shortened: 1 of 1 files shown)


tensorflow_addons/image/transform_ops_test.py
```diff
@@ -121,7 +121,7 @@ def test_transform_data_types(self):
         for dtype in _DTYPES:
             image = tf.constant([[1, 2], [3, 4]], dtype=dtype)
             self.assertAllEqual(
-                np.array([[4, 4], [4, 4]]).astype(dtype.as_numpy_dtype()),
+                np.array([[4, 4], [4, 4]]).astype(dtype.as_numpy_dtype(0)),
```
WindQAQ: I wonder if using `dtype.as_numpy_dtype` is enough, which yields a `numpy.dtype` like `tf.int32` -> `np.int32`.

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/optimizer_v2/adagrad_test.py#L77

tensorflow_addons/image/transform_ops_test.py
```diff
@@ -140,7 +140,7 @@ def test_zeros(self):
                     image = tf.zeros(shape, dtype)
                     self.assertAllEqual(
                         transform_ops.rotate(image, angle),
-                        np.zeros(shape, dtype.as_numpy_dtype()))
+                        np.zeros(shape, dtype.as_numpy_dtype(0)))
```
WindQAQ: Sorry for my mistake. Please modify this place too.

Issue #778, opened 2019-12-18
tfa.text.crf.crf_log_likelihood produce log likelihood > 0 for sequences with length 1


Ghostvv:
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): `mac os`
- TensorFlow version and how it was installed (source or binary): `2.0` from pip
- TensorFlow-Addons version and how it was installed (source or binary): `0.6.0` from pip
- Python version: `3.7`
- Is GPU used? (yes/no): no

**Describe the bug**

If there an example in a batch, which true sequence length is 1, log-likelihood produced by `tfa.text.crf.crf_log_likelihood` can be positive for these examples.

Ghostvv:
created a fix: https://github.com/tensorflow/addons/pull/779

Squadrick:
@Ghostvv Thanks for the fix!

PR #779, opened 2019-12-18
fix crf for sequences of length 1

Ghostvv:
fixes https://github.com/tensorflow/addons/issues/778, where sequences of true length 1 got positive likelihood


Ghostvv:
it seems to me that the test that failed `attention_wrapper_test` is not related to my code changes

seanpmorgan:
Yeah not related to this PR... it's also failing on nightly tests:
https://github.com/tensorflow/addons#nightly-tests

At first it looked like just a flaky test but it might be something more.

PR #796, opened 2019-12-20
Mark filters_test as flaky

seanpmorgan:
This test occasionally fails just outside of tolerance. Marking as flaky:
https://docs.bazel.build/versions/master/be/common-definitions.html#test.flaky


Squadrick:
@seanpmorgan Any idea what in filters_test leads to the flakiness?

PR #798, opened 2019-12-20
Add labeler workflow

seanpmorgan:


seanpmorgan:
So this works, but we're getting an issue due to:
https://github.com/actions/labeler/issues/12

There are some workarounds which i'll look into later.

Issue #787, opened 2019-12-19
Add nightly tests for windows/macos


seanpmorgan:
Currently we only test our nightlies on linux:
https://github.com/tensorflow/addons/blob/master/.travis.yml#L17

It should be relatively simple to enable tests for macos/windows, with the one caveat that `tf-nightly` is not published for windows.

PR #799, opened 2019-12-20
Add github action to test with macos

seanpmorgan:
MacOS piece of #787.


seanpmorgan:
@tensorflow/sig-addons-maintainers ready for review.

I'll submit a  PR to include the test badge on the repo after the first nightly is completed.

tools/ci_testing/addons_cpu.sh
```diff
@@ -44,7 +44,7 @@ export TF_NEED_CUDA=0
 export PYTHON_BIN_PATH=`which python`
 ls -alh $PYTHON_BIN_PATH
 # Use default configuration here.
-yes 'y' | ./configure.sh
+echo 'y' | ./configure.sh
```
facaiy: A little curious: why use echo instead here?

seanpmorgan: The windows shell was hanging when using 'yes' to step through configure.sh. Because linux/mac/windows use `addons_cpu.sh` this was switched.

https://github.com/tensorflow/addons/blob/master/tools/ci_build/builds/release_windows.sh#L31

seanpmorgan: It's a good call-out though, having to pass a single echo is a bit brittle since it only works when `TF_NEED_CUDA` is set. Open to changes for sure

Issue #743, opened 2019-12-06
Drop Python2 support


seanpmorgan:
TF2.1 will be the last package to support python2. We should follow suit and remove support as well.

Some things that we can change off the top of my head:
- [ ]  Removal of `__future__` boilerplate
- [ ]  Type checking (to start maybe just for examples)
- [x]  Clean-up build

gabrieldemarmiesse:
I'd be happy to take care of it. I've done it a few time and I know a lot of tools to make it easier.

Ping me when you'd like to drop support. I suppose it'll be when we drop support for the TF2.1.x branch right? So some time around march/april 2020?

seanpmorgan:
> I'd be happy to take care of it. I've done it a few time and I know a lot of tools to make it easier.
> 
> Ping me when you'd like to drop support. I suppose it'll be when we drop support for the TF2.1.x branch right? So some time around march/april 2020?

That'd be great, thanks @gabrieldemarmiesse! There are a couple of peculiarities with the build that I can modify separately from the python code updates.

As far as timeline we plan to cut our 0.7 branch and publish before 2020. After that we'll drop support on master so sometime in early January.

ulf1:
also replace 

```
return dict(list(base_config.items()) + list(config.items()))
```

by 

```
return {**base_config, **config}
```

gabrieldemarmiesse:
Good call, I'm on it.

gabrieldemarmiesse:
Pyupgrade on all python files:
#878 
#878
#878
#878
#878
#878

gabrieldemarmiesse:
We can close this issue. We'll make another issue to track the typing of the API.

PR #804, opened 2019-12-21
Improve  contributing documentation for updated bazel

seanpmorgan:
Py3 is the default test ran in bazel now, but configure.sh will install reqs to `python`. 

In the very near future, we'll drop py2 support and `configure.sh` can install reqs to python3. #743


seanpmorgan:
Closing as it's included in #808

Issue #335, opened 2019-07-06
Request for examples: Seq2seq


seanpmorgan:
I believe there are many seq2seq examples already published, but it makes sense that we at least have one or two in our examples section for user experience.

Related:
https://google.github.io/seq2seq/
https://github.com/tensorflow/addons/blob/master/examples/README.md

kazemnejad:
I can pick this up if it is ok.

Some ideas to start the example based on:
- For machine translation:
  - On toy English-to-German dataset (I think German-to-English has better user experience since we can demonstrate the outputs better) from http://opennmt.net/OpenNMT-tf/quickstart.html which contains 10k tokenized sentences.
  - Trivial seq2seq LSTM architecture with Attention Mechanism 
  - Use beam decoding to generate outputs
  - Maybe BLEU calculation

seanpmorgan:
> I can pick this up if it is ok.
> 
> Some ideas to start the example:
> 
> * For machine translation:
>   
>   * On toy English-to-German dataset (I think German-to-English has better user experience since we can demonstrate the outputs better) from http://opennmt.net/OpenNMT-tf/quickstart.html which contains 10k tokenized sentences.
>   * Trivial seq2seq LSTM architecture with Attention Mechanism
>   * Use beam decoding to generate outputs
>   * Maybe BLEU calculation

That'd be great! Thanks and welcome to Addons!

PyExtreme:
Hi @seanpmorgan @kazemnejad , Is it still an  _Open Issue_? 
Asking because I am interested to work on it.

kazemnejad:
Hi @PyExtreme, Thank you for showing your interest. 
Yes, actually I'm currently working on it. I was waiting for the #375 to be fixed and it got fixed 3 days ago in #503. I think I can submit a PR in the next few days. Thus you can work on that PR if you want.

matthen:
+1 any progress on this? 
I'd appreciate an example of how to create and train a keras decoder with attention

kazemnejad:
Hi @Mainak431,  Sorry for the inconvenience.

[...]

Issue #485, opened 2019-09-08
Testing Jupyter / Colab Notebooks 


PhilipMay:
From last SIG Addons community meeting I picked up this task.

We want to discuss and implement a way to test jupyter / colab notebooks.
All discussions and ideas should go to this issue.

Furthermore the protocol sais that:
- TF team is looking at ways to test doc strings (not sure if this should go to an other issue)
- May be able to inline doc string test cases (not sure if this should go to an other issue)
- First pass can we just test to make sure the notebook executes
- Be sure to tag @karmel

PhilipMay:
I am not even sure which jupyter / colab notebooks should be tested. I guess these here:
https://github.com/tensorflow/addons/tree/master/examples

Is that right?

PhilipMay:
Should I open a separate issue about docstring testing?

PhilipMay:
Since these are norml jupyter notebooks we can smoke test them by just running them. This is possible with: `nbconvert --ExecutePreprocessor.timeout=None --to notebook --execute <noteboo_name>.ipynb`

Also see here: https://nbconvert.readthedocs.io/en/latest/execute_api.html#executing-notebooks-from-the-command-line

If this fails because some exception is raised the smoke test does not pass. But I am not sure how to integrate this into travis...

What do you think?

seanpmorgan:
> Since these are norml jupyter notebooks we can smoke test them by just running them. This is possible with: `nbconvert --ExecutePreprocessor.timeout=None --to notebook --execute <noteboo_name>.ipynb`
> 
> Also see here: https://nbconvert.readthedocs.io/en/latest/execute_api.html#executing-notebooks-from-the-command-line
> 
> If this fails because some exception is raised the smoke test does not pass.
> Here is are links where they do exactly that:
> 
> 1. example: https://github.com/ghego/travis_anaconda_jupyter
> 2. notebook testing lib: https://github.com/pyviz-dev/nbsmoke
> 3. notebook testing lib: https://github.com/opengeophysics/testipynb
> 
> My suggestion is to test the libs (link 2 and 3) and see if they fit our needs and decide which to use. The ppl. from link 2 (nbsmoke) say they are still "still undergoing heavy development". Lib 3. (testipynb) seems to be more mature to me.
> 
> What do you think?

Thanks for taking this on @PhilipMay. First impression is that either of the two libs looks like it could work for our first pass.

karmel:
[...]

PR #806, opened 2019-12-22
Sequence-to-Sequence NMT

Om-Pandey:
@seanpmorgan with reference to #335 , I have created the tutorial, please check and merge the files. Thanks!


seanpmorgan:
@Om-Pandey Thanks very much for the contribution. Looks very good, but would it be possible to utilize a dataset found in tfds?
https://www.tensorflow.org/datasets/catalog/overview

Because our tutorials will be tested for correctness #485, we need these to run end-to-end.

Om-Pandey:
@seanpmorgan yeah, I saw the datasets, I realize that I will have to change the entire data reading and cleaning process for them to conduct tokenization, even if I keep the model constant. Secondly, the translational datasets on https://www.tensorflow.org/datasets/catalog/overview are too large ≈ 1.5 GB each. I really don't have the kind of compute for it. 😅😬

But, I can assure you that the Dataset I used is pretty authentic([link](http://www.manythings.org/anki/deu-eng.zip)) and that my entire code has been tested on this so in case of testing for correctness there shouldn't be any errors. A link to the dataset has also been given in the notebook for reference. But, I can compile the cells again and re-upload along with results if you want, it will just make the notebook bigger though.

Check 'em:
t2
t1

seanpmorgan:
> Removed a few of errors in display pipeline and removed unnecessary spaces

Thanks @Om-Pandey! So yeah having the results within the notebook is ideal for users to read through. You can remove the output from any cells that you don't think is necessary for understanding.

In order to run the code in Colab, you'll need to download the data. You can run commands on the colab host by prefixing with `!`. For example: `!wget --quiet https://dataset.tar.gz`

If the full notebook can run end-to-end then we can include this as a sort of "test" that the tf-docs team will run at each release.

How long is the total runtime of this example on a Colab instance?

Om-Pandey:
@seanpmorgan yeah, so final changes have been made, I have added a lot of comments, which I think are sufficient for anyone to understand the code properly. As you said, I also coded-in the data downloading and unzipping processes on Colab, in my recent commit.
I just checked the notebook typically takes less than a minute to run, except the variable training time - but it typically takes 22secs per epoch, compute that by whatever number of epochs you find necessary to train. Finally, the testing and final prediction take about a minute too.

John-8704:
@Om-Pandey Will you be implementing Attention Mechanism, Beam Search, Bi-directional encoder-decoder model? It's just that your notebook is not at all using TensorFlow Addons

PR #820, opened 2019-12-31
Rrelu fix

fsx950223:
1.Use public tensorflow api instead of internal api
2.Fix test bug in graph mode(It's impossible to test random in graph mode for now.For details, see [gradient_checker_v2](https://github.com/tensorflow/tensorflow/blob/1cf0898dd4331baf93fe77205550f2c2e6c90ee5/tensorflow/python/ops/gradient_checker_v2.py#L243-L253))


Squadrick:
@fsx950223 Any reason for switching from C++ ops to python ops? Because the native C++ op would be more desirable due to speed. Since they already exist, I don't why we'd revert back to python ops.

fsx950223:
> @fsx950223 Any reason for switching from C++ ops to python ops? Because the native C++ op would be more desirable due to speed. Since they already exist, I don't why we'd revert back to python ops.

The internal api is unstable and c++ op doesn't work on GPU(It returns zeros when it runs on GPU).

Squadrick:
The proper fix would be to figure out why it doesn't work on GPU, and opening an issue of TensorFlow if there's some error on their end. As for the unstable API, again, create an issue. 

In the meantime, include both C++ and Python ops. For now, use the python version, but have a `TODO` and an issue to switch to the C++ version once the issues upstream are fixed. 

Could you run some benchmarks (maybe CPU only since GPU seems to not work) between C++ and Python? If the difference is negligible, we can ditch the C++ version entirely. But until we get some numbers, I'm in favour of fixing the underlying problem with the C++ version before prematurely porting it to Python. 

cc activation maintainers @seanpmorgan @facaiy

fsx950223:
I try it.

Dheeraj R Reddy <notifications@github.com<mailto:notifications@github.com>> 于 2019年12月31日周二 08:04写道：

The proper fix would be to figure out why it doesn't work on GPU, and opening an issue of TensorFlow if there's some error on their end. As for the unstable API, again, create an issue.

In the meantime, include both C++ and Python ops. For now, use the python version, but have a TODO and an issue to switch to the C++ version once the issues upstream are fixed.

Could you run some benchmarks (maybe CPU only since GPU seems to not work) between C++ and Python? If the difference is negligible, we can ditch the C++ version entirely. But until we get some numbers, I'm in favour of fixing the underlying problem with the C++ version before prematurely porting it to Python.

cc activation maintainers @seanpmorgan<https://github.com/seanpmorgan> @facaiy<https://github.com/facaiy>

tensorflow_addons/activations/rrelu_test.py
```diff
@@ -24,54 +24,58 @@
...
-        self.assertAllCloseAccordingToType(result, expect_result)
+
+        training_results = {
+            np.float16: [-0.288330078, -0.124206543, 0, 1, 2],
+            np.float32: [-0.26851666, -0.116421416, 0, 1, 2],
+            np.float64: [-0.3481333923206531, -0.17150176242558851, 0, 1, 2],
+        }
+        for training in [True, False]:
+            with self.subTest(training=training):
+                tf.random.set_seed(SEED)
```
facaiy: Is it necessary?

facaiy:
Thanks for pinging me, Dheeraj :-)

Thank you for the PR, @fsx950223 . I'm fine with the idea: replace c++ op by tf python op. What do you think, @seanpmorgan @WindQAQ ?

tensorflow_addons/activations/rrelu_test.py
fsx950223: Yes, according to https://github.com/tensorflow/tensorflow/issues/33034#issuecomment-539752847

WindQAQ:
> Thanks for pinging me, Dheeraj :-)
> 
> Thank you for the PR, @fsx950223 . I'm fine with the idea: replace c++ op by tf python op. What do you think, @seanpmorgan @WindQAQ ?

As a fallback option, I'm fine with this :P

fsx950223:
The benchmark test doesn't work, could someone give me a hand?

fsx950223:
The benchmark has been added and execute time is acceptable.

PR #823, opened 2020-01-02
Remove tf function where unecessary

### LINKED ISSUE #807: tf.function bug
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04):
- TensorFlow version and how it was installed (source or binary):
- TensorFlow-Addons version and how it was installed (source or binary):
- Python version:
- Is GPU used? (yes/no):

**Describe the bug**
tf.function will trigger tf.function retracing several times if user convert python types into the tf.function which will cause memory leak.

Invalid usage of tf.function leads memory leak 
https://github.com/tensorflow/addons/blob/41a286cc7e2d46913b9e7e1b9f2392d50ea417e3/tensorflow_addons/activations/softshrink.py#L29

Provide a reproducible test case that is the bare minimum necessary to generate the problem.
bazel test -c opt -k --test_timeout 300,450,1200,3600 --crosstool_top=//build_deps/toolchains/gcc7_manylinux2010-nvcc-cuda10.1:toolchain --test_output=all --jobs=1 //tensorflow_addons/activations:softshrink_test

**Other info / logs**
WARNING:tensorflow:9 out of the last 27 calls to <function softshrink at 0x7f37439a28c8> triggered tf.function retracing. Tracing is expensive and the excessive number of tracings is likely due to passing python objects instead of tensors. Also, tf.function has experimental_relax_shapes=True option that relaxes argument shapes that can avoid unnecessary retracing. Please refer to https://www.tensorflow.org/tutorials/customization/performance#python_or_tensor_args and https://www.tensorflow.org/api_docs/python/tf/function for more details.

Include any logs or source code that would be helpful to diagnose the problem. If including tracebacks, please include the full traceback. Large logs and files should be attached.

seanpmorgan:
Closes #807 

Removed from:
* Functions that are simple calls to a custom op
* Functions that are likely to be re-traced


WindQAQ:
Just a note, have to update the README.md about `tf.function` description.

seanpmorgan:
> Thanks for cleanup! My only concern is that `test_unknown_shape` for image tests is designed for `tf.data.Dataset`. If we get rid of `tf.function`, will it always be compatible for `tf.data.Dataset`? Moreover, a old thread [tensorflow/tensorflow#27811](https://github.com/tensorflow/tensorflow/issues/27811) mentioned that `tf.function` will be apply automatically for `tf.data.Dataset.map` for TF2.0.
> 
> A trivial test is to use something like
> 
> ```python
> fn = tf.function(dist_ops.euclidean_dist_transform).get_concrete_function(
>                  tf.TensorSpec(None, tf.uint8))
> ```
> 
> to make sure it would be compatible with `tf.data.Dataset`.
> 
> What do you think, Sean?
> 
> #330 #332

Thanks good catch.. i've added them back. I was also thinking we should have subpackage test suites similar to `keras.layer_test`. I'll create a seperate issue for that.

PR #824, opened 2020-01-02
BLD: fix nightlies

seanpmorgan:
Fix three issues in the nightly build so that all 3 OS whl's will build.

* `configure.sh` depends on env variable so macos_build script needs to run as the current user
* requirements.txt was not present in the build directory
* windows nightly [compile fails](https://travis-ci.org/seanpmorgan/addons/jobs/631914108#L1922) if cmath includes are not present. Strangely this doesn't happen on tf2.1 release candidates. 

Successful build:
https://travis-ci.org/seanpmorgan/addons/builds/632071006


tensorflow_addons/custom_ops/activations/cc/kernels/hardshrink_op.h
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 #define TENSORFLOW_ADDONS_ACTIVATIONS_KERNELS_HARDSHRINK_OP_H_
 
 #define EIGEN_USE_THREADS
+#define _USE_MATH_DEFINES
+#include <cmath>
```
WindQAQ: Just to confirm if this is for windows build? Thanks!

seanpmorgan: Yes. It fails to compile otherwise for a core lib missing the include (log in original post but it's a bit verbose). It didn't happen on the RC, only nightly which is strange.

I can mark an issue to investigate and remove these later if we can get a fix on tf core

seanpmorgan: Also working on windows CI sho it'll be easier to test and review these windows issues

seanpmorgan:
Looked closer at the logs and realized the includes were only necessary for ops using random from core lib. Sorry about that, it was a bit cryptic to see what was triggering the fail.

New passing build:
https://travis-ci.org/seanpmorgan/addons/builds/632274218

tensorflow_addons/custom_ops/activations/cc/kernels/hardshrink_op.h
WindQAQ: Seems that core TF does the same thing recently.

https://github.com/tensorflow/tensorflow/search?q=_USE_MATH_DEFINES&type=Commits

Issue #768, opened 2019-12-14
Verify Tutorial Testing


seanpmorgan:
Per the conversation in #540 the TF-Docs team should be able to test our tutorials during API document generation. 

We'll be adding a broken notebook so we can test the alerting system.

yashk2810:
Let me know when 0.7 is available on pypi, so that I can run the commands.

seanpmorgan:
@yashk2810 0.7 has been released and we added a broken tutorial to test the alerts:
https://github.com/tensorflow/addons/blob/master/docs/tutorials/temp_broken_tutorial.ipynb

yashk2810:
Mark ran the build a couple of days ago and sent it to `addons-testing@tensorflow.org` but the mail didn't go through. Can you confirm that the google group is set up?

yashk2810:
Please see https://groups.google.com/a/tensorflow.org/g/addons/c/F6MrrMfVVBA

seanpmorgan:
@yashk2810 so we never received any notification on the google group:
https://groups.google.com/a/tensorflow.org/forum/#!forum/addons-testing

Perhaps this is because of the permissions regarding who is able to post? I noticed that travis builder seemed to have automatically joined the group:
https://groups.google.com/a/tensorflow.org/forum/#!members/addons-testing

yashk2810:
Please see https://groups.google.com/a/tensorflow.org/g/addons/c/F6MrrMfVVBA. I couldn't post to that group, so Mark posted it here instead.

Is that good enough?

seanpmorgan:
> Please see https://groups.google.com/a/tensorflow.org/g/addons/c/F6MrrMfVVBA. I couldn't post to that group, so Mark posted it here instead.
> 
> Is that good enough?

Yes the post from Mark allows us to fix the issues, but we would like this notification to be automated

yashk2810:
>  but we would like this notification to be automated
Making the notification automatic is a difficult task (for sending emails externally). Whenever the notebook fails, we will email you, so I think it should be fine.

seanpmorgan:
> > but we would like this notification to be automated
> 
> Making the notification automatic is a difficult task (for sending emails externally). Whenever the notebook fails, we will email you, so I think it should be fine.

Okay that is sufficient as long as Docs team can support that. Please let us know if/when the test code is open-sourced so we can manage this internally for convenience.

PR #832, opened 2020-01-04
Add broken tutorial do docs for testing verification

seanpmorgan:
For verification as part of #768


WindQAQ:
umm, should we merge this now to test the system? does that work only on master branch?

seanpmorgan:
> umm, should we merge this now to test the system? does that work only on master branch?

Yeah it'll test master after we instruct them to pull the API docs. Perhaps just approve and then whomever cuts the branch can merge this prior?

PR #836, opened 2020-01-05
Add Novograd Optimizer

shreyashpatodia:
This PR adds in the `Novograd Optimizer` for `tensorflow/addons`.

I have tried to keep my implementation of things as similar to the [original implementation](https://github.com/NVIDIA/OpenSeq2Seq/blob/master/open_seq2seq/optimizers/novograd.py) as possible.


tensorflow_addons/optimizers/novograd.py
```diff
@@ -0,0 +1,165 @@
...
+
+        grad = grad / (tf.sqrt(v_t) + self.epsilon)
+        grad = tf.cond(grad_averaging,
+                       lambda: grad * coefficients['one_minus_beta_1_t'],
+                       lambda: grad)
+        grad = tf.cond(tf.greater(weight_decay, 0),
+                       lambda: grad + weight_decay * var,
+                       lambda: grad)
+        m = self.get_slot(var, 'm')
+        return training_ops.resource_apply_momentum(
```
shreyashpatodia: I would have liked to use Adam here but from what I understand of Novograd it doesn't always use grad_averaging which makes it inconvenient to use Adam.

tensorflow_addons/optimizers/novograd.py
```diff
@@ -0,0 +1,246 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Novograd for TensorFlow."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow as tf
+from tensorflow.python.training import training_ops
```
WindQAQ: Seems that private API is the only way to import this. @seanpmorgan Sean, what do you think about this?

tensorflow_addons/optimizers/novograd.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+@tf.keras.utils.register_keras_serializable(package='Addons')
+class Novograd(tf.keras.optimizers.Optimizer):
```
WindQAQ: Shouldn't this be `NovoGrad`?

https://github.com/NVIDIA/OpenSeq2Seq/blob/master/open_seq2seq/optimizers/novograd.py#L30
https://nvidia.github.io/OpenSeq2Seq/html/optimizers.html

tensorflow_addons/optimizers/novograd.py
```diff
@@ -0,0 +1,246 @@
...
+    Networks](https://arxiv.org/pdf/1905.11286.pdf)
+
+    NovoGrad is a first-order SGD-based algorithm, which computes second
+    moments per layer instead of per weight as in Adam. Compared to Adam,
+    NovoGrad takes less memory, and has been found to be more numerically
+    stable. More specifically we compute (for more information on the
+    computation please refer to this
+    [link](https://nvidia.github.io/OpenSeq2Seq/html/optimizers.html):
+
+    ```
```
WindQAQ: remove ``` here

tensorflow_addons/optimizers/novograd.py
```diff
@@ -0,0 +1,246 @@
...
+        1/sqrt(v_t) (set weight_decay > 0 to use this0:
+            m_t <- beta_1 * m_{t-1} +
+                   [(g_t / (sqrt(v_t) + epsilon)) + (w_d * w_{t-1})]
+        4. weight decay + exponential moving average from Adam:
+            m_t <- beta_1 * m_{t-1} +
+                   [(1 - beta_1) * ((g_t / (sqrt(v_t + epsilon)) +
+                   (w_d * w_{t-1}))]
+    Weight update:
+        w_t <- w_{t-1} - lr_t * m_t
+    ```
```
WindQAQ: ditto

tensorflow_addons/optimizers/novograd.py
```diff
@@ -0,0 +1,246 @@
...
+    """
+
+    def __init__(self,
+                 learning_rate=0.001,
+                 beta_1=0.9,
+                 beta_2=0.999,
+                 epsilon=1e-8,
+                 weight_decay=0.0,
+                 grad_averaging=False,
```
WindQAQ: I find the default arguments are very different from the original implementation. Is it because we use adam instead of momentum optimizer (SGD with momentum)?

tensorflow_addons/optimizers/novograd.py
```diff
@@ -0,0 +1,246 @@
...
+                 amsgrad=False,
+                 name='Novograd',
+                 **kwargs):
+        r"""Construct a new RAdam optimizer.
```
WindQAQ: NIT: `RAdam` -> `NovoGrad`

tensorflow_addons/optimizers/novograd.py
```diff
@@ -0,0 +1,246 @@
...
+        coefficients = ((apply_state or {}).get((var_device, var_dtype))
+                        or self._fallback_apply_state(var_device, var_dtype))
+        weight_decay = self._get_hyper('weight_decay')
+        grad_averaging = self._get_hyper('grad_averaging')
+
+        v = self.get_slot(var, 'v')
+        g_2 = tf.reduce_sum(tf.square(tf.cast(grad, tf.float32)))
+        v_t = tf.cond(
+            tf.equal(self.iterations,
+                     0), lambda: g_2, lambda: v * coefficients['beta_2_t'] +
```
WindQAQ: Is it possible to put `tf.equal(iterations, 0)` in one line, which is more readable.

tensorflow_addons/optimizers/novograd.py
```diff
@@ -0,0 +1,246 @@
...
+        if self.amsgrad:
+            vhat = self.get_slot(var, 'vhat')
+            vhat_t = vhat.assign(
+                tf.maximum(vhat, v_t), use_locking=self._use_locking)
+            grad = grad / (tf.sqrt(vhat_t) + self.epsilon)
+        else:
+            grad = grad / (tf.sqrt(v_t) + self.epsilon)
+        grad = tf.cond(
+            tf.greater(weight_decay,
+                       0), lambda: grad + weight_decay * var, lambda: grad)
```
WindQAQ: Is it possible to put `tf.greater(weight_decay, 0)` in one line, which is more readable.

tensorflow_addons/optimizers/novograd.py
```diff
@@ -0,0 +1,246 @@
...
+                        or self._fallback_apply_state(var_device, var_dtype))
+        weight_decay = self._get_hyper('weight_decay')
+        grad_averaging = self._get_hyper('grad_averaging')
+
+        v = self.get_slot(var, 'v')
+        g_2 = tf.reduce_sum(tf.square(tf.cast(grad, tf.float32)))
+        # v is just a scalar and does not need to involve sparse tensors.
+        v_t = tf.cond(
+            tf.equal(self.iterations,
+                     0), lambda: g_2, lambda: v * coefficients['beta_2_t'] +
```
WindQAQ: Is it possible to put `tf.equal(iterations, 0)` in one line, which is more readable.

tensorflow_addons/optimizers/novograd.py
```diff
@@ -0,0 +1,246 @@
...
+        if self.amsgrad:
+            vhat = self.get_slot(var, 'vhat')
+            vhat_t = vhat.assign(
+                tf.maximum(vhat, v_t), use_locking=self._use_locking)
+            grad = grad / (tf.sqrt(vhat_t) + self.epsilon)
+        else:
+            grad = grad / (tf.sqrt(v_t) + self.epsilon)
+        grad = tf.cond(
+            tf.greater(weight_decay,
+                       0), lambda: grad + weight_decay * var, lambda: grad)
```
WindQAQ: Is it possible to put `tf.greater(weight_decay, 0)` in one line, which is more readable.

tensorflow_addons/optimizers/novograd_test.py
```diff
@@ -0,0 +1,140 @@
...
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow as tf
+import numpy as np
+
+from tensorflow_addons.utils import test_utils
+from tensorflow_addons.optimizers import Novograd
```
WindQAQ:
```python
import numpy as np
import tensorflow as tf

from tensorflow_addons.optimizers import Novograd
```

tensorflow_addons/optimizers/novograd_test.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+@test_utils.run_all_in_graph_and_eager_modes
+class NovogradTest(tf.test.TestCase):
```
WindQAQ: Could we also add a test method for `grad_averaging=True`?

tensorflow_addons/optimizers/novograd_test.py
```diff
@@ -0,0 +1,140 @@
...
+            update = optimizer.apply_gradients(grads_and_vars)
+            self.evaluate(tf.compat.v1.global_variables_initializer())
+            for _ in range(iterations):
+                self.evaluate(update)
+
+        self.assertAllClose(var_0.read_value(), expected[0], atol=2e-4)
+        self.assertAllClose(var_1.read_value(), expected[1], atol=2e-4)
+
+    def test_dense_sample(self):
+
```
WindQAQ: could you remove this empty line?

tensorflow_addons/optimizers/novograd.py
shreyashpatodia: The default arguments I've added in are more in line with the default arguments provided to Adam across different libraries. I didn't see any point in moving away from the convention of having `beta_1` and `beta_2` be 0.9 and 0.999 respectively. In my tests novoograd with 0.95 and 0.98 was not significantly better than novograd with 0.9 and 0.999 so I went with the convention.

tensorflow_addons/optimizers/novograd.py
shreyashpatodia: Fixed.

tensorflow_addons/optimizers/novograd.py
shreyashpatodia: Done.

tensorflow_addons/optimizers/novograd.py
shreyashpatodia: Done.

tensorflow_addons/optimizers/novograd.py
shreyashpatodia: Done.

tensorflow_addons/optimizers/novograd.py
shreyashpatodia: Done.

tensorflow_addons/optimizers/novograd_test.py
shreyashpatodia: Done

tensorflow_addons/optimizers/novograd_test.py
shreyashpatodia: Done.

shreyashpatodia: Done.

tensorflow_addons/optimizers/novograd_test.py
shreyashpatodia: Done.

tensorflow_addons/optimizers/novograd.py
WindQAQ: Got it! Thanks for the survey.

WindQAQ: @shreyashpatodia, one minor thing is that default `epsilon` for `tf.keras.optimizers.Adam` is `1e-7` instead of `1e-8`.

shreyashpatodia: Changed.

shreyashpatodia:
@WindQAQ making the `tf.equal` params come on the same line seemed to have broken sanity-checks. I reverted back the change, for now.

WindQAQ:
> @WindQAQ making the `tf.equal` params come on the same line seemed to have broken sanity-checks. I reverted back the change, for now.

I'm fine with this :P Let's trigger the CI again.

Hello @valendin, could you take a look at this to see if you have further problems? Thanks!

tensorflow_addons/optimizers/novograd.py
WindQAQ: Hi @shreyashpatodia, could we add some comments here? Like
```python
# TODO: Find public API alternatives to these
from tensorflow.python.training import training_ops
```

shreyashpatodia: Done.

PR #842, opened 2020-01-08
Add multi-device testing utils

Squadrick:


Squadrick:
All the test cases pass which shouldn't happen. Test with using `tf.keras.Model()` and doing a `fit()` under the strategy scope.

Squadrick:
@tensorflow/sig-addons-maintainers  The new `test_utils` works. What do you think about the API and usage?

tensorflow_addons/utils/test_utils.py
```diff
@@ -53,6 +53,75 @@ def use_gpu():
...
+
+def run_distributed(num_devices):
+    def decorator(f):
+        if inspect.isclass(f):
+            raise TypeError("`run_distributed` only supports test methods. "
+                            "Did you mean to use `run_all_distributed`?")
+
+        def decorated(self, *args, **kwargs):
+            logical_devices = create_virtual_devices(num_devices)
+            strategy = tf.distribute.MirroredStrategy(logical_devices)
```
seanpmorgan: Is there a way we can make the distribution strategy configurable? This might be useful for certain types of bugs

Squadrick: Nope, each strategy uses different constructor params. I can have them pass a class to the decorated and instantiate that with `logical_devices` and throw if there's an exception.

Squadrick: I'll leave the support for other strategies as a TODO, and have this in the repo for now. `MirroredStrategy` is the most widely used in the community as well.

Squadrick:
Unrelated test breakage. Rerunning tests.

PR #843, opened 2020-01-09
FIX: GELU name

WindQAQ:
Correct the abbreviation from GeLU to GELU, which is used in the original [paper](https://arxiv.org/abs/1606.08415) and the community.


WindQAQ:
cc @AakashKumarNain

AakashKumarNain:
Thanks @WindQAQ . I still don't understand how we missed this for so long..lol

PR #845, opened 2020-01-09
Fix bug in "xy" mode for bilinear interpolation

djl11:
Fixed incorrect clipping of query indices for non-square images.
Also modified tests to use non-square images.

Additional note: a single tf.reverse(query_points, -1) operation in an if statement could simplify this function for "ij" vs "xy" mode.

### DIFF
--- a/tensorflow_addons/image/dense_image_warp.py
+++ b/tensorflow_addons/image/dense_image_warp.py
@@ -114,6 +114,7 @@ def interpolate_bilinear(grid, query_points, indexing="ij", name=None):
         floors = []
         ceils = []
         index_order = [0, 1] if indexing == "ij" else [1, 0]
+        x_axis_width = width if indexing == "ij" else height
         unstacked_query_points = tf.unstack(query_points, axis=2, num=2)
 
         for dim in index_order:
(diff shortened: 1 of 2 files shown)


djl11:
@googlebot I signed it!

seanpmorgan:
> For non-square images, the current code creates out-of-bounds pixel queries during the tf.gather operation. This changes fixes this.
> 
> Additional note: using non-square images in unit tests would prevent these kinds of errors.
> 
> Also: a single tf.reverse(query_points, -1) operation in an if statement could simplify this function for "ij" vs "xy" mode.

Thanks @djl11! Could you please add a test case as you describe.

djl11:
Apologies, my original commit actually did not fix the problem. My newest commit does though. I have also added a new test, which fails for the original code but passes with the modification.

djl11:
@googlebot I consent.

seanpmorgan:
> @googlebot I consent.

@djl11 don't worry about the CLA I'll override it. In the future to fix the lint errors just run:
`bash tools/run_docker.sh -c 'make code-format'`

djl11:
@seanpmorgan okay great, wasn't sure about the lint errors, thanks for the advice!
I'll leave it to you for now then. Cheers!

PR #847, opened 2020-01-10
0.7 Release

seanpmorgan:
Pushed binaries to test pypi:
https://test.pypi.org/project/tensorflow-addons/#files

I checked and py27 is not supported for TF so we'll only be releasing python3 windows wheels. I've made a note to discuss whether or not we need to generate these API docs anymore. I believe the docs team does it themselves and they make these reviews quite difficult.


seanpmorgan:
Failing sanity is for auto generated doc file name

PR #857, opened 2020-01-10
Thresholded Linear Unit

AakashKumarNain:
This activation unit is introduced in Filter Response Normalization

### DIFF
--- a/tensorflow_addons/layers/tlu.py
+++ b/tensorflow_addons/layers/tlu.py
@@ -1,4 +1,4 @@
-# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+# Copyright 2020 The TensorFlow Authors. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/tensorflow_addons/layers/tlu.py
+++ b/tensorflow_addons/layers/tlu.py
@@ -20,7 +20,6 @@ from __future__ import division
 from __future__ import print_function
 
 import tensorflow as tf
-from tensorflow.keras import backend as K
 
 
 @tf.keras.utils.register_keras_serializable(package='Addons')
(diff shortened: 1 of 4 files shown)


Squadrick:
@AakashKumarNain Any reason to put this under `layers` as opposed to `activations`? Also, could you update the `README.md`?

tensorflow_addons/layers/tlu.py
```diff
@@ -0,0 +1,116 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
Squadrick: Nit: 2019 -> 2020

tensorflow_addons/layers/tlu_test.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
Squadrick: Nit: 2019 -> 2020

tensorflow_addons/layers/tlu.py
AakashKumarNain: LOL. Yeah I will change that.

AakashKumarNain:
@Squadrick we have two learnable parameters for this activation, similar to `PReLU`. It makes more sense to build a layer for this and if we want to include it in the `activations` package, then the activation function should initalize an instance of this layer first and then call the activation function on the inputs. Lemme know what you think

Squadrick:
@AakashKumarNain Yeah, you're right. I'd forgotten that `tf.keras` expects activations to be stateless. Let's not bother with creating anything under `activations`. There's a precedent with core `tf.keras` making PReLU a layer.

tensorflow_addons/layers/tlu.py
```diff
@@ -0,0 +1,116 @@
...
+            initializer=self.tau_initializer,
+            regularizer=self.tau_regularizer,
+            constraint=self.tau_constraint)
+        if self.affine:
+            self.alpha = self.add_weight(
+                shape=param_shape,
+                name='alpha',
+                initializer=self.alpha_initializer,
+                regularizer=self.alpha_regularizer,
+                constraint=self.alpha_constraint)
```
Squadrick: It would be better to set `aggregation` and `synchronization` for `add_weight()` as well. This will make it suitable for distributed training out of the box, instead of having to deal with that in the future. [Link](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer?version=nightly#add_weight) to docs.

`synchronization=AUTO`, and `aggregation=MEAN` should be ideal.

AakashKumarNain: Cool. Will do that.

AakashKumarNain:
> @AakashKumarNain Also test for serialization.

Can you elaborate on this? Any example?

tensorflow_addons/layers/tlu.py
AakashKumarNain: `aggregation=MEAN` Shouldn't we leave this to the user for custom training loops?

Squadrick: Nope, it's more related to how you expect the variable to behave. You would only take the mean of these variables, similar to how you would take the mean of the weights of the conv layer, etc. 

I can't think of a situation where you would set this to either `FIRST_REPLICA_ONLY` or `SUM`. But if you do, let me know, and we can reconsider setting this to `MEAN` by default.

Squadrick:
> > @AakashKumarNain Also test for serialization.
> 
> Can you elaborate on this? Any example?

https://github.com/tensorflow/addons/blob/2bcf8c874b18b88f454a758e0f64b959c8543fda/tensorflow_addons/layers/wrappers_test.py#L99

AakashKumarNain:
Will this suffice?

```python
def test_serialization(self):
    tlu = TLU(affine=True, alpha_initializer='ones', tau_initializer='ones')
    serialized_tlu = tf.keras.layers.serialize(tlu)
    new_layer = tf.keras.layers.deserialize(serialized_tlu, custom_objects={'TLU':TLU})
    self.assertEqual(tlu.get_config(), new_layer.get_config())
```

tensorflow_addons/layers/tlu.py
AakashKumarNain: Hmm agreed

Squadrick:
@AakashKumarNain Yeah, that works. It'll work without the `custom_objects` as well.

AakashKumarNain:
@seanpmorgan @Squadrick I think we can merge it now

AakashKumarNain:
Thanks

AakashKumarNain:
I forgot to update the README. Will do it later.

PR #858, opened 2020-01-10
sanity checks for Github Actions

### LINKED ISSUE #852: Move sanity check to GitHub Actions
The sanity check in particular is of highest value to move to GitHub actions. Its the most likely fail for incoming PRs, but often requires a maintainer to add the kokoro label in order to give feedback.

ulf1:
https://github.com/tensorflow/addons/issues/852

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -52,14 +52,10 @@ jobs:
             echo export BAZEL_PATH=/d/a/addons/addons/bazel-%BAZEL_VERSION%-windows-x86_64.exe  >>D:\a\_temp\msys\msys64\home\runneradmin\.bash_profile
             echo export PATH=/c/hostedtoolcache/windows/Python/${{ matrix.python-version }}/x64/:$PATH  >>D:\a\_temp\msys\msys64\home\runneradmin\.bash_profile
             msys2do ./tools/ci_testing/addons_cpu.sh
-  sanity-cpu:
-    name: Sanity Check
+  sanity-check:
+    name: CI Sanity Check
     runs-on: ubuntu-18.04
-    strategy:
-      matrix:
-        python-version: ['3.7.5']
     steps:
-      - uses: actions/checkout@v1
       - name: CI sanity check py${{ matrix.python-version }}
         run: |
           set -x -e
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -53,11 +53,11 @@ jobs:
             echo export PATH=/c/hostedtoolcache/windows/Python/${{ matrix.python-version }}/x64/:$PATH  >>D:\a\_temp\msys\msys64\home\runneradmin\.bash_profile
             msys2do ./tools/ci_testing/addons_cpu.sh
   sanity-check:
-    name: CI Sanity Check
+    name: Sanity Check
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v1
-      - name: CI sanity check py${{ matrix.python-version }}
+      - name: Sanity check
         run: |
           set -x -e
           bash tools/run_docker.sh -c 'make sanity-check'
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -54,10 +54,8 @@ jobs:
             msys2do ./tools/ci_testing/addons_cpu.sh
   sanity-cpu:
     name: Python3 CPU sanity-check
-    runs-on: 3.7.5-slim-buster
+    runs-on: ubuntu-18.04
     steps:
-      - name: Install docker
-        run: apt-get install -y docker-ce
       - name: CI sanity check py${{ matrix.python-version }}
         run: |
           set -x -e
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -56,6 +56,7 @@ jobs:
     name: CI Sanity Check
     runs-on: ubuntu-18.04
     steps:
+      - uses: actions/checkout@v1
       - name: CI sanity check py${{ matrix.python-version }}
         run: |
           set -x -e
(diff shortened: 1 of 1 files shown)


seanpmorgan:
@ulf1 Thanks for getting this going. So we only need to run sanity check once and it's better done from a linux machine with docker. 

Try looking at how TF-IO is doing their builds and then have it run:
`bash tools/run_docker.sh -c 'make sanity-check'`

https://github.com/tensorflow/io/blob/master/.github/workflows/build.yml#L156

seanpmorgan:
@ulf1 try creating a new test that runs on `ubuntu-18.04` then the run_docker command should work without much other setup

.github/workflows/ci_test.yml
```diff
@@ -52,3 +52,22 @@ jobs:
...
+      - uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: CI sanity check py${{ matrix.python-version }}
+        run: |
+          set -x -e
+          python -m pip install wheel setuptools
+          wget --quiet -nc https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-installer-darwin-x86_64.sh
+          chmod +x bazel-${BAZEL_VERSION}-installer-darwin-x86_64.sh
+          ./bazel-${BAZEL_VERSION}-installer-darwin-x86_64.sh --user
```
seanpmorgan: Should be able to remove this. The shell script will install all necessary dependencies in the docker container

ulf1: it works now. summary:
- use `runs-on: ubuntu-18.04`
- adjust bazel for linux `...-installer-linux-x86_64.sh`
- call `bash tools/run_docker.sh -c 'make sanity-check'`

.github/workflows/ci_test.yml
```diff
@@ -52,3 +52,22 @@ jobs:
             echo export BAZEL_PATH=/d/a/addons/addons/bazel-%BAZEL_VERSION%-windows-x86_64.exe  >>D:\a\_temp\msys\msys64\home\runneradmin\.bash_profile
             echo export PATH=/c/hostedtoolcache/windows/Python/${{ matrix.python-version }}/x64/:$PATH  >>D:\a\_temp\msys\msys64\home\runneradmin\.bash_profile
             msys2do ./tools/ci_testing/addons_cpu.sh
+  sanity-cpu:
+    name: Python3 CPU sanity-check
```
seanpmorgan: Can you just name this `Sanity Check`

.github/workflows/ci_test.yml
```diff
@@ -13,7 +13,7 @@ env:
 
 jobs:
   macos-test:
-    name: MacOS CPU Python3
+    name: MacOS CPU Python3 unit-test
```
seanpmorgan: I think `unit-test` isn't needed. It matches better with the other test names if it's removed.

.github/workflows/ci_test.yml
seanpmorgan: Yes it runs, but those 4 lines are unnecessary and take time. The sanity check is ran inside of the docker container not on host. Please remove

.github/workflows/ci_test.yml
```diff
@@ -52,3 +52,22 @@ jobs:
...
+      - uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: CI sanity check py${{ matrix.python-version }}
+        run: |
+          set -x -e
+          python -m pip install wheel setuptools
+          wget --quiet -nc https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-installer-linux-x86_64.sh
+          chmod +x bazel-${BAZEL_VERSION}-installer-linux-x86_64.sh
+          ./bazel-${BAZEL_VERSION}-installer-linux-x86_64.sh --user
```
seanpmorgan: Please remove L69-72 as they're not needed. After that looks good to merge. Thanks!

.github/workflows/ci_test.yml
```diff
@@ -52,3 +52,22 @@ jobs:
...
+    name: Sanity Check
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python-version: ['3.7.5']
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
```
seanpmorgan: This can be removed as well.

.github/workflows/ci_test.yml
ulf1: Oh yes right (－‸ლ)

.github/workflows/ci_test.yml
```diff
@@ -52,3 +52,15 @@ jobs:
             echo export BAZEL_PATH=/d/a/addons/addons/bazel-%BAZEL_VERSION%-windows-x86_64.exe  >>D:\a\_temp\msys\msys64\home\runneradmin\.bash_profile
             echo export PATH=/c/hostedtoolcache/windows/Python/${{ matrix.python-version }}/x64/:$PATH  >>D:\a\_temp\msys\msys64\home\runneradmin\.bash_profile
             msys2do ./tools/ci_testing/addons_cpu.sh
+  sanity-cpu:
+    name: Sanity Check
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python-version: ['3.7.5']
```
seanpmorgan: Please remove the python strategy matrix (L58-60)

.github/workflows/ci_test.yml
```diff
@@ -52,3 +52,15 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v1
+      - name: CI sanity check py${{ matrix.python-version }}
```
seanpmorgan: Just name this `CI sanity check`

seanpmorgan:
@ulf1 Super close... just need to keep `uses: actions/checkout@v1` which is used to download this repo onto the host. After that it should be good to go.

PR #861, opened 2020-01-12
Dropping the python 2 builds.

gabrieldemarmiesse:


gabrieldemarmiesse:
@googlebot I signed it!

seanpmorgan:
Thanks very much @gabrieldemarmiesse!

tools/run_docker.sh
```diff
@@ -24,21 +24,18 @@ fi
...
-        p) PYTHON=$OPTARG;;
         c) COMMAND=$OPTARG;;
         h)
             echo -n "usage: run_docker.sh [-s SYSTEM] [-d DEVICE] "
-            echo "[-p PYTHON] -c string"
+            echo " -c string"
             echo "available commands:"
             echo "    -s    select operating system: 'linux'"
             echo "    -d    select device: 'cpu', 'gpu'"
-            echo "    -p    select python version: 'py2', 'py3'"
```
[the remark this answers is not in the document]
gabrieldemarmiesse: In which file are the kokoro builds configured? I can't find them.

.travis.yml
```diff
@@ -42,11 +42,10 @@ jobs:
 
     # Windows Builds
     - stage: build
-      name: "Build on Windows for Python 2.7 3.5 3.6 3.7"
+      name: "Build on Windows for Python 3.5 3.6 3.7"
       os: windows
       language: shell
       before_install:
-        - choco install python --version 2.7.11
```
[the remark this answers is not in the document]
gabrieldemarmiesse: Done

tools/run_docker.sh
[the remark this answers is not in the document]
seanpmorgan: They're commands ran internally from Google that we don't have access to. It makes sense for us to have a script they call, but currently we just give them the command to run (written in that issue). After these PRs merge the only test they'll control is the GPU testing.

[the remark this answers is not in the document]
gabrieldemarmiesse: > It makes sense for us to have a script they call

That'd be a lot more flexible on our side :)

CONTRIBUTING.md
```diff
@@ -89,6 +89,8 @@ GPU Docker: `docker run --runtime=nvidia --rm -it -v ${PWD}:/addons -w /addons g
 
 Configure:
 ```
+# Temporary until we remove py2 support
+ln -sf /usr/bin/python3.6 /usr/bin/python && rm /usr/bin/python2
```
seanpmorgan: Would like to be able to remove this. The problem is that `configure.sh` will install using `python`:
https://github.com/tensorflow/addons/blob/master/configure.sh#L98

If we make that `python3` we should be able to remove this instruction. Bazel >1.0 will by default run tests/builds on python3 I believe

gabrieldemarmiesse:
> Thanks very much @gabrieldemarmiesse!

I'm happy to help :)

CONTRIBUTING.md
gabrieldemarmiesse: I removed it. The issue is that in `configure.sh`, I cannot call `python3` because on windows it doesn't work like that. There is only `python` available, and `python3` doesn't exist. I believe we need to keep the symlink until `python` calls python3 by default on ubuntu.

README.md
```diff
@@ -61,9 +59,9 @@ import tensorflow_addons as tfa
 #### Linux Build Matrix
 | Version    | Compatible With |Python versions  | Compiler  | cuDNN | CUDA | 
 |:----------------------- |:---|:---------- |:---------|:---------|:---------|
-| tensorflow-addons-0.7.0 | tensorflow>=2.1.0 | 2.7, 3.5-3.7 | GCC 7.3.1 | 7.6 | 10.1 |
-| tensorflow-addons-0.6.0 | tensorflow==2.0.0 <br> tensorflow-gpu==2.0.0 | 2.7, 3.5-3.7 | GCC 7.3.1 | 7.4 | 10.0 |
-| tensorflow-addons-0.5.2 | tensorflow==2.0.0 <br> tensorflow-gpu==2.0.0 | 2.7, 3.5-3.7 | GCC 7.3.1 | 7.4 | 10.0 |
```
seanpmorgan: So I'm not sure we should remove the py2.7 from the build matrix. Those wheels are available for download, just going forward master branch and subsequent releases will not support it.

seanpmorgan: Thoughts?

gabrieldemarmiesse: oops, my bad.

WindQAQ: I would think we should remain it unchanged, but add some comments about supporting py3 only on nightly build section.

seanpmorgan:
This should be unblocked and just needs a conflict handled on the central README

gabrieldemarmiesse:
Sure thing, I'm on it.

CONTRIBUTING.md
gabrieldemarmiesse: We don't have any way to removing it without breaking the build. Windows doesn't understand the `python3` command and on ubuntu `/usr/bin/python` is still py2 for the time being.

seanpmorgan:
Related:

PR #869, opened 2020-01-13
Lazy loading of shared libraries.

### LINKED ISSUE #855: Lazy Load Custom Ops
A major issue we have with Addons is that it is compatible with PyPi TensorFlow, but has ABI incompatibilities with TF that was built from source or installed with conda depending on the compiler and settings:
https://github.com/tensorflow/addons/issues/676
https://github.com/tensorflow/addons/issues/753
https://github.com/tensorflow/addons/issues/574

The ultimate fix for this will be the ABI stable kernel and ops:
https://github.com/tensorflow/community/pull/133

However, as a mitigation it may be worthwhile to lazy load custom-ops when they are used:
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/activations/gelu.py#L23

At the moment all custom-ops are loaded during init. Seeing as >80% of addons does not use a custom-op it is likely that many users are experiencing this issue even when they don't plan on using a custom op. This is certainly a hacky "fix" and doesn't solve the issue but I thought it was worth discussing as I feel this is a major issue regarding adoption.

gabrieldemarmiesse:
Fixes #855

I don't know how to make a test for that though. If possible we need to check that an so isn't loaded when doing `import tensorflow_addons`.

### DIFF
--- a/tensorflow_addons/utils/resource_loader.py
+++ b/tensorflow_addons/utils/resource_loader.py
@@ -41,7 +41,7 @@ def get_path_to_datafile(path):
     return os.path.join(root_dir, path.replace("/", os.sep))
 
 
-class LazyOpLoader:
+class LazySO:
     def __init__(self, relative_path):
         self.relative_path = relative_path
         self._ops = None
--- a/tensorflow_addons/utils/resource_loader.py
+++ b/tensorflow_addons/utils/resource_loader.py
@@ -19,6 +19,8 @@ from __future__ import print_function
 
 import os
 
+import tensorflow as tf
+
 
 def get_project_root():
     """Returns project root folder."""
@@ -37,3 +39,14 @@ def get_path_to_datafile(path):
     """
     root_dir = get_project_root()
     return os.path.join(root_dir, path.replace("/", os.sep))
+
+
+class LazyOpLoader:
+    def __init__(self, relative_path):
+        self.relative_path = relative_path
+        self.op = None
+
+    def get(self):
+        if self.op is None:
+            self.op = tf.load_op_library(get_path_to_datafile(self.relative_path))
+        return self.op
--- a/tensorflow_addons/utils/resource_loader.py
+++ b/tensorflow_addons/utils/resource_loader.py
@@ -44,9 +44,11 @@ def get_path_to_datafile(path):
 class LazyOpLoader:
     def __init__(self, relative_path):
         self.relative_path = relative_path
-        self.op = None
+        self._ops = None
 
-    def get(self):
-        if self.op is None:
-            self.op = tf.load_op_library(get_path_to_datafile(self.relative_path))
-        return self.op
+    @property
+    def ops(self):
+        if self._ops is None:
+            self._ops = tf.load_op_library(
+                get_path_to_datafile(self.relative_path))
+        return self._ops
--- a/tensorflow_addons/activations/gelu.py
+++ b/tensorflow_addons/activations/gelu.py
@@ -18,10 +18,9 @@ from __future__ import division
 from __future__ import print_function
 
 import tensorflow as tf
-from tensorflow_addons.utils.resource_loader import get_path_to_datafile
+from tensorflow_addons.utils.resource_loader import LazyOpLoader
 
-_activation_ops_so = tf.load_op_library(
-    get_path_to_datafile("custom_ops/activations/_activation_ops.so"))
+_activation_so = LazyOpLoader("custom_ops/activations/_activation_ops.so")
 
 
 @tf.keras.utils.register_keras_serializable(package='Addons')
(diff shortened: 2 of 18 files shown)


tensorflow_addons/utils/resource_loader.py
```diff
@@ -37,3 +39,16 @@ def get_path_to_datafile(path):
...
     return os.path.join(root_dir, path.replace("/", os.sep))
+
+
+class LazySO:
+    def __init__(self, relative_path):
+        self.relative_path = relative_path
+        self._ops = None
+
+    @property
+    def ops(self):
```
fsx950223: Override __getattribute__ hook may a better solution.

gabrieldemarmiesse: @fsx950223 thanks for the review! Can you expand on why it may be a better solution ?

fsx950223: No need to modify the original code this way。
For example,
```
    return _activation_ops_so.addons_gelu(x, approximate)
```

gabrieldemarmiesse: I see what you mean, it's clever :)

Though i'm more in favor of having explicit code rather than having smart code. It's more maintainable and readable.

Squadrick:
LGTM

seanpmorgan:
Thanks again @gabrieldemarmiesse !

AlexWang1900:
HI!
I am using this fix in tfa-nightly-0.8.0.dev20200205
on Kaggle Notebook , 

>!pip install tfa-nightly
>import tensorflow_addons as tfa
 I got this :
ImportError                               Traceback (most recent call last)
<ipython-input-11-f11e95d3c435> in <module>
----> 1 from tensorflow_addons import image as Timage

/opt/conda/lib/python3.6/site-packages/tensorflow_addons/__init__.py in <module>
     19 
     20 # Local project imports
---> 21 from tensorflow_addons import activations
     22 from tensorflow_addons import callbacks
     23 from tensorflow_addons import image

/opt/conda/lib/python3.6/site-packages/tensorflow_addons/activations/__init__.py in <module>
     15 """Additional activation functions."""
     16 
---> 17 from tensorflow_addons.activations.gelu import gelu
     18 from tensorflow_addons.activations.hardshrink import hardshrink
     19 from tensorflow_addons.activations.lisht import lisht

/opt/conda/lib/python3.6/site-packages/tensorflow_addons/activations/gelu.py in <module>
     17 
     18 from tensorflow_addons.utils import types
---> 19 from tensorflow_addons.utils.resource_loader import LazySO
     20 
     21 _activation_so = LazySO("custom_ops/activations/_activation_ops.so")

ImportError: cannot import name 'LazySO'

gabrieldemarmiesse:
@seanpmorgan do you have an idea of what can be causing this?

seanpmorgan:
> @seanpmorgan do you have an idea of what can be causing this?

Hmmmm this is puzzling. So as a sanity check here is a notebook using today's nightly without that issue:
https://colab.research.google.com/drive/1kK9on1C31gy7_PyTqHg4B6XrFs39d2EC

Could you provide the full docker image name of what you're using @AlexWang1900 . I'm thinking this is related to #987 since I see no reason this is happening from the python side (no circular import, inits look correct, etc.)

I tried to test this myself but conda still doesn't have a TF2.1 version so perhaps it was built from source in this docker image.

AlexWang1900:
@seanpmorgan 
it can be reproduced by running one kaggle notebook with option in settings,docker:  latest availible 

it is with TF2.1,so I think the docker image you can find it here:
[](https://github.com/Kaggle/docker-python
)
Building a new image
./build
Flags:

--gpu to build an image for GPU.
--use-cache for faster iterative builds.

Running the image
For the GPU image:

# Run the image built locally:
docker run --runtime nvidia --rm -it kaggle/python-gpu-build /bin/bash

Tensorflow custom pre-built wheel
A Tensorflow custom pre-built wheel is used mainly for:

Faster build time: Building tensorflow from sources takes ~1h. Keeping this process outside the main build allows faster iterations when working on our Dockerfiles.
Building Tensorflow from sources:

Increase performance: When building from sources, we can leverage CPU specific optimizations
Is required: Tensorflow with GPU support must be built from sources
The Dockerfile and the instructions can be found in the tensorflow-whl folder/.

seanpmorgan:
@AlexWang1900 thanks for the info. So I tried this in a GPU kaggle kernel and did not get the same error as you did:
https://www.kaggle.com/seanpmorgan/testtfa/

As expected when trying to run a custom-op we get the issue in #987 issue, but importing tfa and using other parts of it should work fine. Is there something different about my kernel than yours?

AlexWang1900:
@seanpmorgan 
I run it again and it works fine now ! many thanks!!!!
there should be no difference between your kernel and mine.
only difference is the tfa version ,yesterday it came with  tfa-nightly-0.8.0.dev20200205
today it is tfa-nightly-0.8.0.dev20200206

PR #889, opened 2020-01-15
CLN: remove buildifier

WindQAQ:
Accidentally pushed in #882.


gabrieldemarmiesse:
or we could maybe add this file to the gitignore to ensure that this doesn't happen again? What do you think?

Issue #851, opened 2020-01-10
Move tests and releases to GitHub Actions


seanpmorgan:
GitHub Actions are now available and provide better integration for releasing / testing. We'll need to keep kokoro for GPU tests but in time we should transition what we can.

- [x] Test Cases
- [ ] Releases

seanpmorgan:
https://github.com/tensorflow/community/blob/master/sigs/addons/RELEASE.md Will need to be updated once this is complete.

PR #892, opened 2020-01-15
Removed travis.yml.

gabrieldemarmiesse:
It seems that we don't use travis anymore. (I'm not sure, maybe @seanpmorgan can clarify). 

If so, we can remove the travis.yml.


seanpmorgan:
Currently we still build nightlies and releases using Travis. We're half way through transitioning:
https://github.com/tensorflow/addons/issues/851

So that will need be be completed before this PR.

gabrieldemarmiesse:
I'll close it for now. It's not needed to have separate pull requests to fix #851

PR #894, opened 2020-01-15
Added types and typeguard.

gabrieldemarmiesse:
This is a proof of concept about how to improve the user experience by catching error as soon as possible.

There are multiple runtime type checkers nowadays. I choose typeguard because:

* it's still maintained.
* The error messages are quite good compared to others.

Linked to #743


gabrieldemarmiesse:
How do we add a dependency to tf addons? I would expect adding it to the setup.py's `install_requires` to work.

tensorflow_addons/layers/normalizations.py
```diff
@@ -16,8 +16,11 @@
 # =============================================================================
 
 import logging
+from typegard import typechecked
```
seanpmorgan: typo `typeguard`

seanpmorgan:
> How do we add a dependency to tf addons? I would expect adding it to the setup.py's `install_requires` to work.

Yeah that is the correct way. It's being installed in the logs:
"Successfully installed ...  tf-nightly-2.2.0.dev20200115 typeguard-2.7.1 urllib3-1.25.7 werkzeug-0.16.0"

Just a typo I believe.

gabrieldemarmiesse:
I feel really stupid haha

ulf1:
Are we supposed to add type hints? Is it optional? Or is it forbidden?

gabrieldemarmiesse:
Type hints are optional in python. You can see the full rational behind type hints in the pep: https://www.python.org/dev/peps/pep-0484/

tensorflow_addons/utils/__init__.py
```diff
@@ -0,0 +1 @@
+from . import types
```
facaiy: nit: is it necessary?

gabrieldemarmiesse: Can you explain your question? I don't really understand what you mean.

tensorflow_addons/layers/normalizations.py
```diff
@@ -16,8 +16,11 @@
 # =============================================================================
 
 import logging
+from typeguard import typechecked
```
facaiy: please put it behind `import tensorflow as tf`

tensorflow_addons/utils/__init__.py
facaiy: I mean we can keep the file empty

facaiy: or set the `__all__` variable if necessary

tensorflow_addons/layers/normalizations.py
facaiy: or move it into `types.py`

tensorflow_addons/utils/__init__.py
gabrieldemarmiesse: ho I see what you mean. Let me change that.

tensorflow_addons/layers/normalizations.py
gabrieldemarmiesse: Sure thing

seanpmorgan:
This seems like a great start toward adding typing. Should this PR also include a github action to verify correctness?

gabrieldemarmiesse:
Our unit tests are supposed to check for correctness. The decorator is active all the time, meaning that if in our unit tests, the types were not correct, the CI would have failed already.

Maybe we don't have enough tests cases in our unit tests, is that what you mean? Or did you think about static type checking?

seanpmorgan:
> Our unit tests are supposed to check for correctness. The decorator is active all the time, meaning that if in our unit tests, the types were not correct, the CI would have failed already.
> 
> Maybe we don't have enough tests cases in our unit tests, is that what you mean? Or did you think about static type checking?

Yeah static checking with mypy. I believe that can be a separate discussion though as you mentioned this would fail in CI which should be sufficient.

gabrieldemarmiesse:
Last time I tried to integrate mypy in my project, it failed because I didn't type enough stuff. Maybe it has changed since. In any case, as you said, it can be done in a separate PR/issue.

gabrieldemarmiesse:
>  Do you have any thoughts on the best way to track the progress of this? 

No idea :(

gabrieldemarmiesse:
> Do you have any thoughts on the best way to track the progress of this?

Actually I have a good idea. I'll do it when I have the time.

Issue #898, opened 2020-01-15
Some proposals to improve the developers' experience


gabrieldemarmiesse:
Just my 2 cents to lower the entry bar for new contributors, there was already some good work done like moving the CI to github actions, using 4 spaces for tabs, removing py2 support, running tests with docker and I must forgot some more. 

I understand that a lot of this stuff came from tensorflow and that it's why it was so complexe. 

I was wondering what are the maintainers opinions on those ideas:

* Use pytest instead of `tf.test.TestCase`. Pytest is used more widely and has a LOT of plugins that we could use. Move the test files into a specific directory to separate them from the source (often `tests/`).

* Replace bash files by python files (especially the complicated ones, when it's less than 10 lines it's fine). Readability counts.

* Remove BUILD files when they're note necessary, and reduce the usage of bazel as much as possible. For example, it's not clear for someone who is used to make python packages why https://github.com/tensorflow/addons/blob/master/tensorflow_addons/callbacks/BUILD is necessary. Addons already has a good structure, Python imports should work out of the box.

* Running flake8 in the CI (instead of the bash script running pep8 and pylint).

* Removing the formatter (yapf) and recommend running autopep8 to fix pep8 errors. We can also give instructions on how to run autopep8 as a pre-commit.

AakashKumarNain:
I agree that it would be easier to go with `pytest` but as @WindQAQ  as long as we need to test things in graph mode, IMO `tf.test.TestCase` is much better. 

`Move the test files into a specific directory to separate them from the source (often tests/).`

I agree on this. Test files should be in a separate directory. My only concern is that if we move tests to a separate directory, should we place all files in a single directory or should we have sub-directories for each module. For example:

```
tests/
       --layers/
       --activations/
```

gabrieldemarmiesse:
@AakashKumarNain , the common practice is to make the tests/ directory have the same structure as the source directory. Meaning tests/ will have the same structure as tensorflow_addons/

@WindQAQ You're right, I see now that contrib uses a lot of sessions and tf.compat. Do we want to remove that? It seems it's leftovers from TF1 right?

Squadrick:
> @WindQAQ You're right, I see now that addons uses a lot of sessions and tf.compat. Do we want to remove that? It seems it's leftovers from TF1 right?

Wrapping a function in `@tf.function` makes it behave like a graph. So we definitely don't want to remove that. There's no way around `tf.test.TestCase`.

WindQAQ:
[...]

PR #904, opened 2020-01-16
Used flake8 + no formatter instead of pep8 + pylint + yapf.

gabrieldemarmiesse:
Linked to #898

We can discuss here the changes. It's easier to comment on an implementation than to discuss in an issue an implementation that doesn't exist yet.

If we merge this, we should afterwards open a ticket to:
1) make sure the two TODOs in the `.flake8` file are taken care of (fairly easy and quick)
2) Removing all the `# yapf: disable` and `# pylint: disable`  which will become obsolete.

It seems there is an unrelated failure on Windows.


WindQAQ:
Thanks for bringing this up! To prevent this from merging, I mark it as WIP.

facaiy:
Hi, Gabriel, thanks for the PR. I'm wondering why we should remove python formatter here?

gabrieldemarmiesse:
Hi @facaiy , thanks for the question!

I've made the experience with [Sacred](https://github.com/IDSIA/sacred). What happened in practice is that a lot of first time contributors just wanted to modify one or two lines and didn't run the formatter, even though it was documented.

Because of that, many small pull request were failing the CI because the formatting was wrong. But those people respected pep8, so it's not as if the code was unreadable.

The idea is that in open source, a lot of first timers/beginners want to change only a few lines, and having to run a formatter on top of it is not something that bring us much value compared to pep8, and it makes their first experience with the repo worst. It's also fustrating for reviewers who can't merge the pull request because the build is failing.

If we already enforce pep8, we have to think about what is the value that a formatter bring us on top of it. I believe the added value isn't much, especially for beginners, but I can understand if the other maintainers have a different opinion.

When working on internal projects, I'm all for using a formatter. I'm especially a big fan of black :)

gabrieldemarmiesse:
@tensorflow/sig-addons-maintainers What are your opinions on this?

kyleabeauchamp:
I like having a formatter

seanpmorgan:
So I agree that the current formatting has caused un-needed burden (in some cases running `make code-format` doesn't solve the issue). However, I think a repository of this many contributors and maintainers benefits from having a formatter.

I would be in favor of replacing the current setup with a single authority in `black`. We could verify easily setup a github action to check compliance:
https://github.com/lgeiger/black-action

Few line changes that are pep8 compliant will likely not be reformatted using black. Thoughts?

gabrieldemarmiesse:
Let's do that, I'll also add some instructions in the readme on how to make black and flake8 run as pre-commit hooks :)

I'll also make a separate github action for flake8, it's easier to understand from a beginners' perspective.

gabrieldemarmiesse:
This pull request is already quite big, and the setup of black necessitate to NUKE THE REPO.

So I'd prefer to do the setup of black in a separate PR.

.flake8
```diff
@@ -0,0 +1,32 @@
...
+    W291
+    # expected 2 blank lines after class or function definition, found 1
+    E305
+
+
+#imported but unused in __init__.py, that's ok.
+per-file-ignores = **/__init__.py:F401
+
+# TODO: reduce the max line length to 90-ish
+max-line-length = 210
```
seanpmorgan: Rationale for 210? Since we're going with black should we use their default 88?

.gitignore
```diff
@@ -1,9 +1,3 @@
-# Ignore all hidden files
-.*
```
seanpmorgan: Not sure this is preferable to allowing .flake8 with a `!` in gitignore

STYLE_GUIDE.md
```diff
@@ -17,28 +17,8 @@ void formatted_code_again;
...
-# yapf: enable
-```
-
-```python
-# pylint: disable=protected-access
-foo._protected_member
-# pylint: enable=protected-access
-```
-
+Addons uses [flake8](http://flake8.pycqa.org/en/latest/) check pep8 compliance and
```
seanpmorgan: nit : to check pep8 compliance?

.flake8
gabrieldemarmiesse: Some lines are too long in this repo. I didn't want to make them shorter in this PR to avoid having a PR with too many diffs.

I'll change it in the same PR applying black to the repo. Black will make them shorter for me. I won't have to do it manually.

.gitignore
gabrieldemarmiesse: As you wish

STYLE_GUIDE.md
gabrieldemarmiesse: Sure, let me fix that

gabrieldemarmiesse:
Thanks! I'll try to add black as smoothly as possible (I don't want every opened pull requests to have conflicts).

PR #905, opened 2020-01-17
refact with Generator

fsx950223:
1.Remove "--benchmarks=all" will skip the benchmark test. Previous behavior will skip all test cases.
2.Refact rrelu with Generator which is introduced in [Random number generation](https://www.tensorflow.org/guide/random_numbers)
3.Add graph test support


fsx950223:
How could I disable flake8 error ```E722 do not use bare 'except'```

gabrieldemarmiesse:
@googlebot I consent

gabrieldemarmiesse:

tensorflow_addons/activations/rrelu.py
```diff
@@ -26,6 +31,7 @@ def rrelu(
     upper: Number = 0.3333333333333333,
     training: Optional[bool] = None,
     seed: Optional[int] = None,
+    gs: Optional[tf.random.experimental.Generator] = None,
```
gabrieldemarmiesse: Do we have a policy about the use of experimental features in TensorFlow? What are the conditions in which we should accept the usage?

For example, I'd be ok if we use experimental features in tests, but not really in user-facing code. That's because it's ok if the build breaks because of an experimental feature, but it's not ok if the code the users are using is breaking.

@fsx950223  Maybe we can wait until the Generator is not experimental anymore?

fsx950223: Yes, I will wait until stable API is released

bhack: https://www.tensorflow.org/versions/r2.2/api_docs/python/tf/random/Generator

seanpmorgan: Stable API is available on TF2.2 which is what master builds against. Please merge master into this PR and should be good to update

tensorflow_addons/activations/rrelu.py
```diff
@@ -14,18 +14,22 @@
...
 def rrelu(
-    x: types.TensorLike,
+    x: TensorLike,
     lower: Number = 0.125,
     upper: Number = 0.3333333333333333,
-    training: Optional[bool] = None,
-    seed: Optional[int] = None,
+    training: bool = None,
+    seed: Number = None,
+    gs: tf.random.Generator = None,
```
gabrieldemarmiesse, line 32: Please use `Optional` here. Having None as the default value doesn't make the type optional automatically. I think it should but it doesn't :(

tensorflow_addons/activations/rrelu.py
```diff
@@ -14,18 +14,22 @@
 # ==============================================================================
 
 import tensorflow as tf
-from tensorflow_addons.utils.types import Number
-from tensorflow_addons.utils import types
-from typing import Optional
+from tensorflow_addons.utils.types import TensorLike, Number
+
+try:
+    tf.no_gradient("StatefulUniform")
```
gabrieldemarmiesse, line 20: I believe this deserves a comment as it's not obvious at all what is the purpose of this line.

fsx950223: Yes, I'm curiously the op doesn't register the gradient.

gabrieldemarmiesse: I'm not sure I understand. You mean the tests will fail if this line is not used?

fsx950223: I test it with latest tf version and it seems the op has registered the gradient and I remove the ```tf.no_gradient```.

tensorflow_addons/activations/rrelu.py
```diff
@@ -14,18 +14,18 @@
...
 
 @tf.keras.utils.register_keras_serializable(package="Addons")
 def rrelu(
-    x: types.TensorLike,
+    x: TensorLike,
     lower: Number = 0.125,
     upper: Number = 0.3333333333333333,
     training: Optional[bool] = None,
     seed: Optional[int] = None,
+    gs: tf.random.Generator = None,
```
gabrieldemarmiesse, line 28:
```suggestion
    gs: Optional[tf.random.Generator] = None,
```

tensorflow_addons/activations/rrelu.py
```diff
@@ -44,19 +44,25 @@ def rrelu(
...
         seed: `int`, this sets the operation-level seed.
+        gs: A `Generator`. Default value is tf.random.get_global_generator().
     Returns:
         result: A `Tensor`. Has the same type as `x`.
     """
     x = tf.convert_to_tensor(x)
+    lower = tf.cast(lower, x.dtype)
+    upper = tf.cast(upper, x.dtype)
+    if gs is None:
+        gs = tf.random.get_global_generator()
```
gabrieldemarmiesse, line 55: Here we affect the global context, unlike before were the random global context wasn't affected. This is a breaking change potentially, especially if it's used with `reset_from_seed`. Also from the [documentation](https://www.tensorflow.org/api_docs/python/tf/random/get_global_generator) :

> This function will create the global generator the first time it is called, and the generator will be placed at the default device at that time, so one needs to be careful when this function is first called. Using a generator placed on a less-ideal device will incur performance regression.

So I believe that the behavior the rrelu should have when `gs` is None is the old one. We call tf.random.uniform and set the seed. 

If the user provides his/her generator, then we use that (gs.uniform(...))

If the user provides a seed and a generator, we throw a `ValueError` because it makes no sense, the user is likely mistaken, with a nice error message of course.

fsx950223: > If the user provides a seed and a generator, we throw a `ValueError` because it makes no sense, the user is likely mistaken, with a nice error message of course.

I remove gs or seed. But it casues unstable result when I test the code.

fsx950223:
cc @gabrieldemarmiesse

tensorflow_addons/activations/rrelu.py
```diff
@@ -44,19 +44,28 @@ def rrelu(
         training: `bool`, indicating whether the `call`
         is meant for training or inference.
         seed: `int`, this sets the operation-level seed.
+        gs: A `Generator`. Default value is tf.random.get_global_generator().
```
gabrieldemarmiesse, line 47: We should change this line, it's not accurate anymore.

Also we should change the argument name to `random_number_generator` or `rng` whatever you prefer. It's more common to use this name.

tensorflow_addons/activations/rrelu.py
```diff
@@ -44,19 +44,28 @@ def rrelu(
...
-            tf.shape(x), minval=lower, maxval=upper, dtype=x.dtype, seed=seed
-        )
+        if gs is None:
+            alpha = tf.random.uniform(
+                tf.shape(x), minval=lower, maxval=upper, dtype=x.dtype, seed=seed
+            )
+        else:
+            if seed is not None:
+                raise ValueError("Either seed or gs should be specific")
+            alpha = gs.uniform(tf.shape(x), minval=lower, maxval=upper, dtype=x.dtype)
```
gabrieldemarmiesse, line 67:
```suggestion
                raise ValueError(
                    "Either seed or gs should be specified. Not both at the same time."
                    )
            alpha = gs.uniform(tf.shape(x), minval=lower, maxval=upper, dtype=x.dtype)
```

tensorflow_addons/activations/rrelu.py
```diff
@@ -44,19 +44,28 @@ def rrelu(
...
     if training:
-        alpha = tf.random.uniform(
-            tf.shape(x), minval=lower, maxval=upper, dtype=x.dtype, seed=seed
-        )
+        if gs is None:
+            alpha = tf.random.uniform(
+                tf.shape(x), minval=lower, maxval=upper, dtype=x.dtype, seed=seed
+            )
+        else:
+            if seed is not None:
```
gabrieldemarmiesse, line 65: In term of control flow, I believe something flatter would be easier to read. Something like:

```python
if gs is not None and seed is not None:
    raise ValueError(...)
elif gs is not None:
    gs.uniform(...)
else:
    tf.random.uniform(..., seed=seed)
```

tensorflow_addons/activations/tests/rrelu_test.py
```diff
@@ -23,43 +23,37 @@
 SEED = 111111
 
 
+def rrelu_wrapper(lower, upper, training):
```
gabrieldemarmiesse, line 26: This function seems unused.

tensorflow_addons/activations/tests/rrelu_test.py
```diff
@@ -23,43 +23,37 @@
...
-
-    tf.random.set_seed(SEED)
-    result = rrelu(x, lower, upper, training=training, seed=SEED)
+    training_results = {
+        np.float16: [-0.3826, -0.165, 0, 1, 2],
+        np.float32: [-0.282151192, -0.199812651, 0, 1, 2],
+        np.float64: [-0.25720977, -0.1221586, 0, 1, 2],
+    }
+    gs = tf.random.Generator.from_seed(SEED)
+    result = rrelu(x, lower, upper, training=training, seed=None, gs=gs)
```
gabrieldemarmiesse, line 46: We shouldn't modify an existing test when testing a new option. It should be a new standalone test. 

In the new test, rather than hardcoding the results, we could check that the generator works as expected, meaning that it doesn't affect the global random state, but is still reproductible. An example would be (pseudo code):

```python
def test1():
    """gs should ensure reproducibility without affecting the global random state."""
    tf.set_seed(0)
    a = tf.random.uniform()
    gs = tf.random.Generator.from_seed(8448)
    rrelu_result = rrelu(..., gs=gs)

    tf.set_seed(0)
    gs = tf.random.Generator.from_seed(8448)
    assert rrelu(..., gs=gs) == rrelu_result
    assert a == tf.random.uniform()
```

What do you think?

fsx950223: I revert old test cases.

PR #907, opened 2020-01-17
Fix the macos build.

gabrieldemarmiesse:
I don't know how to fix the Windows one though.

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
-        python-version: ['3.7.5']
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_test.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
-        python-version: ['3.7.5']
+        python-version: ['3.7.6']
```
WindQAQ: Could we also make the commands on nightly/CI the same?
https://github.com/tensorflow/addons/blob/master/.github/workflows/macos_nightly.yml#L11-L29

gabrieldemarmiesse: Done

Issue #893, opened 2020-01-15
Clarifying the CI and supported versions of tensorflow/python


gabrieldemarmiesse:
Hello all, 

I'm trying to see how we can improve the developers experience in tensorflow/addons. I'm confused about a few things and I hope maintainers can answer my questions to help me make addons better :)

* Can we add a new row in the [linux build matrix](https://github.com/tensorflow/addons#linux-build-matrix) in the README about the master branch? Like what TF/Python/Cuda version do we target? It's currently unclear from the CI.

* Do we sync tf addons releases on TF releases? Do we want to?

* Do we want our SO files to work with tensorflow-cpu ? Currently pure Python objects work well with it.

* Is it necessary to use tf nightly in the unit tests locally and in the CI? Can we use the latest stable version? Usually in the CI, a best practice is to pin the dependencies' versions to avoid having the master branch build breaking suddently. We can then do the upgrade of the dependencies in pull requests which confines the upgrade issues in separate branches and can be worked on without being in a rush.

* In the CI, do we build against the most recent version of python that we support, or the oldest? I believe it would be good to have both to ensure forward and backward compatibility.

Thanks for answering my questions! :)

seanpmorgan:
> * Can we add a new row in the [linux build matrix](https://github.com/tensorflow/addons#linux-build-matrix) in the README about the master branch? Like what TF/Python/Cuda version do we target? It's currently unclear from the CI.

Yes I think that's a good idea.

> * Do we sync tf addons releases on TF releases? Do we want to?

No, but in the past we've used new features (keras serialization) so that required us to wait until TF2.1 release. TF-Core is switching to quarterly releases and I think a faster cadence makes sense for Addons.

> * Do we want our SO files to work with tensorflow-cpu ? Currently pure Python objects work well with it.

Yes. [From that error in the other issue](https://github.com/tensorflow/addons/issues/864#issuecomment-573857046) I believe we may have to re-evaluate our CUDA linking. Certainly worth working through.


> * Is it necessary to use tf nightly in the unit tests locally and in the CI? Can we use the latest stable version? Usually in the CI, a best practice is to pin the dependencies' versions to avoid having the master branch build breaking suddently. We can then do the upgrade of the dependencies in pull requests which confines the upgrade issues in separate branches and can be worked on without being in a rush.

[...]

Issue #864, opened 2020-01-12
Tensorflow-cpu is ignored when installing tensorflow-addons


gabrieldemarmiesse:
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): ubuntu 18.04
- TensorFlow version and how it was installed (source or binary): tensorflow-cpu 2.1.0
- TensorFlow-Addons version and how it was installed (source or binary): 0.7.0
- Python version: 3.7
- Is GPU used? (yes/no): no

**Describe the bug**

In an environement where tensorflow-cpu is already installed, installing tensorflow-addons should not download and install tensorflow (now being the gpu version by default). 

**Code to reproduce the issue**

```
FROM python:3.7

RUN pip install tensorflow-cpu
RUN pip install tensorflow-addons
RUN pip freeze | grep tensorflow
```

result:

tensorflow (gpu) is still being installed (400MB download) and the installation of tensorflow is now duplicated: 

```
tensorflow==2.1.0
tensorflow-addons==0.7.0
tensorflow-cpu==2.1.0
tensorflow-estimator==2.1.0
```

A solution would be to remove the tensorflow dependency in the setup.py and let people install the version of tensorflow they want. It's not a perfect solution but this is the best one we found, at least for keras-tuner and autokeras. (https://github.com/keras-team/keras-tuner/pull/211 and https://github.com/keras-team/autokeras/pull/860)

The problem is similar to libraries that depend on Pillow and users have already pillow-SIMD installed.

seanpmorgan:
Hi @gabrieldemarmiesse this is a good point and previously we used to just do a TF version check during import. Since https://github.com/tensorflow/community/blob/master/rfcs/20190816-tf-project-versioning.md#versioning-policy however we moved toward pip dependencies. 

This is certainly a problem, but now that `tensorflow` package for TF 1.15 and TF 2.1 supports CPU/GPU I'm hoping that `tensorflow-cpu` usage will significantly drop off. 

It looks as though `tensorflow-cpu` is orders of magnitude less common than `tensorflow` (they used to be the same package):
https://pypistats.org/packages/tensorflow
https://pypistats.org/packages/tensorflow-cpu

gabrieldemarmiesse:
I believe that if we don't support tensorflow-cpu, we are forgetting a lot of users who don't have access to a high bandwidth. 

[...]

PR #912, opened 2020-01-19
Target tensorflow stable instead of tensorflow-nightly.

gabrieldemarmiesse:
Linked to #893

I'll try to list here the pros and cons so that we can decide if we go forward with this or not:

Pros:

* The CI can't fail out of nowhere because or a bug/API change on the Tensorflow side. The CI is more deterministic than before.
* The bump of the tensorflow version can be done in a isolated manner, in a pull request.
* We can make a minimal Dockerfile used for testing and tooling locally, and include bazel and tensorflow inside. Contributors won't have to download tf-nightly every day. Build once, run every day :)
* We don't have to support a single version of tensorflow, we can support a range of them and test against the oldest and the most recent in the CI: (eg: >=2.0.0,<2.2.0)
* If the dependency doesn't change every day, it means that the unit tests are reproductible. We can then use tools like `git biscect` to detect the commit which introduced a bug in the master branch.

Cons:

* We can't use the latest features available in tf-nightly.
* We have to bump the version number manually because it will be pinned, before it was done automatically.
* People using TensorFlow nightly might not be able to use the latest version of the addons master branch if TensorFlow inserted a regression or an API change.

To be honest, all the points made above (pros and cons) are pretty weak because tensorflow have strong API garantees. Those are generalities about pinning the version numbers in CIs. 

The only strong points might be the ones about tensorflow bugs and about not having to download tf-nightly every day when using Docker locally to test things.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-tf-nightly
+tensorflow==2.1.0
--- a/.github/workflows/macos_nightly.yml
+++ b/.github/workflows/macos_nightly.yml
@@ -2,7 +2,6 @@ name: macos-nightly
 
 on:
   schedule:
-    # Shortly after tf-nightly is released
     - cron:  '0 11 * * *'
 
 env:
--- a/.github/workflows/ubuntu_nightly.yml
+++ b/.github/workflows/ubuntu_nightly.yml
@@ -3,7 +3,6 @@ name: ubuntu-cpu-nightly
 
 on:
   schedule:
-    # Shortly after tf-nightly is released
     - cron:  '0 11 * * *'
 
 env:
--- a/.github/workflows/windows_nightly.yml
+++ b/.github/workflows/windows_nightly.yml
@@ -2,7 +2,6 @@ name: windows-nightly
 
 on:
   schedule:
-    # Shortly after tf-nightly is released
     - cron:  '0 11 * * *'
 
 env:
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ import tensorflow_addons as tfa
 
 #### Nightly Builds
 There are also nightly builds of TensorFlow Addons under the pip package
-`tfa-nightly`, which is built against `tf-nightly`. Nightly builds
+`tfa-nightly`, which is built against the latest stable version of TensorFlow. Nightly builds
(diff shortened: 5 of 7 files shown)


facaiy:
@seanpmorgan @tensorflow/sig-addons-maintainers what's your opinion?

seanpmorgan:
Seeing as there should be no regressions inserted into TF (barring experimental features which we do occasionally use), I'd be in favor of our unit tests being built against a stable TF release. 

As a check against drifting though, I would like to setup a periodic build/test against tf-nightly though so come release time we don't suddenly find out about several breaking issues.

Does that seem agreeable?

requirements.txt
```diff
@@ -1 +1 @@
-tf-nightly
\ No newline at end of file
+tensorflow==2.1.0
```
seanpmorgan: How would we handle this if #864 goes through? Make this file into `requirements-test.txt`?

gabrieldemarmiesse:
> As a check against drifting though, I would like to setup a periodic build/test against tf-nightly though so come release time we don't suddenly find out about several breaking issues.

Isn't it the primary purpose of release candidates? We'll have several weeks to adapt before the final release :)

seanpmorgan:
> > As a check against drifting though, I would like to setup a periodic build/test against tf-nightly though so come release time we don't suddenly find out about several breaking issues.
> 
> Isn't it the primary purpose of release candidates? We'll have several weeks to adapt before the final release :)

Fair. I guess I'm just still scarred from trying to keep up with breaking changes during TF2 beta. That should be sufficient and if we find out that it's not we can re-look into how we deal with that.

gabrieldemarmiesse:
for example, we can say that we're compatible with tf 2.1.0-2.2.0 and test against that. Then, when the release candidate for 2.3.0 comes, someone can make a PR to bump the version number in the CI. The CI will then run on 2.1.0 and 2.3.0rc0

Then on 2.1.0 and 2.3.0rc1 
then on 2.1.0 and 2.3.0.

I believe it's simpler than to setup periodic builds :) it's also easier for open source beginners to understand :) :)

seanpmorgan:
@gabrieldemarmiesse small conflict since merging the typegaurd PR

PR #921, opened 2020-01-23
CLN: Move build pip pkg script from root directory

seanpmorgan:
Bit of a legacy placement from tensorflow/custom-op. Makes sense to me to move this away from users who do not intend to build anything themselves.


gabrieldemarmiesse:
Thanks for the pull request!

PR #925, opened 2020-01-24
Enforce type info

### LINKED ISSUE #924: Add a test in the CI to ensure that typing information is added to the public API
To ensure that all new public functions/classes added are typed and old ones will be typed at some point.

I'll make a PR. This is mostly a reminder.

gabrieldemarmiesse:
Fix #924

Follow up on #894

This PR provides us with a TODO list of typing. It will be linked to an issue.

When a contributors adds a feature, he/she can either type his/her feature's public API (class constructor of function signature) or add his/her feature to the TODO list.

If we want to completely enfore typing, we can ask contributors to avoid adding something to the exception list.

I added also an option to install tf addons in editable mode without having to build (useful for the CI and local developement).


.github/workflows/ci_test.yml
```diff
@@ -66,3 +66,17 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-python@v1
+        with:
+          python-version: '3.7'
+      - name: install tensorflow and typeguard
+        run: pip install tensorflow-cpu typeguard
+      - name: install tf addons
+        run: TF_ADDONS_NO_BUILD=1 pip install --no-deps -e .
```
seanpmorgan: We should add some information about this env variable in CONTRIBUTING.md

tools/ci_build/verify/check_typing_info.py
```diff
@@ -0,0 +1,219 @@
...
+
+import tensorflow_addons
+
+EXAMPLE_URL = 'https://github.com/tensorflow/addons/blob/fa8e966d987fd9b0d20551a666e44e2790fdf9c7/tensorflow_addons/layers/normalizations.py#L73'
+TUTORIAL_URL = 'https://docs.python.org/3/library/typing.html'
+
+
+# TODO: add types and remove all elements from
+# the exception list.
+EXCEPTION_LIST = [
```
seanpmorgan: I like this system as an iterative way to add typing :) 

Probably worth describing this system in the docs?

.github/workflows/ci_test.yml
gabrieldemarmiesse: Done

tools/ci_build/verify/check_typing_info.py
gabrieldemarmiesse: Done

PR #926, opened 2020-01-24
Add AverageModelCheckpoint

### LINKED ISSUE #903: MovingAverage cannot be used with callbacks.ModelCheckpoint
Mostly, I save the model at the point of highest val_accuracy by `callbacks.ModelCheckpoint`, but the `MovingAverage` optimizer doesn't seem to support `callbacks.ModelCheckpoint`, i.e., we cannot save the model with moving average in the middle of training processing!

I think we should add MovingAverage into callbacks, and save the moving average model in the middle of training.

**Relevant information**
- Are you willing to contribute it (yes/no):
- Are you willing to maintain it going forward? (yes/no):
- Is there already an implementation in another framework? (if so, where): https://github.com/alno/kaggle-allstate-claims-severity/blob/master/keras_util.py

callbacks

all tf.keras users

**Any other info.**

Squadrick:
Closes #903


wuliytTaotao:
A question about AverageModelCheckpoint, when saving the moving average weights, we should not change the model.trainable_weights for we may keep training later. Does AverageModelCheckpoint change the model's weights after saving function?

tensorflow_addons/callbacks/average_model_checkpoint.py
```diff
@@ -0,0 +1,40 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# =============================================================================
+
+import tensorflow as tf
+from tensorflow_addons.optimizers.average_wrapper import AveragedOptimizerWrapper
+
+
+class AverageModelCheckpoint(tf.keras.callbacks.ModelCheckpoint):
```
shun-lin: NIT: add some comments here for the class. Otherwise LGTM :)

Squadrick: > A question about AverageModelCheckpoint, when saving the moving average weights, we should not change the model.trainable_weights for we may keep training later. Does AverageModelCheckpoint change the model's weights after saving function?

@wuliytTaotao I'll use a parameter that allows both options:
1. Assign the moving average weights to the model, and save them.
2. Keep the old non-averaged weights, but the saved model uses the average weights. 

That will cover both scenarios of the user.

Squadrick:
> A question about AverageModelCheckpoint, when saving the moving average weights, we should not change the model.trainable_weights for we may keep training later. Does AverageModelCheckpoint change the model's weights after saving function?

@wuliytTaotao I'll use a parameter that allows both options:
1. Assign the moving average weights to the model, and save them.
2. Keep the old non-averaged weights, but the saved model uses the average weights. 

That will cover both scenarios of the user.

seanpmorgan:
@Squadrick Thanks!! LGTM. Would you mind merging master into this. I believe we'll need to add type checking, but want to verify the CI

Squadrick:
@seanpmorgan Added type info, and it should be good to merge.

PR #931, opened 2020-01-25
Typing the lamb optimizer.

gabrieldemarmiesse:
Linked to #894


gabrieldemarmiesse:
That's definitly strange. 

I read the docs a second time, and it says:

> The decorator also respects the optimized mode setting so it does nothing when the interpreter is running in optimized mode.

Do you think it's related?

tensorflow_addons/optimizers/lamb.py
```diff
@@ -31,15 +34,16 @@ class LAMB(tf.keras.optimizers.Optimizer):
...
     def __init__(self,
-                 learning_rate=0.001,
-                 beta_1=0.9,
-                 beta_2=0.999,
-                 epsilon=1e-6,
-                 weight_decay_rate=0.0,
-                 exclude_from_weight_decay=None,
-                 exclude_from_layer_adaptation=None,
-                 name='LAMB',
+                 learning_rate: Union[FloatTensorLike, Callable] = 0.001,
```
seanpmorgan: Looks good. Did we ever figure out why this was only failing for linux?

gabrieldemarmiesse: Nope haha, my overall position on this is that we're very early in the history of tooling atound typing. We can try typeguard now, but if we see that it doesn't work well, we can drop it until the field is more mature.

PR #932, opened 2020-01-25
Enabling black progressively.

gabrieldemarmiesse:
With this method, we have a big todo list and we can start formatting files with black when there are no pull requests pending and working on those files.

Hooo boy this pull request was painful. Lesson learned, [pre-commit](https://pre-commit.com/) and customization don't play well together.

I removed the exception for dotted files in the gitignore because it didn't work to just write:
```
\!.pyproject.toml
```


STYLE_GUIDE.md
```diff
@@ -15,11 +15,25 @@ void formatted_code_again;
 ```
 
 #### Python
-Python code should conform to [PEP8](https://www.python.org/dev/peps/pep-0008/).
 
 Addons uses [flake8](http://flake8.pycqa.org/en/latest/) to check pep8 compliance and 
 code analysis.
 
+Addon use [Black](https://black.readthedocs.io/en/stable/) to format our code.
```
seanpmorgan: nit: Addons

STYLE_GUIDE.md
```diff
@@ -15,11 +15,25 @@ void formatted_code_again;
...
+Install them with:
+```
+pip install flake8 black
+```
+
+Be sure to run them both before you push your commits, otherwise the CI will fail!
+
+```
+python -m black ./
+python -m flake8
```
seanpmorgan: So at some point I would like to have this down to a single check. Once the rather large black reformatting is complete do you see a need for flake8? I believe as a subset of pep8 we should be fine (minus the line length)

STYLE_GUIDE.md
gabrieldemarmiesse: When reviewing pull requests and adding a nit, I think it would be nice to use the "suggest changes" feature of Github :) it makes faster iteration cycles.

gabrieldemarmiesse:
```suggestion
Addons use [Black](https://black.readthedocs.io/en/stable/) to format our code.
```

STYLE_GUIDE.md
gabrieldemarmiesse: Flake8 is an addition of multiple tools. If we enforce black, then we should disable pycodestyle. See http://flake8.pycqa.org/en/latest/user/error-codes.html

I'd preferto use flake8 with pycodestyle disabled. We get useful checks and we can add plugins to flake8 in the future. 

I'll disable pycodestyle in another pull request if you don't mind :)

.gitignore
```diff
@@ -1,10 +1,3 @@
-# Ignore all hidden files
-.*
-
-# Include travis.yml
-\!.travis.yml
-\!.flake8
```
seanpmorgan: I think it's okay to allow hidden files. If we see this as an issue we can revisit. It was a larger issue before when we were planning on doc files living within the repo that were auto generated.

PR #940, opened 2020-01-26
Moving tests into a separate directory and run them with unittest, not bazel

gabrieldemarmiesse:
To try simplify the developer's workflow and flexibility, we can run the tests without bazel.

As a bonus, we don't need to declare the tests now in the BUILD files.

Linked to #898


seanpmorgan:
Hmmmm, I'm less sold on this simplifying developer workflow. We'll still require users to run a `bazel build` prior to testing any of the python tests requiring custom ops. Currently the BUILD files specify which tests have a build dependency. This means its possible to call `py_test` targets without building the custom ops if they don't need it, but also allows users to build and test custom ops by building a single kernel.

With this setup we'll have to instruct users to always build prior to testing. IMO there isn't a major issue with using Bazel as the test runner. There's the overhead of a binary download but once bazel is installed it's never needed again.

The pros I see:
- [ ] Friendlier to pure-python developers
- [ ] Ability to use unittest plugins

The cons:
- [ ] Splinters our test rig into using 2 frameworks (bazel & unittest)
- [ ] Unable to run single tests for custom ops
- [ ] Some contributions require updating BUILD files while others don't

Squadrick:
@gabrieldemarmiesse Does unittest have caching support for tests? Haven't used unittest before, and not re-running unchanged tests during development is a pretty useful feature of bazel.

gabrieldemarmiesse:
I see, I didn't consider enough the C++/cuda use case. I'll try to find other ways to make it easier for python beginners to get started with the repo.

PR #941, opened 2020-01-26
Added a cache to the CI.

gabrieldemarmiesse:


seanpmorgan:
Reusing cache'd bazel builds is a good idea. But in this case I'm not sure the cache is being re-used by anything. Each OS only runs a single python version so as seen in the CI test times I don't believe there is any affect here?

gabrieldemarmiesse:
Yeah it doesn't really work great, I'll drop that for now.

PR #943, opened 2020-01-26
Added a small pre-commit hook for easier running of tools.

gabrieldemarmiesse:
This should make it easier for users to run the formatting tools we need.


pre-commit.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
seanpmorgan: Perhaps we move this into `tools` to prevent clutter on the project root?

gabrieldemarmiesse: Good point. Done.

PR #939, opened 2020-01-26
Add files via upload

failure-to-thrive:
#854

PR #945, opened 2020-01-27
[WIP] BLD: Use python3 in configuration

seanpmorgan:
Previously we kept the configuration as `python` since windows doesn't have a python3 symlink. I would prefer not leaving having the `ln -sf /usr/bin/python3.6 /usr/bin/python && rm /usr/bin/python2` in our instructions since a large majority of our developers run in the docker container.

SIG IO has done a similar thing:
https://github.com/tensorflow/io/blob/master/.github/workflows/build.yml#L203

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -42,7 +42,7 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
-        python-version: ['3.7']
+        python-version: ['3.7.6']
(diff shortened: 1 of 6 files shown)


seanpmorgan:
NOTE: This may be obsolete if we go with: https://github.com/tensorflow/addons/pull/939

seanpmorgan:
The more I think about it the more #939 makes sense. Closing this PR.

PR #947, opened 2020-01-27
BLD: Fix github actions build

seanpmorgan:
Expected this to take an iteration or two...

* Windows failed due to not specifying bash as shell
* Mac failed due to using pyenv (this PR replaces with build matrix)
* Linux failed due to pypa action not picking up the token correctly

### DIFF
--- a/tools/ci_build/builds/release_macos.sh
+++ b/tools/ci_build/builds/release_macos.sh
@@ -18,43 +18,27 @@ set -e -x
 # No GPU support for MacOS
 export TF_NEED_CUDA="0"
 
-PYTHON_VERSIONS="3.5.6 3.6.6 3.7.4"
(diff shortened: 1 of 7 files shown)


tools/ci_build/builds/release_macos.sh
```diff
@@ -18,43 +18,27 @@ set -e -x
 # No GPU support for MacOS
 export TF_NEED_CUDA="0"
 
-PYTHON_VERSIONS="3.5.6 3.6.6 3.7.4"
```
gabrieldemarmiesse: For the future usages: where are specified the target python versions for the wheels now?

gabrieldemarmiesse: My bad, it's in the github actions file.

PR #959, opened 2020-01-27
Add triplet hard loss with soft margin

HauserA:
Added batch hard loss with soft margin version as in [In Defense of the Triplet Loss for Person Re-Identification](https://arxiv.org/abs/1703.07737).

The difference to triplet_semihard_loss is the hard negative and hard positive mining and the option for a soft margin with the softplus-function instead of a hard margin.


HauserA:
@googlebot I signed it!

tensorflow_addons/losses/triplet.py
```diff
@@ -175,3 +227,43 @@ def get_config(self):
...
+        super().__init__(name=name, reduction=tf.keras.losses.Reduction.NONE)
+        self.margin = margin
+        self.soft = soft
+
+    def call(self, y_true, y_pred):
+        return triplet_hard_loss(y_true, y_pred, self.margin, self.soft)
+
+    def get_config(self):
+        config = {
+            "margin": self.margin,
```
WindQAQ: Should we also serialize `self.soft` here?

HauserA: Yes

PR #961, opened 2020-01-28
FIX: GitHub build

seanpmorgan:
Feedback cycle on this is painfully slow. If this doesn't work tomorrow then I'll modify the yaml so we can troubleshoot on each push.

PyPi Github action is only for linux. We could work around this by uploading /downloading the artifacts and pushing once. But I'd like to see the builds succeed and publish prior to optimizing so as to change as little as possible.


gabrieldemarmiesse:
Thanks for the PR!

Issue #964, opened 2020-01-28
TimeStopping Callback fail in verbose mode


RomainBrault:
**System information**

OS Platform and Distribution (Ubuntu 19.04):
TensorFlow version and how it was installed (pip==2.1.0):
TensorFlow-Addons version and how it was installed (pip==0.7.0):
Python version: 3.7
Is GPU used? (yes):

**Describe the bug**

When the fit method stop before the timer, it raises with

```AttributeError: 'TimeStopping' object has no attribute 'stopped_epoch'```

**Code to reproduce the issue**

Use tqdm callback

**Other info / logs**

Proposed fix

```python
...

    def on_epoch_end(self, epoch, logs={}):
        self.stopped_epoch = epoch
        if time.time() >= self.stopping_time:
            self.model.stop_training = True

...

    def on_train_end(self, logs=None):
        if self.verbose > 0:
            formatted_time = datetime.timedelta(seconds=self.seconds)
            msg = "Stopping at epoch {} after training for {}".format(
                self.stopped_epoch + 1, formatted_time
            )
            print(msg)
...

```

shun-lin:
Hi @RomainBrault thank you so much for catching the bug! I apologize I forgot to consider the obvious case where training stopped before proposed time, your proposed fix looks good to me (putting `self.stopped_epoch = epoch`) outside of `'if` statement in `on_epoch_end` method. If you have time, may you submit a simple PR for it? (If not I can do it) Thank you so much for your contribution!

failure-to-thrive:
I would recommend some kind of flag to distinguish whether `on_train_end` was called due to the "timeout" or by any other reason.  @RomainBrault, it will fix your "cosmetic problem", too. If no one doesn't mind, I can get down to this.

shun-lin:
> I would recommend some kind of flag to distinguish whether `on_train_end` was called due to the "timeout" or by any other reason. @RomainBrault, it will fix your "cosmetic problem", too. If no one doesn't mind, I can get down to this.

Hi @failure-to-thrive, great idea and I think it would be helpful! I think one thing to consider is to see what `EarlyStopping` Callback is doing in similar case (training finished regularly without `EarlyStopping` effect took place and `verbose` was set to `1`), so when we handle that case here it's similar. @failure-to-thrive thanks for your contribution in advance!

failure-to-thrive:
[...]

PR #967, opened 2020-01-29
Quick Fix for Time_Stopping Callback Bug

shun-lin:
Discovered by @RomainBrault in https://github.com/tensorflow/addons/issues/964, the time_stopping callback breaks when the training ends before the proposed time is passed, due to `self.stopped_epoch` been undefined when that case happened. This small PR fixed it. Thanks @RomainBrault for catching the bug and proposed fix!

### DIFF
--- a/tensorflow_addons/callbacks/time_stopping.py
+++ b/tensorflow_addons/callbacks/time_stopping.py
@@ -36,14 +36,15 @@ class TimeStopping(Callback):
 
         self.seconds = seconds
         self.verbose = verbose
+        self.stopped_epoch = 0
 
     def on_train_begin(self, logs=None):
         self.stopping_time = time.time() + self.seconds
 
     def on_epoch_end(self, epoch, logs={}):
+        self.stopped_epoch = epoch
         if time.time() >= self.stopping_time:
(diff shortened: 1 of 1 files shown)


RomainBrault:
Hello,

Thanks for the fix!

Now there is a 'cosmetic' problem: when training stop before the maximum time allowed it will always print the maximum time. Thus the verbose mode is not so informative about the time. I would suggest to remove the formatted_time or print an elapsed time instead.

But this could be corrected in another PR as it is not a bug.

shun-lin:
> Hello,
> 
> Thanks for the fix!
> 
> Now there is a 'cosmetic' problem: when training stop before the maximum time allowed it will always print the maximum time. Thus the verbose mode is not so informative about the time. I would suggest to remove the formatted_time or print an elapsed time instead.
> 
> But this could be corrected in another PR as it is not a bug.

Thanks for the suggestion, @failure-to-thrive's proposed solution in the original issue would solve the cosmetic problem you suggested. It would be another PR by @failure-to-thrive, thanks so much for the suggestion!

shun-lin:
@seanpmorgan may you review this small PR? Thanks!

Issue #965, opened 2020-01-28
Difficult to use tensorflow-addons with tf-nightly builds


ghop02:
It's a bit tricky to use the tf-nightly builds and have a dependency on tensorflow addons because of the pinned tensorflow version.  

Is there a way to have the setup.py file for the nightly builds be slightly more lenient as to the version? It's a bit tricky because tf-nightly is an entirely different package.

**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): 
- TensorFlow version and how it was installed (source or binary): tensorflow nightly (pip)
- TensorFlow-Addons version and how it was installed (source or binary): (0.7.1)
- Python version: 3.7

**Code to reproduce the issue**
```
pip install tf-nightly
pip install tfa-nightly (or tensorflow-addons)
# installs tensorflow 2.1.0 which downgrades gast
```

seanpmorgan:
Hi @ghop02! Thanks for reporting. So we should have a soon incoming solution to this which is to remove the hard installation dependencies:
https://github.com/tensorflow/addons/issues/864
https://github.com/tensorflow/addons/issues/952

I plan to work on this later today when I have some time to rework the builds. If you need to unblock today you can `pip install tfa-nightly --no-deps`

ghop02:
Great, figured there was something in the works!

And never knew about `--no-deps`, will probably abuse this now for many carefully constructed brittle python environments 😄

seanpmorgan:
TFA Nightlies starting tomorrow will have no pip dependency. 0.8 release is upcoming and will also be dependency free

PR #974, opened 2020-01-30
BLD: Remove hard installation dep

### LINKED ISSUE #864: Tensorflow-cpu is ignored when installing tensorflow-addons

seanpmorgan:
Closes #864 #965


seanpmorgan:
Since this is major request and large change we should consider a 0.8 release after we finish PRs that are finalizing.

tensorflow_addons/__init__.py
```diff
@@ -14,6 +14,47 @@
...
+    """Attempt to import tensorflow, and ensure its version is sufficient.
+    Raises:
+      ImportError: if either tensorflow is not importable or its version is
+      inadequate.
+    """
+    try:
+        import tensorflow as tf
+    except ImportError:
+        # Print more informative error message, then reraise.
+        print(
```
gabrieldemarmiesse: I'm not sure this message is needed. A "no module named tensorflow" is already very clear.

If we don't need to print this message, then we don't need to break the flake8 rules and we can also move this check in the "utils" directory to avoid cluttering the init file.

build_deps/build-requirements.txt
```diff
@@ -0,0 +1 @@
+tensorflow==2.1.0
```
gabrieldemarmiesse: Maybe we should remove that too to be able to build with tensorflow, tensorflow-cpu, tf-nightly?

tensorflow_addons/__init__.py
seanpmorgan: Agree that "no module named tensorflow" is sufficient. I'm not sure that we can avoid violating flake8 here though. We'll need to verify the TF version prior to importing all of the TFA submodules so it'll have a import not at top regardless.

PTAL at most recent push

build_deps/build-requirements.txt
seanpmorgan: So that's a bit tricky. At the moment I believe we depend on some of the TF-GPU framework for our build (at least when compiling CUDA kernels). Certainly willing to look at a PR to change that but would rather merge this first.

Only developers should be using the build scripts so no risk of a general user running into an issue with this.

gabrieldemarmiesse: Sure thing. If it's not trivial, let's do that later on.

tensorflow_addons/__init__.py
gabrieldemarmiesse: True, I'll review when I get home. Thanks for the pull request!

configure.py
```diff
@@ -102,6 +102,9 @@ def generate_shared_lib_name(namespec):
 with open("requirements.txt") as f:
     _REQUIRED_PKG = f.read().splitlines()
 
+with open("build_deps/build-requirements.txt") as f:
+    _REQUIRED_PKG.extend(f.read().splitlines())
+
```
gabrieldemarmiesse: Can you add a comment explaining why we don't use the "-r" option of pip? Thanks!

build_deps/check_deps.py
```diff
@@ -26,8 +26,9 @@ def check_dependencies(requirement_file_name):
     except DistributionNotFound as e:
         print(e)
         sys.exit(1)
-    sys.exit(0)
 
 
 if __name__ == "__main__":
     check_dependencies("requirements.txt")
+    check_dependencies("build_deps/build-requirements.txt")
+    sys.exit(0)
```
gabrieldemarmiesse: It's the last line of the file. Not sure it's needed there.

configure.py
seanpmorgan: Sure, the reason is just so we can print the dependencies in L116. I can add a comment for that, but wanted to make sure you don't see a better method. 

Currently `check_deps` does not return what package is missing (and will require a refactor since its called 2x and can exit on the first one. The current setup is not ideal, but I'm not really sure its worth the trouble to change it.

gabrieldemarmiesse: Yeah, this PR is important. I don't want to block it for something like that. Let's just remove the exit(0) and I'll merge it.

seanpmorgan: Done. Thanks

gabrieldemarmiesse:
Thanks for the pull request, looks good to me!

PR #978, opened 2020-01-30
Typing optimizers

autoih:


tensorflow_addons/optimizers/average_wrapper.py
```diff
@@ -16,11 +16,17 @@
...
 import tensorflow as tf
+from typeguard import typechecked
 
 
 class AveragedOptimizerWrapper(tf.keras.optimizers.Optimizer, metaclass=abc.ABCMeta):
+    @typechecked
     def __init__(
-        self, optimizer, sequential_update=True, name="AverageOptimizer", **kwargs
+        self,
+        optimizer,
```
gabrieldemarmiesse:
```suggestion
        optimizer: tf.keras.optimizers.Optimizer,
```

autoih: Thanks @gabrieldemarmiesse.

tensorflow_addons/optimizers/cyclical_learning_rate.py
```diff
@@ -15,19 +15,24 @@
...
+    @typechecked
     def __init__(
         self,
-        initial_learning_rate,
-        maximal_learning_rate,
-        step_size,
+        initial_learning_rate: Union[FloatTensorLike, Callable],
+        maximal_learning_rate: Union[FloatTensorLike, Callable],
+        step_size: FloatTensorLike,
         scale_fn,
```
gabrieldemarmiesse:
```suggestion
        scale_fn: Callable,
```

autoih: Got it, thanks @gabrieldemarmiesse.

tensorflow_addons/optimizers/average_wrapper.py
seanpmorgan: Since Keras accepts optimizer class or string names of optimizers this will need to be a Union:
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/optimizers/lookahead_test.py#L150

autoih: Got it! Thanks @seanpmorgan.

tensorflow_addons/optimizers/weight_decay_optimizers.py
```diff
@@ -170,7 +175,8 @@ def _resource_apply_sparse(self, grad, var, indices):
             return super()._resource_apply_sparse(grad, var, indices)
 
 
-def extend_with_decoupled_weight_decay(base_optimizer):
+def extend_with_decoupled_weight_decay(
+        base_optimizer: tf.keras.optimizers) -> tf.keras.optimizers:
```
gabrieldemarmiesse: tf.keras.optimizer is a module, not a class. I think you mean `tf.keras.optimizers.Optimizer`. Also you can use the typechecked decorator here since there is no risk for this function to be converted using autograph.

seanpmorgan: Is this another case to Union string and optimizer since Keras accepts either or?

gabrieldemarmiesse: Nope, not here. There is a subclassing done just after.

autoih: Yes, thanks @gabrieldemarmiesse. I've tried, but have the following error msg, `TypeError: type of argument "base_optimizer" must be tensorflow.python.keras.optimizer_v2.optimizer_v2.OptimizerV2; got tensorflow.python.keras.optimizer_v2.gradient_descent.SGD instead`. I'm not sure why can't pass the test.

gabrieldemarmiesse: I get it now! It's a class, not an optimizer.

That's a strange API but ok. I don't even know how to put typing hints. I'm looking for something like `typing.class` but I can't find it.

autoih: I am searching for too, but no idea yet, should I use `tf.keras.optimizers` instead?

WindQAQ: Hi, have you tried `Type[tf.keras.optimizers.Optimizer]`?

https://docs.python.org/3/library/typing.html#typing.Type

gabrieldemarmiesse: Thanks @WindQAQ you're right!

autoih: Thanks @WindQAQ. It may work 👍

PR #979, opened 2020-01-30
Nicer API for configure.py

gabrieldemarmiesse:
It's not because we're not users that we can't have a nice API :p


gabrieldemarmiesse:
I have no idea what's going on with the sanity check.

seanpmorgan:
> I have no idea what's going on with the sanity check.

Will take a look in a bit. I'm also going to post an issue tonight about breaking up or refactoring sanity check now that we have a new test setup

seanpmorgan:
So passing in `echo y` saying yes to build with GPU since there is no more prompt for install deps. The other CI tests are getting around this since they explicitly set `TF_NEED_CUDA` env to 0. 

Should be able to just export that appropriately in sanity check and then we can remove the `echo y` from all of these.

gabrieldemarmiesse:
I see! I don't get why but I though passing "y" said yes to both questions. At least I'll be the last making this mistake :)

PR #981, opened 2020-01-31
Layers: typing cleanup

autoih:


tensorflow_addons/layers/normalizations.py
```diff
@@ -311,6 +311,7 @@ class InstanceNormalization(GroupNormalization):
         (https://arxiv.org/abs/1607.08022)
     """
 
+    @typechecked
```
gabrieldemarmiesse: I think we're going to get warnings if we do that. There are no types, so it's not necessary to do any check. If the variables were listed explicitly and typed, yes it would make sense to add the decorator

autoih: Thanks @gabrieldemarmiesse.

PR #1004, opened 2020-02-02
fix WeightNormalization

### LINKED ISSUE #949: WeightNormalization doesn't work with @tf.function in an edge case
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): macOS 10.15.2 (19C57)
- TensorFlow version and how it was installed (source or binary): 2.1.0 via pip
- TensorFlow-Addons version and how it was installed (source or binary): 0.8.0-dev via pip
- Python version: 3.7.4
- Is GPU used? (yes/no): no

**Describe the bug**

When the train function is decorated with @tf.function, using multiple WeightNormalization layers in a multi-output model will cause the training process hanging on without any meaningful output logs.

The model would work if:
- Comment out the @tf.function
- Using only one WeightNormalization layer
- Using only one output in the InnerModel

I don't really know what happen, I post the full code to reproduce the issue. The issue also reproducible on Linux (ubuntu 16.04) with GPU (RTX 2080, CUDA 10.1) enabled.

```python
import tensorflow as tf
import numpy as np

class InnerModel(tf.keras.Model):
    def __init__(self):
        super().__init__()
        self.model1 = tf.keras.Sequential([
            tfa.layers.WeightNormalization(
                tf.keras.layers.Conv2D(3, 1),
            ),
        ])
        self.model2 = tf.keras.Sequential([
            tfa.layers.WeightNormalization(
                tf.keras.layers.Conv2D(3, 1),
            ),
        ])

    def call(self, inputs):
        out1 = self.model1(inputs)
        out2 = self.model2(out1)
        return out1, out2

class OuterModel(tf.keras.Model):
    def __init__(self):
        super().__init__()
        self.model1 = InnerModel()
        self.model2 = InnerModel()
        self.downsample = tf.keras.layers.AveragePooling2D(2, 1, 'same')

    def call(self, inputs):
        init_out = self.model1(inputs)
        inputs = self.downsample(inputs)
        final_out = self.model2(inputs)
        return init_out, final_out

def loss_fn(out):
    loss = 0.
    for i in range(len(out)):
        for j in range(len(out[i])):
            loss += tf.reduce_mean(out[i][j])
    return loss

net = OuterModel()
opt = tf.keras.optimizers.Adam(1e-4)

@tf.function
def train_step(inputs):
    with tf.GradientTape() as tape:
        out = net(inputs)
        loss = loss_fn(out)
    grads = tape.gradient(loss, net.trainable_variables)
    opt.apply_gradients(zip(grads, net.trainable_variables))

def train():
    for i in range(20):
        x = np.random.randn(1, 32, 32, 3)
        train_step(x)
        print(f'step {i}')

train()
```

**Other info / logs**

```
2020-01-27 15:35:40.604091: I tensorflow/core/platform/cpu_feature_guard.cc:142] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX2 FMA
2020-01-27 15:35:40.624043: I tensorflow/compiler/xla/service/service.cc:168] XLA service 0x7fc57e1f3f50 initialized for platform Host (this does not guarantee that XLA will be used). Devices:
2020-01-27 15:35:40.624082: I tensorflow/compiler/xla/service/service.cc:176]   StreamExecutor device (0): Host, Default Version
WARNING:tensorflow:Layer outer_model is casting an input tensor from dtype float64 to the layer's dtype of float32, which is new behavior in TensorFlow 2.  The layer has dtype float32 because it's dtype defaults to floatx.

If you intended to run this layer in float32, you can safely ignore this warning. If in doubt, this warning is likely only an issue if you are porting a TensorFlow 1.X model to TensorFlow 2.

To change all layers to have dtype float64 by default, call `tf.keras.backend.set_floatx('float64')`. To change just this layer, pass dtype='float64' to the layer constructor. If you are the author of this layer, you can disable autocasting by passing autocast=False to the base Layer constructor.
```

fsx950223:
fix #949

Is it format correct?

### DIFF
--- a/tensorflow_addons/layers/wrappers.py
+++ b/tensorflow_addons/layers/wrappers.py
@@ -61,7 +61,7 @@ class WeightNormalization(tf.keras.layers.Wrapper):
         self._track_trackable(layer, name='layer')
         self._init_critical_section = tf.CriticalSection(name='init_mutex')
         self.is_rnn = isinstance(self.layer, tf.keras.layers.RNN)
-
+        self.temp_kernel=None
         if self.data_init and self.is_rnn:
             logging.warning(
                 "WeightNormalization: Using `data_init=True` with RNNs "
@@ -119,31 +119,34 @@ class WeightNormalization(tf.keras.layers.Wrapper):
                     self._naked_clone_layer.activation = None
 
         self.built = True
-
+        
+    @tf.function(autograph=False)
     def call(self, inputs):
         """Call `Layer`"""
 
         def _do_nothing():
             return tf.identity(self.g)
 
+
         def _update_weights():
             # Ensure we read `self.g` after _update_weights.
             with tf.control_dependencies(self._initialize_weights(inputs)):
                 return tf.identity(self.g)
-
+        
         g = self._init_critical_section.execute(lambda: tf.cond(
             self._initialized, _do_nothing, _update_weights))
-
+            
         with tf.name_scope('compute_weights'):
             # Replace kernel by normalized weight variable.
             kernel = tf.nn.l2_normalize(self.v, axis=self.kernel_norm_axes) * g
-
+            
             if self.is_rnn:
                 self.layer.cell.recurrent_kernel = kernel
                 update_kernel = tf.identity(self.layer.cell.recurrent_kernel)
             else:
                 self.layer.kernel = kernel
                 update_kernel = tf.identity(self.layer.kernel)
+                
 
             # Ensure we calculate result after updating kernel.
             with tf.control_dependencies([update_kernel]):
@@ -215,7 +218,7 @@ class WeightNormalization(tf.keras.layers.Wrapper):
         kernel = tf.Variable(
             tf.nn.l2_normalize(self.v, axis=self.kernel_norm_axes) * self.g,
             name='recurrent_kernel' if self.is_rnn else 'kernel')
-
+        
         if self.is_rnn:
             self.layer.cell.recurrent_kernel = kernel
         else:
--- a/tensorflow_addons/layers/wrappers.py
+++ b/tensorflow_addons/layers/wrappers.py
@@ -19,7 +19,7 @@ import tensorflow as tf
 from typeguard import typechecked
 
 
-@tf.keras.utils.register_keras_serializable(package='Addons')
+@tf.keras.utils.register_keras_serializable(package="Addons")
 class WeightNormalization(tf.keras.layers.Wrapper):
     """This wrapper reparameterizes a layer by decoupling the weight's
     magnitude and direction.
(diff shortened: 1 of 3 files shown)


tensorflow_addons/layers/wrappers_test.py
```diff
@@ -165,9 +164,14 @@ def test_removal(self, base_layer_fn, input_shape):
...
 
+            if wn_layer.is_rnn:
+                kernel = wn_layer.layer.cell.recurrent_kernel
+                remove_kernel = wn_removed_layer.cell.recurrent_kernel
+            else:
+                kernel = wn_layer.layer.kernel
+                remove_kernel = wn_removed_layer.kernel
             self.evaluate(tf.compat.v1.global_variables_initializer())
-            self.assertAllClose(
-                self.evaluate(wn_removed_output), self.evaluate(wn_output))
```
Squadrick: Why remove this check?

tensorflow_addons/layers/wrappers.py
```diff
@@ -92,34 +93,36 @@ def build(self, input_shape):
...
+                layer_config["config"]["trainable"] = False
+                self._naked_clone_layer = tf.keras.layers.deserialize(layer_config)
                 self._naked_clone_layer.build(input_shape)
                 self._naked_clone_layer.set_weights(self.layer.get_weights())
                 if not self.is_rnn:
                     self._naked_clone_layer.activation = None
 
         self.built = True
 
+    @tf.function(autograph=False)
```
Squadrick: I'm guessing this fixes the issue. Can you add a test to check for this?

fsx950223: Sorry, I don't know how to write such test case

tensorflow_addons/layers/wrappers_test.py
fsx950223: ```test_removal``` failed and I believe check kernel is more reasonable.

fsx950223: Old ```test_removal``` failed and I believe check kernel is more reasonable.

seanpmorgan:
Small flake8 error:
`./tensorflow_addons/layers/wrappers_test.py:165:17: F841 local variable 'wn_removed_output' is assigned to but never used`

fsx950223:
> Small flake8 error:
> `./tensorflow_addons/layers/wrappers_test.py:165:17: F841 local variable 'wn_removed_output' is assigned to but never used`

Could I remove CriticalSection? I believe there is no need to guarantee atomic and I get pass test cases if I remove CriticalSection.

Squadrick:
See https://github.com/failure-to-thrive/addons/commit/ec008b35e89d810bc7d7b41960664f98859336ea

cc @failure-to-thrive

fsx950223:
Decorate ```call``` only with ```tf.function``` manually fix the magical bug.

failure-to-thrive:
> Manually decorate `call` only with `tf.function` fix the magical bug.
> As `tf.function` has its own [variable creator](https://github.com/tensorflow/tensorflow/blob/cf7fcf164c9846502b21cebb7d3d5ccf6cb626e8/tensorflow/python/eager/def_function.py#L436) may fix the issue.

That way every instance of `WeightNormalization` class has the AutoGraph for its `call` turned on by default. Is it OK?

fsx950223:
@failure-to-thrive Every keras layer has autograph default, what's your meaning?

fsx950223:
It seems @failure-to-thrive has better solution.

failure-to-thrive:
> It seems @failure-to-thrive has better solution.

Thanks! If everything else is OK, I'm ready to PR.

PR #1006, opened 2020-02-02
changedthreadpool->ParallelFor instead of shard

### LINKED ISSUE #1000: Use threadpool->ParallelFor instead of Shard
As [Shard](https://github.com/tensorflow/tensorflow/blob/e4f4579e42155801f91156a53970d54a2ba0f54f/tensorflow/core/util/work_sharder.h#L26) is marked as DEPRECATED.
The custom op implementations should use threadpool->ParallelFor instead of Shard.
For example:
``` cpp
    auto worker_threads = *(context->device()->tensorflow_cpu_worker_threads());
    Shard(worker_threads.num_threads, worker_threads.workers, oN * oH * oW,
          cost_per_pixel, work);
```
**Relevant information**
- Are you willing to contribute it (yes/no):yes
- Are you willing to maintain it going forward? (yes/no):yes
- Is there already an implementation in another framework? (if so, where):

**Any other info.**

PSY27:
fixes #1000


PSY27:
@googlebot I signed it!

WindQAQ:
Closes via #1005.

Issue #991, opened 2020-01-31
Refactor Sanity Check


seanpmorgan:
The term sanity check is a bit legacy from the TF core repo. In the past month we've done some major revamping our of repo and it's time to update this to align.

At the moment we break out flake8, black, and typing into separate CI tests:
https://github.com/tensorflow/addons/blob/master/.github/workflows/ci_test.yml#L60-L102

Personally I like this from a PR perspective because it limits the logs you need to read through to find your problem. I did want to discuss this though since it does make this list of test cases in PRs quite lengthy.

If we do keep this setup, I think it makes sense to break up the sanity check into its respective tests:
* C++ code format
* Bazel code format / no-build test
* Python check filename test

An alternative option would be to consolidate Lint tests under a single github action. For example we could add black and flake8 to https://github.com/tensorflow/addons/blob/master/tools/ci_build/code_format.sh

I think there is a number of ways to go about this so opening this issue to discuss and then we can implement.

gabrieldemarmiesse:
For the CI, I believe that fast feedback is very important for the developer's experience. Black action takes  around 15s to run, flake8 too, the typing check takes 40s and I also agree with @seanpmorgan about logs clarity which is important for first contributors to addons.

I'd be in favor of a tool that we can run and do all the checks at once locally, but in the CI, those checks should stay separate to improve the time it takes to do a cycle of push/fix.

gabrieldemarmiesse:
I've made 4 pull requests to extract the 4 checks from the sanity check. Once they are all merged, I'll work on the tool to run them locally (docker + git hook).

gabrieldemarmiesse:
Can we completely assume that users have access to docker locally? It would make the pre-commit hook and local testing much easier to implement.

seanpmorgan:
> Can we completely assume that users have access to docker locally? It would make the pre-commit hook and local testing much easier to implement.

Yes I think we can safely assume this for developers.

gabrieldemarmiesse:
Ho yeah, fasten your seat belts it's going to awesome really fast!

PR #1008, opened 2020-02-02
Moved the check_file_name.py outside sanity check.

gabrieldemarmiesse:
Linked to #991 , let's break down the sanity check and make a new one afterwards once the CI has them all into separate checks.

It runs in 4 seconds. Quite good for the developer's feedback.


.github/workflows/ci_test.yml
```diff
@@ -100,6 +100,16 @@ jobs:
         run: TF_ADDONS_NO_BUILD=1 pip install --no-deps -e .
       - name: Run type check
         run: python tools/ci_build/verify/check_typing_info.py
+  case-insensitive-filesystem:
+    name: Works on case insensitive FS
```
seanpmorgan: Mind just renaming this slightly for clarity in the CI test names that are displayed. `Verify python filenames work` or `Verify python filenames work on case sensitive FS`

I believe it should be `case sensitive` btw so might need to rework the comment in the `check_file_name` file

gabrieldemarmiesse: Are you sure it's case sensitive? Having two filenames with the same name but different cases doesn't cause any issue in a case sensitive filesystem like linux, but will cause an issue in case insensitive filesystems like Windows.

seanpmorgan: Apologies, you're correct. Too early in the morning :)

seanpmorgan:
So currently we still have sanity_check in the docs:
https://github.com/tensorflow/addons/blob/master/CONTRIBUTING.md#development-tips

gabrieldemarmiesse:
I'll make a new sanity check/pre-commit as soon as this PR is merged. I like to keep them small for ease of review.

PR #1010, opened 2020-02-02
Is#964

failure-to-thrive:


shun-lin:
Linking https://github.com/tensorflow/addons/issues/964

PR #1019, opened 2020-02-03
Add typing matthews correlation coefficient, f1 score, multilabel confusion matrix, r square 

autoih:


tensorflow_addons/utils/types.py
```diff
@@ -41,4 +41,4 @@
 Constraint = Union[None, dict, str, Callable]
 
 TensorLike = Union[List[Union[Number, list]], Number, np.ndarray, tf.Tensor]
-FloatTensorLike = Union[tf.Tensor, float, np.float16, np.float32, np.float64]
+FloatTensorLike = Union[tf.Tensor, float, np.float16, np.float32, np.float64, str]
```
seanpmorgan: Why add string to FloatTensorLike? Seems like that breaks the typing pattern and probably better to Union a FloatTensorLike and String if it's required.

autoih: Thanks, @seanpmorgan. Yes, I think there is a default value, i.e., None, for dtype, if no `str` in FloatTensorLike, an error msg `TypeError: type of argument "dtype" must be one of (Tensor, float, float16, float32, NoneType); got str instead`.  May I have your suggestion?

AakashKumarNain: Wouldn't this work?
`FloatTensorLike = Optional[tf.Tensor, float, np.float16, np.float32, np.float64]`

The `Optional[x]` works as `Union[x, None]`

autoih: Thanks @AakashKumarNain. Yes, according to https://docs.python.org/3/library/typing.html#typing.Optional, it should work, but after have testing, shows the error. What would be your suggestions?

AakashKumarNain: This is weird. I will look into it

AakashKumarNain: @autoih this would work:
`FloatTensorLike = Optional[Union[tf.Tensor, float, np.float16, np.float32, np.float64]]`

I quickly tested it here:
https://colab.research.google.com/drive/10OjpjSrLTw3-LQoXPZM4xsz2wYqnq_Ue

AakashKumarNain: @seanpmorgan thoughts?

gabrieldemarmiesse: Shouldn't `dtype` be `Union[tf.DType, str]`?

AakashKumarNain: @gabrieldemarmiesse in the colab?

autoih: Followed the https://github.com/tensorflow/addons/pull/1026 fixed the issue.

gabrieldemarmiesse: I'm happy to help :)

gabrieldemarmiesse:
Thanks a lot!

PR #1023, opened 2020-02-05
FIX: Remove trailing comma from double star

seanpmorgan:
Windows/MacOS builds failed last night on Py35 for trailing comma.

Related: https://github.com/psf/black/issues/419

But going forward we should probably test oldest python and newest python in CI


gabrieldemarmiesse:
Thanks for the pull request!

PR #764, opened 2019-12-13
FIX: Protect WeightNorm init with a mutex

Squadrick:
Closes #722

Issue #722, opened 2019-11-27
tfa.layers.WeightNormalization fails with Assert when called by multiple threads


dominikroblek:
### Issue description

`tfa.layers.WeightNormalization` fails with the following Tensorflow assert failure when used by multiple threads (e.g., when the same layer wrapped by `tfa.layers.WeightNormalization` is part of two parallel execution flows in a Tensorflow graph while reusing the variables):
```
(0) Invalid argument: assertion failed: [The layer has been initialized.] [Condition x == y did not hold element-wise:] [x (discriminator/discriminator_1/n_layer_discriminator/sequential_21/grouped_conv1d_3/weight_normalization_304/cond/assert_equal_1/ReadVariableOp:0) = ] [1] [y (discriminator/discriminator_1/n_layer_discriminator/sequential_21/grouped_conv1d_3/weight_normalization_304/cond/assert_equal_1/y:0) = ] [0]
[[{{node discriminator/discriminator_1/n_layer_discriminator/sequential_21/grouped_conv1d_3/weight_normalization_304/cond/assert_equal_1/Assert/AssertGuard/Assert}}]]
```

### Proposed solution

Add the following line of code to the `tfa.layers.WeightNormalization.build` function:
```
self._initialized_critical_section = tf.CriticalSection(name='initialized_mutex')
```

Replace `g = tf.cond(self._initialized, _do_nothing, _update_weights)` in the `tfa.layers.WeightNormalization.call` function with the following lines of code:
```
def _get_initialized_weights():
    return tf.cond(self._initialized, _do_nothing, _update_weights)
g = self._initialized_critical_section.execute(_get_initialized_weights)
```

WindQAQ:
Hi @dominikroblek, thanks for the report! Are you willing to fill a PR and add some unittests for it? Thank you again.

failure-to-thrive:
What I see from the assert message (_[The layer has been initialized.]_) - it has nothing to do with the race conditions of parallel execution. It's a "multiple initializations" problem. CriticalSection won't help. Atomic flag (test-and-set lock mechanism) should have been there. And a piece of code to reproduce the problem, in the first place.

PR #1025, opened 2020-02-05
Share critical section across WeightNormalization layers

### LINKED ISSUE #949: WeightNormalization doesn't work with @tf.function in an edge case

failure-to-thrive:
fixes #949


fsx950223:
Why different layers' variables should share same mutex?

failure-to-thrive:
> Why different layers' variables should share same mutex?

According to #722:

> when the same layer wrapped by tfa.layers.WeightNormalization is part of two parallel execution flows

fsx950223:
> > Why different layers' variables should share same mutex?
> 
> According to #722:
> 
> > when the same layer wrapped by tfa.layers.WeightNormalization is part of two parallel execution flows


> > Why different layers' variables should share same mutex?
> 
> According to #722:
> 
> > when the same layer wrapped by tfa.layers.WeightNormalization is part of two parallel execution flows

Need a test case to check it.

failure-to-thrive:
> Need a test case to check it.

Totally agree with you! Without a test case it's impossible to find out what **really** happened in the first place. @dominikroblek should provide it. Otherwise, we are merely adding bugs to another bugs. :)

seanpmorgan:
@failure-to-thrive could you please fix the merge conflict so we can test and merge this. The original addition of this should have definitely came with a test case but I would prefer to fix this dead-lock issue even though we haven't been able to re-create the original issue.

Thanks very much for this!

failure-to-thrive:
Unfortunately, currently there is no solution. Any option spawns new glitches. Moreover, if we look close to the original #722 issue, there might be better solution then Critical Sections. Unfortunately, OP didn't provide a test case to clearly understand what goes wrong. So I suggest to revert the #764 PR back as well as close this PR. Should the #722 like issue arise again, we will fix it in the better way.

seanpmorgan:
> Unfortunately, currently there is no solution. Any option spawns new glitches. Moreover, if we look close to the original #722 issue, there might be better solution then Critical Sections. Unfortunately, OP didn't provide a test case to clearly understand what goes wrong. So I suggest to revert the #764 PR back as well as close this PR. Should the #722 like issue arise again, we will fix it in the better way.

Thanks... agreed there's no need for this complexity without a valid test case that shows how it breaks.

PR #1026, opened 2020-02-05
Added type of dtype and fix unused parameter and fix broken master branch

gabrieldemarmiesse:
See https://github.com/tensorflow/tensorflow/blob/dea01bb56459c35b77730ff1261c6306b75314d4/tensorflow/python/framework/dtypes.py#L607 for the type of dtype.

I don't usually like to do multiple things in the same pull request but here I am... :p

### DIFF
--- a/tensorflow_addons/metrics/f_scores.py
+++ b/tensorflow_addons/metrics/f_scores.py
@@ -71,9 +71,9 @@ class FBetaScore(tf.keras.metrics.Metric):
         beta=1.0,
         threshold=None,
         name="fbeta_score",
-        dtype=tf.float32,
+        dtype=None,
     ):
-        super().__init__(name=name)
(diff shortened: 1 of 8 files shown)


tensorflow_addons/metrics/f_scores.py
```diff
@@ -64,16 +67,17 @@ class FBetaScore(tf.keras.metrics.Metric):
...
         num_classes,
         average=None,
         beta=1.0,
         threshold=None,
         name="fbeta_score",
-        dtype=tf.float32,
+        dtype: types.AcceptableDTypes = None,
     ):
-        super().__init__(name=name)
+        super().__init__(name=name, dtype=dtype)
```
gabrieldemarmiesse: the `dtype` parameter was unused. I also changed the defaut value to be consistant with keras (keras uses K.floatx() when None is given).

Issue #1024, opened 2020-02-05
Update CI tests to run oldest and newest python


seanpmorgan:
So I think this is worth discussing before implementing. We had a break last night because our CI didn't test py35:
#1023 

We can update this to test most recent and oldest python versions, but it'll ~double the CI test time for those jobs. Ultimately we'll be notified of a failure on the nightly CI builds so there's a safety net but comes at the cost of missing nightlies for the day.

ulf1:
Is it possible to tell the sanity-check or formatter for trailing commas in lists?

gabrieldemarmiesse:
@ulf1 running the sanity check with Python 3.5 should be enough.

PR #1027, opened 2020-02-05
Use py35 instead of 36 as much as possible.

gabrieldemarmiesse:
Python being really good at backward compatibility, I think all our tests should run with the oldest version of python we support and some selected tests should run with the most recent version that we support.

See #1024


seanpmorgan:
@gabrieldemarmiesse Thanks. I think this approach is a good middle ground and should have us sufficiently covered during CI tests

gabrieldemarmiesse:
I totally forgot to un-draft it.

PR #1035, opened 2020-02-06
Add a test for the documentation.

gabrieldemarmiesse:
For now it's not super fast to use docker build in the CI, but I tried without and for building the docs, it's not much faster.

Wait until [docker buildx](https://github.com/docker/buildx) becomes stable and we're going to have super fast CI builds with remote caching!


gabrieldemarmiesse:
Maybe the version of tensorflow docs should be pinned?

PR #1038, opened 2020-02-07
Remove dev from r0.8 branch

seanpmorgan:
Release process is extremely simple now. After merging I'll publish a release and after confirming it all works I'll update https://github.com/tensorflow/community/blob/master/sigs/addons/RELEASE.md


AakashKumarNain:
Looks good. So, we are testing things on the stable versions of TF now and not on nightly, right?

gabrieldemarmiesse:
Yes, we use the stable version.

AakashKumarNain:
Cool. Thanks

PR #1028, opened 2020-02-06
Clean-up of the docker tools for local dev.

gabrieldemarmiesse:

PR #1044, opened 2020-02-07
correct code-format name and remove mkdir

autoih:


CONTRIBUTING.md
```diff
@@ -54,7 +54,7 @@ us**
 ## Development Tips
 Try these useful commands below:
 
-* Format code automatically: `bash tools/run_docker.sh -c 'make code-format'`
```
gabrieldemarmiesse: It's already done in #1028

tools/pre-commit.sh
```diff
@@ -2,6 +2,5 @@
 
 set -e
 
-export DOCKER_BUILDKIT=1
```
gabrieldemarmiesse: This was done to avoid long lines. It makes it more readable.

tools/docker/Dockerfile_formatting
```diff
@@ -8,7 +8,6 @@ RUN bash buildifier.sh
 COPY tools/ci_build/install/clang-format.sh ./clang-format.sh
 RUN bash clang-format.sh
 
-RUN mkdir /addons
```
gabrieldemarmiesse: I didn't know doing WORKDIR created a directory! :) Good to know, it'll save us some typing in the future.

CONTRIBUTING.md
autoih: Thanks @gabrieldemarmiesse.

tools/pre-commit.sh
autoih: I see. Just wanna be consistent wiht `run_sanity_check.sh`, that's why I remove it. I'll add it back. Thanks @gabrieldemarmiesse.

Issue #906, opened 2020-01-17
Problems with CohenKappa metrics


rere-rere:
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Ubuntu 18.04
- TensorFlow version and how it was installed (source or binary):  2.1.0
- TensorFlow-Addons version and how it was installed (source or binary): 0.7.0
- Python version: 3.6.9
- Is GPU used? (yes/no): yes

**Describe the bug**

Hello,
I would to use tfa.metrics.CohenKappa from tensorflow_addons.
I have a problem when I wanted to use it. I have a function where I create a basic convolution network, and I would like to use this metrics.

However, when I do that, it raised an exception 

> ValueError: Number of samples in `y_true` and `y_pred` are different

So I checked in the code, and it's seam that the shape of the two Tensor are the not the same : 

`Tensor("Cast:0", shape=(None, None), dtype=int64)
Tensor("Cast_1:0", shape=(None, 5), dtype=int64)`

I wanted to know, how I can I precise the shape of the y_pred in order to have the same shape as the y_true.

**Code to reproduce the issue**

```
def convolution(categories=5, shape_x=224, shape_y=224, channels=3):
    model = tf.keras.models.Sequential([
      tf.keras.layers.Conv2D(10, kernel_size=(5, 5), strides=(1, 1), activation=tf.nn.relu, use_bias=True, input_shape=(shape_x, shape_y, channels)),
      tf.keras.layers.MaxPooling2D(pool_size=(2, 2), strides=(1, 1), padding='valid'),
      tf.keras.layers.Conv2D(10, kernel_size=(5, 5), strides=(1, 1), activation=tf.nn.relu, use_bias=True),
      tf.keras.layers.MaxPooling2D(pool_size=(2, 2), strides=(1, 1), padding='valid'),
      tf.keras.layers.Flatten(),
      tf.keras.layers.Dense(128, activation=tf.nn.relu),
      tf.keras.layers.Dense(categories, activation=tf.nn.softmax)
    ])
    model.compile(optimizer='adam', loss='mse', metrics=[tfa.metrics.CohenKappa(num_classes=5)])
    return model
```

**Other info / logs**

>   File "/home/rere/Project/Aptos/aptos2019-blindness-detection/model.py", line 38, in convolution
>     model.compile(optimizer='adam', loss='mse', metrics=[tfa.metrics.CohenKappa(num_classes=5)])
[...]

PR #1047, opened 2020-02-08
Fix kappa

AakashKumarNain:
This PR fixes the problems mentioned in #906 

Highlights:

1. Works with/without keras model.
2. Can handle sparse labels for multi-class classification.
3. Can handle the cases where user want to treat the problem as a regression problem


AakashKumarNain:
@facaiy @seanpmorgan @WindQAQ This is the best I could think of to keep things as simple and clean as possible. But if you feel we need to change something, I am open to suggestions.

AakashKumarNain:
@googlebot I fixed it.

AakashKumarNain:
@googlebot I fixed it.

tensorflow_addons/metrics/cohens_kappa.py
```diff
@@ -67,18 +67,26 @@ def __init__(self,
...
+          weightage: (optional) Weighting to be considered for calculating
             kappa statistics. A valid value is one of
-            [None, 'linear', 'quadratic']. Defaults to `None`.
-          dtype: (Optional) Data type of the metric result.
-            Defaults to `None`.
+            [None, 'linear', 'quadratic']. Defaults to `None`
+          sparse_lables: (bool) Valid only for multi-class scenario.
+            If True, ground truth labels are expected tp be integers
+            and not one-hot encoded
+          reg_values: (bool) If set, that means the problem is being treated
```
seanpmorgan: nit: Could we make this `regression (bool)` . `reg_values` seems a bit underspecified since `reg` could mean regression, regularization, etc. Granted there wouldn't be regularization in this metric but its in the vocabulary.

tensorflow_addons/metrics/cohens_kappa.py
```diff
@@ -89,8 +97,18 @@ def __init__(self,
         if weightage not in (None, 'linear', 'quadratic'):
             raise ValueError("Unknown kappa weighting type.")
 
+        if num_classes == 2:
+            self._update = self._update_binary_class_model
```
seanpmorgan: Could you add a test case for `categorical_crossentropy` on a two class problem? I believe this will have issues.

tensorflow_addons/metrics/cohens_kappa.py
AakashKumarNain: Makes sense

tensorflow_addons/metrics/cohens_kappa.py
AakashKumarNain: You want test cases for binary case, right?

AakashKumarNain:
Thanks @seanpmorgan for the feedback. I have renamed that variable and added more tests to cover all the cases.

tensorflow_addons/metrics/cohens_kappa.py
AakashKumarNain: Done. Please check

seanpmorgan:
We'll add this to 0.8.2 patch release

AakashKumarNain:
Cool. Thanks for the valuable feedback

PR #1049, opened 2020-02-08
DOC: Update bld matrix

seanpmorgan:
* Removed 0.5.2 since there is no reason to use that instead of 0.6
* Added 0.8.1
* Updated 0.7.0 to 0.7.1

Do we think it is sufficient to just label the tensorflow version? tf-gpu or tf-cpu could be added but IMO it can be inferred.


gabrieldemarmiesse:
> tf-gpu or tf-cpu could be added but IMO it can be inferred.

Yes I don't think it's necessary.

PR #1058, opened 2020-02-09
Making the tests faster.

gabrieldemarmiesse:
Keep in mind that the total time vary a lot in the CI, so it might not be fair to compare using that. This is why I compare the speed of the steps:

| Step | before (s)| after (s)|
| ------------- |:-------------:| -----:|
| Downloading docker image | 74 | 22 |
| installing tf | 93 | 54|
| installing bazel | 0 | 8|

The speed of compiling and testing should in theory be the same.

I also think it's nice to start from scratch to build and run because:

* it helps other to understand how to build tensorflow addons from source locally. We don't start from a big docker image with who-knows-what already installed in it. 
* Here we have full control over the python version and bazel version for example. 

We might want to do the same for the gpu and releases.

### DIFF
--- a/tools/ci_testing/addons_cpu.sh
+++ b/tools/ci_testing/addons_cpu.sh
@@ -14,6 +14,8 @@
 # limitations under the License.
 #
 # ==============================================================================
+# usage: bash tools/ci_testing/addons_cpu.sh [--no-deps]
+
 set -x
 
 # Make sure we're in the project root path.
@@ -42,9 +44,9 @@ export TF_NEED_CUDA=0
 
 # Check if python3 is available. On Windows VM it is not.
 if [ -x "$(command -v python3)" ]; then
-    python3 ./configure.py
+    python3 ./configure.py $1
   else
-    python ./configure.py
+    python ./configure.py $1
 fi
 
 cat ./.bazelrc
--- a/tools/ci_testing/addons_cpu.sh
+++ b/tools/ci_testing/addons_cpu.sh
@@ -16,7 +16,13 @@
 # ==============================================================================
 # usage: bash tools/ci_testing/addons_cpu.sh [--no-deps]
 
-set -x
+set -x -e
+
+
+if [ "$1" != "--no-deps" ] && [ "$1" != "" ]; then
+  echo Wrong argument $1
+  exit 1
+fi
 
 # Make sure we're in the project root path.
 SCRIPT_DIR=$( cd ${0%/*} && pwd -P )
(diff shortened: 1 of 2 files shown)


seanpmorgan:
> Keep in mind that the total time vary a lot in the CI, so it might not be fair to compare using that. This is why I compare the speed of the steps:
> 
> Step	before (s)	after (s)
> Downloading docker image	74	22
> installing tf	93	54
> installing bazel	0	8
> The speed of compiling and testing should in theory be the same.
> 
> I also think it's nice to start from scratch to build and run because:
> 
> * it helps other to understand how to build tensorflow addons from source locally. We don't start from a big docker image with who-knows-what already installed in it.
> * Here we have full control over the python version and bazel version for example.
> 
> We might want to do the same for the gpu and releases.

So for the release we need that specific docker image since it has devtoolset-7 and will build manylinux2010 compatible wheels. For testing I believe this is fine, but will need to review when I have a little time.

seanpmorgan:
> > Keep in mind that the total time vary a lot in the CI, so it might not be fair to compare using that. This is why I compare the speed of the steps:
> > Step	before (s)	after (s)
> > Downloading docker image	74	22
> > installing tf	93	54
> > installing bazel	0	8
> > The speed of compiling and testing should in theory be the same.
> > I also think it's nice to start from scratch to build and run because:
> > 
> > * it helps other to understand how to build tensorflow addons from source locally. We don't start from a big docker image with who-knows-what already installed in it.
> > * Here we have full control over the python version and bazel version for example.
> > 
> > We might want to do the same for the gpu and releases.
> 
> So for the release we need that specific docker image since it has devtoolset-7 and will build manylinux2010 compatible wheels. For testing I believe this is fine, but will need to review when I have a little time.

To expand on this, we could (and it would be nice) to create our own image. But there are some things we have to get correct.

tools/ci_testing/addons_cpu.sh
```diff
@@ -42,9 +44,9 @@ export TF_NEED_CUDA=0
 
 # Check if python3 is available. On Windows VM it is not.
 if [ -x "$(command -v python3)" ]; then
-    python3 ./configure.py
+    python3 ./configure.py $1
   else
-    python ./configure.py
```
seanpmorgan: Would you mind adding a check of the passed arg and return error if it's not `no-deps`

gabrieldemarmiesse: Sure thing

gabrieldemarmiesse:
> To expand on this, we could (and it would be nice) to create our own image. But there are some things we have to get correct.

Sure thing. Building it is a first step :)

tools/ci_testing/addons_cpu.sh
gabrieldemarmiesse: This file would actually deserve to be a python file. Maybe I'll open a pr to convert it in the future.

PR #1059, opened 2020-02-09
add typing skip_gram function and remove typing info file

autoih:


gabrieldemarmiesse:
Please keep the file, it'll need to be there when people add new features to the API :)

gabrieldemarmiesse:
You can just have an empty list for exceptions.

tensorflow_addons/text/skip_gram_ops.py
```diff
@@ -19,27 +19,30 @@
...
-    seed=None,
-    name=None,
-):
+    input_tensor: TensorLike,
+    min_skips: FloatTensorLike = 1,
+    max_skips: FloatTensorLike = 5,
+    start: FloatTensorLike = 0,
+    limit: FloatTensorLike = -1,
+    emit_self_as_target: bool = False,
+    vocab_freq_table: AcceptableDTypes = None,
```
gabrieldemarmiesse: It seems to be a lookup table, not a data type.

autoih: Thanks @gabrieldemarmiesse. I added `tf.lookup.KeyValueTensorInitializer` in `AcceptableDTypes`.  Maybe just place here directly.

gabrieldemarmiesse: Indeed. Having `vocab_freq_table=tf.float32` wouldn't be correct for exemple.

autoih: Got it. Thanks @gabrieldemarmiesse.

gabrieldemarmiesse:
Thanks a lot!

Issue #951, opened 2020-01-27
[Feature] Add MultiHeadAttention Layer


cgarciae:
Currently `tf.keras.layers` doesn't have a `MultiHeadAttention` module which is a bit odd given the popularity Transformers have today and that Attention is All You Need and BERT came out of Google. Maybe this layer could be added to this repo initially at least and maybe migrate to the official in the future. 

I have an implementation based on the code in the BERT repo I could share which I tweaked a little bit to make it easier to use if you are open to it. Either way, it would be good for the community to have easy access to this layer.

Squadrick:
@cgarciae Thanks for bringing this up, and it would great if you could create a PR with your implementation.

cgarciae:
@Squadrick WIll do.

bhack:
@cgarciae Just cause your are interested in the topic, take a look at https://github.com/Nandan91/ULSAM

seanpmorgan:
This was merged. Closing but feel free to keep discussion going in this issue.

Issue #992, opened 2020-01-31
Unified Sub-package Testing


seanpmorgan:
Similar to [Keras layer_test](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/testing_utils.py#L75) it would be nice to have a suite of tests that all addons in a subpackage are verified against.

We've seen a number of times an addon which looked correct but later failed for serialization or something along those lines since it was missed in review. Maintainers try our best on reviews, but its volunteer work and things will always get through.

Any thoughts on the complexities of implementing this? If not we can move forward with tracking the implementation of these.

gabrieldemarmiesse:
I can look into this. As you said, it requires some digging.

hyang0129:
I could try designing a common test for optimizers. 

If I understand correctly, all subclasses of optimizers should have common behaviors (eg. applies gradients, serializes, etc). So given these common behaviours, we could have a test that identifies all the classes in the optimizers subpackage and tests them the same way. 

For example, we get a list of optimizers, instantiate them with default parameters, create a dummy model and compile with MSE loss, fit the model for x steps, serialize, deserialize, etc. Would this be a test that the contributor has to add or would it be better if this test automatically finds all optimizers in the subpackage?

seanpmorgan:
> I could try designing a common test for optimizers.
> 
> If I understand correctly, all subclasses of optimizers should have common behaviors (eg. applies gradients, serializes, etc). So given these common behaviours, we could have a test that identifies all the classes in the optimizers subpackage and tests them the same way.
> 
> For example, we get a list of optimizers, instantiate them with default parameters, create a dummy model and compile with MSE loss, fit the model for x steps, serialize, deserialize, etc. Would this be a test that the contributor has to add or would it be better if this test automatically finds all optimizers in the subpackage?

That would be great!  I really like the idea of the test finding and testing all optimizers, but want to make sure the test case has a good UX when reading the test logs (e.g if we could utilize pytest parameterize that'd be great.) Feel free to open a PR and we can iterate on how we enforce the test.

hyang0129:
@seanpmorgan could you take a look at #2208

hyang0129:
@seanpmorgan please review #2208. It should work well as a standardized test for optimizers.

seanpmorgan:
Leaving this open because this could apply to other subpackages as well. If anyone is interested in this, let's build it out as a task list so we can easily keep track.

hyang0129:
[...]

PR #1062, opened 2020-02-10
MultiHeadAttention Layer

cgarciae:
Implementation of `MultiHeadAttention` as presented in [Attention Is All You Need](https://arxiv.org/abs/1706.03762) and discussed in #951. Uses `tf.einsum` to generalize dot-product to multiple heads.

Missing:

* [x] Documentation
* [x] Tests
* [x] Parameters for weight initializers, regularizers, constaints, etc.
* [x] config method 

References:
* [tensorflow/models](https://github.com/tensorflow/models/blob/master/official/nlp/modeling/layers/attention.py)


tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,128 @@
...
+                name="projection_bias",
+                shape=[output_size],
+                # initializer=self.bias_initializer,
+                # regularizer=self.bias_regularizer,
+                # constraint=self.bias_constraint,
+            )
+        else:
+            self.projection_bias = None
+
+        self.dropout = tf.keras.layers.Dropout(self.dropout_rate)
```
guillaumekln: The dropout layer should be used in the call method.

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,128 @@
...
+        # I = input features
+        # O = output features
+
+        query = inputs[0]
+        key = inputs[1]
+        value = inputs[2]
+        attention_mask = inputs[3] if len(inputs) == 4 else None
+
+        # Linear transformations
+        query = tf.einsum("...NI , HIO -> ...NHO", query, self.query_kernel)
```
guillaumekln: Question: do you use einsum for code simplicity or because it is more efficient than matmul in this case (or both)?

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,128 @@
...
+        # instead of their product saves some computation
+        depth = tf.cast(tf.shape(query)[-1], tf.float32)
+        query /= tf.sqrt(depth)
+
+        # Calculate dot product attention
+        logits = tf.einsum("...NHO,...MHO->...HNM", key, query)
+
+        # apply mask
+        if attention_mask is not None:
+            logits += -10e9 * (1.0 - attention_mask)
```
guillaumekln: Maybe `attention_mask` should be casted to float32.

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,128 @@
...
+                # constraint=self.bias_constraint,
+            )
+        else:
+            self.projection_bias = None
+
+        self.dropout = tf.keras.layers.Dropout(self.dropout_rate)
+
+        super().build(input_shape)
+
+    def call(self, inputs):
```
guillaumekln: When this layer is used in a decoder (e.g. GPT-2, Transformer), it is common to cache the key/value projections of the previous timesteps. Do you confirm it is not in the scope of this PR?

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # Linear transformations
+        query = tf.einsum("...NI , HIO -> ...NHO", query, self.query_kernel)
+        key = tf.einsum("...MI , HIO -> ...MHO", key, self.key_kernel)
+        value = tf.einsum("...MI , HIO -> ...MHO", value, self.value_kernel)
+
+        # Scale dot-product, doing the division to either query or key
+        # instead of their product saves some computation
+        depth = tf.cast(tf.shape(query)[-1], tf.float32)
+        query /= tf.sqrt(depth)
```
guillaumekln: I think the scaling value could be computed statically as the depth dimension is known.

tensorflow_addons/layers/multihead_attention.py
cgarciae: Correct.

tensorflow_addons/layers/multihead_attention.py
cgarciae: Nice catch :)

tensorflow_addons/layers/multihead_attention.py
cgarciae: Yeah, in `tensorflow/models` they have that as a separate layer, I think we should also treat it separately in a different PR.

tensorflow_addons/layers/multihead_attention.py
cgarciae: That is right, it is `self.head_size`, let me change it.

tensorflow_addons/layers/multihead_attention.py
cgarciae: Its mostly like this right now because I based on the `tensorflow/models` implementation which used `EinsumDense` which did it like this. More concretely, I am not sure `matmul` whould preserve the `H` dimension so you would have to do some juggling which `reshape`, which `einsum` it simpler. Don't know about performance though.

guillaumekln: Ok, thanks for the info!

tensorflow_addons/layers/multihead_attention.py
guillaumekln: Makes sense, thanks!

cgarciae:
@guillaumekln thanks for the comments! I've updated the code to address some of them.

tensorflow_addons/layers/multihead_attention.py
AakashKumarNain: The performance difference isn't that much. `einsum` is much better for this use case

cgarciae:
@guillaumekln @AakashKumarNain @facaiy @seanpmorgan 

Code should be ready for review :) Only 2 things are missing:

* [ ] Finish docstring
* [ ] Fix Github CI Issues

cgarciae:
Anyone know what is wrong with `flake8`? I am not getting any error from `flake8` locally.

cgarciae:
@ulf1 I see, thanks! `black` is automatically adding that comma :( I'll disable "format on save" to remove it.

gabrieldemarmiesse:
@googlebot I consent

gabrieldemarmiesse:

cgarciae:
The `pre-commit.sh` script modified a bunch of files unrelated to this PR, possibly the ones added during the merge by @gabrieldemarmiesse .

gabrieldemarmiesse:
I'll look into it and push the fix to your branch. Thanks for the heads up :)

gabrieldemarmiesse:
@cgarciae give me a sec, I'll fix it :)

cgarciae:
@gabrieldemarmiesse sorry, hadn't seen your comment, I'll stop trying to fix the CI issues so I don't mess up your work.

gabrieldemarmiesse:
No worries. Sorry for the big changes, we completed the black formatting in a rush and maybe we should have waited, seeing how much of a mess it has caused.
I'll fix that, but I may have to squash your commits.

gabrieldemarmiesse:
Hoo boy, that was scary! I'm never doing that again 😄

gabrieldemarmiesse:
you can do `git fetch origin` and `git reset --hard origin/mha` to get the changes locally.

cgarciae:
Thanks @gabrieldemarmiesse ! All CI tests are passing :D

I've been thinking about giving support to self attention using an API compatible with Sequential. I think we could create the `MultiHeadSelfAttention` class which inherits from `MultiHeadAttention` and just overload the `call` method like this:

```python
def call(self, x, **kwargs):
    return super().__call__([x, x, x], **kwargs)
```

gabrieldemarmiesse:
@cgarciae I'm happy to help. I'll let one of the other maintainers answer you and take care of this pull request. I'm just the cleaning dude, don't mind me.

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,253 @@
...
+class MultiHeadAttention(tf.keras.layers.Layer):
+    r"""
+    MultiHead Attention layer.
+
+    Defines the MultiHead Attention operation as defined in [Attention Is All You Need](https://arxiv.org/abs/1706.03762) which takes in a `query`, `key` and `value` tensors returns the dot-product attention between them:
+
+        ```python
+        mha = MultiHeadAttention(head_size=128, num_heads=128)
+
+        query = tf.random.uniform((32, 20, 200)) # (B, N, Dq)
```
qlzh727: I would expand all the "B, N, Dq" here to batch, timestep, and depth of query?

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,253 @@
...
+        key = tf.random.uniform((32, 15, 300)) # (B, M, Dk)
+        value = tf.random.uniform((32, 15, 400)) # (B, M, Dv)
+
+        attention = mha([query, key, value]) # (B, N, Dv)
```
qlzh727: I think keras layer supports keyword argument now, so you don't necessary to pass in all the values as a list.

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,253 @@
...
+
+import tensorflow as tf
+
+
+@tf.keras.utils.register_keras_serializable(package="Addons")
+class MultiHeadAttention(tf.keras.layers.Layer):
+    r"""
+    MultiHead Attention layer.
+
+    Defines the MultiHead Attention operation as defined in [Attention Is All You Need](https://arxiv.org/abs/1706.03762) which takes in a `query`, `key` and `value` tensors returns the dot-product attention between them:
```
qlzh727: Can u please break this long line?

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,253 @@
...
+        query = tf.random.uniform((32, 20, 200)) # (B, N, Dq)
+        key = tf.random.uniform((32, 15, 300)) # (B, M, Dk)
+
+        attention = mha([query, key]) # (B, N, Dk)
+        ```
+
+    Arguments
+        head_size: int, dimensionality of the `query`, `key` and `value` tensors after the linear transformation.
+        num_heads: int, number of attention heads.
+        output_size: int, dimensionality of the output space, if `None` then the input dimension of `value` or `key` will be used, default `None`.
```
qlzh727: Having None as default for int is bit dangerous, u will need to do explicit None check down below to distingish between None and 0.

qlzh727: And also wrap this long line at 80 chars.

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,253 @@
...
+        dropout_rate: float, `rate` parameter for the dropout layer that is applied to attention after softmax, default `0`.
```
qlzh727: Let's remove the "rate" and call it dropout, which is consistent with all other keras layers.

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,253 @@
...
+        use_projection_bias: bool, whether to use a bias term after the linear output projection.
```
qlzh727: Please specify the default value for all the following params.

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,253 @@
...
+        self.kernel_constraint = tf.keras.constraints.get(kernel_constraint)
+        self.bias_initializer = tf.keras.initializers.get(bias_initializer)
+        self.bias_regularizer = tf.keras.regularizers.get(bias_regularizer)
+        self.bias_constraint = tf.keras.constraints.get(bias_constraint)
+
+    def build(self, input_shape):
+
+        num_query_features = input_shape[0][-1]
+        num_key_features = input_shape[1][-1]
+        num_value_features = input_shape[2][-1]
```
qlzh727: This might not be available when value is not provided.

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,253 @@
...
+                name="projection_bias",
+                shape=[output_size],
+                initializer=self.bias_initializer,
+                regularizer=self.bias_regularizer,
+                constraint=self.bias_constraint,
+            )
+        else:
+            self.projection_bias = None
+
+        self.dropout = tf.keras.layers.Dropout(self.dropout_rate)
```
qlzh727: This shouldn't be in build(), but in init().

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,253 @@
...
+
+        super().build(input_shape)
+
+    def call(self, inputs, training=None):
```
qlzh727: There might should be param for masking as well, if you layer support input masking.

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,253 @@
...
+        # ------------------------
+        # N = query elements
+        # M = key/value elements
+        # H = heads
+        # I = input features
+        # O = output features
+
+        query = inputs[0]
+        key = inputs[1]
+        value = inputs[2]
```
qlzh727: Since the value could be optional, this will probably raise error.

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,253 @@
...
+        attention_mask = inputs[3] if len(inputs) == 4 else None
```
qlzh727: Masking should be passed in from kwarg "masking", rather than from the input list.

tensorflow_addons/layers/multihead_attention.py
cgarciae: Yes! Let me make it explicit.

tensorflow_addons/layers/multihead_attention.py
cgarciae: I wanted to do that but `input_shapes` in `build` is still a list of shape of the first argument in `call`, if there is another way of passing the shapes of the other arguments to `build` I'd like to implement it like that.

tensorflow_addons/layers/multihead_attention.py
cgarciae: Both `build` and `compute_output_shape` check on `self.output_size`:
```python
output_size = self.output_size if self.output_size else num_value_features
```

tensorflow_addons/layers/multihead_attention.py
cgarciae: Ah yes, its named [mask](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer#__call__)

tensorflow_addons/layers/multihead_attention.py
cgarciae: You're absolutely right!

cgarciae:
@qlzh727 thanks for the feedback! I changes to address you comments :)

tensorflow_addons/layers/multihead_attention.py
qlzh727: Ah, good point. Then we I think the current approach is probably the only way.

tensorflow_addons/layers/multihead_attention.py
qlzh727: yea, my point is if user specify output_size = 0, which is not None, it will be ignored by ur code and fallback to use num_value_features.

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,274 @@
...
+        # I = input features
+        # O = output features
+
+        query = inputs[0]
+        key = inputs[1]
+        value = inputs[2] if len(inputs) > 2 else key
+
+        # verify shapes
+        if mask is not None:
+            assert len(mask.shape) >= 2, "'mask' must have atleast 2 dimensions"
```
gabrieldemarmiesse: In code which will be used by user, never user plain assserts. Raise an error. In this case a `ValueError` would be appropriate. Assert are only for unit tests. AssertErrors are too general. It's similar to raising a plain `Exception` which is considered bad practice too.

cgarciae: @gabrieldemarmiesse I researched this a while ago and didn't find a consensus regarding the use of `assert`, but `ValueError` seems more appropriate.

cgarciae:
All test pass! :tada: :tada: :tada:

cgarciae:
@Squadrick I added a small commit, I don't know if that ticked off the `kokoro` stuff.

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,282 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
qlzh727: 2020.

tensorflow_addons/layers/multihead_attention.py
```diff
@@ -0,0 +1,282 @@
...
+
+    def compute_output_shape(self, input_shape):
+        num_value_features = (
+            input_shape[2][-1] if len(input_shape) > 2 else input_shape[1][-1]
+        )
+        output_size = (
+            self.output_size if self.output_size is not None else num_value_features
+        )
+
+        return input_shape[0][:-1] + (output_size,)
```
qlzh727: I think the output structure is changed when return_attn_coef is set. Can u update this as well, and add a test?

tensorflow_addons/layers/multihead_attention_test.py
```diff
@@ -0,0 +1,166 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
qlzh727: 2020

tensorflow_addons/layers/multihead_attention_test.py
```diff
@@ -0,0 +1,166 @@
...
+        self.assertEqual(output.shape[0], batch_size)
+        self.assertEqual(output.shape[1], q.shape[1])
+        self.assertEqual(output.shape[2], v.shape[2])
+
+        if tf.executing_eagerly():
+            attn_coef = attn_coef.numpy()
+            mask = mask.numpy()
+
+            self.assertTrue(((attn_coef != 0) == mask).all())
+
```
qlzh727: For layer behavior test, could u add:
1. small model with fit/eval/predict
2. from_config/to_config
3. save_model and load back.

tensorflow_addons/layers/multihead_attention.py
cgarciae: Done!

tensorflow_addons/layers/multihead_attention.py
cgarciae: Done!

tensorflow_addons/layers/multihead_attention_test.py
cgarciae: Done!

tensorflow_addons/layers/multihead_attention_test.py
cgarciae: Added tests for all of these.

cgarciae:
@qlzh727 the changes you requested where made.

tensorflow_addons/layers/multihead_attention_test.py
```diff
@@ -0,0 +1,338 @@
...
+        model.predict(np.random.uniform(size=(10, 42, 13)))
+
+        model.evaluate(
+            x=np.random.uniform(size=(20, 42, 13)),
+            y=np.random.randint(0, 10, size=(20,)),
+            batch_size=10,
+        )
+
+
```
seanpmorgan: Could you add a simple test that uses keras.layer_test:
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/utils/test_utils.py#L32

Similar to:
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/layers/sparsemax_test.py#L59

This utility verifies a number of things (such as ser/deser) and we'd like to use uniform testing for subpackages as much as possible. See #992

tensorflow_addons/layers/__init__.py
```diff
@@ -24,3 +24,4 @@
 from tensorflow_addons.layers.sparsemax import Sparsemax
 from tensorflow_addons.layers.tlu import TLU
 from tensorflow_addons.layers.wrappers import WeightNormalization
+from tensorflow_addons.layers.multihead_attention import MultiHeadAttention
```
seanpmorgan: nit: please alphabetize

tensorflow_addons/layers/multihead_attention_test.py
cgarciae: @seanpmorgan the `layer_test` docs say its a

> Test routine for a layer with a single input and single output.

Not sure if I can use it because `MutiHeadAttention` takes 3 inputs.

seanpmorgan: Ah, you're correct no need then and you've implemented the serialization tests outside of this anyway.

cgarciae:
@seanpmorgan Yeah, happy to maintain it. Added entry to CODEOWNERS.

tensorflow_addons/layers/__init__.py
cgarciae: Done

tensorflow_addons/layers/multihead_attention_test.py
```diff
@@ -0,0 +1,338 @@
...
+if __name__ == "__main__":
+    tf.test.main()
```
seanpmorgan: Late request: 
Could you change this to `sys.exit(pytest.main([__file__]))` as per #1263. We're aiming to utilize pytest for its extensions.

Will require some imports

cgarciae: Done

seanpmorgan:
@cgarciae Sorry for the conflicts. Could you resolve and then LGTM

cgarciae:
@seanpmorgan no problem. Conflicts solved!

seanpmorgan:
> LGTM thanks for this great contribution! Will leave the PR open for another day in case any other the other reviewers have any issues.

@qlzh727 please let us know if the changes you requested are sufficient. I believe they were addressed.

PR #1075, opened 2020-02-11
reduce docker image size

autoih:
Clean cache to reduce image size


gabrieldemarmiesse:
Thanks for the pull request

It's not clear that this slight decrease in readability is worth the few hundred MB of disk space that we're going to save because of the `--no-cache`. 

Unless there is a strong reason, like increase speed for the tests or saving 1~2 GB of disk space, I would prioritize readability over disk space.

If you're familiar with the dockerfiles, something you could help with is finding a way to centralize the version numbers of packages. Currently they're all over the place and when we'll need to bump them, it will be easy to forget one. Maybe putting them in multiple requirement.txt files?

autoih:
Understood. Thanks @gabrieldemarmiesse.

PR #1077, opened 2020-02-11
Changes default value of Yogi hyper-parameters

manzilz:
It is found initial_accumulator_value=1e-6 works better for a range of tasks. Thus switching the default value.


seanpmorgan:
Hi @manzilz since you're the author of the paper I'm good with just your word for it :smile: 
If there are an references to support this though please post for user reference.

PR #1081, opened 2020-02-12
Update to optionally run gpu tests without buildkit

### LINKED ISSUE #1078: Buildkit not support on GPU instance
https://source.cloud.google.com/results/invocations/e8d3f59e-f9b4-4345-9141-4684ec456fc8/targets/tensorflow_addons%2Fubuntu%2Fgpu%2Fpy3%2Fcontinuous/log

I believe we can just make a flag (with default being DOCKER_BUILDKIT=1). Then we can just update:
https://github.com/tensorflow/addons/blob/master/tools/run_google_cloud_tests.sh

seanpmorgan:
Closes #1078 and returns our CI

* Removes run_docker script since it's no longer used.


gabrieldemarmiesse:
I think a better API would be to have an environment variable called `DOCKER_BUILDKIT`. But fixing the master branch takes priority over fanciness :)

PR #1088, opened 2020-02-13
Add tf.Variable into TensorLike type

howl-anderson:
Add tf.Variable into TensorLike type (`tensorflow_addons/utils/types.py`)

PR #1091, opened 2020-02-14
Add tf.SparseTensor into TensorLike type

howl-anderson:
Add `tf.SparseTensor` into `TensorLike` type (`tensorflow_addons/utils/types.py`) as @autoih @gabrieldemarmiesse mentioned in #1088 .

### DIFF
--- a/tensorflow_addons/utils/types.py
+++ b/tensorflow_addons/utils/types.py
@@ -42,7 +42,13 @@ Constraint = Union[None, dict, str, Callable]
 Activation = Union[None, str, Callable]
 
 TensorLike = Union[
-    List[Union[Number, list]], tuple, Number, np.ndarray, tf.Tensor, tf.Variable
+    List[Union[Number, list]],
+    tuple,
+    Number,
+    np.ndarray,
+    tf.Tensor,
+    tf.SparseTensor,
+    tf.Variable,
 ]
 FloatTensorLike = Union[tf.Tensor, float, np.float16, np.float32, np.float64]
 AcceptableDTypes = Union[tf.DType, np.dtype, type, int, str, None]


howl-anderson:
@gabrieldemarmiesse @autoih Can you review this PR for me? Thank you!

fsx950223:
Some operations don't support ```SparseTensor```.
It's not a good idea to add it into ```TensorLike```.

howl-anderson:
IMHO, what @autoih @gabrieldemarmiesse want is a general type that can check if a variable is a tensor, can safely be converted to a tensor or can be used as a tensor. 

In TF there is a function `tf.is_tensor` (https://www.tensorflow.org/api_docs/python/tf/is_tensor) do a similar job:

its doc:

> If is_tensor(x) returns True, it is safe to assume that x is a tensor or can be converted to a tensor using ops.convert_to_tensor(x).

its code:

```python
  return (isinstance(x, tensor_like._TensorLike) or  # pylint: disable=protected-access
          ops.is_dense_tensor_like(x) or
          getattr(x, "is_tensor_like", False))
```

**NOTE**: The code's behavior does not exactly match the content of the document (not`ops.convert_to_tensor` but `ops.is_dense_tensor_like`)

`tf.is_tensor` using `isinstance(x, tensor_like._TensorLike)` to test if a variabe `x` is a tensor or not. Since `tf.SparseTensor` is a subclass of `tensor_like._TensorLike`, so to `tf.is_tensor` fucntion `tf.SparseTensor` is a tensor.

to `tf.is_tensor` fucntion,`tf.Variable` is a tensor too, because `ops.is_dense_tensor_like(x)` return True for x is a `tf.Variable`.

`list`, `tuple` and `np.array` etc can be converted to a tensor by `ops.convert_to_tensor`, so it can be considered as a more general tensor-like type. 

Above is my personal opinion, cc @autoih @gabrieldemarmiesse @fsx950223 for further discussion.

fsx950223:
> IMHO, what @autoih @gabrieldemarmiesse want is a general type that can check if a variable is a tensor, can safely be converted to a tensor or can be used as a tensor.
> 
> In TF there is a function `tf.is_tensor` (https://www.tensorflow.org/api_docs/python/tf/is_tensor) do a similar job:
> 
> its doc:
> 
> > If is_tensor(x) returns True, it is safe to assume that x is a tensor or can be converted to a tensor using ops.convert_to_tensor(x).
> 
> its code:
> 
> ```python
>   return (isinstance(x, tensor_like._TensorLike) or  # pylint: disable=protected-access
>           ops.is_dense_tensor_like(x) or
>           getattr(x, "is_tensor_like", False))
> ```
> 
> **NOTE**: The code's behavior does not exactly match the content of the document (not`ops.convert_to_tensor` but `ops.is_dense_tensor_like`)
> 
> `tf.is_tensor` using `isinstance(x, tensor_like._TensorLike)` to test if a variabe `x` is a tensor or not. Since `tf.SparseTensor` is a subclass of `tensor_like._TensorLike`, so to `tf.is_tensor` fucntion `tf.SparseTensor` is a tensor.
> 
> to `tf.is_tensor` fucntion,`tf.Variable` is a tensor too, because `ops.is_dense_tensor_like(x)` return True for x is a `tf.Variable`.
> 
> `list`, `tuple` and `np.array` etc can be converted to a tensor by `ops.convert_to_tensor`, so it can be considered as a more general tensor-like type.
> 
> Above is my personal opinion, cc @autoih @gabrieldemarmiesse @fsx950223 for further discussion.

Yes, add ```DenseTensorLike``` is a better solution.

gabrieldemarmiesse:
When flagging wrong types, it's worse and more fustrating for users to get an error like this: "Expected ... but you used a sparse tensor" when they know that the function could work with spase tensors than if they pass a wrong type and the type checker doesn't do anything.

Later on, maybe we can work into differentiating functions accepting only dense tensors and others, but for the moment it's a good move.

In short, it's preferable to authorize too many things than not enough things.

PR #1094, opened 2020-02-14
[WIP] trying to reproduce bug with test_utils run distributed

hyang0129:


Squadrick:
@hyang0129 Could you please explain the bug? I'm the author of `run_distributed`, maybe I can help.

hyang0129:
> @hyang0129 Could you please explain the bug? I'm the author of `run_distributed`, maybe I can help.

I believe the bug occurs when attempting to set logical devices a 2nd time in the test session. 

This can be reproduced by running any model.fit before running your tests. It may also extend to other keras operations, such as initializing the weights. 

Here is file contains a simple network training function. It appears that trying to call that training function after calling it once before tf.test.main() creates the problem.  https://github.com/tensorflow/addons/blob/5f4c46b59eb779d5720d6890b63ac39682fef4e7/tensorflow_addons/utils/test_utils_test.py

I also suspect that applying the method wrapper run_distributed(1) and run_distributed(2) to two different functions will cause the same error. I'm not sure when you would need to do this. 

Although I have not reproduced this, I have experienced issues where the model.fit function is called within a test that is not wrapped with run_distributed and one that is wrapped with run_distributed.

PR #1096, opened 2020-02-14
Cache clean in dockerfile to avoid invalidating layers

autoih:
Modified some parts of dockderfiles: put pagckages together to install and clean cache. To clean cache is if any of the files changed, that invalidates all later layers: we’ll need to re-run pip install. This can keep the correctness, and somehow have a smaller image size even it's not big. 
ref: https://www.docker.com/blog/intro-guide-to-dockerfile-best-practices/

### DIFF
--- a/tools/docker/cpu_tests.Dockerfile
+++ b/tools/docker/cpu_tests.Dockerfile
@@ -7,7 +7,7 @@ RUN git clone https://github.com/abhinavsingh/setup-bazel.git
 RUN bash ./setup-bazel/setup-bazel.sh 1.1.0
 
 COPY requirements.txt ./
-RUN pip install -r requirements.txt
+RUN pip install --no-cache-dir -r requirements.txt
 
 COPY ./ /addons
 WORKDIR addons
--- a/tools/docker/sanity_check.Dockerfile
+++ b/tools/docker/sanity_check.Dockerfile
@@ -9,7 +9,7 @@ RUN touch /ok.txt
 # -------------------------------
 FROM python:3.6 as black-test
 
-RUN pip install --no-cache-dir black==19.10b0
+RUN pip install black==19.10b0
 COPY ./ /addons
 RUN black --check /addons
 RUN touch /ok.txt
@@ -17,11 +17,12 @@ RUN touch /ok.txt
 # -------------------------------
 FROM python:3.6 as public-api-typed
 
-RUN pip install --no-cache-dir tensorflow-cpu==2.1.0 typeguard==2.7.1
-RUN pip install --no-cache-dir git+https://github.com/gabrieldemarmiesse/typed_api.git@0.1.1
+RUN pip install tensorflow-cpu==2.1.0 \
+			typeguard==2.7.1 \
+			git+https://github.com/gabrieldemarmiesse/typed_api.git@0.1.1
 
 COPY ./ /addons
-RUN TF_ADDONS_NO_BUILD=1 pip install --no-deps -e /addons
+RUN TF_ADDONS_NO_BUILD=1 pip install --no-cache-dir --no-deps -e /addons
 RUN python /addons/tools/ci_build/verify/check_typing_info.py
 RUN touch /ok.txt
 
@@ -36,7 +37,7 @@ RUN touch /ok.txt
 # -------------------------------
 FROM python:3.5 as valid_build_files
 
-RUN pip install --no-cache-dir tensorflow-cpu==2.1.0
+RUN pip install tensorflow-cpu==2.1.0
 
 RUN apt-get update && apt-get install sudo
 RUN git clone https://github.com/abhinavsingh/setup-bazel.git
@@ -80,7 +81,7 @@ RUN touch /ok.txt
 # docs tests
 FROM python:3.6 as docs_tests
 
-RUN pip install --no-cache-dir tensorflow-cpu==2.1.0 typeguard==2.7.1
+RUN pip install tensorflow-cpu==2.1.0 typeguard==2.7.1
 
 COPY tools/docs/doc_requirements.txt ./
 RUN pip install --no-cache-dir -r doc_requirements.txt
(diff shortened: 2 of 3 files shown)


gabrieldemarmiesse:
Thanks for the pull request :) 

Most of the advices written in the link are valid only when the version are not pinned and that one wants to ship the resulting image (decrease size). Here we don't really care about the resulting image size, since we're not going to ship it to anybody. The most important is build speed.

tools/docker/sanity_check.Dockerfile
```diff
@@ -17,12 +17,12 @@ RUN touch /ok.txt
 # -------------------------------
 FROM python:3.6 as public-api-typed
 
-RUN pip install tensorflow-cpu==2.1.0
-RUN pip install typeguard==2.7.1
-RUN pip install git+https://github.com/gabrieldemarmiesse/typed_api.git@0.1.1
+RUN pip install tensorflow-cpu==2.1.0 \
+			typeguard==2.7.1 \
+			git+https://github.com/gabrieldemarmiesse/typed_api.git@0.1.1
```
gabrieldemarmiesse: If I change the version of `typed_api`, it's going to download tensorflow again, which is not really a wanted behavior since it's so big.

tools/docker/sanity_check.Dockerfile
```diff
@@ -81,11 +81,10 @@ RUN touch /ok.txt
 # docs tests
 FROM python:3.6 as docs_tests
 
-RUN pip install tensorflow-cpu==2.1.0
-RUN pip install typeguard==2.7.1
+RUN pip install tensorflow-cpu==2.1.0 typeguard==2.7.1
```
gabrieldemarmiesse: Same here, tensorflow is ~100MB, if we change the typeguard version, we don't want to download tensorflow again.

tools/docker/sanity_check.Dockerfile
```diff
@@ -17,12 +17,12 @@ RUN touch /ok.txt
...
-RUN pip install tensorflow-cpu==2.1.0
-RUN pip install typeguard==2.7.1
-RUN pip install git+https://github.com/gabrieldemarmiesse/typed_api.git@0.1.1
+RUN pip install tensorflow-cpu==2.1.0 \
+			typeguard==2.7.1 \
+			git+https://github.com/gabrieldemarmiesse/typed_api.git@0.1.1
 
 COPY ./ /addons
-RUN TF_ADDONS_NO_BUILD=1 pip install --no-deps -e /addons
+RUN TF_ADDONS_NO_BUILD=1 pip install --no-cache-dir --no-deps -e /addons
```
gabrieldemarmiesse: When doing `pip install -e` it only creates a symlink, and since we use `--no-deps`, no other package is installed, so the cache won't be used in any case.

tools/docker/cpu_tests.Dockerfile
```diff
@@ -7,7 +7,7 @@ RUN git clone https://github.com/abhinavsingh/setup-bazel.git
 RUN bash ./setup-bazel/setup-bazel.sh 1.1.0
 
 COPY requirements.txt ./
-RUN pip install -r requirements.txt
+RUN pip install --no-cache-dir -r requirements.txt
```
gabrieldemarmiesse: This is not really necessary since we don't care about a few dozens of MB more  in image size. We're not going to push it/pull it.

gabrieldemarmiesse:
I'll close the pull request since there are little benefits to merge it. If we were shipping the image or not pinning the versions, it would be merged.

autoih:
Thanks @gabrieldemarmiesse for the explanation.

tools/docker/cpu_tests.Dockerfile
autoih: Thanks @gabrieldemarmiesse. This part is about the built image `layer`:  if the `requirements.txt` modified, the image will not be an updated one since there is cache here. Only if we are sure the file won't be changed. May I know more from your suggestions, @gabrieldemarmiesse?

tools/docker/sanity_check.Dockerfile
autoih: Gotcha! Thanks @gabrieldemarmiesse.

tools/docker/sanity_check.Dockerfile
autoih: Yes, in this case(if `typed_api` changed), do you think it's better to clean cache for the `RUN pip install  typedapi==0.2.0` with `--no-cache-dir`, otherwise, the image won't be updated.

tools/docker/cpu_tests.Dockerfile
gabrieldemarmiesse: The pip cache is not present if the docker cache is invalidated. You can try and see for yourself :)

tools/docker/sanity_check.Dockerfile
gabrieldemarmiesse: The image will be updated. Try and see for yourself :)

autoih: Oh, yes, if you assigned the version number, it'll be updated. But, for `pathlib` and `pyyaml`, in `doc_requirement.txt`,  do we also assign version number?

tools/docker/cpu_tests.Dockerfile
autoih: Oh, yes, if you assigned the version number, it'll be updated. But, for `pathlib` and `pyyaml`, in `doc_requirement.txt`,  do we also assign version number?

gabrieldemarmiesse: Pathlib is actually not needed in the requirements since it's in the standard library since py3. For pyyaml, it might be worth pinning but we don't get many benefits since it's a very stable library.

gabrieldemarmiesse: Pinning is more for libraries where we might expect things to change frequently.

autoih: Thanks, @gabrieldemarmiesse. I'll remove `pathlib`. Thanks a lot again for the comments!

PR #1110, opened 2020-02-19
Update macos build version

### LINKED ISSUE #1108: 0.8.2, 0.8.1 version is not compatible with Mac OS version < 10.15
I'm using Mac OS 10.14.6 version and I cannot install 0.8.2 or 0.8.1 version with pip
Seems 0.8.2 is only compatible with Mac OS version 10.15.x
Could you make 0.8.2 version of tensorflow-addons compatible with Mac OS 10.14.x?

**Relevant information**
- Was it part of tf.contrib? (if so, where): I'm not sure

anyone who are using Mac OS 10.13, 10.14 version

**Any other info.**

seanpmorgan:
Closes #1108

### DIFF
--- a/.github/workflows/macos_nightly.yml
+++ b/.github/workflows/macos_nightly.yml
@@ -48,6 +48,8 @@ jobs:
           bash tools/ci_testing/install_bazel_macos.sh $BAZEL_VERSION
           bazel build \
             -c opt \
+            --copt -mmacosx-version-min=10.13 \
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse:
Thanks for the quick fix!

nbowon:
Thanks for quick fix. 
But I still cannot install 0.8.2 in Mac OS 10.14.6
is this not deployed? 
and can I know when are you plan to deploy this?

Issue #1116, opened 2020-02-20
Stop using tensorflow-testing docker images and start using hub.docker.com/tensorflow images


gabrieldemarmiesse:
The master branch is broken because the docker image used to test on gpus is the one used to test the tensorflow source code itself. This can be considered private API. 

Instead, to build our manylinux2010 wheels and test our code, we should use the official way recommended by tensorflow (the same way external users do it).

Here is the official guide: https://github.com/tensorflow/custom-op
The docker images we should use are from the [official dockerhub repo](https://hub.docker.com/r/tensorflow/tensorflow):

```
tensorflow/tensorflow:2.1.0-custom-op-gpu-ubuntu16  # gpu
tensorflow/tensorflow:2.1.0-custom-op-ubuntu16          # cpu
```

Those images are "manylinux2010 compatible" sort of, since it's ubuntu.

By doing that, our testing environment will become stable again, and we shouldn't suffer sudden build breaks.

seanpmorgan:
Thanks for the issue! I agree using a `nosla` docker image is not the way we should be doing this.

So couple of things:
1) I'm not convinced of the reason GPU builds are breaking today. Almost looks like a pip networking issue. Not sure why this has just started today. The error:
```
ERROR: Could not find a version that satisfies the requirement tensorflow==2.1.0 (from -r build-requirements.txt (line 1)) (from versions: none)
ERROR: No matching distribution found for tensorflow==2.1.0 (from -r build-requirements.txt (line 1))
```

2) We used to use `tensorflow/tensorflow:custom-op-ubuntu16` but the images are rarely updated (something we can probably fix with more communication) and we had an issue when TF incremented their cuDNN version. We can build from a base image, but we can't pragmatically pull in a new cuDNN since this requires an NVIDIA account.

3) The new `tensorflow/tensorflow:2.1.0-custom-op-gpu-ubuntu16` looks promising and wasn't available before. This should align the correct cuDNN in the container, but I wonder what the release process of this is. Come 2.2 release candidates how long will it take for us to get a new docker image? I suppose we can use the old one provided CUDA and cuDNN match. It also has the needed devtoolset-7 & devtoolset-8 for building manylinux2010 (`/dt7/` in the contianer)

seanpmorgan:
@av8ramit Do you happen to know why our GPU tests are failing for the above error? I'm not sure why (as of today) it's unable to locate a TF 2.1.0 from pypi

gabrieldemarmiesse:
[...]

PR #1117, opened 2020-02-20
HOTFIX: pin docker container pip install to py36

seanpmorgan:
Not really a good way to handle this. But it returns CI and then we can go from there.


seanpmorgan:
So python3 is still symlinked to py36 in the container. pip3 is for py38 though. This is messy and we should clean it up with #1116

gabrieldemarmiesse:
> So python3 is still symlinked to py36 in the container. pip3 is for py38 though.

This is evil haha

PR #1119, opened 2020-02-21
Fix label of macos version in wheels

### LINKED ISSUE #1108: 0.8.2, 0.8.1 version is not compatible with Mac OS version < 10.15

seanpmorgan:
Closes #1108

### DIFF
--- a/.github/workflows/macos_nightly.yml
+++ b/.github/workflows/macos_nightly.yml
@@ -56,7 +56,7 @@ jobs:
             --test_output=errors \
             build_pip_pkg
 
-          bazel-bin/build_pip_pkg artifacts --nightly
+          bazel-bin/build_pip_pkg artifacts --plat-name macosx_10_13_x86_64 --nightly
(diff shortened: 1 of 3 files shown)


seanpmorgan:
@tensorflow/sig-addons-maintainers Fixed this. PTAL when time allows.

gabrieldemarmiesse:
Should this have been backported?

seanpmorgan:
> Should this have been backported?

Ah yeah it should have. I think let's just make sure it works on today's nightly and then we can do it manually.

PR #1121, opened 2020-02-21
fix state size to be a TensorShape instead of a Tensor

georgesterpu:
tf.keras.layers.RNN expects the state size to be an integer or TensorShape, while AttentionWrapper currently returns Tensors for two of its sub-states.

additional info: https://github.com/tensorflow/tensorflow/pull/36796


guillaumekln:
Could you also add the test case that is failing without this change?

georgesterpu:
> Could you also add the test case that is failing without this change?

sure @guillaumekln

georgesterpu:
My test snippet is:
```python
import tensorflow as tf

cell = tf.keras.layers.LSTMCell(3)
mechanism = tfa.seq2seq.LuongAttention(units=3)
cell = tfa.seq2seq.AttentionWrapper(cell, mechanism)
layer = tf.keras.layers.RNN(cell)

@tf.function
def test():
    seq_len = tf.random.uniform(shape=(), minval=2, maxval=10, dtype=tf.int32)
    data = tf.ones(shape=(seq_len, seq_len, 3))
    mechanism.setup_memory(data)
    return layer(data)

test()
```
Should I modify `test_attention_state_with_keras_rnn` recently introduced by @qlzh727 , or create a new test case? The problem only occurs in graph mode, when the sequence length is variable.

guillaumekln:
If you can minimally update `test_attention_state_with_keras_rnn` that should work (e.g. set a sequence length that is unknown in graph mode).

georgesterpu:
@guillaumekln @qlzh727 
Could you please take a look ?
I can see two test cases in `beam_search_decoder_test` failing, but I am not sure if they are caused by this PR. Besides, the BeamSearchDecoder with an AttentionWrapper cell works locally with this change in tf 2.1 and tfa 0.8.2

guillaumekln:
The tests are passing on master but not on your branch, so it is clearly related.

From a quick look, I don't think `self._alignments_size.shape` is correct because here `self._alignments_size` is a scalar tensor.

gabrieldemarmiesse:
@googlebot I consent

georgesterpu:
Of course I got it wrong, thanks !
So when `self._alignments_size` is a `Tensor`, do we need to build a `TensorShape` with one dimension equal to the value of the `Tensor` ?

gabrieldemarmiesse:

guillaumekln:
> So when self._alignments_size is a Tensor, do we need to build a TensorShape with one dimension equal to the value of the Tensor ?

If the tensor value is not statically known this will not work because TensorShape expects Python values.

At the moment I don't see a good workaround. To me it seems that Keras should be updated to support a tensor `state_size`.

georgesterpu:
@guillaumekln 
Would something like this work ?
```python
        if isinstance(self._alignments_size, int):
            return self._alignments_size
        else:
            return self._alignments_size.shape
        return tf.TensorShape([None])
```

georgesterpu:
> At the moment I don't see a good workaround. To me it seems that Keras should be updated to support a tensor `state_size`.

@qlzh727 advised in a different PR to change the AttentionWrapper instead:
https://github.com/tensorflow/tensorflow/pull/36796#pullrequestreview-361508451

guillaumekln:
Ah right, we don't need the actual dimension values here so a TensorShape with None should work.

georgesterpu:
@guillaumekln all checks passed

guillaumekln:
Thanks!

georgesterpu:
thanks @guillaumekln and @gabrieldemarmiesse for your help

Issue #1122, opened 2020-02-21
Test the release process and wheels at each commit.


gabrieldemarmiesse:
It's difficult to modify the release code because it's not run in the CI at each commit. We also sometimes have bugs that appear only in the wheel. My plan for fast feedback without making the tests too long:

at PR and commits on master:
* test for py35 on linux, macos, windows, test py37 for linux only (forward compatibility)
* produce the wheels after the tests (the cache is there, so ~10s)
* we group the wheels together (`upload-wheels`)
* We add a IF statement at the action `pypa/gh-action-pypi-publish@master` to disable it.
* We download the wheels in a fresh environements on windows/linux/max and install them, along with tensorflow, install them and run a simple `import tensorflow_addons` check (to check dependencies and python files are all there). 
* Also run a very short sanity check where a simple tensor is created and passed through a single activation function (test that a single `.so` is loaded correctly).

---> ~1-2 minutes longer than the current tests

At each release:
We do exactly the same but
* We also test for py36, py37 on linux, macos, windows
* We also run the action `pypa/gh-action-pypi-publish@master`.

---> a bit shorter than the current version because we create the wheels after the tests, when the cache is hot.

In practice, it means we have only 3 yaml files in github actions: the backport, the sanity checks and the "tests+releases+nighly" file.

This should remove a lot of code duplication too.

seanpmorgan:
Good to close?

gabrieldemarmiesse:
It's missing a small step:

> We download the wheels in a fresh environements on windows/linux/max and install them, along with tensorflow, install them and run a simple import tensorflow_addons check (to check dependencies and python files are all there).
Also run a very short sanity check where a simple tensor is created and passed through a single activation function (test that a single .so is loaded correctly).

seanpmorgan:
> It's missing a small step:
> 
> > We download the wheels in a fresh environements on windows/linux/max and install them, along with tensorflow, install them and run a simple import tensorflow_addons check (to check dependencies and python files are all there).
> > Also run a very short sanity check where a simple tensor is created and passed through a single activation function (test that a single .so is loaded correctly).

:smile:  arguably the most important part

PR #1125, opened 2020-02-21
Do release at each commit (but not the upload)

gabrieldemarmiesse:
Fix part of #1122 . I believe the first step is to enable running EVERYTHING at each commit, and then, with other pull requests, we can reduce code duplication in the yaml and make the tests faster.


gabrieldemarmiesse:
I'll close that, the running time is not reasonable.

PR #1127, opened 2020-02-21
We now test the release code at each commit.

gabrieldemarmiesse:
#1125 but better.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -130,7 +130,7 @@ jobs:
           TF_NEED_CUDA: 0
           BAZEL_VC: "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/"
         shell: bash
-        if: github.event_name == 'release' || ${{ matrix.python-version }} == '3.5'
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse:
For windows and macos, I grouped the tests and building of the wheels in the same job to have a hot cache.

.github/workflows/release.yml
```diff
@@ -156,6 +127,7 @@ jobs:
             build_pip_pkg
           bazel-bin/build_pip_pkg artifacts
       - uses: actions/upload-artifact@v1
+        if: github.event_name == 'release' || matrix.python-version == '3.5'
```
seanpmorgan: Wondering why we would need to upload-artifact for CI testing?

gabrieldemarmiesse: To test the small job where we gather actifacts. If you see a simpler way, I'm open to ideas.

seanpmorgan: Gotcha. Better to test that than not test I suppose, but I do think the most likely cause of a failure there would be timeout or something.

gabrieldemarmiesse:
It's going to make it easier to switch the docker images :)

Issue #1123, opened 2020-02-21
Group the pinned version numbers together


gabrieldemarmiesse:
For our tools, we pin the version numbers (tensorflow, tensorflow-cpu, black, soon isort, bazel, typeguard...).

Those version numbers are often duplicated across the project, in various dockerfiles, yamls, requirements files... we should have them only in one single place place, at least the goal is to make it easy to bump version numbers.

gabrieldemarmiesse:
Can be closed now.

PR #1131, opened 2020-02-22
Grouped tests dependencies together.

gabrieldemarmiesse:
Fix part of #1123

Note that it's quite verbose now, but Dockerfiles will allow the syntax:

```
RUN --mount=requirements.txt pip install -r requirements.txt
```

In the future. It's going to make it cleaner.


gabrieldemarmiesse:
The conflicts are fixed

requirements.txt
```diff
@@ -1 +1 @@
-typeguard
+typeguard==2.7.1
```
seanpmorgan: Seeing as this requirements file gets added to the pip dependencies, should we relax this to be `typeguard>=`? I worry that this will create conflicts with other packages that utilize typeguard.

gabrieldemarmiesse: I totally agree. Can I do that in another PR? I'm going to relax the requirements on the minor version for each dependency too. Ex pip install tensorflow~=2.1

PR #1136, opened 2020-02-22
Remove nightly files.

gabrieldemarmiesse:
Part of #1122

Please be careful when reviewing. We don't want to upload nightlys instead of stable, this is quite sensitive.


.github/workflows/release.yml
```diff
@@ -61,12 +65,14 @@ jobs:
       matrix:
         python-version: ['3.5', '3.6', '3.7']
     steps:
+      - if: github.event_name == 'schedule'
+        run: echo "::set-env name=NIGHTLY_FLAG::--nightly"
```
seanpmorgan: As just a sanity check that [this method](https://help.github.com/en/actions/reference/development-tools-for-github-actions#set-an-environment-variable-set-env) is working correctly could you do a quick commit to echo `$NIGHTLY_FLAG` for all three OS's? I would just modify the nightly flag to be added for py35 and then we can revert that commit.

gabrieldemarmiesse: Sure thing

seanpmorgan:
LGTM. Can rollback the debug commit

seanpmorgan:
So it looks like we never were checking that tests pass before uploading wheels in the release script. Could/should we add `manylinux-cpu-release-test` to the `needs` field of the upload job? That'll prevent uploads in case there's a break in testcases for some reason.

gabrieldemarmiesse:
Sure thing

Issue #1114, opened 2020-02-20
Adding a pure python implementation corresponding to simple C++/Cuda ops.


gabrieldemarmiesse:
**Describe the feature and the current behavior/state.**

For custom ops, we have two implementations: C++ and Cuda.
I propose that __for simple ops__ , meaning less than 10 python lines, we implement a pure Python version too. I see several use cases:

* We can more easily test the correctness of the C++ and Cuda implementation. In the test suite, we generate random tensors and check that the results is the same for the three implementations. Since python code is easy to read, we have a readable reference implementation that's easier to check.
* It helps with users who want to understand the source code.
* It should also help all users who have issues with custom ops. It's better for those users to have a slow implementation, than an implementation which doesn't run at all. It also helps for people who want to run those ops in specialized hardware (ex: ROCm). We've had a number of issues related to users having to deal with that: #1109 #1099 #990 #987 .... We should provide a flag for them which they can activate to use the pure python implementation. For users of tensorflow addons on windows and mac, they'll be able to use their gpu.

**Relevant information**
- Are you willing to contribute it (yes/no): yes
- Are you willing to maintain it going forward? (yes/no): yes
- Is there a relevant academic paper? (if so, where): no
- Is there already an implementation in another framework? (if so, where): no
- Was it part of tf.contrib? (if so, where): no

**Which API type would this fall under (layer, metric, optimizer, etc.)**

Activation, image, layers, seq2seq, text

**Who will benefit with this feature?**

People having problems with custop ops, the maintainers because the test suite will be more robust.

**Any other info.**

We should make a pull request for a simple op see how to implement it. Especially how do we access the different implementations from the user side and tests side. Good candidates would be activations functions.

failure-to-thrive:
Frankly, it's not clear for me why TFA activations/... need C++ code. All those functions can be expressed with regular Tensorflow ops, gaining the same benefits such as a GPU placement. What did I miss?

gabrieldemarmiesse:
https://github.com/tensorflow/addons/pull/913 can give some background and discussion that we had concerning this.

seanpmorgan:
I really like the idea of this and it's ability to test python vs custom-op implementations is much needed. I'm a little squeemish regarding the flag to do this though. Do you imagine it would be a flag in for example the activation API or some global setting?

gabrieldemarmiesse:
I've no idea what kind of flage we should use for the public API. Ideas welcome :)

georgesterpu:
[...]

Issue #784, opened 2019-12-19
Build GPU custom Ops for windows


seanpmorgan:
Currently we only compile/use the CPU version of custom ops in Addons for windows. Users should still be able to leverage GPU ops from TF-Core but this is something we want to build/enable.

I don't think this is a huge lift since the gpu toolchain works on nix and windows, but needs some testing.

allen108108:
So, we can't use tensorflow-addons in Win10 system  now ,right ?

seanpmorgan:
> So, we can't use tensorflow-addons in Win10 system now ,right ?

You can compile from source. See the README. It will only support CPU addons ops (but thats the majority of them). 

Alternatively we'll have a 0.7 release in the very near future to pip install.

allen108108:
Will you release the GPU version in future ?

seanpmorgan:
Yes that is what this issue is here to track. 

Just to be clear, prior to this issue being completed the addons package will still utilize GPU for underlying TF ops (matmul, etc.). But there will not be GPU implementations for the custom CUDA kernels in this repository... though they'll have CPU fallback.

allen108108:
Thanks for answering ~~~~

seanpmorgan:
This has been unblocked:
https://github.com/tensorflow/custom-op/pull/43#event-2902430632

seanpmorgan:
Given the recent news: https://developer.nvidia.com/cuda/wsl
Is this feature still requested from the community?

seanpmorgan:
TensorFlow Addons is transitioning to a minimal maintenance and release mode. New features will not be added to this repository. For more information, please see our public messaging on this decision:
[TensorFlow Addons Wind Down](https://github.com/tensorflow/addons/issues/2807)

Please consider sending feature requests / contributions to other repositories in the TF community with a similar charters to TFA:
[Keras](https://github.com/keras-team/keras)
[Keras-CV](https://github.com/keras-team/keras-cv)
[Keras-NLP](https://github.com/keras-team/keras-nlp)

PR #1137, opened 2020-02-23
Added an option to use the pure python implementation.

gabrieldemarmiesse:
See #1114 

What is left to do:
*  Add try-except. If there are problems with loading the .so
fallback to the pure python implementation and print a warning
explaining the TF_ADDONS_PY_OPS and addons versions compatible with 
the current tf.
May also flip the TF_ADDONS_PY_OPS switch automatically?
* Recommend that contributors making new ops add pure python equivalent in the CONTRIBUTING

I think macos and windows users will be happy to use the activations functions on gpu :)

I also believe that the description of TF addons discouraged Windows and MacOS users to use addons when they want to use the gpu too. That's not good, because the majority of the components of addons work on windows and mac with a gpu.


gabrieldemarmiesse:
Bazel is really not easy for python devs.
When locally, I'm doing
```
TF_ADDONS_NO_BUILD=1 pip install -e ./
```

I can do:
```
from tensorflow_addons.activations import hardshrink
```
and it works.

Is there a flag to make bazel respect basic python rules? This is really fustrating to spend so much time just to add a single python file.

Squadrick:
@gabrieldemarmiesse yeah, that happens with bazel due to the way it resolves path imports. Haven't found a fix for it yet.

seanpmorgan:
@googlebot I consent.

tensorflow_addons/activations/BUILD
```diff
@@ -16,6 +16,7 @@ py_library(
         "tanhshrink.py",
     ],
     data = [
+        "//tensorflow_addons:options.py",
```
seanpmorgan: Agree that is this painful and unituitive. Sorry for any wasted time. We could raise this with the bazel team, but I believe this is fundamental to how this resolving works.

seanpmorgan:
> * Add try-except. If there are problems with loading the .so
>   fallback to the pure python implementation and print a warning
>   explaining the TF_ADDONS_PY_OPS and addons versions compatible with
>   the current tf.

This should be doable since it throws a python error instead of a crash. `tensorflow.python.framework.errors_impl.NotFoundError`

>   May also flip the TF_ADDONS_PY_OPS switch automatically?
> * Recommend that contributors making new ops add pure python equivalent in the CONTRIBUTING

Good idea, but not sure this is always feasible so maybe dependant on the subpackage. For example, image kernels may be hard to reproduce without depending on OpenCV or something.

> I think macos and windows users will be happy to use the activations functions on gpu :)

Agree, but I think macos users are out of luck unless they're using a ROCm supported TF. AFAIK there hasn't been an NVIDIA GPU on macos system for a long time. I could be wrong on the current state of this:
https://github.com/ROCmSoftwarePlatform/tensorflow-upstream

tensorflow_addons/activations/BUILD
gabrieldemarmiesse: Yeah it's linked to how bazel works and that makes sense for monorepos like tensorflow. I'm not convinced the added value of making the wheels + running the tests with bazel outweight the costs in addons. It makes totally sense to use it to build the SO though. I know we had a similar conversation a while ago, maybe I'll open an issue so that we can discuss it more. Anyway thanks for the fix! It helps a lot!

gabrieldemarmiesse:
I'll implement the fallback in this PR. We can change the CONTRIBUTING in another one

gabrieldemarmiesse:
Here is the warning message:

```
2020-02-25 16:58:17.446948: W tensorflow/stream_executor/platform/default/dso_loader.cc:55] Could not load dynamic library 'libcuda.so.1'; dlerror: libcuda.so.1: cannot open shared object file: No such file or directory
2020-02-25 16:58:17.447611: E tensorflow/stream_executor/cuda/cuda_driver.cc:313] failed call to cuInit: UNKNOWN ERROR (303)
2020-02-25 16:58:17.448055: I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:156] kernel driver does not appear to be running on this host (CRE1-L11685): /proc/driver/nvidia/version does not exist
2020-02-25 16:58:17.458128: I tensorflow/core/platform/cpu_feature_guard.cc:143] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX2 FMA
2020-02-25 16:58:17.507655: I tensorflow/core/platform/profile_utils/cpu_utils.cc:102] CPU Frequency: 2712000000 Hz
2020-02-25 16:58:17.510852: I tensorflow/compiler/xla/service/service.cc:168] XLA service 0x7fffe1fe08d0 initialized for platform Host (this does not guarantee that XLA will be used). Devices:
2020-02-25 16:58:17.511249: I tensorflow/compiler/xla/service/service.cc:176]   StreamExecutor device (0): Host, Default Version
/mnt/c/Users/gdemarmi/Desktop/projects/addons/tensorflow_addons/options.py:48: RuntimeWarning: Traceback (most recent call last):
  File "/mnt/c/Users/gdemarmi/Desktop/projects/addons/tensorflow_addons/activations/hardshrink.py", line 49, in hardshrink
    return _hardshrink_custom_op(x, lower, upper)
  File "/mnt/c/Users/gdemarmi/Desktop/projects/addons/tensorflow_addons/activations/hardshrink.py", line 58, in _hardshrink_custom_op
    return _activation_so.ops.addons_hardshrink(x, lower, upper)
  File "/mnt/c/Users/gdemarmi/Desktop/projects/addons/tensorflow_addons/utils/resource_loader.py", line 49, in ops
    self._ops = tf.load_op_library(get_path_to_datafile(self.relative_path))
  File "/home/gdemarmi/softwares/python/anaconda/lib/python3.7/site-packages/tensorflow/python/framework/load_library.py", line 58, in load_op_library
    lib_handle = py_tf.TF_LoadLibrary(library_filename)
tensorflow.python.framework.errors_impl.NotFoundError: /mnt/c/Users/gdemarmi/Desktop/projects/addons/tensorflow_addons/custom_ops/activations/_activation_ops.so: cannot open shared object file: No such file or directory


The hardshrink C++/CUDA custom op could not be loaded. 
For this reason, Addons will fallback to an implementation written 
in Python with public TensorFlow ops. There worst you might experience with
this is a moderate slowdown on GPU. There can be multiple 
reason for this loading error, one of them may be an ABI incompatibility between 
the TensorFlow installed on your system and the TensorFlow used to compile
TensorFlow Addons' custom ops. The stacktrace generated when loading the 
shared object file was displayed above.

If you want this warning to disappear, either make sure the TensorFlow installed
is compatible with this version of Addons, or tell TensorFlow Addons to 
prefer using Python implementations and not custom C++/CUDA ones. You can do that 
by changing the TF_ADDONS_PY_OPS flag
either with the environment variable:

TF_ADDONS_PY_OPS=1 python my_script.py

or in your code, after your imports:

import ...
import ...

tfa.options.TF_ADDONS_PY_OPS = True


  warnings.warn(warning_msg, RuntimeWarning)
```

tensorflow_addons/activations/BUILD
seanpmorgan: Yeah I think this can be discussed further. Typical contributions to Addons will/should fall into a subpackage that should have the BUILD already setup. This PR is a change that affects the entirety of the project so it's a bit more complicated. I agree we should encourage contributors to make these types of changes but we can look at and cons again to see if its worth splitting the C++ test/builds from python tests/builds.

tensorflow_addons/options.py
```diff
@@ -0,0 +1,49 @@
...
+    else:
+        TF_ADDONS_PY_OPS = True
+
+
+FALLBACK_WARNING_TEMPLATE = """{}
+
+The {} C++/CUDA custom op could not be loaded.
+For this reason, Addons will fallback to an implementation written
+in Python with public TensorFlow ops. There worst you might experience with
+this is a moderate slowdown on GPU. There can be multiple
```
seanpmorgan: C++ custom ops are still faster than their python op equivalents from what we've seen in profiling GELU, so this isn't limited to GPU

tensorflow_addons/options.py
```diff
@@ -0,0 +1,49 @@
...
+import warnings
+import traceback
+
+try:
+    TF_ADDONS_PY_OPS = bool(int(os.environ["TF_ADDONS_PY_OPS"]))
+except KeyError:
+    if platform.system() == "Linux":
+        TF_ADDONS_PY_OPS = False
+    else:
+        TF_ADDONS_PY_OPS = True
```
seanpmorgan: So MacOS and Windows should support CPU custom ops without issue. Similar to the question above about priority is this the way we want to handle this?

gabrieldemarmiesse: https://github.com/tensorflow/addons/pull/913

I suspect that on CPU, pure python might be faster. I need to run some speed tests though.

Windows user expect ops to work on gpu because it's the most common use case. Since I don't know if there is any way to know if the op is going to be executed on gpu or cpu, I can't decide at the last moment if I can use pure Python or not.

If you have any suggestion on how to make it work on GPU on windows and use the custom C++ op when using the CPU, I'm open to ideas.

tensorflow_addons/options.py
gabrieldemarmiesse: Do we have benchmarks? Nice! 😃  I'm really interested, benchmarking was an issue last time we needed it in #913 . Can you link the issue/PR?

seanpmorgan: https://github.com/tensorflow/tensorflow/pull/33945#issue-336001657

gabrieldemarmiesse: It seems those benchmarks were only run on gpu. Furthermore, it's missing the `.numpy()` to force the execution of the ops because gpu ops are executed lazily. Benchmarking is kind of hard as we saw in #913 . I really hope we have a proper benchmarking tool in the future.

tensorflow_addons/options.py
seanpmorgan: Thoughts:

1) GPU Ops on Windows is an achievable task, just requires a bit of a time commitment: https://github.com/tensorflow/addons/issues/784
2) When you say pure python might be faster you mean with XLA right? I have a hunch the majority of users will not be applying that themselves, but at least they have the option to.
3) You make a good point regarding windows users not being able to select custom op instead of GPU. It is possible to do something like this with device placement but unweildy.

I'm good with macOS / Windows using python ops when they're available as default since it makes this quite a bit more concise. When windows GPU ops are added would we consider editing this option switch?

tensorflow_addons/options.py
gabrieldemarmiesse: Actually, for activation functions, for big tensors (to remove the overhead of function calls) it seems pure python (even eager mode!) is faster. XLA is much much faster.

See https://colab.research.google.com/drive/1LTx3vMpA1fLCESKl-_WrLIp0Fq1zYZL0

It might be worth opening an issue to investigate if we should drop some C++ implementations because they give no benefit at all (apart from being slightly faster for small tensors).

seanpmorgan: Nice find! I think it's a great idea but we should look into standardizing our methods of benchmarking. I will bring it up at the next monthly meeting to see if TF core team has standard process  (though this is not blocking us from moving forward)

As @WindQAQ mentioned when testing GPU we'll want to make sure we're not benchmarking the time to move from GPU back to CPU which `.numpy()` will do:
https://colab.research.google.com/drive/12ende9xXMSywP2lOKWrFJBwaDDHbkYXh

But the tests you posted above are on CPU.

gabrieldemarmiesse: Yeah, we've no way to actually benchmark for real on gpu. That's a shame. Either we measure the time moving back and forth between the gpu and cpu with `.numpy()`, either the ops are executed lazily and you have enormous tensors processed in nanoseconds which is not realistic at all XD

gabrieldemarmiesse: Current numbers tend to agree with what is currently in the readme, I'll minimize this section. We can always change what is said if #1156 gives us new elements.

README.md
```diff
@@ -105,9 +105,40 @@ TF-Addons. In order to achieve these we require that our additions
 conform to established API patterns seen in core TensorFlow.
 
 #### GPU/CPU Custom-Ops
-A major benefit of TensorFlow Addons is that there are precompiled ops. Should 
-a CUDA 10.1 installation not be found then the op will automatically fall back to 
-a CPU implementation.
+A major benefit of TensorFlow Addons is that there are precompiled ops.
+Custom ops work on GPU only on Ubuntu. This is why on Windows and MacOS, 
+If available, Addons will fallback to a pure TensorFlow Python implementation for custom-ops.
```
seanpmorgan:
```suggestion
A major benefit of TensorFlow Addons is that there are precompiled ops for CPU/GPU.
Currently however, GPU custom ops only work for Linux distributions. For this reason Windows and MacOS will fallback to pure TensorFlow Python implementations whenever possible.
```

README.md
```diff
@@ -105,9 +105,40 @@ TF-Addons. In order to achieve these we require that our additions
...
+
+The order of priority in MacOS/Windows:
+1) Pure TensorFlow + Python implementation (work on cpu+gpu)
+2) C++ implementation for CPU
+
+The order of priority for Ubuntu:
```
seanpmorgan:
```suggestion
The order of priority for Linux:
```

README.md
```diff
@@ -105,9 +105,40 @@ TF-Addons. In order to achieve these we require that our additions
...
+```
+
+or in your code:
+
+```
+import tensorflow_addons as tfa
+tfa.options.TF_ADDONS_PY_OPS=True
+```
+
+This variable default to `True` on Windows and Mac, and `False` for Linux.
```
seanpmorgan:
```suggestion
This variable will default to `True` on Windows and Mac, and `False` for Linux.
```

gabrieldemarmiesse:
I'm unsure about serialization. I think we need another opinion on this one.

gabrieldemarmiesse:
Sorry about the spam. My bot is training.

Squadrick:
> Almost LGTM. Minor edits.
> 
> One last question I have is the implications of serializing and deserializing these ops. For example, if someone trained a model with custom-ops and then someone else loaded using py_ops what guarantees do we have that they behave the same? I suppose as long as we have test cases ensuring they're the same?

That's the only way I can see it work. Also, in `verify_funcs_are_equivalent`, `atol=1e-4` is a significant error bound that can't be blamed on fp-approx, the [default value for `atol` is 1e-6](https://www.tensorflow.org/api_docs/python/tf/test/TestCase?version=nightly#assertAllCloseAccordingToType), and I think we should aim for lower.

PR #1146, opened 2020-02-24
add tanhshrink_py

fsx950223:
Related #1114

### DIFF
--- a/tensorflow_addons/activations/tanhshrink_test.py
+++ b/tensorflow_addons/activations/tanhshrink_test.py
@@ -24,26 +24,6 @@ from tensorflow_addons.utils import test_utils
 
 @test_utils.run_all_in_graph_and_eager_modes
 class TanhshrinkTest(tf.test.TestCase, parameterized.TestCase):
-    # @parameterized.named_parameters(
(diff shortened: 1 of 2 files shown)


tensorflow_addons/activations/tanhshrink_test.py
```diff
@@ -18,30 +18,51 @@
...
+
+    # @parameterized.named_parameters(("float32", np.float32), ("float64", np.float64))
+    # def test_theoretical_gradients(self, dtype):
+    #     # Only test theoretical gradients for float32 and float64
+    #     # because of the instability of float16 while computing jacobian
+    #     x = tf.constant([-2.0, -1.0, 0.0, 1.0, 2.0], dtype=dtype)
+
+    #     theoretical, numerical = tf.test.compute_gradient(tanhshrink, [x])
+    #     self.assertAllCloseAccordingToType(theoretical, numerical, atol=1e-4)
+
```
gabrieldemarmiesse: Maybe we could actually remove all that? The check with random tensors is much more robust. What do you think?

fsx950223: Good idea. I remove them later.

PR #1154, opened 2020-02-25
Added python implementation for GeLU

abhichou4:
Refer #1114


abhichou4:
Pardon my noob mistakes here. I'll try to be more careful.

tensorflow_addons/activations/gelu.py
```diff
@@ -49,3 +50,13 @@ def _gelu_grad(op, grad):
     return _activation_so.ops.addons_gelu_grad(
         grad, op.inputs[0], op.get_attr("approximate")
     )
+
+
+def _gelu_py(x: types.TensorLike, approximate: bool = True) -> tf.Tensor:
+    x = tf.convert_to_tensor(x)
+    if approximate:
+        pi = tf.cast(math.pi, x.dtype)
+        coeff = tf.cast(0.044715, x.dtype)
```
gabrieldemarmiesse: Are those cast necessary? Would it work without them? (I've seriously no idea since there are multiple ways of running a TF function)

If so, could you add a comment just above explaining why? Because it might not be obvious for many readers.

gabrieldemarmiesse:
> Pardon my noob mistakes here. I'll try to be more careful.

No worries, it won't appear in the master branch because we squash the commits before merging. 

If you want to know how to open multiple pull requests in a clean way, you can refer to the document we mention in the CONTRIBUTING: https://github.com/gabrieldemarmiesse/getting_started_open_source

tensorflow_addons/activations/gelu.py
abhichou4: The cast to coeff is not necessary. But if pi is not caste an `InvaidArgument` error when input is of dtype `tf.float64`. Input dtype can be changed to `tf.float32` as an alternative too. 
https://colab.research.google.com/drive/1FazobvLpBoHVMY59cK1im8VlOym6iee5

gabrieldemarmiesse: Thanks for the explanation. I'll merge this pull request then.

PR #1159, opened 2020-02-27
Fix subclasses of CyclicalLearningRate

lcsfrey:


lcsfrey:
@googlebot I signed it!

PR #1160, opened 2020-02-27
Removed erroneous scale_fn from CLR subclasses

lcsfrey:
The current CyclicalLearningRate subclasses (TriangularCyclicalLearningRate, Triangular2CyclicalLearningRate, and ExponentialCyclicalLearningRate) are responsible for generating their own scale_fn. So they should not have scale_fn in their function arguments.


lcsfrey:
Woops, my bad. It's fixed!

PR #1167, opened 2020-02-27
Notify proxy maintainers

### LINKED ISSUE #1145: Find a way to label PRs with the relevant code owners.
Currently we assign maintainers for incoming PRs with [CODEOWNERS](https://github.com/tensorflow/addons/blob/master/.github/CODEOWNERS). However, we're limited to tagging users with write access.

Seeing as a majority of our maintainers are submodule and not repository maintainers, it would be nice to devise a strategy for tagging submodule maintainers (even if it's just an @). Possibly would need to be a bot.

gabrieldemarmiesse:
Fix #1145


gabrieldemarmiesse:
Hooray!

gabrieldemarmiesse:
I suppose we need to remove all people who have write access, right? They already get a notification.

seanpmorgan:
> I suppose we need to remove all people who have write access, right? They already get a notification. and also tensorflow/sig-addons-maintainers since it doesn't seem to trigger anything: [#1137 (comment)](https://github.com/tensorflow/addons/pull/1137#issuecomment-592021710)

Thanks this is fantastic!

1) Yeah we'll probably want to remove the duplicate notifications. Do you have a method in mind for doing this? Filtering members of sig-addons-maintainers?

Subsequent PR:
1) Yeah sig-addons-maintainers doesn't actually do anything. We can remove that catch all.
2) We'll need to update the CODEOWNERS to match the README docs. Then we can remove the maintainers from the README so we don't have duplicated work on PRs.

gabrieldemarmiesse:
The easier is to hard-code in the python file the people with write access. This is not public information so I cannot fetch it automatically. I'll add a filter somewhere.

We want to minimize spamming. So the bot will only notify when the pull request is opened. Never after. It's also simpler code-wise.

.github/workflows/notify_codeowners.py
```diff
@@ -0,0 +1,173 @@
...
+BLACKLIST = [
+    "tensorflow/sig-addons-maintainers",
+    "facaiy",
+    "seanpmorgan",
+    "squadrick",
+    "shun-lin",
+    "windqaq",
+    "qlzh727",
+    "guillaumekln",
+    "helinwang",
```
seanpmorgan: Doesn't have write access. Updated the confusing codeowners in another repo

gabrieldemarmiesse: Done. Please merge this PR before updating the codeowners. A check runs in the CI to verify that CODEOWNERS is valid.

Issue #1143, opened 2020-02-23
Make the tests faster


gabrieldemarmiesse:
**Describe the bug**

Currently, the tests take around 600s. But only a few tests take the majority of this time. The short term solution is to make those specific tests shorter, the long term solution is to put a specific solution in place to monitor the changes in test time and have automatic report of new long tests/tests which became much longer after a change.

Here are the current number from the linux build in the master branch:

```
//tensorflow_addons/activations:activations_test                         PASSED in 2.6s
//tensorflow_addons/activations:gelu_test                                PASSED in 2.4s
//tensorflow_addons/activations:hardshrink_test                          PASSED in 3.9s
//tensorflow_addons/activations:lisht_test                               PASSED in 2.6s
//tensorflow_addons/activations:mish_test                                PASSED in 2.7s
//tensorflow_addons/activations:rrelu_test                               PASSED in 2.2s
//tensorflow_addons/activations:softshrink_test                          PASSED in 4.1s
//tensorflow_addons/activations:sparsemax_test                           PASSED in 11.2s
//tensorflow_addons/activations:tanhshrink_test                          PASSED in 2.4s
//tensorflow_addons/image:connected_components_test                      PASSED in 5.3s
//tensorflow_addons/image:dense_image_warp_test                          PASSED in 40.9s
//tensorflow_addons/image:distance_transform_ops_test                    PASSED in 2.8s
//tensorflow_addons/image:distort_image_ops_test                         PASSED in 4.4s
//tensorflow_addons/image:filters_test                                   PASSED in 56.8s
//tensorflow_addons/image:interpolate_spline_test                        PASSED in 6.8s
//tensorflow_addons/image:resampler_ops_test                             PASSED in 9.7s
//tensorflow_addons/image:sparse_image_warp_test                         PASSED in 14.2s
//tensorflow_addons/image:transform_ops_test                             PASSED in 66.1s
//tensorflow_addons/image:translate_ops_test                             PASSED in 5.5s
//tensorflow_addons/image:utils_test                                     PASSED in 4.2s
//tensorflow_addons/layers:gelu_test                                     PASSED in 4.2s
//tensorflow_addons/layers:layers_wrappers_test                          PASSED in 37.3s
//tensorflow_addons/layers:maxout_test                                   PASSED in 6.0s
//tensorflow_addons/layers:normalizations_test                           PASSED in 24.0s
//tensorflow_addons/layers:optical_flow_test                             PASSED in 27.0s
[...]

PR #1170, opened 2020-02-27
[Test] Speed up moving_average_test

Squadrick:
Increase error: 1e-3 -> 5e-3
Reduce number of samples: 50000 -> 5000
Use a simpler optimizer: adam -> sgd
Reduce epochs: 10 -> 5

CPU Test time: 35s -> 8s

Related to #1143


gabrieldemarmiesse:
Failure is unrelated. Thanks for the pull request!

PR #1173, opened 2020-02-27
[Test] Speed up transform_ops_test

Squadrick:
`_test_grad` was the most expensive call. Reduce the number of calls by removing useless differing image sizes.

Local test: 55s -> 21s 

Related to #1143


gabrieldemarmiesse:

PR #1175, opened 2020-02-27
Have faster filters_test 

autoih:
filters_test: 56.8s -> 41s (by replacing divide op)
related to [1143](https://github.com/tensorflow/addons/issues/1143)


Squadrick:
Wowzah, didn't think that would make such a big difference. Excellent catch, this is brilliant.

PR #1176, opened 2020-02-27
Update codeowners

### LINKED ISSUE #1168: Update CODEOWNERS to match documentation
Once #1167 merges, we'll finally have an automatic system for notifying submodule maintainers. 
We can remove the maintainers from each subpackage README so as not to duplicate work on PRs. Will also want to modify the contribution instructions.

seanpmorgan:
Closes #1168 

* Single location of maintainers
* Enables #1145 
* Directs users to auto-generated documentation when possible

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -47,7 +47,7 @@
 /tensorflow_addons/image/transform_ops*.py @mels630
 /tensorflow_addons/image/translate_ops*.py @sayoojbk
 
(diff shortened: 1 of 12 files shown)


gabrieldemarmiesse:
@seanpmorgan could we merge the bot before? It will then check your CODEOWNERS and  report if it is valid (typos can happen)

gabrieldemarmiesse:
I think my bot doesn't like tabs. It should handle multiple spaces without problem.

PR #1181, opened 2020-02-28
update to make it consistent with the parent.

pkanwar23:
details in https://github.com/tensorflow/tensorflow/issues/36398


PhilJd:
Thanks for the fix! How about using **kwargs instead? I just skimmed over the code base and it seems that LossScaleOptimizer is the only optimizer having this argument, so this would break the other optimizers.

gabrieldemarmiesse:
This pull request has been left without activity for a while. If there are new changes, feel free to re-open it.

PR #1184, opened 2020-02-28
Remove all mentions of nosla.

### LINKED ISSUE #1116: Stop using tensorflow-testing docker images and start using hub.docker.com/tensorflow images

gabrieldemarmiesse:
Fix #1116

A "search in the whole projects" tells me that those private docker images are not used anymore.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -104,9 +104,9 @@ bash tools/run_gpu_tests.sh
 
 It is recommend that tests are ran within docker images, but should still work on host.
 
-CPU Docker: `docker run --rm -it -v ${PWD}:/addons -w /addons gcr.io/tensorflow-testing/nosla-ubuntu16.04-manylinux2010 /bin/bash`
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
The bot is drunk. It will be fixed with #1183

PR #1186, opened 2020-02-28
Enable full editable mode workflow

gabrieldemarmiesse:
The python `pip install -e ./` is very useful for numerous reasons, quick install, running plain script files with an editable version of tf addons and a very important one I believe: graphical debuggers (pycharm, visual studio code). 

Before, we could do `TF_ADDONS_NO_BUILD=1 pip install -e ./` and we would get an editable version of tf addons. The only problem was that since bazel wasn't used, no `.so` file was present, and thus it was impossible to use functions which used custom ops.

I made a script to bypass this limitation, allowing pycharm and visual studio code to debug tf addons code without problem, even when running tests involving custom ops.

I wonder if there is a more practical way to do that than listing all the `.so` files.


tools/docker/sanity_check.Dockerfile
```diff
@@ -106,6 +106,30 @@ RUN TF_ADDONS_NO_BUILD=1 pip install --no-deps -e .
...
+COPY tools/docker/finish_bazel_install.sh ./
+RUN bash finish_bazel_install.sh
+
+
+COPY ./ /addons
+WORKDIR /addons
+RUN python configure.py --no-deps
+RUN bash tools/install_so_files.sh
+RUN TF_ADDONS_NO_BUILD=1 pip install --no-deps -e .
+RUN python -c "from tensorflow_addons.activations import gelu; print(gelu(0.2))"
```
seanpmorgan: Same as the other PR could we switch to lisht

tools/install_so_files.sh
```diff
@@ -0,0 +1,9 @@
+bazel build //tensorflow_addons/...
+cp ./bazel-bin/tensorflow_addons/custom_ops/activations/_activation_ops.so ./tensorflow_addons/custom_ops/activations/
+cp ./bazel-bin/tensorflow_addons/custom_ops/image/_distort_image_ops.so ./tensorflow_addons/custom_ops/image/
+cp ./bazel-bin/tensorflow_addons/custom_ops/layers/_correlation_cost_ops.so ./tensorflow_addons/custom_ops/layers/
+cp ./bazel-bin/tensorflow_addons/custom_ops/text/_parse_time_op.so ./tensorflow_addons/custom_ops/text/
+cp ./bazel-bin/tensorflow_addons/custom_ops/seq2seq/_beam_search_ops.so ./tensorflow_addons/custom_ops/seq2seq/
+cp ./bazel-bin/tensorflow_addons/custom_ops/image/_image_ops.so ./tensorflow_addons/custom_ops/image/
+cp ./bazel-bin/tensorflow_addons/custom_ops/image/_resampler_ops.so ./tensorflow_addons/custom_ops/image/
+cp ./bazel-bin/tensorflow_addons/custom_ops/text/_skip_gram_ops.so ./tensorflow_addons/custom_ops/text/
```
seanpmorgan: Hmmm anyway we can automate this? I don't think there should be other .so files other than the ones we want to copy?

tools/docker/sanity_check.Dockerfile
gabrieldemarmiesse: Sure thing

tools/install_so_files.sh
gabrieldemarmiesse: I'd like to automate this but I have no idea how to. Maybe we can merge this and open an issue? Maybe someone else has a good idea.

gabrieldemarmiesse: I used a precise wildcard pattern because if not, we end up with other so like `tensorflow_addons/custom_ops/activations/lib_activation_ops_gpu.so`.  Are those useful?

Issue #1156, opened 2020-02-26
Benchmark activations using different implementations.


gabrieldemarmiesse:
See https://github.com/tensorflow/addons/pull/1137#discussion_r384586158

We now have multiple implementations for activations on CPU and GPU, It might be worth benchmarking them because it's not obvious which ones are faster. On CPU, for big tensors, pure python in eager mode seems to be faster than custom C++ ops. 

See https://colab.research.google.com/drive/1LTx3vMpA1fLCESKl-_WrLIp0Fq1zYZL0
for benchmarking on cpu (when running the notebook, make sure the gpu is not available in colab).

For GPU and CPU, 4 implementations should be tested:
* Custom kernel (C++ or CUDA)
* Pure python in eager mode
* Pure python with tf.function
* Pure python with XLA

We should also test for big and small tensors (the best would be to have plots, with 4 curves, x axis being the number of elements in the input tensor, y axis being the speed (number of elements processed by sec). Don't fortget to use `.numpy()` to force the execution (ops might get executed lazily).

To obtain the result of a %timeit, use the `-o` flag:

```
timeit_object_with_results = %timeit -o my_func(my_tensor)
```

If everything could be delivered in a pretty notebook on colab with the link in this issue (no pull request), it would be surper duper awesome 😃

seanpmorgan:
Just a note I'll be bringing this up at the monthly meeting to see if TF core team has a standard process for this. As previously mentioned calling `.numpy()` will transfer results from GPU to CPU so we need a way to test without that. See:
https://colab.research.google.com/drive/12ende9xXMSywP2lOKWrFJBwaDDHbkYXh

gabrieldemarmiesse:
> Just a note I'll be bringing this up at the monthly meeting to see if TF
core team has a standard process for this.

Is this possible to join this meeting? At least as observer?

seanpmorgan:
> Is this possible to join this meeting? At least as observer?


Of course! Makes me think we're not properly advertisting this meeting.. its open to all though you need to be subscribed to the mailing list in order to join.
https://github.com/tensorflow/addons#community
https://groups.google.com/a/tensorflow.org/forum/#!forum/addons
https://docs.google.com/document/d/1kxg5xIHWLY7EMdOJCdSGgaPu27a9YKpupUz2VTXqTJg/edit

failure-to-thrive:
[...]

PR #1189, opened 2020-02-28
Test wheels in a fresh environement

### LINKED ISSUE #1122: Test the release process and wheels at each commit.

gabrieldemarmiesse:
Fix #1122


.github/workflows/release.yml
```diff
@@ -54,32 +54,50 @@ jobs:
...
+        with:
+          name: ${{ runner.os }}-${{ matrix.python-version }}-wheel
+          path: ./wheel
+      - name: Test wheel
+        if: github.event_name == 'schedule' || github.event_name == 'release' || matrix.python-version == '3.5'
+        shell: bash
+        run: |
+          pip install tensorflow~=2.1.0
+          pip install wheel/tensorflow_addons-*.whl
+          python -c "import tensorflow_addons as tfa; print(tfa.activations.gelu(0.2))"
```
seanpmorgan: Since gelu is a possible candidate for moving to TF core. Could we replace with another custom-op? Say [gather_tree](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/seq2seq/beam_search_decoder.py#L34) As seq2seq is likely to stay in Addons for the foreseeable future.

gabrieldemarmiesse: I used `lisht` because it's simpler to setup than `gather_tree`.

seanpmorgan: Thanks, I'm good with that. Depending on results of #1156 lisht custom-op may be removed at some point so lets just keep that in mind as those results get reported.

PR #1115, opened 2020-02-20
Revert WeightNorm mutex

seanpmorgan:
* Revert b62bbde

See https://github.com/tensorflow/addons/pull/1025 for more details

PR #1190, opened 2020-02-29
fix wrapper test

fsx950223:
Related #1115.

### DIFF
--- a/tensorflow_addons/layers/wrappers_test.py
+++ b/tensorflow_addons/layers/wrappers_test.py
@@ -163,12 +163,11 @@ class WeightNormalizationTest(tf.test.TestCase, parameterized.TestCase):
             with tf.control_dependencies([wn_output]):
                 wn_removed_layer = wn_layer.remove()
                 wn_removed_output = wn_removed_layer(sample_data)
-
-            self.evaluate(tf.compat.v1.initialize_variables(wn_removed_layer.variables))
-            self.assertAllClose(
-                self.evaluate(wn_removed_output), self.evaluate(wn_output)
-            )
-            self.assertTrue(isinstance(wn_removed_layer, base_layer.__class__))
+                self.evaluate(tf.compat.v1.initialize_variables(wn_removed_layer.variables))
+                self.assertAllClose(
+                    self.evaluate(wn_removed_output), self.evaluate(wn_output)
+                )
+                self.assertTrue(isinstance(wn_removed_layer, base_layer.__class__))
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


tensorflow_addons/layers/wrappers_test.py
```diff
@@ -156,19 +156,22 @@ def test_removal(self, base_layer_fn, input_shape):
...
+                base_layer = base_layer_fn()
+                wn_layer = wrappers.WeightNormalization(base_layer, data_init)
+                wn_output = wn_layer(sample_data)
+                self.evaluate(tf.compat.v1.global_variables_initializer())
+                with tf.control_dependencies([wn_output]):
+                    wn_removed_layer = wn_layer.remove()
+                    wn_removed_output = wn_removed_layer(sample_data)
+
+                self.evaluate(
+                    tf.compat.v1.initialize_variables(wn_removed_layer.variables)
```
fsx950223: Use ```initialize_variables``` instead of ```global_variables_initializer``` fix the test bug.

fsx950223:
cc @seanpmorgan @Squadrick.

gabrieldemarmiesse:
@failure-to-thrive also, who is familiar with the matter at hand. If you have the time to, could you review? Thanks a lot :)

tensorflow_addons/layers/wrappers_test.py
gabrieldemarmiesse: Shouldn't we do the same line 163 if `global_variables_initializer` causes issues? Can you also detail a bit more why this fixes the problem? I'll assign myself to this pull request. It's a quite important bug to fix.

gabrieldemarmiesse:
I tested locally and I can confirm the flaky test is fixed by this pull request. If someone wants to try it out:

```
TF_ADDONS_NO_BUILD=1 pip install -e ./
pip install  git+https://github.com/dropbox/pytest-flakefinder.git   # a cool tool from dropbox
pytest --flake-finder -v -k "test_removal_" tensorflow_addons/layers/wrappers_test.py
```

It will run the tests in a loop. Open another terminal to run this command twice and you'll see the frequency of failed tests increase (~one failure every 3 seconds). With this pull request, I can't see a single test failing.

failure-to-thrive:
Sorry to be late. My 2 cents: [`tf.compat.v1.initialize_variables` has been deprecated](https://www.tensorflow.org/api_docs/python/tf/compat/v1/initialize_variables). Although it is not a big deal as for now.
Another 2 center: `tf.compat.v1.global_variables_initializer` displacement is a good catch despite whether it fixes the bug or not. Who knows what else in the test's computational graph it reinitializes. Test infrastructure might have its own variables.

gabrieldemarmiesse:
@failure-to-thrive, I agree with you about all this and I hope we can reduce the number of deprecated APIs we use.

PR #1191, opened 2020-02-29
Make release_linux.sh configurable for python version

### LINKED ISSUE #1179: Make release_linux.sh configurable for python version
Currently our release for linux works quite differently than MacOS / Windows. All python versions are built serially within a docker container:
https://github.com/tensorflow/addons/blob/master/tools/ci_build/builds/release_linux.sh#L36

We should instead pass in an argument for which python version to build with. This would make it so our CI tests can build a single wheel file during the build check:
https://github.com/tensorflow/addons/blob/master/.github/workflows/release.yml#L32-L46

Currently the test times are:
- MacOS: ~9min
- Windows: ~14min
- Linux: ~23min

The current system also hardcodes the python versions in this script and is a maintenance burden when we update workflow files in the future.

gabrieldemarmiesse:
Fix #1179

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,18 +32,23 @@ jobs:
   manylinux-release-wheel:
     name: Build release wheels for manylinux2010
     runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python-version: ['3.5', '3.6', '3.7']
     steps:
       - if: github.event_name == 'schedule'
         run: echo "::set-env name=NIGHTLY_FLAG::--nightly"
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
         with:
-          python-version: 3.5
+          python-version: ${{ matrix.python-version }}
       - name: Build manylinux2010 wheels
+        env:
+          PY_VERSION: ${{ matrix.python-version }}
         run: |
           docker run -e TF_NEED_CUDA=1 -v ${PWD}:/addons -w /addons \
             tensorflow/tensorflow:2.1.0-custom-op-gpu-ubuntu16 \
-            tools/ci_build/builds/release_linux.sh $NIGHTLY_FLAG
+            tools/ci_build/builds/release_linux.sh $PY_VERSION $NIGHTLY_FLAG
 
           sudo apt-get install patchelf
           python3 -m pip install -U auditwheel==2.0.0
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -43,6 +43,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Build manylinux2010 wheels
+        if: github.event_name == 'schedule' || github.event_name == 'release' || matrix.python-version == '3.5'
         env:
           PY_VERSION: ${{ matrix.python-version }}
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -32,18 +32,23 @@ jobs:
...
         run: echo "::set-env name=NIGHTLY_FLAG::--nightly"
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
         with:
-          python-version: 3.5
+          python-version: ${{ matrix.python-version }}
       - name: Build manylinux2010 wheels
+        env:
+          PY_VERSION: ${{ matrix.python-version }}
         run: |
```
seanpmorgan: Can we add an if statement similar to the others:
```
 if: github.event_name == 'schedule' || github.event_name == 'release' || matrix.python-version == '3.5'
```
Probably best to do this after we confirm that we're correctly building 3.5/3.6/3.7

tools/ci_build/builds/release_linux.sh
```diff
@@ -29,48 +29,22 @@ export CUDA_TOOLKIT_PATH="/usr/local/cuda"
...
-done
-
-# Clean up
-rm get-pip.py
+ln -sf $(which python$1) /usr/bin/python
+python -m pip install --upgrade pip
+python -m pip install --upgrade setuptools
+
+#Link TF dependency
+python ./configure.py --quiet
```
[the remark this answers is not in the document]
gabrieldemarmiesse: Sure thing, I see the artifacts are wrong, but I don't understand why. Isn't a symlink enough?

tools/ci_build/builds/release_linux.sh
```diff
@@ -29,48 +29,23 @@ export CUDA_TOOLKIT_PATH="/usr/local/cuda"
...
-
-# Clean up
-rm get-pip.py
+ln -sf $(which python$1) /usr/bin/python
+ln -sf $(which python$1) /usr/bin/python3
+python -m pip install --upgrade pip
+python -m pip install --upgrade setuptools
+
+#Link TF dependency
+python ./configure.py --quiet
```
seanpmorgan:
```suggestion
ln -sf $(which python$1) /usr/bin/python3
python3 -m pip install --upgrade pip
python3 -m pip install --upgrade setuptools

#Link TF dependency
python3 ./configure.py --quiet
```

tools/ci_build/builds/release_linux.sh
```diff
@@ -29,48 +29,22 @@ export CUDA_TOOLKIT_PATH="/usr/local/cuda"
...
-done
-
-# Clean up
-rm get-pip.py
+ln -sf $(which python$1) /usr/bin/python3
+python3 -m pip install --upgrade pip
+python3 -m pip install --upgrade setuptools
+
+#Link TF dependency
+python3 ./configure.py --quiet
```
seanpmorgan:
```suggestion
python3 --version
python3 ./configure.py
```

seanpmorgan: Debugging why that fails

gabrieldemarmiesse:
https://stackoverflow.com/questions/44967202/pip-is-showing-error-lsb-release-a-returned-non-zero-exit-status-1/44967506 May solve the issue

PR #1193, opened 2020-03-01
Public function to register custom ops

### LINKED ISSUE #1151: Public API to force load custom ops
Currently, it is inconvenient to load in Python a SavedModel that includes Addons custom ops. Consider the example below:

* `save.py`

```python
import tensorflow as tf

class Model(tf.keras.Model):
    @tf.function(input_signature=(tf.TensorSpec(shape=[None, 32], dtype=tf.float32),))
    def call(self, x):
        return tfa.activations.gelu(x)

model = Model()
tf.saved_model.save(model, '/tmp/model', signatures=model.call)
```

* `load.py`

```python
import tensorflow as tf
tf.saved_model.load("/tmp/model")
```

The load will fail because Addons custom ops are not registered to the TensorFlow runtime. This is expected as we first have to invoke `tf.load_op_library` on the custom ops.

However, with the new work on lazy loading #855 it got harder to force this op registration. For this model, the user should run the following which relies on internal APIs:

```python
from tensorflow_addons.activations.gelu import _activation_so
_activation_so.ops
```

If the custom ops are not loaded during the main import (i.e. during `import tensorflow_addons`), then the package should expose a public API that registers all custom ops.

Any thoughts?

gabrieldemarmiesse:
Fix #1151

I'll open an issue to track the registering of keras objects with this function. It's there for future backward compatibility.


seanpmorgan:
Haven't had time for a full review, but at first glance looks like a good strategy. @guillaumekln could you take a look when time allows and see if this covers the issue from your perspective.

tensorflow_addons/register.py
```diff
@@ -0,0 +1,53 @@
+import glob
+import os
+from pathlib import Path
+
+import tensorflow as tf
+
+from tensorflow_addons.utils.resource_loader import get_project_root
+
+
+def register_all(keras_objects: bool = True, custom_kernels: bool = True) -> None:
```
guillaumekln: Maybe you should add docstring here to describe what it does and when it is required?

tensorflow_addons/register.py
```diff
@@ -0,0 +1,53 @@
...
+            "use custom ops in your model, you can skip registering custom ops with "
+            "`tfa.register_all(custom_kernels=False)`".format(shared_object)
+        ) from e
+
+
+def _get_all_shared_objects():
+    custom_ops_dir = os.path.join(get_project_root(), "custom_ops")
+    all_shared_objects = glob.glob(custom_ops_dir + "/**/*.so", recursive=True)
+    all_shared_objects = [x for x in all_shared_objects if Path(x).is_file()]
+    return list(all_shared_objects)
```
guillaumekln: Looks like you don't need to copy the list here.

gabrieldemarmiesse:
Thanks a lot for the review! Let me fix all that.

guillaumekln:
@seanpmorgan Do you want to take a second look before merging?

gabrieldemarmiesse:
This pull request will be discussed at the meeting tonight I believe.

seanpmorgan:
@gabrieldemarmiesse Conflicts and then LGTM. Per discussion we'll want to work toward registering Keras objects in a similar manner

gabrieldemarmiesse:
> Per discussion we'll want to work toward registering Keras objects in a similar manner

Let's do that in another pull request to avoid making the review harder.

PR #1195, opened 2020-03-02
Clean tutorials 

### LINKED ISSUE #1104: Clean up tutorials
So TFA is currently included in the TF2.x environment within colab. This is great news, but since it's an old release we'll want to upgrade the package prior to importing as @failure-to-thrive pointed out:
https://github.com/tensorflow/addons/issues/927#issuecomment-583868873

We've created an issue to see if we can get up to date installations but while we wait for that we should make sure our current tutorials work:
https://github.com/googlecolab/colabtools/issues/1005

Another thing that would improve the maintainer burden would be if we could drop the required versions from the pip upgrades. Recent addons should be more or less backwards compatible so I don't think this is still necessary. An edge case would be for a very new tutorial that was added prior to release. In that case we could install `tfa-nightly`

abhichou4:
Closes #1104


gabrieldemarmiesse:
Thanks a lot!

PR #1199, opened 2020-03-02
Proof that hamming_test.py was not run by bazel

gabrieldemarmiesse:
This is just a proof of concept, I don't have the expertise to fix the problem.


gabrieldemarmiesse:
This is just a proof of concept, I don't have the expertise to fix the problem.

gabrieldemarmiesse:
Closing as the issue is enough to track the problem.

Issue #1171, opened 2020-02-27
test_utils.run_distributed fails on GPU


Squadrick:
The error is: 

```
tensorflow.python.framework.errors_impl.UnknownError: Not enough GPUs to create virtual devices. num_gpus_to_use: 1 #virtual_devices: 2
```
This happens since number of physical GPUs != number of virtual GPUs.

The line which throws the error is here: [`gpu_device.cc`](https://github.com/tensorflow/tensorflow/blob/feacb9f25458882d1f15aec153848017ccb5c162/tensorflow/core/common_runtime/gpu/gpu_device.cc#L782). 

The [documentation for `tf.config.set_logical_device_configuration`](https://www.tensorflow.org/api_docs/python/tf/config/set_logical_device_configuration) seems to run that bit of code successfully. 

Relevant log: https://source.cloud.google.com/results/invocations/606014e4-04b9-4cd5-a907-1b7a2661375a/targets/tensorflow_addons%2Fubuntu%2Fgpu%2Fpy3%2Fpresubmit/log

hyang0129:
@Squadrick I'm willing to take a look at this one. #969 needs to run distributed tests.

gabrieldemarmiesse:
Closing this as it's going to be tracked in #1682

PR #1209, opened 2020-03-02
[WIP] Testutils run distributed fix

hyang0129:
see #1171


hyang0129:
@Squadrick  please add Ubuntu GPU Python3 CI

hyang0129:
@seanpmorgan I thought that the gpu test would stay, but it seems that it gets removed for some reason. I'll ping you again once the code is actually ready to test on GPU.

hyang0129:
@seanpmorgan please add gpu test label

hyang0129:
I've read through the error logs on the gpu run and it appears that everything except combining run_graph_eager and run_distributed works. 

The error in graph execution seems to be OP_REQUIRES failed at nccl_ops.cc:104 : Unknown: Error invoking NCCL: unhandled cuda error. 

I believe this is related to an earlier warning NCCL is not supported when using virtual GPUs, fallingback to reduction to one device. This warning always appears when using run distributed with virtual GPUs. 

My guess is that the graph mode is missing a hook to change NCCL GPU to NCCL CPU when using virtual GPUs. It maybe tries to build a graph with NCCL GPU, but it cannot access that op because the GPUs are virtualized.

tensorflow_addons/utils/test_utils.py
```diff
@@ -80,17 +80,79 @@ def create_virtual_devices(
...
+        """
+            % (len(logical_devices), num_devices)
+        )
+    elif len(logical_devices_out) > num_devices:
+        raise ArithmeticError(
+            """List comprehension failure.
+        Expected the following python code to return true, when x, i, n are integers:
+        x <= len([i for i in range(n)][:x])
+        """
+        )
```
Squadrick: Use an `assert` instead of a conditional. More importantly, this is a useless check, since we can assume that python isn't broken.

Squadrick:
@hyang0129 I've re-triggered the GPU tests, so I can see the logs as well. 

It might be more suitable to open a bug report in TF-core.

tensorflow_addons/utils/test_utils.py
hyang0129: makes sense I'll remove that check and chance the others to asserts

hyang0129:
> @hyang0129 I've re-triggered the GPU tests, so I can see the logs as well.
> 
> It might be more suitable to open a bug report in TF-core.

Can we proceed with by creating a catch block for GPUs when using run distributed in graph context? 

It seems to run fine in eager. So if it detects GPU and graph execution, it will log a warning and skip or pass the test? 

I also have no idea when a bug logged on the main TF package would be resolved and I don't have enough experience with the underlying TF ops and stuff to fix it myself.

Squadrick:
> Can we proceed with by creating a catch block for GPUs when using run distributed in graph context?

Yes, that's a good enough temp fix. 

> It seems to run fine in eager. So if it detects GPU and graph execution, it will log a warning and skip or pass the test?

Log a warning, and skip the test. 

> I also have no idea when a bug logged on the main TF package would be resolved and I don't have enough experience with the underlying TF ops and stuff to fix it myself.

No worries. We can stick to the above escape until it's resolved. Once an issue has been created in TF-core, we can add a TODO to our util to revert back to standard behaviour once it's resolved.

hyang0129:
@Squadrick @seanpmorgan please add the GPU tag

hyang0129:
@seanpmorgan @Squadrick  it seems that the wheel build process changed sometime in the last 6 days. 

The build wheels process now runs tests differently than the bazel tests. In seems that the tf context is initialized before tests are run or do not get reset between tests. 

This is very odd because the bazel tests work without issue. This also seems to bypass the check to see if logical devices have been set up already.

seanpmorgan:
@googlebot I consent.

seanpmorgan:
>  it seems that the wheel build process changed sometime in the last 6 days.
> 
> The build wheels process now runs tests differently than the bazel tests. In seems that the tf context is initialized before tests are run or do not get reset between tests.
> 
> This is very odd because the bazel tests work without issue. This also seems to bypass the check to see if logical devices have been set up already.

Hmmm thanks for bringing this up. I'm thinking it's related to running pytest as the runner:
https://github.com/tensorflow/addons/blob/master/tools/testing/addons_cpu.sh#L41

But its not clear to me why. Just merged the most recent CI which splits that out as a single test in CI so it'll be easier to debug

gabrieldemarmiesse:
So a bit of background to help here:

Pytest doesn't restart the python interpreter between tests files. I believe that bazel does. When running the gpu tests in the CI, there is only one worker to avoid out of memory errors. If that's an issue, we can always fallback to run the bazel command in kokoro, we're compatible with both bazel and pytest so that should be fine.

gabrieldemarmiesse:
@googlebot I consent.

hyang0129:
> So a bit of background to help here:
> 
> Pytest doesn't restart the python interpreter between tests files. I believe that bazel does. When running the gpu tests in the CI, there is only one worker to avoid out of memory errors. If that's an issue, we can always fallback to run the bazel command in kokoro, we're compatible with both bazel and pytest so that should be fine.
> 
> If I understand your issue correctly, you count on the python interpreter being shut down do force the cleanup of the virtual devices, right? Is it possible to do the cleanup in python without shutting down the interpreter?
> 
> Update: I tested locally on CPU this pull request. It has nothing to do with shutting down the interpreter. This pull request works fine with pytest when using a single worker `pytest -v tensorflow_addons/utils`, but it fails when using multiple processes: `pytest -v -n 3 tensorflow_addons/utils`. Since the tests are running with multiple workers in the CI, this explains that.
> This also explains why it works with bazel. Bazel can't execute multiple test functions from the same file concurrently, but pytest can. So bazel executing your file with a single worker fixes the issue.
> 
> If possible, we should aim to have tests that work even if multiple processes are running in the same time, at least when running cpu tests.

I would like that as well. Do you have any suggestions? 

I believe the issue arises from the set logical device configuration command, which is cannot be altered as part of this PR (it is the only way to set up virtual devices for running tests in distributed mode). 

My guess is that pytest in multi worker mode is capable of creating independent test sessions in tensorflow, but not able to interact with the physical and logical device configurations independently. There may be something external to the tf session level that determines device configurations. 

Based on the logs and the fact that it works in single worker mode, I believe that the context is shared between test sessions in multi worker mode. For this to be true there would have to be some mechanism to reset the context between tests in single worker mode, that does not fire in multi worker mode. 

If that is the case, then we really only have two options. 

1. Make set logical device configurations work in a multi worker environment. 
2. Run tests that interact with device configurations always in single worker mode. 

The first option may be the most proper, but that requires changes to code in tensorflow. We might be stuck with option 2 for the near future. 

How should I proceed on this?

gabrieldemarmiesse:
Worst case scenario, if this issue concern only a few tests, they can be run without multiprocessing separatly from the main test suite. Pytest-xdist has also options to control how are executed each test. So we can look into that too. But if we have many tests which won't run be able to run in parralel, that's a problem.

hyang0129:
> Worst case scenario, if this issue concerns only a few tests, they can be run without multiprocessing separatly from the main test suite. Pytest-xdist has also options to control how are executed each test. So we can look into that too. But if we have many tests which won't run be able to run in parralel, that's a problem.

Agreed. 

> Worst case scenario, if this issue concerns only a few tests, they can be run without multiprocessing separatly from the main test suite. Pytest-xdist has also options to control how are executed each test. So we can look into that too. But if we have many tests which won't run be able to run in parralel, that's a problem.

Is there a test config file that I should update to indicate that the tests for testutils should run with a single worker?

gabrieldemarmiesse:
@hyang0129  does that make sense to run those tests on CPU? I'm asking because currently, it the CI, gpu tests run with a single worker, but CPU tests run with multiple workers. So if we disable those tests on cpu, this would work, right?

hyang0129:
> @hyang0129 does that make sense to run those tests on CPU? I'm asking because currently, it the CI, gpu tests run with a single worker, but CPU tests run with multiple workers. So if we disable those tests on cpu, this would work, right?

Yes that makes sense. However, doesn't that imply that distributed tests will only run in GPU mode?

gabrieldemarmiesse:
Yes they would only run in gpu mode, is that an issue?

hyang0129:
> Yes they would only run in gpu mode, is that an issue?

That should be fine. I am not aware of any real world application of distributed CPU training, so if we only test for distributed GPU, that should cover the distributed use cases.

gabrieldemarmiesse:
@hyang0129 , I'll let you do the necessary changes. We can review once the CI is passing.

gabrieldemarmiesse:
Closing this pull request because it's been superseded by #1770

Issue #987, opened 2020-01-31
Custom Op Linux ABI Incompatibility: Undefined Symbol


seanpmorgan:
Creating this issue to consolidate the information. As seen in https://github.com/tensorflow/addons/issues/676 https://github.com/tensorflow/addons/issues/574

## Overview:
The TF C++ API for kernel implementations are not ABI stable. Because of this we compile our pip package to be manylinux2010 compatible with GCC 7.3.1. This means that TF-Addons custom ops will work on pip installed TF, but are not guaranteed to work with TF built from source or installed from conda

This should not affect our macOS and Windows versions.

## Temporary "fixes"
We implemented a lazy loading of custom-ops so majority of TFA functionality is not affected by this issue:
https://github.com/tensorflow/addons/pull/869

## Long term solution
[RFC 133](https://github.com/tensorflow/community/pull/133) is in the works to make a stable C ABI.

whatdhack:
I am seeing the undefined symbol problem with a custom built TF 1.14 python whl.  What  bazel build option should I use beside  --cxxopt="-D_GLIBCXX_USE_CXX11_ABI=0"  ? How do I specify compatibility with manylinux2010 compatible with GCC 7.3.1

seanpmorgan:
> I am seeing the undefined symbol problem with a custom built TF 1.14 python whl. What bazel build option should I use beside --cxxopt="-D_GLIBCXX_USE_CXX11_ABI=0" ? How do I specify compatibility with manylinux2010 and GCC 7.3.1

TensorFlow Addons only supports TF2.x. We have an issue for build from source compatibility but it would still only support TF2+ https://github.com/tensorflow/addons/issues/770

whatdhack:
Addons are similar to CustomOps in 1.x , right ?  So has the same issue.

PR #1212, opened 2020-03-03
Run linux tests before building the wheel.

gabrieldemarmiesse:


tools/ci_build/builds/release_linux.sh
```diff
@@ -43,6 +43,12 @@ python3 -m pip install --upgrade setuptools
 python3 --version
 python3 ./configure.py
 
+## Run bazel test command. Double test timeouts to avoid flakes.
+bazel test -c opt -k \
+    --jobs=auto --test_timeout 300,450,1200,3600 \
+    --test_output=errors --local_test_jobs=8 \
+    //tensorflow_addons/...
```
seanpmorgan:
```suggestion
    --crosstool_top=//build_deps/toolchains/gcc7_manylinux2010-nvcc-cuda10.1:toolchain \
    //tensorflow_addons/...
```

seanpmorgan: Whats happening is the test is building the custom-ops without the devtoolset-7 toolchain. The tests pass locally, but when combined with tensorflow from pip we'll get the #987 issue.

Because the test builds the ops already the build will not compile them again using the correct toolchian

gabrieldemarmiesse: Thanks for the explanation!

PR #1213, opened 2020-03-03
Upload the nightlies at every commit

### LINKED ISSUE #1144: Proposal: Upload the nightlies at each commit on the master branch
We already build the wheels already anyway. It's just missing the push to pypi. Yes/No?

gabrieldemarmiesse:
Fix #1144

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -99,7 +99,7 @@ jobs:
           ls -la dist/
           sha256sum dist/*.whl
       - uses: pypa/gh-action-pypi-publish@master
-        if: github.event_name == 'push' || github.event_name == 'release'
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -38,21 +36,21 @@ jobs:
         python-version: ['3.5', '3.6', '3.7']
       fail-fast: false
     steps:
-      - if: github.event_name == 'schedule'
+      - if: github.event_name == 'push'
```
seanpmorgan: So i believe this will trigger on release branches as well. Make it push && master?

gabrieldemarmiesse: Nice catch. Let me change that.

PR #1215, opened 2020-03-04
Fix test-release for nightlies

seanpmorgan:
Last night's release failed:
https://github.com/tensorflow/addons/runs/485118606?check_suite_focus=true

This was because the wheel file did not match `tensorflow_addons-*.whl` since its `tfa-*`. A single wheel artifact is downloaded so a wildcard should work here.


gabrieldemarmiesse:
Thank you for the fix!

Issue #336, opened 2019-07-06
Request for example: Weight Decay Optimizers / Super Convergence


seanpmorgan:
A valuable set of optimizers that are available in Addons.

Related info:
https://www.fast.ai/2018/07/02/adam-weight-decay/
https://github.com/tensorflow/addons/blob/master/examples/README.md

SSaishruthi:
I can take up this one if it is ok. Is it ok if we build this example based on `lazy_adam` example?
@seanpmorgan

seanpmorgan:
> I can take up this one if it is ok. Is it ok if we build this example based on `lazy_adam` example?
> @seanpmorgan

Yeah we can use that as a starting point. I would like to include some info from that Fast AI blog (credit them where appropriate)... and it would be great to show the convergence speed against standard adam optimizer

SSaishruthi:
Sure. I will keep you posted about the progress.

SSaishruthi:
@seanpmorgan Can you please assign this issue to me? I am working on it

seanpmorgan:
TensorFlow Addons is transitioning to a minimal maintenance and release mode. New features will not be added to this repository. For more information, please see our public messaging on this decision:
[TensorFlow Addons Wind Down](https://github.com/tensorflow/addons/issues/2807)

Please consider sending feature requests / contributions to other repositories in the TF community with a similar charters to TFA:
[Keras](https://github.com/keras-team/keras)
[Keras-CV](https://github.com/keras-team/keras-cv)
[Keras-NLP](https://github.com/keras-team/keras-nlp)

PR #1217, opened 2020-03-04
Add AdamW tutorial.

us:
Related issue: #336


us:
I worked on colab but I couldn't find the save notebook with outputs. How can I fix this problem?

seanpmorgan:
> I worked on colab but I couldn't find the save notebook with outputs. How can I fix this problem?

So after running the notebook if you select "Save a copy in GitHub" the full notebook with results should save:
https://github.com/tensorflow/addons/tree/master/docs/tutorials#how-to-contribute 

Please let us know if that documentation is out of date or isn't working for some reason.

us:
Thanks for answering Sean, I missed it on doc, I'm doing it right now.

us:
@seanpmorgan I fixed it. The first commit is a mistake, so should I delete this commit?

us:
@seanpmorgan should I change or add anything?

docs/tutorials/optimizers_adamw.ipynb
```diff
@@ -0,0 +1,692 @@
+{
```
seanpmorgan: <p>I worry  that saying "all the popular Deep Learning frameworks (Tensorflow, Pytorch) have implemented Adam with weight decay wrong." will become out of date (if it is not already). Could you remove that comment please.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/addons/pull/1217/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

seanpmorgan: <p>Can you remove the output of this cell. Displaying the results in graphical form below is sufficient.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/addons/pull/1217/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

seanpmorgan: <p>ditto</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/addons/pull/1217/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

seanpmorgan: <p>Results don't seem to be showing the benefit of the optimizer. Do we think this is do to the number of epochs? Perhaps try training for longer to see. If not and we believe the implementation is correct I'm still good with merging this tutorial as an guide to use it.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/addons/pull/1217/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

us: If TensorFlow's adam use weight decay automatically then we should disable but how can I do it?
The difference between `adam` and `adamw` becomes evident at the end of the graphic. So I will try to increase epoch size.

us: Now, I think the last commit showed the benefit of AdamW better.

MHStadler:
The AdamW paper comes with 3 additional important insights for the use of Adam, beyond just the decoupled weight decay:
 - Despite being an adaptive method, it can benefit from a LR schedule (a cosine schedule in the case of their paper - wd schedule is now fixed: https://github.com/tensorflow/addons/pull/2029)
 - L2 regularization does not work as well for Adam as it does for SGD
 - WD should be normalized to be more stable across short/long runs (see Appendix B.1)

So, if you're still looking to improve the graphs, I think this type of comparison would better illustrate the potential of AdamW:
 - SGD with L2
 - Adam with L2
 - AdamW with cosine decay 

What you'd hope to see is that AdamW with the changes from the paper closes the gap between traditional Adam with L2 penalty & SGD with momentum & L2 penalty

bhack:
AdamW now is Keras:
https://github.com/keras-team/keras/blob/master/keras/optimizers/optimizer_experimental/adamw.py

You can add this tutorial at https://github.com/keras-team/keras-io

Thanks

PR #1218, opened 2020-03-05
Set a single variable for release testing version

seanpmorgan:
Sets a single variable for us to edit when we increase our minimum python version. Also adds some context to why it's in the `if` check.

Possibly could think of a better variable name?


gabrieldemarmiesse:
OLDEST_PY_VERSION?

Issue #1205, opened 2020-03-02
poincare_test.py is not run by bazel


gabrieldemarmiesse:
**Describe the bug**

This time it's just that it's not declared in the BUILD.

**Code to reproduce the issue**

Provide a reproducible test case that is the bare minimum necessary to generate the problem.

```
bazel test //tensorflow_addons/layers:poincare_test.py
```

```
INFO: Writing tracer profile to '/home/gdemarmi/.cache/bazel/_bazel_gdemarmi/c0c8233ef0c2491af30539fc61a9ec68/command.profile.gz'
ERROR: Skipping '//tensorflow_addons/layers:poincare_test.py': no such target '//tensorflow_addons/layers:poincare_test.py': target 'poincare_test.py' not declared in package 'tensorflow_addons/layers'; however, a source file of this name exists.  (Perhaps add 'exports_files(["poincare_test.py"])' to tensorflow_addons/layers/BUILD?) defined by /mnt/c/Users/gdemarmi/Desktop/projects/addons/tensorflow_addons/layers/BUILD
ERROR: no such target '//tensorflow_addons/layers:poincare_test.py': target 'poincare_test.py' not declared in package 'tensorflow_addons/layers'; however, a source file of this name exists.  (Perhaps add 'exports_files(["poincare_test.py"])' to tensorflow_addons/layers/BUILD?) defined by /mnt/c/Users/gdemarmi/Desktop/projects/addons/tensorflow_addons/layers/BUILD
INFO: Elapsed time: 0.372s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
FAILED: Build did NOT complete successfully (0 packages loaded)
```

**Other info / logs**

When actually running the tests in the file, we get this error:

```
(proof_of_concept_pytest) /mnt/c/Users/gdemarmi/Desktop/projects/addons $ pytest -v tensorflow_addons/layers/poincare_test.py
/home/gdemarmi/softwares/python/anaconda/lib/python3.7/site-packages/pep8.py:110: FutureWarning: Possible nested set at position 1
  EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')
================================================================================== test session starts ==================================================================================
platform linux -- Python 3.7.5, pytest-5.0.1, py-1.8.0, pluggy-0.13.0 -- /home/gdemarmi/softwares/python/anaconda/bin/python
cachedir: .pytest_cache
[...]

PR #1222, opened 2020-03-05
Enable testing of poincare_test.py

gabrieldemarmiesse:
Related to #1205 but doesn't close it.

Skipping the failing test. Should be fixed later on.

### DIFF
--- a/tensorflow_addons/layers/BUILD
+++ b/tensorflow_addons/layers/BUILD
@@ -52,6 +52,18 @@ py_test(
 )
 
 py_test(
+    name = "poincare_test",
+    size = "small",
+    srcs = [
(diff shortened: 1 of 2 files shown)


rahulunair:
@gabrieldemarmiesse  can you please add reasons on why the test is failing here as well?

gabrieldemarmiesse:
#1205 explains the problem. The whole test file was not run by bazel before it was not declared in the BUILD file. Now in this pull request, I focus on re-enabling passing tests. Fixing the failing one is outside the scope of this pull request and I'll let someone else do it because I'm not familiar with this feature. If you want to see the traceback of the failing test, I added it in my bug report here: #1205

Issue #1200, opened 2020-03-02
hamming_test.py is not run at all with bazel


gabrieldemarmiesse:
See #1199

**Describe the bug**

The file is not run with bazel. When run with another test runner, it fails.

**Code to reproduce the issue**

See #1199

**Other info / logs**

```
(proof_of_concept_pytest) /mnt/c/Users/gdemarmi/Desktop/projects/addons $ pytest ./tensorflow_addons/metrics/hamming_test.py
/home/gdemarmi/softwares/python/anaconda/lib/python3.7/site-packages/pep8.py:110: FutureWarning: Possible nested set at position 1
  EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')
================================================================= test session starts ==================================================================
platform linux -- Python 3.7.5, pytest-5.0.1, py-1.8.0, pluggy-0.13.0
rootdir: /mnt/c/Users/gdemarmi/Desktop/projects/addons
plugins: arraydiff-0.3, cov-2.7.1, doctestplus-0.2.0, flake8-1.0.4, forked-1.0.2, openfiles-0.4.0, pep8-1.0.6, remotedata-0.3.2, xdist-1.27.0, typeguard-2.7.1
collected 7 items

tensorflow_addons/metrics/hamming_test.py ..FFFFs                                                                                                [100%]

======================================================================= FAILURES =======================================================================
_________________________________________________________ HammingMetricsTest.test_mc_4_classes _________________________________________________________

self = <tensorflow.python.framework.test_util.ErrorLoggingSession object at 0x7f15102d2d50>
fn = <function BaseSession._do_run.<locals>._run_fn at 0x7f1510306b90>
args = ({}, [<tensorflow.python.pywrap_tensorflow_internal.TF_Output; proxy of <Swig Object of type 'TF_Output *' at 0x7f1510277450> >], [], None, None)
message = 'Container localhost does not exist. (Could not find resource: localhost/total)\n\t [[{{node AssignAddVariableOp}}]]'
m = <re.Match object; span=(81, 111), match='[[{{node AssignAddVariableOp}}'>

    def _do_call(self, fn, *args):
      try:
>       return fn(*args)

/home/gdemarmi/softwares/python/anaconda/lib/python3.7/site-packages/tensorflow_core/python/client/session.py:1367:
[...]

PR #1225, opened 2020-03-05
Enable 2 tests out of 6 in hamming_test.py

gabrieldemarmiesse:
Related to #1200 but does not fix it. It's just enable some correct tests in the CI

### DIFF
--- a/tensorflow_addons/metrics/hamming_test.py
+++ b/tensorflow_addons/metrics/hamming_test.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 # ==============================================================================
 """Tests Hamming metrics."""
+import unittest
 
 import tensorflow as tf
 from tensorflow_addons.metrics import HammingLoss, hamming_distance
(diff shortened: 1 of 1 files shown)


tensorflow_addons/metrics/hamming_test.py
```diff
@@ -39,6 +40,7 @@ def update_obj_states(self, obj, actuals, preds):
     def check_results(self, obj, value):
         self.assertAllClose(value, self.evaluate(obj.result()), atol=1e-5)
 
+    @unittest.skip("Failing. See https://github.com/tensorflow/addons/issues/1200")
```
seanpmorgan: @gabrieldemarmiesse  Just wondering what benefits we see for these PR's to use unittest.skip vs. the self method:
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/resampler_ops_test.py#L100

Don't really have a strong preference for one over the other, but it does prevent an additional import

gabrieldemarmiesse: I don't have a strong preference either. I think they're equivalent. Do you want me to change them to `self.skip`?

seanpmorgan: If you don't mind it'll be consistent with the other skips we've implemented

gabrieldemarmiesse: Sure thing

PR #1233, opened 2020-03-06
Fix bug where each wheel has a different version number.

gabrieldemarmiesse:
Because they're not built at the exact same time in github actions, the 9 wheels we upload have a different version number. Here we use the last commit time. It's also then easier to track back the commit used from the version number.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,17 +27,18 @@ of the community).
 import os
 import sys
 
-from datetime import datetime
 from setuptools import find_packages
 from setuptools import setup
 from setuptools.dist import Distribution
 from setuptools import Extension
 
+BASE_DIR = os.path.dirname(os.path.abspath(__file__))
+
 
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -27,33 +27,38 @@
...
 from setuptools import find_packages
 from setuptools import setup
 from setuptools.dist import Distribution
 from setuptools import Extension
 
+BASE_DIR = os.path.dirname(os.path.abspath(__file__))
+
+
+def get_last_commit_time() -> str:
+    from git import Repo
```
seanpmorgan: okay to have this un-decalred dependency here since this should only be called using `--nightly` flag from the release script?

gabrieldemarmiesse: Yes, that was my intention. It's only to build the nightlies that we need GitPython.

PR #1239, opened 2020-03-07
Add a pip cache

gabrieldemarmiesse:
Using Jason's recommendation of using an environement variable to pin the pip cache path on all os.

The key is very simple and the cache won't be uploaded if there is already a tar present with this key. If we change our requirements, the cache will not be updated automatically, but it's fine because the cache artifacts only last 90 days and can be deleted manually if needed.

Using a hash in the key would complicate the build a lot because we can't get the hash of requirements files in the second steps where there is no checkout to avoid polluting the build environement with source files.

### DIFF
--- a/.github/workflows/make_wheel_Linux.sh
+++ b/.github/workflows/make_wheel_Linux.sh
@@ -1,6 +1,8 @@
 set -e -x
 
 docker run -e TF_NEED_CUDA=1 -v ${PWD}:/addons -w /addons \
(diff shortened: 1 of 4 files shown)


gabrieldemarmiesse:
The cache doesn't work for linux but at least it doesn't throw any error.

gabrieldemarmiesse:
Closing as the speed benefit is not clear. And since we're modifying the CI quite a bit with the arrival of tf 2.2.0, it's better if we don't complexify it.

PR #1243, opened 2020-03-07
Enable pytest and bazel tests.

gabrieldemarmiesse:
Used the idea of Jason and @Squadrick to make bazel run pytest.

It's possible to lauch the tests with the pytest command and with the bazel command. In the build, tests are all started by pytest but I made a single new build (py35, linux) to test with bazel and ensure we're still compatible.

#### The 25 slowest tests are printed in the CI logs
It's just an option from the command line; `--durations=25`.

#### The code coverage is printed in the CI logs
Thanks to the `--cov=tensorflow_addons` command line option (needs `pip install pytest-cov`).
Because we're now people who like fancy things, we can even have a [beautiful HTML report](https://pytest-cov.readthedocs.io/en/latest/reporting.html) by just adding an option from the command line: `--cov-report html`.

#### We can use `assert` all the time.
No need to remember specific ones. Note that this is only available if at the bottom of the file, `sys.exit(pytest.main([__file__]))` is used instead of `tf.test.main()`. This is possible on all files, and we should enable it progressively by changing the last line. One example is in this pull request: see `tensorflow_addons/seq2seq/basic_decoder_test.py`.

#### We don't have to subclass `tf.test.TestCase`, but we can if we need `self`.
See `tensorflow_addons/register_test.py` where simple functions are used. 

#### We can run individual methods instead of the whole file.
`pytest ./tensorflow_addons/layers/wrappers_test.py::WeightNormalizationTest::test_removal_Conv2D`

#### Automatic post mortem debugging 
`--pdb` start the interactive Python debugger on errors or KeyboardInterrupt (useful when tests are stuck).

#### We still need to remember to declare the test in the BUILD file
Otherwise, `pytest ./tensorflow_addons` will run it, but not `bazel test`.

#### We still need to remember to use the boilerplate at the bottom of the test file
Otherwise, `pytest ./tensorflow_addons` will run it, but not `bazel test`.
We should use `sys.exit(pytest.main([__file__]))` from now on (bazel will run the file with pytest when this is written, instead of unittest).

#### Pytest prevents us from using certain functions.
On Windows, for some unknown reason, `tf.compat.v1.test.get_temp_dir()` crash with pytest.
Black magic 😨 

If this pull request is too big, I can extract some smaller pull requests from it to reduce the size.

And thanks everyone for the great feedback in the meeting and in github!


.github/workflows/release.yml
```diff
@@ -19,6 +19,20 @@ env:
...
 
 jobs:
+  test-with-bazel:
+    name: Test with bazel
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v1
+        with:
+          python-version: 3.5
```
seanpmorgan: Use `OLDEST_PY_VERSION`

tools/ci_build/builds/release_linux.sh
```diff
@@ -43,12 +43,7 @@ python3 -m pip install --upgrade setuptools
 python3 --version
 python3 ./configure.py
 
-## Run bazel test command. Double test timeouts to avoid flakes.
-bazel test -c opt -k \
-    --jobs=auto --test_timeout 300,450,1200,3600 \
-    --test_output=errors --local_test_jobs=8 \
-    --crosstool_top=//build_deps/toolchains/gcc7_manylinux2010-nvcc-cuda10.1:toolchain \
-    //tensorflow_addons/...
+bash tools/ci_testing/addons_cpu.sh
```
seanpmorgan: I would think this is an issue. `addons_cpu` will compile the custom-ops without any particular toolchain:
https://github.com/tensorflow/addons/blob/master/tools/ci_testing/addons_cpu.sh#L60-L64

I thought the compile at the subsequent build since they're already created, but they won't be compatible with tf from pypi. However the test-release-wheel is passing, so I wanted to verify with you that this isn't reverting to python_op fallback?

tools/ci_testing/addons_gpu.sh
```diff
@@ -43,19 +29,8 @@ export CUDA_TOOLKIT_PATH="/usr/local/cuda"
 export TF_CUDNN_VERSION="7"
 export CUDNN_INSTALL_PATH="/usr/lib/x86_64-linux-gnu"
 
-# Check if python3 is available. On Windows VM it is not.
-if [ -x "$(command -v python3)" ]; then
-    python3 ./configure.py
-  else
-    python ./configure.py
-fi
```
seanpmorgan: Removing this because GPU isn't available for windows? I'm okay with that but something we'll need to add back once that is complete.

tools/ci_build/builds/release_linux.sh
seanpmorgan: Could we add a `bazel clean --expunge` after this just for sanity.

tools/ci_testing/addons_cpu.sh
```diff
@@ -24,43 +24,18 @@ if [ "$1" != "--no-deps" ] && [ "$1" != "" ]; then
...
-
-## Run bazel test command. Double test timeouts to avoid flakes.
-${BAZEL_PATH:=bazel} test -c opt -k \
-    --jobs=${N_JOBS} --test_timeout 300,450,1200,3600 \
-    --test_output=errors --local_test_jobs=8 \
-    //tensorflow_addons/...
-
-exit $?
+bash tools/install_so_files.sh
+$PYTHON_BINARY -m pytest --cov=tensorflow_addons -v --durations=25 -n auto ./tensorflow_addons
```
seanpmorgan: Could we test with custom_ops and python_op fallbacks? That can be a seperate PR but want to debug to make sure they both work.

tools/ci_build/builds/release_linux.sh
gabrieldemarmiesse: Sure thing. I'll make sure the right toolchaine is used.

tools/ci_testing/addons_gpu.sh
gabrieldemarmiesse: Yup, i agree.

tools/ci_testing/addons_cpu.sh
gabrieldemarmiesse: That should be done at the op level, not really at the project-wide tests level. Since we have python variables to control that, we should have no problems testing the fallback in individual tests. 

I can open a separate pull request as I believe this is independant from this PR

seanpmorgan:
@gabrieldemarmiesse So this is failing only for py37 since the push:
https://github.com/tensorflow/addons/runs/495036979

`ModuleNotFoundError: No module named '_sqlite3'` Does that make any sense to you?

rainwoodman:
Does the pytest support allow declaring tests with pytest.marks.parameterized?

A lot of our tests use nested loops for parametrization can probably read cleaner converted to parameterized.

PR #1252, opened 2020-03-08
Exposed the pure python implementation of mish and softshrink.

gabrieldemarmiesse:


tensorflow_addons/activations/softshrink_test.py
```diff
@@ -18,15 +18,18 @@
...
+)
 from tensorflow_addons.utils import test_utils
 
 
 @test_utils.run_all_in_graph_and_eager_modes
 class SoftshrinkTest(tf.test.TestCase, parameterized.TestCase):
     def test_invalid(self):
         with self.assertRaisesOpError("lower must be less than or equal to upper."):
-            y = softshrink(tf.ones(shape=(1, 2, 3)), lower=2.0, upper=-2.0)
+            y = _softshrink_custom_op(tf.ones(shape=(1, 2, 3)), lower=2.0, upper=-2.0)
```
seanpmorgan: What's preventing this from being configurable so we can test custom and not custom? The regex for error?
Not blocking this PR but something we might want to aim for being able to do.

PR #1256, opened 2020-03-08
Create pythonapp.yml

Wdel06:


seanpmorgan:
Hi @Wdel06 this PR seems to be a mistake? It comes from the starter-workflows:
https://github.com/actions/starter-workflows/blob/master/ci/python-app.yml

And we already implement flake8 testing:
https://github.com/tensorflow/addons/blob/master/.github/workflows/ci_test.yml#L17

PR #1260, opened 2020-03-09
Create pythonapp.yml

Wdel06:
#1186


seanpmorgan:
@Wdel06 please do not spam the repositories pull requests. This PR appears to be a mistake, but if there is some rationale for what you're trying to do please respond in https://github.com/tensorflow/addons/pull/1256

PR #1262, opened 2020-03-09
Created a directory tools/testing to gather all tools for testing.

gabrieldemarmiesse:
There is also a plan to create `tools/release`


seanpmorgan:
So we recently had the Google team use `tools/ci_testing/addons_gpu.sh` as a script we could configure without needing changes again. Do we have a strong preference for `testing` over `ci_testing`? It's better I agree but the most important part for me is that they're all in the same location.

Alternatively we can make another request to internal team but that comes with the broken build for some temporary amount of time

gabrieldemarmiesse:
Actually, the google cloud tests run `tools/run_google_cloud_tests.sh` which is general enough and doesn't need any moving.

seanpmorgan:
> Actually, the google cloud tests run `tools/run_google_cloud_tests.sh` which is general enough and doesn't need any moving.

Ah you're correct. Way to think ahead :)

gabrieldemarmiesse:
I totally planned this pull request 2 months ago

seanpmorgan:
I'm seeing 4 references that need to be updated:
image

So `run_cpu_tests.sh` is not called in our CI. Is there a reason for that or way to consolidate?

gabrieldemarmiesse:
Let me change all that.

> So run_cpu_tests.sh is not called in our CI. Is there a reason for that or way to consolidate?

Work in progress :) lots of cleaning to do

PR #1263, opened 2020-03-09
Make bazel run all files with pytest instead of unittest

### LINKED ISSUE #1177: Use pytest as our test runner
See   #1174.

We have multiple approches to this change and I believe it would be beneficial to discuss it before implementing anything.

@Squadrick would you be able when you have the time to make a summary of how you see the migration? I may not have fully undertood your plan. I'll also make a summary of how I see a possible pytest adoption afterwards so I can highlight the differences and the points where we're going in the same direction. 
Thank you in advance. 

@seanpmorgan We might need to discuss it in the monthly meeting.

gabrieldemarmiesse:
This will allow us to use plain asserts and pytest's features without breaking bazel.

Closes #1177

### DIFF
--- a/tensorflow_addons/activations/activations_test.py
+++ b/tensorflow_addons/activations/activations_test.py
@@ -53,4 +53,4 @@ class ActivationsTest(tf.test.TestCase):
 
 
 if __name__ == "__main__":
-    tf.test.main()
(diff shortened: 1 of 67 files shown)


gabrieldemarmiesse:
Sorry everyone for the spam :(

PR #1264, opened 2020-03-09
Remove fast-fail for easier debug.

gabrieldemarmiesse:
Fast-fail was disabled only for the first jobs. When testing the wheels or uploading it, it's important to have the results of all jobs.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -94,7 +94,6 @@ jobs:
       matrix:
         os: ['macOS', 'Windows', 'Linux']
         python-version: ['3.5', '3.6', '3.7']
-      fail-fast: false
     steps:
       - uses: actions/download-artifact@v1
         if: github.event_name == 'push' || github.event_name == 'release' || matrix.python-version == env.OLDEST_PY_VERSION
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -93,6 +94,7 @@ jobs:
       matrix:
         os: ['macOS', 'Windows', 'Linux']
         python-version: ['3.5', '3.6', '3.7']
+      fail-fast: false
```
seanpmorgan: Mostly LGTM, but wondering why we wouldn't want to fail fast on the pypi uploads? 

Failing fast doesn't  guarantee protection from uploading some but not all of the wheels;  though its possible it could stop it. I'm not seeing a benefit of preventing fast-fail on pypi upload?

gabrieldemarmiesse: Oups my bad. I was mostly focused on testing the wheels in a fresh environement because I had a bug in this section earlier. Let me revert that.

.github/workflows/release.yml
```diff
@@ -93,6 +94,7 @@ jobs:
...
     steps:
```
gabrieldemarmiesse:
```suggestion
        python-version: ['3.5', '3.6', '3.7']
    steps:
```

PR #1274, opened 2020-03-09
Added more info in the CONTRIBUTING.md

gabrieldemarmiesse:
More and more love for the contributors.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -175,11 +175,34 @@ bazel test -c opt -k \
 `<package>` can be any package name like `metrics` for example.
 `<py_test_name>` can be any test name given by the `BUILD` file or `*` for all tests of the given package.
 
-### Install in editable mode
+### Setup your development environment
 
 If you're just modifying Python code (as opposed to C++/CUDA code), 
 then you don't need to use Bazel to run your tests. 
 And you don't need to compile anything.
+
+#### Optional but recommended, creating a virtual environment
+
+If you want to work in 
+a [virtualenv](https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/):
+
+```
+pip install virtualenv
+venv my_dev_environement
+source my_dev_environement/bin/activate  # Linux/macos/WSL2
+.\my_dev_environement\Scripts\activate   # PowerShell
+```
+
+If you want to work in 
+a [conda environment](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html):
+```
+conda create --name my_dev_environement
+conda activate my_dev_environement
+```
+
+
+#### Install TensorFlow Addons in editable mode
+
 Just run from the root:
 
 ```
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -187,13 +189,90 @@ pip install -e ./
...
 ```bash
-python -m unittest tensorflow_addons/rnn/cell_test.py
+python configure.py
+python configure.py --no-deps   # if you don't want any dependencies installed with pip
+bash tools/install_so_files.sh  # Linux/macos/WSL2
+sh tools/install_so_files.sh    # PowerShell
 ```
 
+Note that you need bazel, a C++ compiler and a NVCC compiler (if you want to test
+Cuda ops).
```
seanpmorgan:
```suggestion
Cuda ops). For that reason, we recommend you [run inside the custom-op docker containers](https://github.com/tensorflow/addons/blob/master/CONTRIBUTING.md#run-manually)
```

CONTRIBUTING.md
```diff
@@ -187,13 +189,90 @@ pip install -e ./
...
+# Open the debugger to inspect variables and execute code when 
+# an exception is raised.
+pytest --pdb tensorflow_addons/ 
+
+# or if you prefer the Ipython debugger
+pytest --pdb --pdbcls=IPython.terminal.debugger:TerminalPdb --capture no tensorflow_addons/
+
+# by defaults print() aren't displayed with pytest
+# if you like to debug with prints (you might get 
+# the outpout scrambled
```
seanpmorgan:
```suggestion
# the output scrambled)
```

Curious as to what kind of scrambling would happen here? Different tests printing overtop of each other or out of order for a single test?

CONTRIBUTING.md
```diff
@@ -175,10 +175,12 @@ bazel test -c opt -k \
...
 `<py_test_name>` can be any test name given by the `BUILD` file or `*` for all tests of the given package.
 
-### Install in editable mode without compiling
+### Install in editable mode
 
 If you're just modifying Python code (as opposed to C++/CUDA code), 
-then you don't need to use Bazel to run your tests.
+then you don't need to use Bazel to run your tests. 
+And you don't need to compile anything.
+You don't even need a GPU because it's rare that something works for CPU but not for GPU.
```
seanpmorgan:
```suggestion
```
Would just remove this line. Don't love tying ourself to this statement, and since they're not modifying a custom-op it's likely not needed to be said. A counter-example to this is we've found optimizers often have issues when using mixed precision, but they'll be notified of the issue during the CI

CONTRIBUTING.md
```diff
@@ -187,13 +189,90 @@ pip install -e ./
...
+interpreter startup. This command needs to be executed only once. 
+Now, anywhere on your system, if you do `import tensorflow_addons`, it's 
+going to import the code in this git repository.
+
+To undo this operation, for example, you want to later on 
+install TensorFlow Addons from PyPI, the release version, do:
+
+```
+pip uninstall tensorflow-addons
+```
```
seanpmorgan: Would caveat this as being installed for a specific python environment. A best practice is that they use conda or virtualenv

gabrieldemarmiesse: Hmmm, I'm used to work in a docker container so I never do that. For people working with a local python installation (outside docker) that's definitely the best course of action to make a virtual environnement indeed. I'll look into that.

Issue #1226, opened 2020-03-05
Migrate AutoAugment and RandAugment to TensorFlow Addons.


dynamicwebpaige:
**Describe the feature and the current behavior/state.**
RandAugment and AutoAugment are both policies for enhanced image preprocessing that are included in EfficientNet, but are still using `tf.contrib`.

https://github.com/tensorflow/tpu/blob/master/models/official/efficientnet/autoaugment.py

The only `tf.contrib` image operations that they use, however, are [rotate](https://www.tensorflow.org/addons/api_docs/python/tfa/image/rotate), [translate](https://www.tensorflow.org/addons/api_docs/python/tfa/image/translate) and [transform](https://www.tensorflow.org/addons/api_docs/python/tfa/image/transform) - all of which have been included in TensorFlow Addons.

**Relevant information**
- Are you willing to contribute it (yes/no): 
No, but am hoping that someone from the community will pick it up (potentially a Google Summer of Code student)?

- Are you willing to maintain it going forward? (yes/no):
Yes

- Is there a relevant academic paper? (if so, where):
AutoAugment Reference: https://arxiv.org/abs/1805.09501
RandAugment Reference: https://arxiv.org/abs/1909.13719

- Is there already an implementation in another framework? (if so, where):
See link above; this would be a standard migration from `tf.contrib`.

- Was it part of tf.contrib? (if so, where):
Yes

**Which API type would this fall under (layer, metric, optimizer, etc.)**
Image

**Who will benefit with this feature?**
Anyone doing image preprocessing, especially for EfficientNet.

bhack:
For the Gsoc probably we need to think more generally about preprocessing.
We have still numpy based operations in https://keras.io/preprocessing/image/.
Also It needs to be investigated if there will be any integration between AutoAugment/Autokeras.

steph-en-m:
> For the Gsoc probably we need to think more generally about preprocessing.
> We have still numpy based operations in https://keras.io/preprocessing/image/.

@bhack can you please elaborate more on what you mean by this. Thanks

bhack:
Many images operations in Keras are still not in addons or in `tf.image` but numpy implemented:
https://github.com/keras-team/keras-preprocessing/tree/master/keras_preprocessing/image
Other image operations are going in [TF.IO](https://github.com/tensorflow/io/issues/825) or TF.graphics?
I think that we need to unify image processing so that eventually ops could benefit from the compiler stack soon or later (MLIR & friends).

[...]

PR #1275, opened 2020-03-09
Blend Image Op  

abhichou4:
Refers #1333 (#1226)
Adding blend op to `tensorflow_addons.image`


seanpmorgan:
> #1226
> For now, I've just added image ops. Here's a [notebook](https://colab.research.google.com/drive/1KjAFBG7uPMJQmmpXWj4BFpMgjuIVQ_p8) to make it easy to test them. I'm working on adding tests for `AutoAugment`, and adding `Hparams` for training from the tensor2tensor repo too. Please let me know if this is what's expected, and how this migration should be broken down into PRs.

Hi @abhichou4 thanks for starting this! So when we moved from tf.contrib we decided not to bring over HParams since it's a new API and that was against our charter. I'm wondering if we can utilize KerasTuner's Hyperparameter object since this will be maintained going forward without our help:
https://github.com/keras-team/keras-tuner/blob/master/kerastuner/engine/hyperparameters.py#L464

lgeiger:
Thanks for getting this started :+1:

>  I'm wondering if we can utilize KerasTuner's Hyperparameter object since this will be maintained going forward without our help

Since HParams are only used in very few places here, I think a normal `dict` or a small `dataclass` should do the job here, without needing an extra dependency or additional maintenance overhead.

tensorflow_addons/image/autoaugment.py
```diff
@@ -0,0 +1,644 @@
...
+    policy = [
+        [("TranslateX", 1.0, 4), ("Equalize", 1.0, 10)],
+    ]
+    return policy
+
+
+def blend(image1, image2, factor):
+    """Blend image1 and image2 using 'factor'.
+
+  Factor can be above 0.0.  A value of 0.0 means only image1 is used.
```
fsx950223: These ops could be separated into standalone image ops.

abhichou4:
> Hi @abhichou4 thanks for starting this! So when we moved from tf.contrib we decided not to bring over HParams since it's a new API and that was against our charter. I'm wondering if we can utilize KerasTuner's Hyperparameter object since this will be maintained going forward without our help:
> https://github.com/keras-team/keras-tuner/blob/master/kerastuner/engine/hyperparameters.py#L464

Looking forward to the discussion on the issue. I'll make the necessary changes.

tensorflow_addons/image/autoaugment.py
abhichou4: I thought so too. But referring to [this](https://github.com/tensorflow/addons/issues/1226#issuecomment-595469029), would `AutoAugment` benefit from keeping it in the same module? Please expand on what that comment meant.

bhack: Why single ops are not going in `tfa.image` namespace?

abhichou4: My bad I'll put them in separate modules.

abhichou4: @bhack can you please expand on how adding these ops to `tfa.image` would lead to better scheduling? I'm not aware how that works. How apparent would the overhead be if we were using keras preprocessing ops?

bhack: I meant i.e. `solarize` is useful in `tensorflow_addons.image` so it will be not exclusive to `AutoAugment` and could be reused.
I think that `Keras preprocessing` it is another topic and it needs to find a way to coordinate with `tfa.image` and  `tf.image` over the implemented op. 
I think it is not ideal to spread image processing operations everywhere.

abhichou4: I agree it is leading to a lot of boilerplate code. It would be great if Keras preprocessing could do that.

bhack:
> Hi @abhichou4 thanks for starting this! So when we moved from tf.contrib we decided not to bring over HParams since it's a new API and that was against our charter. I'm wondering if we can utilize KerasTuner's Hyperparameter object since this will be maintained going forward without our help:
> [https://github.com/keras-team/keras-tuner/blob/master/kerastuner/engine/hyperparameters.py#L4](https://github.com/keras-team/keras-tuner/blob/master/kerastuner/engine/hyperparameters.py#L464)

/cc @omalleyt12

tensorflow_addons/image/autoaugment.py
bhack: It Is better that you open and try to review/merge another PR to add this image ops in and image ops in AutoAugment variant [remixmatch/ctaugment](https://github.com/google-research/remixmatch/blob/master/libml/ctaugment.py) in `tensorflow_addons.image`

abhichou4: Sure, I'll just add one op in this PR, make new ones for other ops. Should make it easier to review and ultimately better code. Other discussions can be continued after ops are merged.

tensorflow_addons/image/augment_ops.py
```diff
@@ -0,0 +1,63 @@
...
+
+  Factor can be above 0.0.  A value of 0.0 means only image1 is used.
+  A value of 1.0 means only image2 is used.  A value between 0.0 and
+  1.0 means we linearly interpolate the pixel values between the two
+  images.  A value greater than 1.0 "extrapolates" the difference
+  between the two pixel values, and we clip the results to values
+  between 0 and 255.
+
+  Args:
+    image1: An image Tensor of type uint8.
```
gabrieldemarmiesse: This should be enforced in the function for better UX. Could you add a check and raise a `TypeError` with an error message if the user passes the wrong type?

tensorflow_addons/image/augment_ops.py
```diff
@@ -0,0 +1,63 @@
...
+        return tf.convert_to_tensor(image2)
+
+    image1 = tf.cast(image1, dtype=tf.dtypes.float32)
+    image2 = tf.cast(image2, dtype=tf.dtypes.float32)
+
+    difference = image2 - image1
+    scaled = factor * difference
+
+    # Do addition in float.
+    temp = tf.convert_to_tensor(image1, dtype=tf.dtypes.float32) + scaled
```
gabrieldemarmiesse: I have a doubt about the `tf.convert_to_tensor`, do we need it?

tensorflow_addons/image/augment_ops.py
```diff
@@ -0,0 +1,63 @@
...
+
+    # Interpolate
+    if factor > 0.0 and factor < 1.0:
+        # Interpolation means we always stay within 0 and 255.
+        return tf.cast(temp, tf.dtypes.uint8)
```
gabrieldemarmiesse: It would be nice to round the values before doing the cast, even if it's not in the reference implementation. Users expect image transformations to be done with floats and rounded to the nearest integer, like in other image manipulation libraries.

tensorflow_addons/image/augment_ops_test.py
```diff
@@ -0,0 +1,63 @@
...
+from tensorflow_addons.utils import test_utils
+
+_DTYPES = {
+    tf.dtypes.uint8,
+    tf.dtypes.int32,
+    tf.dtypes.int64,
+    tf.dtypes.float16,
+    tf.dtypes.float32,
+    tf.dtypes.float64,
+}
```
gabrieldemarmiesse: Let's focuse on uint8 at the moment. We can always do pull requests later to extend it to other data types, as it doesn't change the API.

tensorflow_addons/image/augment_ops_test.py
```diff
@@ -0,0 +1,63 @@
...
+            blended = augment_ops.blend(image1, image2, 0.5)
+            self.assertAllEqual(
+                self.evaluate(blended),
+                [
+                    [127, 127, 127, 127],
+                    [127, 127, 127, 127],
+                    [127, 127, 127, 127],
+                    [127, 127, 127, 127],
+                ],
+            )
```
gabrieldemarmiesse: Pinning some numbers is nice, it's definitly worth it. Could you also add a reference implementation in numpy, generate random arrays and blend factors and check that they are equivalent? Thanks a lot.

tensorflow_addons/image/BUILD
```diff
@@ -164,3 +165,12 @@ py_test(
...
 )
+
+py_test(
+    name = "augment_ops_test",
+    size = "medium",
+    srcs = [
+        "augment_ops_test.py",
+    ],
+    main = "augment_ops_test.py",
+)
```
gabrieldemarmiesse:
```suggestion
    main = "augment_ops_test.py",
    deps = [
        ":image",
    ],
)
```

tensorflow_addons/image/__init__.py
```diff
@@ -28,3 +28,4 @@
 from tensorflow_addons.image.transform_ops import rotate
 from tensorflow_addons.image.transform_ops import transform
 from tensorflow_addons.image.translate_ops import translate
+from tensorflow_addons.image.augment_ops import blend
```
bhack: @gabrieldemarmiesse Is blend an `augment_ops` or something else? I.e. in Opencv is a `addWeighted` operation so I don't know if could be in the trasnform group, an `arithmetic` etc. 
I we will add more blending operations we could use `image.blend_ops`.

tensorflow_addons/image/augment_ops.py
abhichou4: I wasn't sure if we expect this to work for all types. Let's make it uint8 for now. Another thing was to checking if it's compatible with `tf.data.Dataset`. Let me know if that's necessary right now.

tensorflow_addons/image/augment_ops.py
abhichou4: You're right. It's not needed since `scaled` is `tf.float32`

tensorflow_addons/image/augment_ops.py
abhichou4: Makes sense.

tensorflow_addons/image/BUILD
abhichou4: The build fails due to a syntax error in `configure.py` and hence then Bazel test too.

tensorflow_addons/image/__init__.py
bhack: Or we could use `image.compose_ios` semantically like [ImageMagick](http://www.imagemagick.org/Usage/compose/)

tensorflow_addons/image/BUILD
abhichou4: @gabrieldemarmiesse if you can help me with the failing Bazel Test. Seems like it's an import error. I've just used pytest to run tests locally. Working on setting up docker but it's taking too long.

gabrieldemarmiesse: Sure, I added the modifications to do in my first message. You need to add the `deps`:

```
    main = "augment_ops_test.py",
    deps = [
        ":image",
    ],
)
```

I know bazel can be quite fustrating. We need to keep it because googlers make heavy use of it.

tensorflow_addons/image/__init__.py
gabrieldemarmiesse: This is not critical, we can always refactor/rename files later on as this does not appear in the public API. Let's keep it like that now, we can change it later on.

bhack: Ok but it is better to give them a more appropriate default name. `compose` or `compositing` for blend is better then `augment_ops`.

tensorflow_addons/image/augment_ops_test.py
```diff
@@ -0,0 +1,79 @@
...
+                    [128, 128, 128, 128],
+                ],
+            )
+
+            image1 = np.random.randint(0, 255, (4, 4, 3), np.uint8)
+            image2 = np.random.randint(0, 255, (4, 4, 3), np.uint8)
+            blended = augment_ops.blend(
+                tf.convert_to_tensor(image1), tf.convert_to_tensor(image2), 0.35
+            )
+            self.assertAllEqual(self.evaluate(blended), blend_np(image1, image2, 0.35))
```
fsx950223: It could compare with test images which stores at ```tensorflow_addons/image/test_data```.

abhichou4: Yes, that would be neat. Thanks!

gabrieldemarmiesse:
@abhichou4 Thanks for fixing the build. Could you keep the tests that you made before too? We never have too many tests.

tensorflow_addons/image/augment_ops_test.py
gabrieldemarmiesse: I'm not fond of using images stored in git to check image ops. It has several drawbacks:

* It's hard to check in code review
* It increases the size of the repository, even if we delete it in another commit because it stays in the history forever.
* Git is not made to handle binary files
* It's hard to check visually that the added image is the one expected because pixels could be slightly off (in value or in position) and we would never notice.

For those reasons, I would advise against using images in git for tests unless absolutly necessary. An easy equivalent numpy function is usually the best to test against.

fsx950223: Good. I change to numpy too.

gabrieldemarmiesse:
@abhichou4 could you remove the test on the image in the git repo? Afterwards we're good to merge this.

gabrieldemarmiesse:
Thanks again for the pull request!

PR #1281, opened 2020-03-10
Better user experience when the version of TF is not right.

### LINKED ISSUE #1277: Segmentation fault when using tfa.rotate in tf-nightly
**System information**
- OS Platform: Linux Ubuntu 18.04.3 LTS
- TensorFlow version: tf-nightly (2.2.0-dev20200310) installed with pip
- TensorFlow-Addons version: 0.8.3 (pip) and tfa-nightly (0.9.0-dev; pip)
- Python version: 3.6.9
- Is GPU used?: yes

**Describe the bug**
I currently have to use tf-nightly because of a separate issue. When I use tfa.rotate I get a segmentation fault. Behavior is the same with tfa 0.8.3 and tfa-nightly.

```
import tensorflow as tf

img = tf.io.read_file(IMG_PATH)
img = tf.image.decode_png(img)
print(img.shape)
img = tfa.image.rotate(img, 90)
```
Output:     
> (128, 128, 3)
> Segmentation fault (core dumped)

I also tried `interpolation="BILINEAR"`. Problem is the same.

gabrieldemarmiesse:
Fix #1277

### DIFF
--- a/tensorflow_addons/utils/ensure_tf_install.py
+++ b/tensorflow_addons/utils/ensure_tf_install.py
@@ -28,7 +28,8 @@ warning_template = """
 This version of TensorFlow Addons requires TensorFlow {required}.
 Detected an installation of version {present}.
 
-While some functions might work, custom ops were not compiled
+While some functions might work, TensorFlow Addons was not tested 
+with this TensorFlow version. Also custom ops were not compiled
 against this version of TensorFlow. If you use custom ops,
 you might get errors (segmentation faults for example).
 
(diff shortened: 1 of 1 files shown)


tensorflow_addons/utils/ensure_tf_install.py
```diff
@@ -17,27 +17,40 @@
...
 # needs to happen before anything else, since the imports below will try to
 # import tensorflow, too.
+
+from distutils.version import LooseVersion
+import warnings
+
+import tensorflow as tf
+
+
+warning_template = """
```
seanpmorgan: Just curious why this was pulled out of the function? Not likely to be an issue but if we don't need to add a variable to global space then I don't see a reason to?

gabrieldemarmiesse: It's just easier to write as I don't have to break the indentation in the function to write the message. If you prefer that I put it in the function, I have nothing against that.

PR #1288, opened 2020-03-11
Alternative to run_all_in_graph_and_eager_mode

gabrieldemarmiesse:
So, replacing this `run_all_in_graph_and_eager_mode` is quite difficult but it's worth it for two reasons:
* It's not a public API, so it can go away at any time. The behavior can change silently.
* We don't need to execute the test methods in graph mode. We only need to execute a small part of the test in graph mode (for the layers, it's the `call` method, for the metrics/losses, it's the `update_state`...). It allows us to enjoy the eager mode outside those selected functions and we don't have to do `self.session` `self.evaluate(tf.compat.v1.initialize_global_variables(...))`. We can just do `.numpy()` to get results. This has the benefit to make testing much easier and bring new contributors who might be put off by this testing which looks like coding in tensorflow 1.x.
* It does not run `tf.function` in eager mode. So it's not actually helping us at all. I think it's just a TF 1.x relict and is useful only to force graph mode when tf.function is not present. See #1289 

So how to do that, in a clean way? Well that's more or less difficult.

From the discussion in #13 , all public functions should have the `@tf.functions` decorator. #807  shows that it's a problem when using python variables, but in this case, we should just use the `input_signature` to convert automatically scalars to tensorflow tensors. See https://www.tensorflow.org/api_docs/python/tf/function . We can even add tests to ensure we don't draw the graph multiple times with the method `get_concrete_function`.

We already use tf.function many time in the codebase, and even when we don't use it, keras enforce it with black magic when subclassing `Metric` or even `Layer`.

Let's then assume that all public functions have the `tf.function` decorator.

Here is my proposal. It uses pytest features and it won't work when subclassing tf.test.TestCase. But I believe we don't need subclassing anyway for most of the tests present in tf.addons. Worst case scenario we can code it again for `tf.test.TestCase` with the absl parametrize decorator.

`conftest.py` must be present when running the tests which means that we need to make bazel aware of it. This is why I linked it to `tensorflow_addons/utils` which is present all the time in the tests.

Some reference:
https://docs.pytest.org/en/latest/fixture.html

### DIFF
--- a/tensorflow_addons/conftest.py
+++ b/tensorflow_addons/conftest.py
@@ -1,16 +1,5 @@
-import tensorflow as tf
-import pytest
+from tensorflow_addons.utils.test_utils import maybe_run_functions_eagerly  # noqa: F401
 
-
-def finalizer():
-    tf.config.experimental_run_functions_eagerly(False)
-
-
-@pytest.fixture(scope="function", params=["eager_mode", "tf_function"])
-def maybe_run_functions_eagerly(request):
-    if request.param == "eager_mode":
-        tf.config.experimental_run_functions_eagerly(True)
-    elif request.param == "tf_function":
-        tf.config.experimental_run_functions_eagerly(False)
-
-    request.addfinalizer(finalizer)
+# fixtures present in this file will be available
+# when running tests and can be referenced with strings
+# https://docs.pytest.org/en/latest/fixture.html#conftest-py-sharing-fixture-functions
--- a/tensorflow_addons/conftest.py
+++ b/tensorflow_addons/conftest.py
@@ -1,0 +1,15 @@
+import tensorflow as tf
+import pytest
+
+
+@pytest.fixture(scope="function", params=["eager_mode", "tf_function"])
+def maybe_run_functions_eagerly(request):
+    if request.param == "eager_mode":
+        tf.config.experimental_run_functions_eagerly(True)
+    elif request.param == "tf_function":
+        tf.config.experimental_run_functions_eagerly(False)
+
+    def finalizer():
+        tf.config.experimental_run_functions_eagerly(False)
+
+    request.addfinalizer(finalizer)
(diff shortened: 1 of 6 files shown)


tensorflow_addons/conftest.py
```diff
@@ -0,0 +1,16 @@
+import tensorflow as tf
```
seanpmorgan: Any reason this wouldn't fit under:
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/utils/test_utils.py ?

tensorflow_addons/conftest.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def finalizer():
+    tf.config.experimental_run_functions_eagerly(False)
+
+
+@pytest.fixture(scope="function", params=["eager_mode", "tf_function"])
+def maybe_run_functions_eagerly(request):
+    if request.param == "eager_mode":
+        tf.config.experimental_run_functions_eagerly(True)
```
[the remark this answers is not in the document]
gabrieldemarmiesse: Could we communicate to the TF team that this flag is really needed on our side? So that it gets out of experimental mode? It's a much needed feature for debugging in general.

tensorflow_addons/conftest.py
gabrieldemarmiesse: We can do it, but then we would have an unused import in conftest.py (fixtures need to be in a conftest.py to be refered with strings in tests.) I don't have a strong preference here so I'll move it to test_utils.py

gabrieldemarmiesse:
> Checking the metrics module it says it'll run as graph but quite difficult for me to see where that is happening:
https://github.com/tensorflow/tensorflow/blob/v2.1.0/tensorflow/python/keras/metrics.py#L218
It's possible that this isn't being controlled with tf.function in which case I think the tests should fail with or without the experimental config so probably a non-issue.


https://github.com/tensorflow/tensorflow/blob/ac303a810d440d56033766651ddeffd3e2ee5f57/tensorflow/python/keras/metrics.py#L156

tensorflow_addons/conftest.py
gabrieldemarmiesse: Done

PR #1289, opened 2020-03-11
[WIP] Should break in eager mode.

gabrieldemarmiesse:
Do not merge. It's just a test. If all the tests are passing, we're not testing the eager mode.


AakashKumarNain:
> If all the tests are passing, we're not testing the eager mode.

Sorry, I didn't get this. We aren't testing eager mode?

gabrieldemarmiesse:
It's work in progress. I'm trying to get rid of `run_all_in_graph_and_eager_mode`. At the same time, I noticed that we were not testing some functions in eager mode. This pull request should not be merged. It's just to prove that if a bug happens only in eager mode, we won't know about it.

PR #1290, opened 2020-03-11
Removing benchmarks from the test suite

gabrieldemarmiesse:
We're not timing the ops anyway and I'm pretty sure we don't care about the numbers produced too. If we need to use those benchmarks later on, we can always bring them back to life from the git history.


fsx950223:
LGTM.Thanks

PR #1292, opened 2020-03-11
Support sample weighting in RSquare

harahu:
This PR implements sample weighting for the R<sup>2</sup> metric implementation `RSquare`. There could be some ambiguity as to what sample weighting should mean in the case of R<sup>2</sup>, but I felt that the most obvious interpretation is the same interpretation that has been chosen in the [scikit-learn implementation](https://github.com/scikit-learn/scikit-learn/blob/95d4f0841/sklearn/metrics/_regression.py#L488) of the same metric, and this is the interpretation I have implemented.

Also of note is a simplification of the result calculation. I can describe a proof of it, if need be, but the new calculation is mathematically equivalent with the old, as is indicated by the succeeding tests.

The expected value in the test that is added in this PR is derived from running the scikit-learn implementation with equivalent values.


harahu:
Here's the reasoning for the result calculation simplification, btw:
proof

gabrieldemarmiesse:
Thank you for the pull request! That's a very nice addition!

tensorflow_addons/metrics/r_square_test.py
```diff
@@ -81,6 +81,20 @@ def test_r2_random_score(self):
...
+        sample_weight = tf.constant([2, 5, 1, 8], dtype=tf.float32)
+        actuals = tf.cast(actuals, dtype=tf.float32)
+        preds = tf.cast(preds, dtype=tf.float32)
+        sample_weight = tf.cast(sample_weight, dtype=tf.float32)
+        # Initialize
+        r2_obj = self.initialize_vars()
+        # Update
+        self.update_obj_states(r2_obj, actuals, preds, sample_weight=sample_weight)
+        # Check results
+        self.check_results(r2_obj, -0.4431257)
```
gabrieldemarmiesse: That's a pretty nice move to align ourselves on scikit-learn on this. What would be nice it to run this test multiple times with random values and run the scikit-learn function to compare with your implementation. In this way we make sure we align with scikit-learn on all types of inputs and that it will stay that way in the future.

To be able to use scikit-learn in the tests, just add scikit-learn in this file: https://github.com/tensorflow/addons/blob/master/tools/install_deps/pytest.txt you can then re-run the tests with docker on cpu to check that everything works.

harahu: Very much agree that these are good ideas. Will update the PR.

harahu: This wasn't as straight forward as I thought. I noticed that this metric only agrees with sklearn for `y_true.shape == (x,)` or `y_true.shape == (x, 1)`, in other words it doesn't properly support multivariate regression. I feel like this is a separate issue, though, and should be dealt with in a separate PR. There are also some other subtle differences, like sklearn returning 0.0 instead of -inf in some cases: https://github.com/scikit-learn/scikit-learn/blob/95d4f0841/sklearn/metrics/_regression.py#L604

I thought the test case you described of running the test multiple times with random values was a good fit for using [hypothesis](https://hypothesis.readthedocs.io/en/latest/), so I played around with that. This can be replaced with another way of generating test data, though.

harahu: Also noticing that hypothesis broke the build, so I might have to reconsider.

gabrieldemarmiesse: > I feel like this is a separate issue, though, and should be dealt with in a separate PR. 

Sure thing. Let's keep this pull request small and focused. 

Can you try to make it work without hypothesis? I'm not against using it, but we have many changes for the test suite planned in the coming weeks, so it should be easier if there are less dependencies. I'll merge it as soon as the tests are passing when comparing to the scikit learn implementation.

Also could you open an issue concerning the discrepancy between scikit-learn and tf addons for the arrays of dimension 2? This will be easier to track the progress on this bug after this pull request is merged. @SSaishruthi can then look into it whenever he/she has the time.

harahu: I do have some plans for how to make it work for vector labels, and intend to implement a fix. There are some design choices in that regard that I might want a discussion with you (I guess both you and @SSaishruthi). An issue might be a good place to do that. You can expect one from me soon.

Will definitely make it work without hypothesis. It is a powerful testing library, but it likely shouldn't be pulled in as an afterthought. It also means I don't have to figure out why it failed and how to fix it 😅.

harahu:
@gabrieldemarmiesse I removed hypothesis and did a numpy only version of the test. The check failure here seems to be due to network issues. Any advice?

gabrieldemarmiesse:
Sorry about that. It happens sometimes with github actions, I'll restart the tests and take a look.

gabrieldemarmiesse:
@harahu by curiosity, after we fixed the bug that you had with the `bash tools/run_cpu_tests.sh`, what kind of workflow did you use? Local install? Running the tests only with docker? Did you have trouble with the pre-commit? What IDE/text editor do you use? What OS do you use? What is easy overall to work with addons? It helps us understand how we can make the contributions to this repo easier.

harahu:
> @harahu out of curiosity, after we fixed the bug that you had with the `bash tools/run_cpu_tests.sh`, what kind of workflow did you use? Local install? Running the tests only with docker? 

Trying to keep things simple I initially ran tests solely through docker. I felt that was fairly slow and tedious, though. Particularly running a full test suite when your changes are fairly localized. To speed up my workflow I ended up going for a local editable install in a virtual environment (I use pyenv to manage both python versions and environments), so that I could more easily run selected tests and also do debugging.

> Did you have trouble with the pre-commit? What IDE/text editor do you use? What OS do you use?

I never actually installed the pre-commit script as a hook. I just run it manually before I push. It worked fine for me with no problems. I use black in other projects so I am pretty used to it. I usually develop in Pycharm (using the nice pytest integration), although I sometimes use vim. As for OS, I switch between macOS and Ubuntu.

> What is easy overall to work with addons? Was it your first pull request in GitHub? It helps us understand how we can make the contributions to this repo easier.

It was fairly straight forward, I felt. I do work with python on a daily basis, though, so I might be a bit blind to barriers of entry that others might notice. This was not my very first pull request, but I haven't had many.

One slight annoyance was how necessary development dependencies are split in multiple files. It would be great if there was a single command I could run in order to install all the dependencies I need.

gabrieldemarmiesse:
Thanks for the comments! It helps us a lot. I'll try to think about how we can streamline all that, especially installing the test dependencies locally (usually, projects use `pip install -e .[tests]`, I feel it's a good idea).

PR #1295, opened 2020-03-13
Add testing module, with an function `check_metric_serialization`

gabrieldemarmiesse:
This is a function that test the serialization of a metric object. Multiple tests are run to determine if the new metric object is a perfect copy of the first.

I made a demo on CohenKappa, and this should be applied to all our metrics. This will strenghten the test suite.

It's not declared in the `__init__.py` because it lacks a docstring, and, as such, should not be public yet. The docstring can be added in another pull request.

Some doc:

* metric: The metric object to serialize and copy.
* y_true: Tuple or numpy array. Use this numpy array when calling .update_state() on both the original metric and the copy. If tuple, a random float32 numpy array is created with its shape being the tuple values
* y_pred: Tuple or numpy array. Use this numpy array when calling .update_state() on both the original metric and the copy. If tuple, a random float32 numpy array is created with its shape being the tuple values
* sample_weights: (optional) Tuple or numpy array. Use this numpy array when calling .update_state() on both the original metric and the copy. If tuple, a random float32 numpy array is created with its shape being the tuple values.
* strict: raise an error if in the constructor, `*args` or `**kwargs` is used.


tensorflow_addons/testing/BUILD
```diff
@@ -0,0 +1,23 @@
...
+    name = "test_serialization",
+    size = "small",
+    srcs = [
+        "test_serialization.py",
+    ],
+    main = "test_serialization.py",
+    deps = [
+        ":testing",
+    ],
+)
```
seanpmorgan, line 23: Please update this to use the new glob strategy.

gabrieldemarmiesse: Yeah, totally forgot. I'm on it.

AakashKumarNain:
Really nice. Thanks

PR #1310, opened 2020-03-14
Multivariate r2

harahu:
I noticed that the `RSquared` metric does not behave in the same way as the `scikit-learn` implementation of the same metric for multivariate regression (in `scikit-learn` known as "multioutput"). First I thought I should mend it to align with the `scikit-learn` implementation that calculates separate scores for each label element and aggregates the scores. Then I started asking myself whether there might be other ways of interpreting R<sup>2</sup> in higher dimensions. Looking at the definition presented in https://en.wikipedia.org/wiki/Coefficient_of_determination, it does look like it might make sense to extend it to higher rank data types, but this can be done in different ways, and I am unsure what would make sense mathematically. The `scikit-learn` approach at least makes some sense to me as it just treats the model as a model for n separate dependent variables at the same time.

I would really like a discussion on how R<sup>2</sup> should be understood in higher dimensions, but if we are to go with the `scikit-learn` interpretation, this PR is one way to go about it. It is slightly annoying that you have to specify the shape of y when initializing the metric object. This could be solved by dynamically setting the shape like it is done in the [MeanTensor metric](https://github.com/tensorflow/tensorflow/blob/v2.1.0/tensorflow/python/keras/metrics.py#L2618-L2628). Although more user friendly, I don't think the code is as clean, so there is a trade-off.

I modified a test to show how this acts for higher dimensional data, currently matching the `multioutput="raw_values"` flavour of the `scikit-learn` implementation.

### DIFF
--- a/tensorflow_addons/metrics/r_square.py
+++ b/tensorflow_addons/metrics/r_square.py
@@ -32,7 +32,7 @@ def _reduce_average(
     """Computes the (weighted) mean of elements across dimensions of a tensor.
   """
     if weights is None:
-        return tf.reduce_mean(input_tensor, axis=None, keepdims=False)
+        return tf.reduce_mean(input_tensor, axis=axis, keepdims=keepdims)
 
     weighted_sum = tf.reduce_sum(weights * input_tensor, axis=axis, keepdims=keepdims)
     sum_of_weights = tf.reduce_sum(weights, axis=axis, keepdims=keepdims)
--- a/tensorflow_addons/metrics/r_square.py
+++ b/tensorflow_addons/metrics/r_square.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 # ==============================================================================
 """Implements R^2 scores."""
+from typing import Tuple
 
 import tensorflow as tf
 from tensorflow.keras.metrics import Metric
@@ -53,15 +54,28 @@ class RSquare(Metric):
 
     @typechecked
     def __init__(
-        self, name: str = "r_square", dtype: AcceptableDTypes = None, **kwargs
+        self,
+        name: str = "r_square",
+        dtype: AcceptableDTypes = None,
+        y_shape: Tuple[int, ...] = (),
+        **kwargs
     ):
         super().__init__(name=name, dtype=dtype, **kwargs)
-        self.squared_sum = self.add_weight("squared_sum", initializer="zeros")
-        self.sum = self.add_weight("sum", initializer="zeros")
-        self.res = self.add_weight("residual", initializer="zeros")
-        self.count = self.add_weight("count", initializer="zeros")
+        self.y_shape = y_shape
+        self.squared_sum = self.add_weight(
+            name="squared_sum", shape=y_shape, initializer="zeros", dtype=dtype
+        )
+        self.sum = self.add_weight(
+            name="sum", shape=y_shape, initializer="zeros", dtype=dtype
+        )
+        self.res = self.add_weight(
+            name="residual", shape=y_shape, initializer="zeros", dtype=dtype
+        )
+        self.count = self.add_weight(
+            name="count", shape=y_shape, initializer="zeros", dtype=dtype
+        )
 
-    def update_state(self, y_true, y_pred, sample_weight=None):
+    def update_state(self, y_true, y_pred, sample_weight=None) -> None:
         y_true = tf.cast(y_true, self._dtype)
         y_pred = tf.cast(y_pred, self._dtype)
         if sample_weight is None:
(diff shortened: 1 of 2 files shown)


harahu:
This PR is incomplete, though. It is really just an idea proposal, so please don't merge before we work out the details.

gabrieldemarmiesse:
Thanks a lot @harahu for this pull request. Mutiple things come to mind:

* I'll add WIP to the title of your pull request to prevent merging
* I'm no expert in RSquared, but if we decide to go against what was decided for scikit-learn, we need a very strong reason. Often consistency if better than a small design improvement (priciple of least astonishment), especially concerning numerical libraries, where behavior changes go unoticed.
* Concerning specifying the shape of y, I am faily certain that Metric, like Layer, will support a `build` method at some point. There is no technical reason why it wouldn't be possible, it's a better UX, and it's been shown to be needed. The question is what do we do in the meantime. We have two possibilities as you stated, either, a better UX, but using a private TF API, either a worse UX, but we don't use a private API. I'm in favor of your current implementation. It's a worse UX for a small number of users, but in the end it's temporary. On our side, it makes our life a lot easier. Using private APIs means that the behavior can change silently, which is really bad and can break the CI or worse, user code without us knowing about it. In the end, we can always deprecate the `y_shape` when the `build` method is officially supported, with a warning and everything, so changing the API is not an issue.

harahu:
Thank you for your thoughts on this @gabrieldemarmiesse! I definitely agree that there should be a high threshold for breaking with the 'scikit-learn' behaviour.

I did ask a question regarding this on [Cross Validated](https://stats.stackexchange.com/questions/454176/is-the-coefficient-of-determination-defined-for-higher-dimensional-dependent-var), hoping to get some feedback. I really am no expert myself.

harahu:
Did not get much response on Cross Validated, but I did discuss this with my colleague @ttvand . The only generalization of R<sup>2</sup> we felt comfortable claiming made sense was where you calculate individual scores per label element. This corresponds with the `multioutput="raw_values"` flavour in in `scikit-learn`. Although I would like that as the default setting in a utopia, I did implement the other supported options from `scikit-learn`, and used the same default `"uniform_average"` to maintain as similar behaviour as possible. Also tested that behaviour remains in line with `scikit-learn` for these different options.

Any further feedback on this, @gabrieldemarmiesse ?

gabrieldemarmiesse:
I believe you made the right call here. We don't need to think about it too much if we just copy what scikit learn does and make tests to ensure the results are the same for the three values. Users will pick whatever they want from what is available. I think that's perfect and it also has the nice property that we (I)  don't need to check the math to merge since scikit-learn does it for us (the review is going to be much faster). Give me some time to review on the technical side of things. 

Thanks again for all your work!

tensorflow_addons/metrics/r_square.py
```diff
@@ -70,23 +107,39 @@ def update_state(self, y_true, y_pred, sample_weight=None):
...
-        self.sum.assign_add(tf.reduce_sum(weighted_y_true))
-        self.squared_sum.assign_add(tf.reduce_sum(tf.multiply(y_true, weighted_y_true)))
+        self.sum.assign_add(tf.reduce_sum(weighted_y_true, axis=0))
+        self.squared_sum.assign_add(
+            tf.reduce_sum(tf.multiply(y_true, weighted_y_true), axis=0)
+        )
         self.res.assign_add(
             tf.reduce_sum(
-                tf.multiply(tf.square(tf.subtract(y_true, y_pred)), sample_weight)
+                tf.multiply(tf.square(tf.subtract(y_true, y_pred)), sample_weight),
```
gabrieldemarmiesse: Same here. would something like `((y_true - y_pred)**2) * sample_weight` work?

tensorflow_addons/metrics/r_square.py
```diff
@@ -70,23 +107,39 @@ def update_state(self, y_true, y_pred, sample_weight=None):
...
+            )
 
-    def reset_states(self):
+    def reset_states(self) -> None:
         # The state of the metric will be reset at the start of each epoch.
-        self.squared_sum.assign(0.0)
-        self.sum.assign(0.0)
-        self.res.assign(0.0)
-        self.count.assign(0.0)
+        self.squared_sum.assign(tf.zeros(shape=self.y_shape, dtype=self.dtype))
```
gabrieldemarmiesse: Is this necessary? I would have expected tensorflow to gracefully broadcast the `0.0`. I'm disapointed if TF doesn't do it automatically.

harahu: Just me being conservative. Reverted.

tensorflow_addons/metrics/r_square.py
harahu: It seems it does.

tensorflow_addons/metrics/r_square.py
```diff
@@ -70,23 +107,39 @@ def update_state(self, y_true, y_pred, sample_weight=None):
         sample_weight = weights_broadcast_ops.broadcast_weights(sample_weight, y_true)
 
         weighted_y_true = tf.multiply(y_true, sample_weight)
-        self.sum.assign_add(tf.reduce_sum(weighted_y_true))
-        self.squared_sum.assign_add(tf.reduce_sum(tf.multiply(y_true, weighted_y_true)))
+        self.sum.assign_add(tf.reduce_sum(weighted_y_true, axis=0))
+        self.squared_sum.assign_add(
+            tf.reduce_sum(tf.multiply(y_true, weighted_y_true), axis=0)
```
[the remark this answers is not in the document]
harahu: Seems that way.

tensorflow_addons/metrics/r_square.py
```diff
@@ -53,15 +68,37 @@ class RSquare(Metric):
...
+            name="squared_sum", shape=y_shape, initializer="zeros", dtype=dtype
+        )
+        self.sum = self.add_weight(
+            name="sum", shape=y_shape, initializer="zeros", dtype=dtype
+        )
+        self.res = self.add_weight(
+            name="residual", shape=y_shape, initializer="zeros", dtype=dtype
+        )
+        self.count = self.add_weight(
+            name="count", shape=y_shape, initializer="zeros", dtype=dtype
```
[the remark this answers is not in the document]
harahu: I though I was going to sneak it in ;) Good eyes!

tensorflow_addons/metrics/r_square.py
```diff
@@ -22,6 +23,20 @@
...
+
+def _reduce_average(
+    input_tensor: tf.Tensor, axis=None, keepdims=False, weights=None
+) -> tf.Tensor:
+    """Computes the (weighted) mean of elements across dimensions of a tensor.
+  """
+    return tf.divide(
+        tf.reduce_sum(tf.multiply(weights, input_tensor), axis=axis, keepdims=keepdims),
+        tf.reduce_sum(weights, axis=axis, keepdims=keepdims),
+    )
```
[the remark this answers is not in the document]
harahu: I think this function reads better now. What do you think?

Me using tf ops everywhere was just conservatism. Seems to work fine with the normal python operators.

PR #1311, opened 2020-03-14
Refactoring configure.py

gabrieldemarmiesse:
Grouping the logic into functions for better readability.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -192,6 +192,7 @@ def get_cuda_version():
         cuda_version = answer or default
     print("> Using CUDA version:", cuda_version)
     print()
+    return cuda_version
 
 
 def get_cudnn_version():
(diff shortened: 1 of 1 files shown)


configure.py
```diff
@@ -26,8 +26,6 @@
 import sys
 import logging
 
-_DEFAULT_CUDA_VERISON = "10.1"
-_DEFAULT_CUDNN_VERSION = "7"
```
seanpmorgan: Very much like the new modular functions, but I'm wondering if removing these 2 from global actually hurts the readability / maintainability. I'm thinking when we update to 10.2+ into the future it's logical to just change variables at the top of the file instead of reading through the whole code to find variables that need updated.

If you have a strong preference here, I'm not tied to that opinion, but my thoughts.

gabrieldemarmiesse: Mhhh. True. The best would actually be to fetch those information from tensorflow and not ask the user at all. But I don't know how to do that.

PR #1319, opened 2020-03-16
Fixing the master branch.

gabrieldemarmiesse:
We need to clone a specific commit to avoid making our CI depend on a moving target. Even when using a hack like in this case.

Hopefully one day we can get rid of this hack. I wish there was a bazel command like `bazel download-jdk`


gabrieldemarmiesse:
Let's do both to make sure. I wish there was a easier way to downloas the jdk to avoid this hack.

PR #1322, opened 2020-03-17
Update cohens kappa maintainer

seanpmorgan:
Hi @AakashKumarNain not sure how this was missed when we updated the CODEOWNERS. Want to confirm that you're still willing to maintain?


AakashKumarNain:
@seanpmorgan yes, I will be maintaining it.

gabrieldemarmiesse:
Should we add a check in the CI to ensure that all files in `tensorflow_addons/` have an individual codeowner?

Issue #1320, opened 2020-03-17
Fix Addons compatibility with TF2.2


seanpmorgan:
**System information**
- TensorFlow version and how it was installed (source or binary): TF2.2

**Describe the bug**
As we prepare for the upcoming TF2.2 release we need to fix a number of bugs that have surfaced when linking with the new version. It may be better to break these out into individual issues, but wanted to make a placeholder issue in the mean time.

See https://github.com/tensorflow/addons/pull/1303/files

gabrieldemarmiesse:
All the bugs are fixed! We're now compatible with TF 2.2. What's left now is to change our CI to build on all OS against tf 2.1 and 2.2.

PR #1323, opened 2020-03-17
fix giou loss test

fsx950223:
Related #1320.
As [Reduction.NONE](https://www.tensorflow.org/versions/r2.2/api_docs/python/tf/keras/losses/Reduction) has changed. Change reduction mode to Reduction.SUM_OVER_BATCH_SIZE for unifying graph and eager reduction behavior.


gabrieldemarmiesse:

PR #1326, opened 2020-03-17
Grammar fix

sharvil:
"bare with us" means something else entirely


sharvil:
Oops. Not signing the CLA, so dropping this patch. Meanwhile, someone else might want to do a quick fix.

PR #1327, opened 2020-03-18
Using the pytest fixture rather than run_all_in_keras_and_eager_mode in cohen_kappa

gabrieldemarmiesse:
Nothing changed except the fixtures and going from method to function.

### DIFF
--- a/tensorflow_addons/metrics/cohens_kappa_test.py
+++ b/tensorflow_addons/metrics/cohens_kappa_test.py
@@ -195,31 +195,35 @@ class CohenKappaTest(tf.test.TestCase):
 
         model.fit(x, y, epochs=1, verbose=0, batch_size=32)
 
-    def test_keras_binary_clasasification_model(self):
-        kp = CohenKappa(num_classes=2)
-        inputs = tf.keras.layers.Input(shape=(10,))
-        outputs = tf.keras.layers.Dense(1, activation="sigmoid")(inputs)
-        model = tf.keras.models.Model(inputs, outputs)
-        model.compile(optimizer="sgd", loss="binary_crossentropy", metrics=[kp])
 
-        x = np.random.rand(1000, 10).astype(np.float32)
-        y = np.random.randint(2, size=(1000, 1)).astype(np.float32)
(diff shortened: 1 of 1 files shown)


tensorflow_addons/metrics/cohens_kappa_test.py
```diff
@@ -195,31 +195,35 @@ def test_keras_multiclass_reg_model(self):
...
-        model = tf.keras.models.Model(inputs, outputs)
-        model.compile(optimizer="sgd", loss="categorical_crossentropy", metrics=[kp])
+    model.fit(x, y, epochs=1, verbose=0, batch_size=32)
 
-        x = np.random.rand(1000, 10).astype(np.float32)
-        y = np.random.randint(5, size=(1000,)).astype(np.float32)
-        y = tf.keras.utils.to_categorical(y, num_classes=5)
 
-        model.fit(x, y, epochs=1, verbose=0, batch_size=32)
+@pytest.mark.xfail(tf.__version__ == "2.2.0-rc0", reason="TODO: Fix this test")
```
AakashKumarNain: Can you elaborate this on this a bit?

AakashKumarNain:
Thanks. Looks good. Just need a bit of clarification on one of the decorators.

tensorflow_addons/metrics/cohens_kappa_test.py
gabrieldemarmiesse: The decorator was already there before. It was added in #1303 . Note that this doesn't mean there is a bug on our side, it may be a bug on the tensorflow side :)

gabrieldemarmiesse: The decorator was already there before. It was added in #1303 . Note that this doesn't mean there is a bug on our side, it may be a bug on the tensorflow side :)

AakashKumarNain: Cool. Thanks

gabrieldemarmiesse:
Thank you for the review!

AakashKumarNain:
You are welcome.

Issue #1324, opened 2020-03-17
Make LossFunctionWrapper public


fsx950223:
**Describe the feature and the current behavior/state.**
We just wrapper loss function by ourselves and public ```LossFunctionWrapper``` will unify the behavior.
**Relevant information**
- Are you willing to contribute it (yes/no): Yes
- Are you willing to maintain it going forward? (yes/no): yes
- Is there a relevant academic paper? (if so, where):
- Is there already an implementation in another framework? (if so, where):Yes
- Was it part of tf.contrib? (if so, where):

**Which API type would this fall under (layer, metric, optimizer, etc.)**
Losses
**Who will benefit with this feature?**
Developers.
**Any other info.**
For example:
```
@tf.keras.utils.register_keras_serializable(package="Addons")
class GIoULoss(LossFunctionWrapper):
    """Implements the GIoU loss function.

    GIoU loss was first introduced in the
    [Generalized Intersection over Union:
    A Metric and A Loss for Bounding Box Regression]
    (https://giou.stanford.edu/GIoU.pdf).
    GIoU is an enhancement for models which use IoU in object detection.

    Usage:

    ```python
    gl = tfa.losses.GIoULoss()
    boxes1 = tf.constant([[4.0, 3.0, 7.0, 5.0], [5.0, 6.0, 10.0, 7.0]])
    boxes2 = tf.constant([[3.0, 4.0, 6.0, 8.0], [14.0, 14.0, 15.0, 15.0]])
    loss = gl(boxes1, boxes2)
    print('Loss: ', loss.numpy())  # Loss: [1.07500000298023224, 1.9333333373069763]
    ```
    Usage with tf.keras API:

    ```python
    model = tf.keras.Model(inputs, outputs)
    model.compile('sgd', loss=tfa.losses.GIoULoss())
    ```

    Args:
      mode: one of ['giou', 'iou'], decided to calculate GIoU or IoU loss.
    """

    @typechecked
    def __init__(
        self,
        mode: str = "giou",
        reduction: str = tf.keras.losses.Reduction.AUTO,
        name: Optional[str] = "giou_loss",
    ):
        super(GIoULoss, self).__init__(
            giou_loss, name=name, reduction=reduction, mode=mode)
```

gabrieldemarmiesse:
I believe that's a good idea. Contributions welcome. It seems our implementation is different from the TF one. It might be worth looking into the difference.

PR #1329, opened 2020-03-18
use loss wrapper

fsx950223:
cc @gabrieldemarmiesse .
Related #1324.

### DIFF
--- a/tensorflow_addons/losses/giou_loss.py
+++ b/tensorflow_addons/losses/giou_loss.py
@@ -15,14 +15,14 @@
 """Implements GIoU loss."""
 
 import tensorflow as tf
-
+from tensorflow.python.keras.losses import LossFunctionWrapper
 from tensorflow_addons.utils.types import TensorLike
 from typing import Optional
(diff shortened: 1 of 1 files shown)


tensorflow_addons/losses/giou_loss.py
```diff
@@ -58,18 +58,9 @@ def __init__(
...
-        self.mode = mode
-
-    def get_config(self):
-        base_config = super().get_config()
-        base_config["mode"] = self.mode
-        return base_config
-
-    def call(self, y_true, y_pred):
-        return giou_loss(y_true, y_pred, mode=self.mode)
+        super(GIoULoss, self).__init__(
```
gabrieldemarmiesse:
```suggestion
        super().__init__(
```

Adding arguments to the `super` function is only needed in python 2.

fsx950223: I remove them later.

gabrieldemarmiesse:

Issue #1328, opened 2020-03-18
[Easy and help welcome] Migrating all tests from run_all_in_graph_and_eager_mode to the new pytest fixture.


gabrieldemarmiesse:
This issue is to track the progression. Help is welcome, we have a lot of work to do but it's fairly easy. You can take example on the pull requests made before.

The idea is that before, using the `@run_all_in_graph_and_eager_mode` meant that we needed to make the tests with the tensorflow 1.x style, with  things like `tf.compat.v1.initialize_global_variables`, `self.evaluate`and other scary functions like that.

Now we use the `@pytest.mark.usefixtures("maybe_run_functions_eagerly")` which will run the test function twice. Once normally, and once with `tf.config.experimental_run_functions_eagerly(True)`. Which means the tests are eager-first. No need to initialize variables, get the result with `.numpy()`. To test the values, use [the numpy testing module](https://docs.scipy.org/doc/numpy/reference/routines.testing.html) instead of the methods of `tf.test.TestCase` like `self.AssertAllClose` and such. 

You can take as reference the two pull request already made:
#1288
#1327

When doing a pull request, please do not do more than 2 tests at once.

The policy is that when we're testing a simple function (for example, using a custom op), no need to use `tf.functions` and no need to use `@pytest.mark.usefixtures("maybe_run_functions_eagerly")` because we're not afraid of some python side effects.
When working with complex functions (for loops, if/else with tensors...) we need to add tf.function and to add the `@pytest.mark.usefixtures("maybe_run_functions_eagerly")` to make sure we don't rely on some pythonic behavior. To avoid having to make complex `input_signature` in `tf.function` we can isolate the sensitive part (if/else/for loop with tensors) in a separate `tf.function` and not decorate the main one.

fsx950223:
I suggest use ```tf.autograph.to_graph``` instead of ```tf.funciotn``` when working with complex functions. There are two reasons:
1.The function will raise errors when I set ```autograph=False```
2.Subgraph will use more resources

gabrieldemarmiesse:
@fsx950223 you're right we need more discussion about when and how to use tf.function. We may sometimes have complexe signatures (when multiple dtypes are accepted for exemple) and it's hard to express with `tf.function` and `tf.autograph.to_graph`. We still need to test the graph mode though (not all the time). So we need to find a solution.

steph-en-m:
[...]

PR #1332, opened 2020-03-18
Moved test_fit_simple_linear_model function out of run_all_in_graph_and_eager_mode.

gabrieldemarmiesse:
I didn't add the fixture to test in eager mode because running this test in eager mode takes 40seconds compared to 5 seconds with tf.function.

See #1328


shreyashpatodia:
Did you just move the code here?

gabrieldemarmiesse:
I took it out of the class to avoid making it run twice in graph mode. We're trying to get rid of the `@run_all_in_graph_and_eager_mode`

shreyashpatodia:
Ah okay. Thanks!

gabrieldemarmiesse:
Thank you for the review!

gabrieldemarmiesse:
Thanks for the reviews @Squadrick, much appreciated!

PR #1334, opened 2020-03-18
Improved LAMB optimizer.

spidyDev:
Using Fused Adam with weight decay kernel.

- Observed ****~100ms** per step** savings while training BERT(large) on single GPU.

### DIFF
--- a/tensorflow_addons/custom_ops/optimizers/BUILD
+++ b/tensorflow_addons/custom_ops/optimizers/BUILD
@@ -1,0 +1,18 @@
+licenses(["notice"])  # Apache 2.0
+
+package(default_visibility = ["//visibility:public"])
+
+load("//tensorflow_addons:tensorflow_addons.bzl", "custom_op_library")
+
+custom_op_library(
+    name = "_optimizer_ops.so",
+    srcs = [
+        "cc/kernels/adam_weightdecay_op.cc",
+        "cc/kernels/adam_weightdecay_op.h",
+        "cc/ops/adam_weightdecay_op.cc",
+    ],
+    cuda_srcs = [
+        "cc/kernels/adam_weightdecay_op.h",
+        "cc/kernels/adam_weightdecay_op_gpu.cu.cc",
+    ],
+)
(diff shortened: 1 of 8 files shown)


gabrieldemarmiesse:
@spidyDev thank you for the pull request! you might want to modify `tools/install_so_files.py` to fix the pytest error.

Also could you provide the speedup in percent of total time? For the CPU and the GPU version? With eager, no eager and XLA? See #1156 

Would it be possible to keep the python version too? See [how we do for activations](https://github.com/tensorflow/addons/blob/69be560e5dd9071669fd5c4fb487cf3973ce876d/tensorflow_addons/activations/hardshrink.py#L45). It's to be able to run the optimizer on TPUs and ROCm. It's also useful because when we compile things against a specific tensorflow version and we then cannot use it on any other tensorflow version. For example, people compiling TF from source won't be able to use your implementation. 

As you can see, using a custom op for something in tensorflow addons has a high UX and maintainance cost. We need to make sure the speedup is worth it before accepting to maintain a custom op.

spidyDev:
@gabrieldemarmiesse  Thanks for all the pointers. Will get back with more info/updates.

spidyDev:
@gabrieldemarmiesse  added some number in the description. Thnx !!

gabrieldemarmiesse:
@tensorflow/sig-addons-maintainers  are the speedup significant enough for merging? Even if we merge, is it worth it to have a C++ version? As only a minority of people in their sane mind would train a model on CPU and expect maximum speed.

gabrieldemarmiesse:
@spidyDev , in the end, we'll keep the proposed CUDA implementation, but we'll prefer to keep the pure python implementation for CPU. @Squadrick proposed himself to review your pull request.

Squadrick:
`training_op_helpers.*` aren't public in TF. Created a PR for making them public: https://github.com/tensorflow/tensorflow/pull/37873. 

This PR is blocked until that is resolved.

spidyDev:
@gabrieldemarmiesse @Squadrick Quick question about the comment "we'll prefer to keep the pure python implementation for CPU". Can you give some pointers how do we do this ? AFAIK , if we do not register a CPU kernel, TF is going to throw error:"No kernel registered"

gabrieldemarmiesse:
@spidyDev , in theory, we'd like to do that but I have no idea how to do it in practice. I hope it's possible. We would need it too for activation functions.

bhack:
@spidyDev Thank for your contribution.

@tomerk As this optimizer is [used by the Tensorflow model SIG](https://github.com/tensorflow/models/blob/master/official/modeling/optimization/optimizer_factory.py#L32), if they are not interested to review these changes I will close the PR.
It is 1 year old and we don't have currently the bandwidth to maintain another custom ops.

bhack:
Thanks you for the contribution. I suggest to move this PR to https://github.com/keras-team/keras-nlp

PR #1335, opened 2020-03-18
Function may be static in optical_flow_test.py

gabrieldemarmiesse:
This is to progressively get rid of the `@run_all_in_graph_and_eager_mode`. I try no to take everything out of the class at once to avoid a big diff.

See #1328


gabrieldemarmiesse:
Thank you for the review!

Issue #1333, opened 2020-03-18
[1/2] Add Image Ops for Data Augmentation 


abhichou4:
Part 2: https://github.com/tensorflow/addons/issues/1353

As part of #1226, relevant image ops need to be added to `tfa.image`. List of ops tracked in this issue (tick the op only once the corresponding PR is *merged*):

- [x] Blend 
- [x] CutOut 
- [ ] Identity
- [x] Translate X, Y
- [x] Shear X, Y
- [ ] Solarize Ops 
- [ ] Posterize
- [ ] AutoContrast
- [x] Equalize 
- [ ] Invert  
- [ ] Color
- [ ] Brightness 
- [x] Sharpness 

Reference: https://github.com/tensorflow/tpu/blob/master/models/official/efficientnet/autoaugment.py

vinayvr11:
Ok So these unchecked options are remaining. But what is PIL does it stands for Pillow or something else. And i have to implement these in autoaugment.py am i right.

abhichou4:
> Ok So these unchecked options are remaining. But what is PIL does it stands for Pillow or something else. And i have to implement these in autoaugment.py am i right.
That's right. The appropriate namescapes and where to place these operations is something that can be discussed here or on your PR. 
/cc @bhack @gabrieldemarmiesse

vinayvr11:
Ok so for now i just need to make a script for these operations and after that we decide that where to put them all. So all these operations are have to be integrated in one script or i have to make a different scripts for all of them.
Thank you

gabrieldemarmiesse:
Yeah, as long as the public API stays the same and is decided first, it's not critical how you name your files.

vinayvr11:
@gabrieldemarmiesse : I have some doubts so we have to use the tensorflow for doing these operations or we are free to use any module. 
And all these methods have to be made in different files. Please correct me if i am wrong.

gabrieldemarmiesse:
You should always use tensorflow to make those ops (so they can run on tpu/cpu/cuda/ROCm and can be exported in a graph and have gradients and many other things... ) and where you put the ops is not critical (meaning we care more about the pull request being merged than about the name of the file, as it can be changed later without impacting the API). You can group ops together in the same file if you feel this is the right move.

Prefer using python only if possible.

vinayvr11:
Ok and what about Tensorflow version shall i need to take care about the version also or i can use any version of Tensorflow.

gabrieldemarmiesse:
[...]

PR #1338, opened 2020-03-19
add cutout image op

fsx950223:
Related #1333

### DIFF
--- a/tensorflow_addons/image/cutout_ops.py
+++ b/tensorflow_addons/image/cutout_ops.py
@@ -18,7 +18,6 @@ import tensorflow as tf
 from tensorflow_addons.utils.types import TensorLike, Number
 from tensorflow_addons.image.utils import from_4D_image, to_4D_image
 from tensorflow.python.keras.utils import conv_utils
-import numpy as np
 
 
 def _get_image_wh(images, data_format):
@@ -67,7 +66,7 @@ def random_cutout(
     Returns:
       An image Tensor.
     """
-    if np.isscalar(mask_size):
+    if tf.equal(tf.rank(mask_size), 0):
         mask_size = [mask_size, mask_size]
     data_format = conv_utils.normalize_data_format(data_format)
     # Sample the center location in the images where the zero mask will be applied.
@@ -121,7 +120,7 @@ def cutout(
       An image Tensor.
     """
     with tf.name_scope("cutout"):
-        if np.isscalar(mask_size):
+        if tf.equal(tf.rank(mask_size), 0):
             mask_size = [mask_size, mask_size]
         data_format = conv_utils.normalize_data_format(data_format)
         image_height, image_width = _get_image_wh(images, data_format)
--- a/tensorflow_addons/image/cutout_ops.py
+++ b/tensorflow_addons/image/cutout_ops.py
@@ -16,25 +16,21 @@
 
 import tensorflow as tf
 from tensorflow_addons.utils.types import TensorLike, Number
-from tensorflow_addons.image.utils import from_4D_image, to_4D_image
 from tensorflow.python.keras.utils import conv_utils
 
 
 def _get_image_wh(images, data_format):
-    if tf.equal(tf.rank(images), 4):
-        if data_format == "channels_last":
-            image_height, image_width = tf.shape(images)[1], tf.shape(images)[2]
-        else:
-            image_height, image_width = tf.shape(images)[2], tf.shape(images)[3]
+    if data_format == "channels_last":
+        image_height, image_width = tf.shape(images)[1], tf.shape(images)[2]
     else:
-        image_height, image_width = tf.shape(images)[0], tf.shape(images)[1]
+        image_height, image_width = tf.shape(images)[2], tf.shape(images)[3]
 
     return image_height, image_width
 
 
 def _norm_params(images, mask_size, data_format):
     mask_size = tf.convert_to_tensor(mask_size)
-    if tf.equal(tf.rank(mask_size), 0):
+    if tf.rank(mask_size) == 0:
         mask_size = tf.stack([mask_size, mask_size])
     data_format = conv_utils.normalize_data_format(data_format)
     image_height, image_width = _get_image_wh(images, data_format)
@@ -50,19 +46,18 @@ def random_cutout(
 ) -> tf.Tensor:
     """Apply cutout (https://arxiv.org/abs/1708.04552) to images.
 
-    This operation applies a (2*pad_size[0] x 2*pad_size[1]) mask of zeros to
+    This operation applies a (2 * mask_height x 2 * mask_width) mask of zeros to
     a random location within `img`. The pixel values filled in will be of the
     value `replace`. The located where the mask will be applied is randomly
     chosen uniformly over the whole images.
 
     Args:
       images: A tensor of shape
-        (num_images, num_rows, num_columns, num_channels)
-        (NHWC), (num_images, num_channels, num_rows, num_columns)(NCHW), (num_rows, num_columns, num_channels) (HWC), or
-        (num_rows, num_columns) (HW).
+        (batch_size, height, width, channels)
+        (NHWC), (batch_size, channels, height, width)(NCHW).
       mask_size: Specifies how big the zero mask that will be generated is that
         is applied to the images. The mask will be of size
-        (2*pad_size[0] x 2*pad_size[1]).
+        (2 * mask_height x 2 * mask_width).
       constant_values: What pixel value to fill in the images in the area that has
         the cutout mask applied to it.
       seed: A Python integer. Used in combination with `tf.random.set_seed` to
@@ -75,23 +70,20 @@ def random_cutout(
     Returns:
       An image Tensor.
     """
+    batch_size = tf.shape(images)[0]
     mask_size, data_format, image_height, image_width = _norm_params(
         images, mask_size, data_format
     )
 
     cutout_center_height = tf.random.uniform(
-        shape=[], minval=0, maxval=image_height, dtype=tf.int32, seed=seed
+        shape=[batch_size], minval=0, maxval=image_height, dtype=tf.int32, seed=seed
     )
     cutout_center_width = tf.random.uniform(
-        shape=[], minval=0, maxval=image_width, dtype=tf.int32, seed=seed
+        shape=[batch_size], minval=0, maxval=image_width, dtype=tf.int32, seed=seed
     )
-    return cutout(
-        images,
-        mask_size,
-        [cutout_center_height, cutout_center_width],
-        constant_values,
-        data_format,
-    )
+
+    offset = tf.transpose([cutout_center_height, cutout_center_width], [1, 0])
+    return cutout(images, mask_size, offset, constant_values, data_format,)
 
 
 def cutout(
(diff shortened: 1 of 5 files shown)


tensorflow_addons/image/cutout_ops_test.py
```diff
@@ -0,0 +1,86 @@
...
+        test_image_file = tf.io.read_file(
+            "tensorflow_addons/image/test_data/Yellow_Smiley_Face.png"
+        )
+        expect_image_file = tf.io.read_file(
+            "tensorflow_addons/image/test_data/Yellow_Smiley_Face_Warp-cutout-20.png"
+        )
+        for channel in [0, 1, 3, 4]:
+            with self.subTest(channel=channel):
+                test_image = tf.image.decode_image(
+                    test_image_file, channels=channel, dtype=tf.uint8
```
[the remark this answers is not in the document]
fsx950223: I have tested ```dtype``` in ```test_different_dtypes``` and it will increase test time if I use it on all test cases.

tensorflow_addons/image/cutout_ops.py
```diff
@@ -0,0 +1,85 @@
...
+            tf.zeros(cutout_shape, dtype=image.dtype), padding_dims, constant_values=1
+        )
+
+        if tf.equal(tf.rank(image), 3):
+            mask = tf.expand_dims(mask, -1)
+            mask = tf.tile(mask, [1, 1, tf.shape(image)[-1]])
+        elif tf.equal(tf.rank(image), 4):
+            mask = tf.expand_dims(mask, 0)
+            mask = tf.expand_dims(mask, -1)
+            mask = tf.tile(mask, [tf.shape(image)[0], 1, 1, tf.shape(image)[-1]])
```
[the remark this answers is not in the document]
fsx950223: Changed. But I wonder about the performance of these ops.

gabrieldemarmiesse:
Thank you for this pull request! I believe we're going to change the API of the reference implementation to make it more user friendly. Here is what I propose:

```python

def cutout(
    images: TensorLike,
    mask_size: TensorLike,
    offset: TensorLike= (0, 0),
    constant_values: Number = 0,
    data_format="channels_last"
) -> tf.Tensor:
"""

Args:
    image: A 4D tensor, (N, H, W, C). 
    mask_size: a tuple (or tensor) with two values. Height and width of the mask (correspond to 2x pad_size). A single scalar means a square mask (like in keras conv and pad layers). Can be a tensor of shape (N, 2) to have different mask sizes in the same batch.
    offset: the offset relative to the center of the image. Default is (0, 0) which means the mask will be in the middle of the image. Can be a tensor of shape (N, 2) to have different offsets in the same batch.
    constant_values: the values used to fill the mask.
    data_format: The data format
"""
```

The rational behind using only 4D tensors is that when the users passes 3D tensors, there is a non-negligable probability that user mistake `NHW` and `HWC`.

It's totally fine to not implement everything. For example, you can do 
```python
if data_format == "channels_first":
    raise NotImplementedError("Channels first is not yet available for cutout. Contributions welcome!")

if tf.rank(mask_size) != 0:
     raise NotImplementedError("Having non-square masks is not supported yet, contributions welcome.")
```


random_cutout should follow a similar pattern.

I believe that it should not change too much your code, it's already really close to the end result.

For the tests, please avoid adding images to the git repo, also we moving away from `tf.test.TestCase` and `@run_all_in_graph_and_eager_mode`. The philosophy now is to use plain pytest tests with the decorator `@pytest.mark.usefixtures("maybe_run_functions_eagerly")` . This enables eager mode in tests which helps a lot in readability. See #1328 .

gabrieldemarmiesse:
@fsx950223 feel free to ping me when you're done with the changes.

fsx950223:
cc @gabrieldemarmiesse

tensorflow_addons/image/cutout_ops.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+    mask_size, data_format, image_height, image_width = _norm_params(
+        images, mask_size, data_format
+    )
+
+    cutout_center_height = tf.random.uniform(
+        shape=[], minval=0, maxval=image_height, dtype=tf.int32, seed=seed
+    )
+    cutout_center_width = tf.random.uniform(
+        shape=[], minval=0, maxval=image_width, dtype=tf.int32, seed=seed
```
gabrieldemarmiesse: If I understand correctly, if a batch of images is passed and random cutout is applied, the mask will be at the same place for all the images in the batch, right? Is this something we want? Maybe users would expect (as an augmentation strategy) to have the mask being at different places for all the images in the batch. 
If the original implementation uses the same mask for all the images in the batch, let's go with this implementation, but the function shouldn't be public. I believe a public `random_cutout` shoud respect the principle of least astonishment. 
What do you think?

fsx950223:
> Thank you for the pull request. We're nearly there, a few changes here and there need to be made.

I will change them later.

fsx950223:
cc @gabrieldemarmiesse

tensorflow_addons/image/cutout_ops.py
AakashKumarNain: Exactly! In a batch the cutout operation has to be random for each image.

tensorflow_addons/image/cutout_ops.py
```diff
@@ -0,0 +1,166 @@
...
+    value `replace`. The located where the mask will be applied is randomly
+    chosen uniformly over the whole images.
+
+    Args:
+      images: A tensor of shape
+        (batch_size, height, width, channels)
+        (NHWC), (batch_size, channels, height, width)(NCHW).
+      mask_size: Specifies how big the zero mask that will be generated is that
+        is applied to the images. The mask will be of size
+        (2 * mask_height x 2 * mask_width).
```
gabrieldemarmiesse: `mask_size` should be the shape of the mask. Here the shape of the true mask is `(2 * mask_height x 2 * mask_width)` which might confuse users. Could you change that to `(mask_height x mask_width)`? You can throw an error if the width and height cannot be divided by 2.

tensorflow_addons/image/cutout_ops.py
```diff
@@ -0,0 +1,166 @@
...
+    a random location within `img`. The pixel values filled in will be of the
+    value `replace`. The located where the mask will be applied is randomly
+    chosen uniformly over the whole images.
+
+    Args:
+      images: A tensor of shape (batch_size, height, width, channels)
+        (NHWC), (batch_size, channels, height, width)(NCHW).
+      mask_size: Specifies how big the zero mask that will be generated is that
+        is applied to the images. The mask will be of size
+        (2 * mask_height x 2 * mask_width).
```
gabrieldemarmiesse: Same as above.

tensorflow_addons/image/cutout_ops.py
```diff
@@ -0,0 +1,166 @@
...
+      offset: A tuple of (height, width)
```
gabrieldemarmiesse: Could you specify all the shapes are are possible? A tuple is possible but a 2D tensor too (batch_size, 2)

tensorflow_addons/image/cutout_ops.py
```diff
@@ -0,0 +1,166 @@
...
+    images: TensorLike,
+    mask_size: TensorLike,
+    offset: TensorLike = (0, 0),
+    constant_values: Number = 0,
+    data_format: str = "channels_last",
+) -> tf.Tensor:
+    """Apply cutout (https://arxiv.org/abs/1708.04552) to images.
+
+    This operation applies a (2 * mask_height x 2 * mask_width) mask of zeros to
+    a random location within `img`. The pixel values filled in will be of the
```
gabrieldemarmiesse:
```suggestion
    a location within `img` specified by the offset. The pixel values filled in will be of the
```

tensorflow_addons/image/cutout_ops_test.py
```diff
@@ -0,0 +1,84 @@
...
+    result_image = random_cutout(
+        test_image, 20, seed=1234, data_format="channels_first"
+    )
+    np.testing.assert_allclose(tf.shape(result_image), tf.shape(expect_image))
+
+
+@pytest.mark.usefixtures("maybe_run_functions_eagerly")
+def test_with_tf_function():
+    test_image = tf.ones([1, 40, 40, 1], dtype=tf.uint8)
+    result_image = tf.function(random_cutout)(test_image, 2)
```
gabrieldemarmiesse: Here we need to make sure that the for loop with the TensorArray works well. To do that, we need to make sure the loop is possible even if the size of the batch is not known when drawing the graph. Could you use `input_signature` and set the shape of `images` to `[None, 40, 40, 1]` ?

tensorflow_addons/image/cutout_ops_test.py
```diff
@@ -0,0 +1,84 @@
...
+    cutout_area = tf.zeros([4, 4], dtype=tf.uint8)
+    cutout_area = tf.pad(cutout_area, ((0, 36), (0, 36)), constant_values=1)
+    expect_image = to_4D_image(cutout_area)
+    np.testing.assert_allclose(tf.shape(result_image), tf.shape(expect_image))
+
+
+if __name__ == "__main__":
```
gabrieldemarmiesse: Could you add one more test to ensure that with random cutout, the masks are at different places in the images of the batch? I'll let you decide what's the easier way of doing that.

tensorflow_addons/image/cutout_ops_test.py
```diff
@@ -0,0 +1,84 @@
...
+import tensorflow as tf
+import numpy as np
+from absl.testing import parameterized
+from tensorflow_addons.image.cutout_ops import cutout, random_cutout
+from tensorflow_addons.image.utils import to_4D_image
+
+
+@parameterized.named_parameters(
+    ("float16", np.float16), ("float32", np.float32), ("uint8", np.uint8)
+)
```
gabrieldemarmiesse:
```suggestion
@pytest.mark.parameterize("dtype", [np.float16, np.float32, np.uint8])
```

tensorflow_addons/image/cutout_ops_test.py
```diff
@@ -0,0 +1,84 @@
...
+def test_different_dtypes(dtype):
+    test_image = tf.ones([1, 40, 40, 1], dtype=dtype)
+    result_image = cutout(test_image, 2, [2, 2])
+    cutout_area = tf.zeros([4, 4], dtype=dtype)
+    cutout_area = tf.pad(cutout_area, ((0, 36), (0, 36)), constant_values=1)
+    expect_image = to_4D_image(cutout_area)
+    np.testing.assert_allclose(result_image, expect_image)
+
+
```
gabrieldemarmiesse:
```suggestion
    np.testing.assert_allclose(result_image, expect_image)
    assert result_image.dtype == dtype


```

fsx950223:
cc @gabrieldemarmiesse .

tensorflow_addons/image/cutout_ops.py
```diff
@@ -30,11 +30,20 @@ def _get_image_wh(images, data_format):
...
-    return mask_size, data_format, image_height, image_width
+    with tf.control_dependencies(
+        [
+            tf.assert_equal(
+                tf.reduce_any(mask_size % 2 != 0),
+                False,
+                "mask_size should be divisible by 2",
+            )
+        ]
+    ):
```
gabrieldemarmiesse: control_dependencies is quite tf 1.x style, I recommend to activate the check only in eager mode. It will also be easier to debug for users.

gabrieldemarmiesse: Your piece of code raised a very important question that I discussed in #1417 . We don't need to fix #1417 for merging your pull request of course.

fsx950223:
> Thanks for this great pull request!

Thanks for your review

PR #1339, opened 2020-03-19
Added helper ops, translate, shear

abhichou4:
Refer #1333

### DIFF
--- a/tensorflow_addons/image/translate_ops_test.py
+++ b/tensorflow_addons/image/translate_ops_test.py
@@ -18,9 +18,10 @@ import sys
 
 import pytest
 import tensorflow as tf
+import numpy as np
 
 from tensorflow_addons.image import translate_ops
-from tensorflow_addons.utils import test_utils
+from PIL import Image
 
 _DTYPES = {
     tf.dtypes.uint8,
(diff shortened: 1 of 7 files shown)


seanpmorgan:
Just want to confirm that none of these helpers should be exported in the __init__ . Any thoughts?

abhichou4:
> Just want to confirm that none of these helpers should be exported in the **init** . Any thoughts?

Sure I don't see their use outside of these functions for now.

sayoojbk:
> > Just want to confirm that none of these helpers should be exported in the **init** . Any thoughts?
> 
> Sure I don't see their use outside of these functions for now.

As @abhichou4  said I don't see much of their use case outside these ops for now. Still up to you @seanpmorgan if you feel they should be included. Maybe @abhichou4 you could add Shear as a new image_ops and change the translate ops function to add a parameter to specify the translation to be done specifically to x or y-axis. Just my thoughts. Rest LGTM ✌️

abhichou4:
> > > Just want to confirm that none of these helpers should be exported in the **init** . Any thoughts?
> > 
> > 
> > Sure I don't see their use outside of these functions for now.
> 
> As @abhichou4 said I don't see much of their use case outside these ops for now. Still up to you @seanpmorgan if you feel they should be included. Maybe @abhichou4 you could add Shear as a new image_ops and change the translate ops function to add a parameter to specify the translation to be done specifically to x or y-axis. Just my thoughts. Rest LGTM v

I'll make the change to specify the axis. Can I move shear to `transform_ops.py`? Seems well suited there.

AakashKumarNain:
For any image op that we add, we should test the implementation results against standard libraries like `PIL` and `skimage`. Thoughts? @seanpmorgan @gabrieldemarmiesse

gabrieldemarmiesse:
>For any image op that we add, we should test the implementation results against standard libraries like PIL and skimage. 

Yes, very much yes. If we have reference implementations, we can generate random numpy arrays and as long as the tests are passing, we don't even have to check the math. It's the best option when it's available.

abhichou4:
> For any image op that we add, we should test the implementation results against standard libraries like `PIL` and `skimage`. Thoughts? @seanpmorgan @gabrieldemarmiesse

The only reason I did not add any tests was because these operations were immediate use cases of transform and translate. But instead of a `numpy` reference, `PIL` or `skimage` would be way better. @gabrieldemarmiesse any thoughts?

gabrieldemarmiesse:
@abhichou4 I see what you mean. Indeed it does not seem critical, but there is very few drawbacks about having too many tests. I don't even know if it's possible to have too many tests. Since you don't change the public API, we can merge this PR and add tests later on.

abhichou4:
No, I would like to add the tests. Can `PIL` can be used for this? Sorry I can't see why we use just numpy.

gabrieldemarmiesse:
You can use PIL, skimage, scikit-learn... whatever you prefer.

AakashKumarNain:
> No, I would like to add the tests. Can `PIL` can be used for this? Sorry I can't see why we use just numpy.

There is a reason. I don't exactly remember the op name but there was an image op in TF/contrib which didn't output the same results as other libraries did. People wasted days debugging that. It is always good to check things against a standard implementation. Sorry, I know it increases the time to merge these PR but I need to point out what I have seen in the past

abhichou4:
> You can use PIL, skimage, scikit-learn... whatever you prefer.

@gabrieldemarmiesse Should I add `scikit-image` and `pillow` to `pytest.txt` to resolve import errors?

abhichou4:
> > No, I would like to add the tests. Can `PIL` can be used for this? Sorry I can't see why we use just numpy.
> 
> There is a reason. I don't exactly remember the op name but there was an image op in TF/contrib which didn't output the same results as other libraries did. People wasted days debugging that. It is always good to check things against a standard implementation. Sorry, I know it increases the time to merge these PR but I need to point out what I have seen in the past

Here `trans

abhichou4:
> There is a reason. I don't exactly remember the op name but there was an image op in TF/contrib which didn't output the same results as other libraries did. People wasted days debugging that. It is always good to check things against a standard implementation. Sorry, I know it increases the time to merge these PR but I need to point out what I have seen in the past

`translate_xy` mirrors the `PIL` public API and the shear operations agree with `skimage.transforms` with appropriate matrices. 

What might be left is making `warp` and `unwarp` compatible with other `dtypes` and adding their tests. After that, we can make it compatible with different ranks (making these `tf.data.Datasets` compatible). These can be added in another PR for the sake of simplicity.

AakashKumarNain:
Thank you for the changes. @gabrieldemarmiesse any other changes required?

tensorflow_addons/image/translate_ops_test.py
```diff
@@ -32,24 +33,53 @@
...
-                [[1, 0, 1, 0], [0, 1, 0, 0], [1, 0, 1, 0], [0, 0, 0, 0]],
-            )
-
-    def test_translations_to_projective_transforms(self):
-        translation = tf.constant([-1, -1], dtype=tf.float32)
-        transform = translate_ops.translations_to_projective_transforms(translation)
-        self.assertAllEqual(self.evaluate(transform), [[1, 0, 1, 0, 1, 1, 0, 0]])
+# @test_utils.run_all_in_graph_and_eager_modes
+# class TranslateOpTest(tf.test.TestCase):
+
```
gabrieldemarmiesse: I think you can remove this.

gabrieldemarmiesse:
Thank you for the pull request! It's a great addition. As a side note, in the future, try to make smaller pull requests, it's easier to review.

abhichou4:
> Thank you for the pull request! It's a great addition. As a side note, in the future, try to make smaller pull requests, it's easier to review.

That's the intent always :smile:
Thanks for the help and review.

PR #1345, opened 2020-03-19
Moving out of tf.test.TestCase in sparsemax_test.py

gabrieldemarmiesse:
See #1328


AndreasMadsen:
I'm not familiar with the new pytest system. I guess this is fine, but I don't understand why only one test was changed. Why not change them all?

gabrieldemarmiesse:
It's to avoid having this pull request being too big and hard to review. I'll make other pull requests to finish the work.

PR #1348, opened 2020-03-19
Test parse time in eager mode

gabrieldemarmiesse:
See #1328

### DIFF
--- a/tensorflow_addons/text/parse_time_op.py
+++ b/tensorflow_addons/text/parse_time_op.py
@@ -26,7 +26,7 @@ _parse_time_so = LazySO("custom_ops/text/_parse_time_op.so")
 tf.no_gradient("Addons>ParseTime")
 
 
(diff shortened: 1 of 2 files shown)


tensorflow_addons/text/parse_time_op.py
```diff
@@ -26,7 +26,7 @@
 tf.no_gradient("Addons>ParseTime")
 
 
-def parse_time(time_string: str, time_format: str, output_unit: str) -> str:
+def parse_time(time_string: str, time_format: str, output_unit: str) -> tf.Tensor:
```
helinwang: The return type is changed. Does it still work for non-eager mode / v1 compat mode?

gabrieldemarmiesse: Sorry, I changed the return type because it was wrong in the first place. The function returns a tensor, not a python string.

helinwang: Ah you are right, thanks!

PR #1359, opened 2020-03-20
removed run_all_in_eager_and_graph_mode in beam search.

gabrieldemarmiesse:
See #1328

### DIFF
--- a/tensorflow_addons/seq2seq/beam_search_decoder_test.py
+++ b/tensorflow_addons/seq2seq/beam_search_decoder_test.py
@@ -22,7 +22,6 @@ import tensorflow as tf
 
 from tensorflow_addons.seq2seq import attention_wrapper
 from tensorflow_addons.seq2seq import beam_search_decoder, gather_tree
-from tensorflow_addons.utils import test_utils
 
 
 class TestGatherTree(tf.test.TestCase):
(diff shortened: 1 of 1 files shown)


tensorflow_addons/seq2seq/beam_search_decoder_test.py
```diff
@@ -530,139 +529,117 @@ def get_probs():
...
+    cell = tf.keras.layers.LSTMCell(cell_depth)
+    initial_state = cell.get_initial_state(batch_size=batch_size, dtype=tf.float32)
+    coverage_penalty_weight = 0.0
+    if has_attention:
+        coverage_penalty_weight = 0.2
+        inputs = tf.compat.v1.placeholder_with_default(
+            np.random.randn(batch_size, decoder_max_time, input_depth).astype(
+                np.float32
+            ),
+            shape=(None, None, input_depth),
```
guillaumekln: In eager mode, `placeholder_with_default` returns fully defined shapes. So we may need to convert part of this test to use a `tf.function` with an input signature. What do you think?

gabrieldemarmiesse: Overall, I'd prefer avoiding `tf.compat.v1` as it's not TF2 idiomatic code. Let me try to remove this part of the diff of this pull request so that we can merge the rest and then I'll open a new pull request for this specific piece of code.

gabrieldemarmiesse: I tried to make it work with a `tf.keras.layers.Input` but that didn't go well. I'll drop this PR for now. I may have better ideas in the future.

I believe that we should test with code that looks like user code, that's why I'd prefer using a input layer rather than a tf.function on this one. But if we use tf.function that's not so bad too if we can keep the error checking code out of the tf.function.

PR #1365, opened 2020-03-22
Fix TQDM Progress bar in TF 2.2.0

lminer:
TQDMProgressBar is currently broken in tensorflow 2.2.0 because of refactor of the way fitting works in tf.keras. This change should fix it so that it works as before.


gabrieldemarmiesse:
Thanks for the fix! Could you add some tests to ensure that we won't intruduce any regressions in the future? You can take a look at how tests are made for layers or activations for example.

bjtho08:
@lminer this fix disables reporting of validation metrics. Maybe you could try and look into this?

gabrieldemarmiesse:
> this fix disables reporting of validation metrics. Maybe you could try and look into this?

Thanks a lot @bjtho08 for catching this before the merge. I added extensive testing, including a test for the issue about the validation metrics you're mentioning. See #1568 .

gabrieldemarmiesse:
@googlebot I consent

gabrieldemarmiesse:
It seems that not all the tests are passing with the proposed fix. Maybe somebody would like to step in to propose a new fix?

seanpmorgan:
Did a small dig into this. The callback's `self.params` are not being set the same way in 2.1 and in 2.2.

2.2 self.params:
```
{'verbose': 0, 'epochs': 1, 'steps': 1}
```

2.1 self.params:
```
{'batch_size': 32, 'epochs': 1, 'steps': 1, 'samples': 5, 'verbose': 0, 'do_validation': False, 'metrics': ['loss', 'acc']}
```

Which is causing the validation metrics to not be reported:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/callbacks.py#L160-L173

This affects more than the validation reporting since we leverage self.params in multiple places. E.g. the number of samples in a batch is no longer available so the batch progress always shows 1 sample per batch. 
`1/1  ██████████ ETA: 00:00s - loss: 0.4079 - acc: 0.2000`.

Using `self.model.metrics_names` won't be a sufficient fix. @lminer  do you have any more information regarding the changes to `keras.fit`?

lminer:
@seanpmorgan Unfortunately I don't have any insight on this. Just noticed the breakage and offered an unsatisfactory fix :) Looking over the updated ProgBarCallback, it seems as if they don't rely on `metrics_names` anymore to decide what to log. Instead they just pop out size, num_steps, and batch and then log everything else regardless of the name: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/callbacks.py#L950

bjtho08:
@lminer Can we not just do the same for TQDM progbar?

gabrieldemarmiesse:
I'm looking into it, and I'll try to propose a pull request when I have a solution.

gabrieldemarmiesse:
Superseded by #1595

linyuanthocr:
Thanks for the update, colab uses 2.2.0.rc1 and has the problem as well. It totally works after the update!

PR #1370, opened 2020-03-24
Test gather tree in eager mode only because it's a custom op

gabrieldemarmiesse:
We only need to run this test in eager mode, it makes it more readable too.


gabrieldemarmiesse:
Thanks a lot!

Issue #1317, opened 2020-03-16
Make a better UX around the issue of tensorflow ABI instability.


gabrieldemarmiesse:
**Describe the feature and the current behavior/state.**

See #1298

Currently, if we compile Tensorflow Addons against tensorflow 2.1.0, we have no garantees that it's going to work with another version of tensorflow. We don't even have a garantee that it will work with the same version if tensorflow has been compiled from source. As such, we should try to find ways of making this as painless as possible to users until Tensorflow provides backward compatibily garantees about the ABI (see tensorflow/community#77, tensorflow/community#101 and  tensorflow/community#133). Ideas welcome.

failure-to-thrive:
Make a [black]list of builds/hashes of `.so`/`.dll`that are seems to be good / proven to be bad. Print out appropriate warnings.

MarkDaoust:
If you have a different version installed, `tfa-nightly` does warn that it was built for TensorFlow 2.1 and may not work.  So steps have already been taken towards @failure-to-thrive's suggestion. It's just not included in the 0.8 package.

gabrieldemarmiesse:
> It's just not included in the 0.8 package.

#1307 was merged. We just need to make a new release. @seanpmorgan , you may want to add this to your todo list.


I propose that we make a 0.9 release, compiled against TF 2.1, and at the same time, we make a 0.10  release, compile against TF 2.2, both with an appropriate warning message similar to the one we had before. We'll need to backport all the bugfixes twice, but with the bot in place it's fairly painless.

gabrieldemarmiesse:
I believe we also need a section in our readme describing the ins and outs of the problem so that users really understand what is going on and why our compatibility with TF is so uncommon. We can put a link to this section in our warning message.

gabrieldemarmiesse:
We have two types of incompatibilities:
1) Python incompatibilities
2) ABI incompatibility.

It's possible for a user to use the python-only features and be completely fine. So I think we should have two warning messages, one at import time, for python incompatibility, and one when loading the first SO, to warn about ABI incompatibility. We except more compatibility in the python world than in the ABI world so the conditions should be different.

PR #1371, opened 2020-03-24
Update SequenceLoss for TensorFlow 2.2 compatibility

guillaumekln:
For previous versions, we removed the `reduction` attribute to force tf.keras to call the main `__call__` method. Otherwise, it calls the `call` method where we have no way of controlling the reduction logic.

In 2.2, tf.keras has a new way of dealing with loss objects (see the `LossContainer` class). It now always calls `__call__` and requires the `reduction` attribute to exist.

This new logic is only enabled for V2 execution mode so this commit also disables the deprecated V1 graph mode for this test.

Related to #1320.

Note: this change is only compatible with TensorFlow 2.2 so I'm opening a draft PR for now.

### DIFF
--- a/tensorflow_addons/seq2seq/loss.py
+++ b/tensorflow_addons/seq2seq/loss.py
@@ -196,8 +196,9 @@ class SequenceLoss(tf.keras.losses.Loss):
         # However, prior to TensorFlow 2.2 tf.keras will actually invokes "call"
         # when the loss object has the "reduction" attribute. So we remove it in
         # this case.
-        if (LooseVersion(tf.__version__) < LooseVersion("2.2.0")
-            and hasattr(self, "reduction")):
+        if LooseVersion(tf.__version__) < LooseVersion("2.2.0") and hasattr(
+            self, "reduction"
+        ):
             delattr(self, "reduction")
 
     def __call__(self, y_true, y_pred, sample_weight=None):
--- a/tensorflow_addons/seq2seq/loss.py
+++ b/tensorflow_addons/seq2seq/loss.py
@@ -14,6 +14,8 @@
 # ==============================================================================
 """Seq2seq loss operations for use in sequence models."""
 
+from distutils.version import LooseVersion
+
 import tensorflow as tf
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse:
I guess the easiest solution is to do a version check before deleting the reduction.

But having a custom reduction prevents us from using a distributed strategy (from what I remember). Would it be possible to use keras' built-in reductions for some use cases (not all)?

guillaumekln:
> I guess the easiest solution is to do a version check before deleting the reduction.

I suppose the project will shortly move to require TensorFlow 2.2, right? Then we could merge this PR as is (after a rebase).

> But having a custom reduction prevents us from using a distributed strategy (from what I remember). Would it be possible to use keras' built-in reductions for some use cases (not all)?

I'm not sure it prevents the use of distribution strategies, but it requires the user to carefully scale the loss (or gradients) based on the global batch size.

We could indeed rely on the built-in reduction for some combinations. But these combinations are actually the ones that can be implemented with `tf.keras.losses.{Sparse}CategoricalCrossentropy`.

gabrieldemarmiesse:
> I suppose the project will shortly move to require TensorFlow 2.2, right? 

To improve the UX, we'd like the python code to be compatible with as many versions of TF as possible. The custom ops will only be compatible with a specific python wheel. See https://github.com/tensorflow/addons/issues/1317#issuecomment-600133789

seanpmorgan:
Thanks @guillaumekln!

Just wanted to note that this is the last component of #1320 and will block a release built with TF2.2 once it is released.

@qlzh727 Could you review when time allows and advise how best to handle this change in 2.2

qlzh727:
@pavithrasv who works on all Keras losses for more input.

guillaumekln:
> Iet's let you do what you think is best here.

I propose to just make a bug fix in this PR and try to not change the behavior. Maybe @pavithrasv has more recommendation on how to deal with custom loss reduction.

gabrieldemarmiesse:
Let's merge it as it's a good step forward. We can open an issue to see how to replace the `__call__` which is a private keras API.

PR #1374, opened 2020-03-24
run keras compatibility test for sequence loss with v2 behavior.

gabrieldemarmiesse:
I runs in full eager mode + with the tf.function enabled by default for keras layers.

@guillaumekln that might be useful for #1371 

`get_test_data` is here to replace the `setup` method. I'll convert to pytest the rest of the tests in this file in another pull request.

`np.testing.assert_allclose(calculated_loss, expected_loss, rtol=1e-6, atol=1e-6)`

`np.testing.assert_allclose` is more strict than the tf counterpart `self.assertAllClose` so I took the rtol and atol from the `self.assertAllClose` to have the same behavior. Otherwise the tests were not passing.

### DIFF
--- a/tensorflow_addons/seq2seq/loss_test.py
+++ b/tensorflow_addons/seq2seq/loss_test.py
@@ -24,6 +24,37 @@ from tensorflow_addons.seq2seq import loss
 from tensorflow_addons.utils import test_utils
 
 
+def get_test_data():
+    batch_size = 2
+    sequence_length = 3
+    number_of_classes = 5
+    logits = [
+        tf.constant(i + 0.5, shape=[batch_size, number_of_classes])
+        for i in range(sequence_length)
+    ]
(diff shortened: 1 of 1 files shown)


tensorflow_addons/seq2seq/loss_test.py
```diff
@@ -24,6 +24,37 @@
 from tensorflow_addons.utils import test_utils
 
 
+def get_test_data():
```
guillaumekln: You could already factorize the data preparation with the other class. Do you plan to do that in the next PR instead?

gabrieldemarmiesse: Yeah, I don't want to make the PR too big. Big pull requests are a pain to review.

guillaumekln: Ok!

PR #1407, opened 2020-03-25
add invert

abhichou4:
Relates to #1333


gabrieldemarmiesse:
Thanks for your work. That seems a bit overkill to expose that in the public API and to have tests for it. It can be just a private function to be used whenever needed by another feature in addons. I'll close this because the invert function should be added with the function that makes use of it to justify its presence.

Feel free to open another pull request with a feature that makes use of inverse.

PR #1416, opened 2020-03-26
Update contributing documentation

seanpmorgan:
* Python3.5 symlink is actually linked to a system python that has complications with setuptools. Bazel uses the symlink python3 anyway and this will be more future proof as 3.5 is deprecated

* Updated custom op installation


gabrieldemarmiesse:
Do we drop python 3.5 in september (official EOL) or before?

seanpmorgan:
> Do we drop python 3.5 in september (official EOL) or before?

I think we'll follow what TF core does. Likely dropping slightly after EOL based on py34

PR #1425, opened 2020-03-26
posterize_ops.py operation for Tensorflow Addons

vinayvr11:
Posterize operations for Tensorflow Addons
Issue Refrence: https://github.com/tensorflow/addons/issues/1333


Squadrick:
You need to update `tensorflow_addons/image/BUILD` and `.github/CODEOWNERS`.

Squadrick:
Also, add a test that compares with `PIL` so we know the result is correct.

bhack:
This is now in https://github.com/tanzhenyu/image_augmentation /cc @tanzhenyu

PR #1430, opened 2020-03-26
Moved test out of run_all_in_graph_and_eager_mode sparse_image_wrap.

gabrieldemarmiesse:
See #1328

### DIFF
--- a/tensorflow_addons/image/sparse_image_warp_test.py
+++ b/tensorflow_addons/image/sparse_image_warp_test.py
@@ -231,42 +231,35 @@ class SparseImageWarpTest(tf.test.TestCase):
                 # than that in the saved png file loaded into target_image.
                 self.assertAllClose(target_image, out_image, atol=2, rtol=1e-3)
 
-    def testThatBackpropRuns(self):
-        """Run optimization to ensure that gradients can be computed."""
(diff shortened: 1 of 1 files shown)


tensorflow_addons/image/sparse_image_warp_test.py
```diff
@@ -231,42 +231,35 @@ def testSmileyFace(self):
...
-                control_point_locations + control_point_displacements,
-                num_boundary_points=3,
-            )
-            loss = tf.reduce_mean(tf.abs(warped_image - image))
-            return loss
-
-        optimizer = tf.keras.optimizers.SGD(
-            learning_rate=0.001, momentum=0.9, clipnorm=1.0
+def test_that_backprop_runs():
+    """Run optimization to ensure that gradients can be computed."""
```
seanpmorgan: Probably need to re-word this doc string now that we're directly checking backprop gradients instead of running an optimization

gabrieldemarmiesse:
```suggestion
    """Making sure the gradients can be computed."""
```

gabrieldemarmiesse: done

PR #1434, opened 2020-03-26
Moved test out of run_in_graph_and_eager_mode in maxout.

gabrieldemarmiesse:
See #1328

### DIFF
--- a/tensorflow_addons/layers/maxout_test.py
+++ b/tensorflow_addons/layers/maxout_test.py
@@ -18,49 +18,45 @@ import sys
 
 import pytest
 import numpy as np
-import tensorflow as tf
 
 from tensorflow_addons.layers.maxout import Maxout
 from tensorflow_addons.utils import test_utils
 
 
-@test_utils.run_all_in_graph_and_eager_modes
-class MaxOutTest(tf.test.TestCase):
-    def test_simple(self):
-        test_utils.layer_test(
-            Maxout, kwargs={"num_units": 3}, input_shape=(5, 4, 2, 18)
-        )
+pytestmark = pytest.mark.usefixtures("maybe_run_functions_eagerly")
 
-    def test_nchw(self):
-        test_utils.layer_test(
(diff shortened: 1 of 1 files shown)


tensorflow_addons/layers/maxout_test.py
```diff
@@ -18,49 +18,45 @@
...
-            input_shape=(None, None, None, None),
-            input_data=inputs,
-        )
-
-    def test_invalid_shape(self):
-        with self.assertRaisesRegexp(ValueError, r"number of features"):
-            test_utils.layer_test(
-                Maxout, kwargs={"num_units": 3}, input_shape=(5, 4, 2, 7)
-            )
+pytestmark = pytest.mark.usefixtures("maybe_run_functions_eagerly")
```
AakashKumarNain: @gabrieldemarmiesse this isn't used anywhere else in this code. Is this required?

gabrieldemarmiesse: Yes, when you use a `pytestmark` variable in a test file run with pytest, the `pytest.mark.*` markers are applied  to all the test functions in the file. I agree that it's counter-intuitive. I like explicit stuff so I would have like it better if pytest had a function for that. Like `apply_to_all_tests_in_module(...)` or something like that. See http://doc.pytest.org/en/latest/example/markers.html#marking-whole-classes-or-modules

AakashKumarNain: Cool. Yeah that is not `pythonic` at all. Any other changes? If not, I will merge it

gabrieldemarmiesse: No other changes. :)

AakashKumarNain: Ok. Thanks

gabrieldemarmiesse:
Thanks for the review!

failure-to-thrive:
Too late, but looks like
```
```
is not required either...

gabrieldemarmiesse:
@failure-to-thrive it's used in the last test.

Generally flake8 catches unused imports in the CI :)

PR #1436, opened 2020-03-26
Moved test out of run_in_graph_and_eager_mode.

gabrieldemarmiesse:
See #1328

### DIFF
--- a/tensorflow_addons/layers/multihead_attention_test.py
+++ b/tensorflow_addons/layers/multihead_attention_test.py
@@ -25,6 +25,7 @@ from tensorflow_addons.layers.multihead_attention import MultiHeadAttention
 from tensorflow_addons.utils import test_utils
 
 
+@pytest.mark.usefixtures("maybe_run_functions_eagerly")
 def test_output_shape():
     batch_size = 10
     num_heads = 8
--- a/tensorflow_addons/layers/multihead_attention_test.py
+++ b/tensorflow_addons/layers/multihead_attention_test.py
@@ -25,25 +25,26 @@ from tensorflow_addons.layers.multihead_attention import MultiHeadAttention
 from tensorflow_addons.utils import test_utils
 
 
+def test_output_shape():
+    batch_size = 10
+    num_heads = 8
+    head_size = 12
+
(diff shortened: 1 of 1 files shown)


cgarciae:
@gabrieldemarmiesse why is `test_output_shape` now outside the class?

cgarciae:
Ah, its just an example? The idea is that we move all test outside without the decorator?

gabrieldemarmiesse:
Long story short, `run_all_in_graph_and_eager_mode` is a private function, so we're trying to get rid of it. It also had a behavior that makes us write tests in the TF 1.x style, which is not very user friendly. We use fixtures as alternative.

cgarciae:
If I read correctly, the idea would be to use `pytest.mark.usefixtures("maybe_run_functions_eagerly")` over each test that previously relied on `run_all_in_graph_and_eager_mode`?

gabrieldemarmiesse:
Not necessarily. By default, tensorflow runs in eager mode. So the tests will run in eager mode. If there are some tf.function, then what is inside runs in graph mode. 

It depends on what is the objective of the test, and what is the feature you're testing. 
If the code written in a layer uses no strange control flow, like for loops or if on tensors, we don't need to run the tests in graph mode. 

If we see that there are some pieces of code where the control flow is tricky, we need to decorate with a tf.function (possibly in the test even, it's not necessary to expose a tf.function to the user) and then we use the fixture `maybe_run_functions_eagerly`. The fixture will run the test twice. Once normally, and once with all the tf.functions disabled (so full eager mode).

We need to remember that by checking with tf.function and eager, we double the running time of the test. So we shouldn't use it everywhere "just to be safe". Having fast tests is very important for the developer's experience.

We're going to add more stuff in the contributing about the fixtures that contributors can use in the tests.

cgarciae:
I see. I mainly see the value of testing on both modes because of `Tensor.shape` behaving differently between eager and graph mode. 

Will address this tonight :)

gabrieldemarmiesse:
I can add the decorator for this one, to ensure that the `call` method works in graph and eager mode.

PR #1439, opened 2020-03-26
Added multiple useful fixtures.

gabrieldemarmiesse:
I added two fixtures, one to run the test two times, one with CPU and one with GPU.

The other fixure I added is when a test is needed to run multiple times with `channels_first` and `channels_last`.

To use fixtures, either use `@pytest.mark.usefixture` for fixtures that don't provide parameters (like gpu and cpu) or you can use directly the name of the function as parameter for the test. Here `data_format`.

More about fixtures: https://docs.pytest.org/en/latest/fixture.html

### DIFF
--- a/tensorflow_addons/layers/optical_flow_test.py
+++ b/tensorflow_addons/layers/optical_flow_test.py
@@ -74,51 +74,50 @@ def _create_test_data(data_format):
     return val_a, val_b
 
 
+@pytest.mark.usefixtures("cpu_and_gpu", "maybe_run_functions_eagerly")
+def test_forward_simple(data_format):
+    # We are just testing where the output has vanishing values.
+    val_a, val_b = _create_test_data(data_format)
+    input_a = tf.constant(val_a, dtype=tf.float32)
+    input_b = tf.constant(val_b, dtype=tf.float32)
+
+    input_a_tensor = tf.convert_to_tensor(input_a, dtype=tf.float32)
+    input_b_tensor = tf.convert_to_tensor(input_b, dtype=tf.float32)
+
+    kernel_size = 1
+    max_displacement = 2
+    stride_1 = 1
+    stride_2 = 2
+    pad = 4
+
+    actual = _forward(
+        input_a_tensor,
+        input_b_tensor,
+        kernel_size=kernel_size,
+        max_displacement=max_displacement,
+        stride_1=stride_1,
+        stride_2=stride_2,
+        pad=pad,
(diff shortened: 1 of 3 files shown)


tensorflow_addons/layers/optical_flow_test.py
```diff
@@ -74,51 +74,50 @@ def _create_test_data(data_format):
...
-            input_a_tensor = tf.convert_to_tensor(input_a, dtype=tf.float32)
-            input_b_tensor = tf.convert_to_tensor(input_b, dtype=tf.float32)
-
-            kernel_size = 1
-            max_displacement = 2
-            stride_1 = 1
-            stride_2 = 2
-            pad = 4
+@pytest.mark.usefixtures("cpu_and_gpu", "maybe_run_functions_eagerly")
+def test_forward_simple(data_format):
```
seanpmorgan: I can't say I find it very intuitive that this is a shared fixture being loaded from conftest. If the documentation will be expanded to describe the pytest method I'm okay with it though.

tensorflow_addons/utils/test_utils.py
```diff
@@ -177,6 +177,23 @@ def maybe_run_functions_eagerly(request):
...
+            yield
+    else:
+        if not tf.test.is_gpu_available():
+            pytest.skip("GPU is not available.")
+        with tf.device("/device:GPU:0"):
+            yield
+
+
+@pytest.fixture(scope="function", params=["channels_first", "channels_last"])
+def data_format(request):
```
seanpmorgan: Ugh, it's a bit unfortunate that `data_format` is the tf / keras terminology for this. MXNet's `layout` or Caffe2's `order` better describe the variable IMO.

No sense in breaking what's established, just wanted to complain :P

tensorflow_addons/layers/optical_flow_test.py
gabrieldemarmiesse: Yes definitly. Pytest has many shortcuts that are impossible to understand unless you read the docs. I can't say I agree with all the design choices, but to me there is no doubt that it's much easier to use than unittest.

We should have a section in the CONTRIBUTING.md explaining this machinery. I agree, it's work in progress.

tensorflow_addons/utils/test_utils.py
gabrieldemarmiesse: Haha, I never noticed. We just have to wait tensorflow 3.0 to change it :p

Issue #361, opened 2019-07-18
Request for example: Focal Loss


seanpmorgan:
It would be nice to have an example for how to use focal loss in an object detection pipeline.

Resources:
https://arxiv.org/pdf/1708.02002.pdf

SSaishruthi:
Would like to work on this

AakashKumarNain:
Not exactly object detection but as this one is a binary classification problem, so the same concept applies here. Check this out: https://www.kaggle.com/aakashnain/diving-deep-into-focal-loss

SSaishruthi:
Thanks @AakashKumarNain

AakashKumarNain:
You are welcome @SSaishruthi . I wanted to add an example when I submitted the `PR` for focal loss but for putting object detection as an example with focal loss requires a little verbose example with many moving parts, so I parked it for later at that point in time

SSaishruthi:
I have a placeholder example PR for this. Will keep experimenting with different use cases. Thanks for the input

bhack:
@seanpmorgan It could be also nice f the Tensorflow google team will start to contribute a little bit in Tensorflow addons repo instead of accumulate single person/small team code snippets in the new just announced Model Garden or in the Google research repos. 
See i.e. https://github.com/tensorflow/models/blob/master/official/vision/detection/modeling/losses.py#L25.
I don't know if we could open a general issue about this topic cause I've mentioned the same things on other tickets.

PR #1440, opened 2020-03-26
Add focal loss tutorial.

us:
Related issue: #361

### DIFF
--- a/docs/tutorials/losses_focalloss.ipynb
+++ b/docs/tutorials/losses_focalloss.ipynb
@@ -1,0 +1,420 @@
+{
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata": {
+    "accelerator": "GPU",
+    "colab": {
+      "name": "Losses_SigmoidFocalCrossEntropy.ipynb",
+      "provenance": [],
+      "collapsed_sections": [],
+      "toc_visible": true,
+      "include_colab_link": true
+    },
+    "kernelspec": {
+      "name": "python3",
+      "display_name": "Python 3"
+    },
+    "language_info": {
+      "codemirror_mode": {
+        "name": "ipython",
+        "version": 3
+      },
+      "file_extension": ".py",
+      "mimetype": "text/x-python",
+      "name": "python",
+      "nbconvert_exporter": "python",
+      "pygments_lexer": "ipython3",
+      "version": "3.6.3"
+    }
+  },
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "view-in-github",
+        "colab_type": "text"
+      },
+      "source": [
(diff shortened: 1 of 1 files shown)


Squadrick:
IIRC, focal loss is specifically formulated to help out for detection tasks (like PASCAL). It was also originally introduced for RetinaNet which wasn't doing simple classification. 

From the paper:

> The Focal Loss is designed to address the one-stage object detection scenario in which there is an extreme imbalance between foreground and background classes during training (e.g., 1:1000).

A tutorial on MNIST classification isn't suitable. It's better to do a simple detection task using a single short detector, or if that's too complex, we can artificially imbalance the dataset to show where focal loss really shines.

us:
Can we try on mnist that is labels are 1 or not.

bhack:
I agree with @squadrick. I think an easy single stage anchor free object detector could be simple enough for a focal loss tutorial.

I.e. see focal loss in https://github.com/xuannianz/keras-CenterNet

us:
@Squadrick @bhack I changed the dataset, now MNIST have just 3 categories 0, 1, and others, so dataset is imbalanced and loss graph are same as graph on paper. 
Training bigger model will cost more time and colab is inefficient for this.(continuous timeout errors, etc.)

docs/tutorials/losses_focalloss.ipynb
```diff
@@ -0,0 +1,467 @@
+{
```
ashutosh1919, line 1: <p>It is good to add <code>compile()</code>  and <code>fit()</code>  in <code>Train and Evaluate</code>  and not in <code>Building The Model</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/addons/pull/1440/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

aaronmondal:
Tbh I feel like this is may be an unnecessary tutorial.

1. From a 'code-point-of-view' the usage of the focal loss is essentially just swapping `sigmoid_cross_entropy_with_logits` against `sigmoid_focal_crossentropy` anyways. The mathematical intricacies which justify using the focal loss over regular crossentropy are not addressed besides the copy of the `SigmoidFocalCrossentropy` docstring.
2. The `tfa.__version__` is outdated by over half a year.
3. IMO one of the mistakes one could make when applying `SigmoidFocalCrossentropy` is wrong scaling of the model outputs depending on `from_logits`. This is not addressed in the tutorial.
4. The parameter `alpha` is not explained and the image from the paper omits `alpha` as well, hence the equation from the image does not actually represent the calculations performed by `SigmoidFocalCrossentropy`.
5. Gradient instability for small `gamma` is not addressed.
5. The `create_model` method in the tutorial defines a `Sequential` model, then builds and trains it as well, i.e. it does not just "create a model". I would also argue that rebuilding the entire model just for the sake of changing a loss hyper-parameter is not good practice, as this could be achieved by re-compiling the model as well.

I agree with @bhack that the most prominent use case for the focal loss is object detection. A tutorial on object detection which *makes use* of the focal loss would probably be a lot more informative. In that case it would also not be such a big problem to be less mathematically precise about the formulation of the focal loss itself.

bhack:
We have now a focal loss in Keras CV. I suggest you to port this in the Keras io repository:
https://github.com/keras-team/keras-io

PR #1452, opened 2020-03-27
add sharpness

abhichou4:
See #1333


tensorflow_addons/image/sharpness_op.py
```diff
@@ -0,0 +1,52 @@
...
+# limitations under the License.
+# ============================================================================
+"""Sharpness Op for color ops"""
+
+import tensorflow as tf
+
+from tensorflow_addons.image.compose_ops import blend
+
+
+def sharpness(image, factor):
```
Squadrick: Why not do something similar to `equalize` ops? 

Have a function called `sharpness_image` that operates on a single image. Then do a `tf.map_fn` in `sharpness` after converting to 4D?

Squadrick:
Also, include the reference implementation (PIL, scikit-image, opencv) from henceforth.

tensorflow_addons/image/sharpness_op.py
abhichou4: I haven't handled different `data_formats`  here as `NCHW` is not implemented in `depthwise_conv2d` for CPU. Batching has been added.

tensorflow_addons/image/tests/color_ops_test.py
```diff
@@ -53,3 +53,24 @@ def test_equalize_channel_first(shape):
...
+@pytest.mark.parametrize("dtype", _DTYPES)
+@pytest.mark.parametrize("shape", [(5, 5, 3), (10, 5, 5, 3)])
+def test_sharpness_dtype_shape(dtype, shape):
+    image = np.ones(shape=shape, dtype=dtype)
+    sharp = color_ops.sharpness(tf.constant(image), 0).numpy()
+    np.testing.assert_equal(sharp, image)
+    assert sharp.dtype == image.dtype
+
+
+@pytest.mark.usefixtures("maybe_run_functions_eagerly")
```
gabrieldemarmiesse, line 67: This is not needed because sharpness is not a function making use of control flow depending on tensors values.

tensorflow_addons/image/tests/color_ops_test.py
```diff
@@ -53,3 +53,24 @@ def test_equalize_channel_first(shape):
...
+def test_sharpness_with_PIL():
```
gabrieldemarmiesse, line 68: Please add a `@pytest.mark.parametrize("factor", [0, 0.25, 0.5, 0.75, 1])`. This will run the test multiple times.

gabrieldemarmiesse:
Awesome, thanks!

PR #1453, opened 2020-03-27
add dtypes to blend

abhichou4:
Small change so operations using `blend` don't have to cast their results to `tf.uint8` before using it. (#1452)

### DIFF
--- a/tensorflow_addons/image/compose_ops_test.py
+++ b/tensorflow_addons/image/compose_ops_test.py
@@ -44,6 +44,7 @@ def blend_np(image1, image2, factor):
     return temp
 
 
+@pytest.mark.usefixtures("maybe_run_functions_eagerly")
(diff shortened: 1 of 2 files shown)


abhichou4:
@gabrieldemarmiesse everything handled in `tf.float32` now and it's also the return `dtype`. Any thoughts?

tensorflow_addons/image/compose_ops_test.py
```diff
@@ -64,13 +70,16 @@ def test_blend(dtype):
         ],
     )
 
-    image1 = np.random.randint(0, 255, (4, 4, 3), np.uint8)
-    image2 = np.random.randint(0, 255, (4, 4, 3), np.uint8)
+    np.random.seed(0)
```
Squadrick: Need to seed `tf` as well.

PR #1454, opened 2020-03-27
Moved test out of run_in_graph_and_eager_mode in optical flow.

gabrieldemarmiesse:


tensorflow_addons/layers/optical_flow_test.py
```diff
@@ -116,102 +115,71 @@ def test_forward_simple(data_format):
...
+
+@pytest.mark.usefixtures("cpu_and_gpu")
+def test_keras(data_format):
+    # Unable to use `layer_test` as this layer has multiple inputs.
+    val_a, val_b = _create_test_data(data_format)
+
+    input_a = tf.keras.Input(shape=val_a.shape[1:])
+    input_b = tf.keras.Input(shape=val_b.shape[1:])
+
+    layer = CorrelationCost(
```
AakashKumarNain: Although it doesn't matter here, still it would be much better if you can rename `layer` to something `corr_cost_layer`. Again, just a suggestion.

Issue #1458, opened 2020-03-27
Make our CI build wheels against TF 2.1 and 2.2


gabrieldemarmiesse:
**Describe the feature and the current behavior/state.**

Currently in the CI, we test against tf 2.2 only on linux with python 3.5 and without releasing the wheels.

We should build wheels against 2.1 and 2.2, and push the wheels for tf 2.1 to pypi, then when 2.2 is stable, we should start pushing the wheels for tf 2.2 to pypi.

**Relevant information**
- Are you willing to contribute it (yes/no): yes
- Are you willing to maintain it going forward? (yes/no): yes
- Is there a relevant academic paper? (if so, where): no
- Is there already an implementation in another framework? (if so, where): no
- Was it part of tf.contrib? (if so, where): no

**Which API type would this fall under (layer, metric, optimizer, etc.)**

CI

**Who will benefit with this feature?**

Users who will be able to use the python part of addons with tf 2.1 and 2.2.

**Any other info.**

There are two parts to this issue, one that we can do now and one where we may have to wait for the custom ops docker image to be released for tf 2.2.

seanpmorgan:
Hmmm would like to carfefully think about how we version this? 0.8.x pinned to TF2.1 and 0.9.x pinned to TF2.2? How long do we support older TF versions?

gabrieldemarmiesse:
That may be a topic to add to the meeting next week. The way I see it:

>  0.8.x pinned to TF2.1 and 0.9.x pinned to TF2.2?

We have two things here. Python API compatibility and ABI compatibility. 

* The ABI is compatible with a single wheel not even a full tensorflow version because when building tf from source or when using a nightly, it's not garanteed that it works. Even TF increasing the minor version can potentially break the ABI. We can see that we can't use the ABI compatibility to dictate what we support overall. It's way too strict.
* The python API compatibility is a lot easier to respect. As long as we use the TF public API, we'll be forward compatible, even with nightlies as long as TF doesn't introduce bugs. For backward compatibility, my take is that as long as it's not painful, we should keep it. Meaning, if someone comes forward and wants to merge a feature that requires a version of tensorflow strictly higher than the oldest one we support, that may be a good time to discuss dropping support for an older version of TF.

Using this principle, we could support way more than 2 TensorFlow versions, it's just more jobs in the build matrix. Similar to the versions of python we support. 

[...]

PR #1462, opened 2020-03-27
Removed the automatic install of python packages in configure.py

gabrieldemarmiesse:
This makes little sense to install TF automatically in the configure.py since we're going to support multiple versions soon.

It also make little sense to install the content of requirements.txt since it's done already when doing `pip install (-e) ./`

First step towards #1458

I had some strange issue with the finish_bazel_install so I removed it. We're going to assume that devs have buildkit available locally for the sanity check and running the tests docker images.

For a list of what is possible with the new dockerfile syntax, see https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/experimental.md


CONTRIBUTING.md
```diff
@@ -244,7 +244,6 @@ If you need a custom C++/Cuda op for your test, compile your ops with
 
 ```bash
 python configure.py
-python configure.py --no-deps   # if you don't want any dependencies installed with pip
 bash tools/install_so_files.sh  # Linux/macos/WSL2
 sh tools/install_so_files.sh    # PowerShell
 ```
```
seanpmorgan: The instructions for running within the docker container yourself probably should be updated to let them know they need to install tensorflow?
```
(addons-nightly) sean-morgan@ubuntu-spmorgan:~/code/tf/seanpmorgan/addons$ docker run --rm -it -v ${PWD}:/addons -w /addons tensorflow/tensorflow:2.1.0-custom-op-ubuntu16
root@d1c464db8012:/addons# python3
Python 3.6.10 (default, Dec 19 2019, 23:04:32) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tensorflow as tf
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'tensorflow'
```

seanpmorgan: This may be less of an issue once we publish a docker developer image

CONTRIBUTING.md
```diff
@@ -305,8 +306,9 @@ quickly, as Bazel has great support for caching and distributed testing.
 To test with Bazel:
 
 ```
-python configure.py
-pip install pytest
+python3 -m pip install tensorflow==2.1.0
+python3 configure.py
+python3-m pip install pytest
```
seanpmorgan:
```suggestion
python3 -m pip install pytest
```

seanpmorgan:
@googlebot I consent.

PR #1463, opened 2020-03-27
Moved test out of run_in_graph_and_eager_mode in giou

autoih:
See #1328

cc @gabrieldemarmiesse

### DIFF
--- a/tensorflow_addons/losses/giou_loss_test.py
+++ b/tensorflow_addons/losses/giou_loss_test.py
@@ -17,7 +17,6 @@
 import sys
 
 import pytest
-from absl.testing import parameterized
 
 import numpy as np
 import tensorflow as tf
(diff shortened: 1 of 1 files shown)


tensorflow_addons/losses/giou_loss_test.py
```diff
@@ -40,62 +39,35 @@ def test_iou(dtype):
...
-        boxes2 = tf.constant([3.0, 4.0, 6.0, 8.0], dtype=dtype)
-        expected_result = tf.constant(1.07500000298023224, dtype=dtype)
-        loss = giou_loss(boxes1, boxes2)
-        self.assertAllCloseAccordingToType(loss, expected_result)
+    loss = giou_loss(boxes1, boxes2)
+    test_utils.assert_allclose_according_to_type(loss, expected_result)
+
+
+@pytest.mark.parametrize("dtype", [np.float16, np.float32, np.float64])
+def test_different_shapes(dtype):
```
gabrieldemarmiesse: It seems a part of the test is missing. Could you run the tests also with the boxes without expand_dims?

autoih: Thanks @gabrieldemarmiesse, but without having expand_dims should already in `test_giou_loss`?

gabrieldemarmiesse: You're right. My bad.

PR #1464, opened 2020-03-27
Moved tests out of run_in_graph_and_eager_mode in contrastive_test

autoih:
cc @gabrieldemarmiesse

### DIFF
--- a/tensorflow_addons/losses/contrastive_test.py
+++ b/tensorflow_addons/losses/contrastive_test.py
@@ -135,10 +135,8 @@ def test_no_reduction():
     #         1.3^2, 0.7^2, max(1 - 1.1, 0)^2, 0.5^2]
     #      = [0.9^2, 0.7^2, 1.3^2, 0.7^2, 0^2, 0.5^2]
     #      = [0.81, 0.49, 1.69, 0.49, 0, 0.25]
-    
-    np.testing.assert_allclose(
-        loss, [0.81, 0.49, 1.69, 0.49, 0.0, 0.25], rtol=1e-5
-    )
+
+    np.testing.assert_allclose(loss, [0.81, 0.49, 1.69, 0.49, 0.0, 0.25], rtol=1e-5)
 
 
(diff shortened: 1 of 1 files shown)


tensorflow_addons/losses/contrastive_test.py
```diff
@@ -121,19 +96,49 @@ def test_non_default_margin(self):
...
+    y_pred = tf.constant([0.1, 0.3, 1.3, 0.7, 1.1, 0.5], dtype=tf.dtypes.float32)
+    loss = cl_obj(y_true, y_pred)
+
+    # Loss = y * (y`)^2 + (1 - y) * (max(m - y`, 0))^2
+    #      = [max(1 - 0.1, 0)^2, max(1 - 0.3, 0)^2,
+    #         1.3^2, 0.7^2, max(1 - 1.1, 0)^2, 0.5^2]
+    #      = [0.9^2, 0.7^2, 1.3^2, 0.7^2, 0^2, 0.5^2]
+    #      = [0.81, 0.49, 1.69, 0.49, 0, 0.25]
+
+    np.testing.assert_array_almost_equal(
```
gabrieldemarmiesse: If it doesn't work with `np.testing.asser_allclose` it may be because self.assertAllClose and `np.testing.assert_allclose` don't have the same default `atol` and `rtol`. You just have to change those parameters to match the ones in `self.assertAllclose`.

autoih: Thanks @gabrieldemarmiesse, since the test is about `Reduction.NONE`, which output is a list/array. What do you think?

gabrieldemarmiesse: I'm not sure I understand what you mean. Can you detail why you chose `assert_array_almost_equal` instead of `assert_allclose`?

autoih: I corrected it by `assert_allclose`. Thanks for the reminder, @gabrieldemarmiesse.

PR #1478, opened 2020-03-29
Add tf 2.2.0rc1 in the build matrix.

### LINKED ISSUE #1458: Make our CI build wheels against TF 2.1 and 2.2

gabrieldemarmiesse:
Needs #1462 to be merged first

Related to #1458

I didn't add python 3.8 in the matrix yet, it can be done separately.


bhack:
Or `rc2`?

gabrieldemarmiesse:
This is planned.

.github/workflows/release.yml
```diff
@@ -39,27 +39,28 @@ jobs:
...
       - if: github.event_name == 'push'
         run: echo "::set-env name=NIGHTLY_FLAG::--nightly"
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: ${{ matrix.py-version }}
       - name: Build wheels
-        if: github.event_name == 'push' || github.event_name == 'release' || matrix.python-version == env.OLDEST_PY_VERSION
+        if: (matrix.os != 'ubuntu-18.04' || matrix.tf-version != '2.2.0rc1') && (github.event_name == 'push' || github.event_name == 'release' || matrix.py-version == env.OLDEST_PY_VERSION)
```
seanpmorgan: Could you explain why we're skipping ubuntu+2.2?
image

gabrieldemarmiesse: Because we don't have the docker image to build it yet :)

seanpmorgan: Ah gotcha. So the docker container for 2.2.0 won't have any changes other than pre-installed TF. GCC and CUDA versions are the same so removing this should work.

I guess the question is how we want to manage this. Would you prefer we wait until a 2.2.0 docker image is released to build/test so that it is explicit?

gabrieldemarmiesse: I prefer to err on the side of caution concerning the CI. Breaking the master branch means nobody can make new pull request and have the CI displaying meaningful results. I know it's very unlikely that we break it by using the 2.1.0 image. But there is very little benefit with working on this before the official image is released. I'd prefer to merge this and when the official image is released, we remove the exception in the yaml and fix potential incompatibilities. This new image will have python 3.8, will be possibly bigger and burst our disk size limit in github actions, I'd prefer we do it one step at a time.

seanpmorgan: Agree

tools/docker/build_wheel.Dockerfile
```diff
@@ -7,8 +8,9 @@ ARG PY_VERSION
 RUN python$PY_VERSION -m pip install --upgrade pip setuptools auditwheel==2.0.0
 
 COPY tools/install_deps/ /install_deps
+ARG TF_VERSION
```
seanpmorgan: Rationale for two `ARG TF_VERSION`?

gabrieldemarmiesse: When using an `ARG` before a from, the ARG is not valid afterwards and needs to be declared again from what I remember. Let me find a piece of docs about that.

gabrieldemarmiesse: See https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

PR #1482, opened 2020-03-29
Add a warning when trying to load a custom op with a uncompatible version of TF.

gabrieldemarmiesse:


gabrieldemarmiesse:
I believe we can add more info in the readme. Let's not bload the warning too much, it's already very descriptive.

tensorflow_addons/utils/resource_loader.py
```diff
@@ -46,5 +52,54 @@ def __init__(self, relative_path):
...
+            "This means that you might get segfaults when loading the custom op, "
+            "or other kind of low-level errors.\n If you do, do not file an issue "
+            "on Github. This is a known limitation."
+            "\n\n"
+            "It might help you to fallback to pure Python "
+            "ops with TF_ADDONS_PY_OPS . To do that, see "
+            "https://github.com/tensorflow/addons#gpucpu-custom-ops "
+            "\n\n"
+            "You can also change the TensorFlow version installed on your system. "
+            "You would need a TensorFlow version equal to or above {} and strictly "
```
seanpmorgan, line 79: Can you add "installed from pip" somewhere in there. I suspect that is just as likely of an issue as incompatible version

gabrieldemarmiesse: I'll add it. But this warning won't appear if someone is building from source. I don't know if there is any way we can detect that. If TF was exposing an API to get info about the build, we could use that. As far as I know, such an API doesn't exist.

PR #1498, opened 2020-03-30
Moved tests out of run_in_graph_and_eager_mode in normalizatoin_test

autoih:
cc @gabrieldemarmiesse


gabrieldemarmiesse:
Thanks again, that's a great help, if I had to do it alone, it would take forever!

autoih:
My pleasure. It's a very good idea using pytest fixture, thanks for the solution.

PR #1499, opened 2020-03-30
Remove the need to declare the tests in the build file in activations/

gabrieldemarmiesse:
Given that all the `tensorflow_addons/activations` tests runs under 3 seconds, it's maybe not needed for bazel users to select exactly which one they want to run.


AakashKumarNain:
Agreed. If the time taken to check all is only 3 seconds, then it is much better to test them all rather than assuming than letting the user figure out how to run test for a single file

PR #1503, opened 2020-03-30
Add python 3.8 to the build matrix

gabrieldemarmiesse:


gabrieldemarmiesse:
It seems to work. You can take a look at https://github.com/tensorflow/addons/pull/1506 which is the same PR but the CI is building everything.

PR #1506, opened 2020-03-30
[do not merge] Try enable all

gabrieldemarmiesse:


seanpmorgan:
Out of curiosity mind removing the linux check so we can see if there are issues building 2.2 in the container. Don't want to proceed with that but interested if there are fails.

gabrieldemarmiesse:
I don't know why this is not working :( We can do more tests later on.

PR #1509, opened 2020-03-31
Fix unhashable decay_var_list

theoldfather:
Fixes an error where the list of decay variables is unhashable because because they are compared without `var.ref()` or `var.experimental_ref()`. This patch is forward compatible with TF 2.2.0


theoldfather:
@googlebot I signed it!

PR #1510, opened 2020-03-31
Fix unhashable decay_var_list

theoldfather:
Fixes an error where the list of decay variables is unhashable because they are compared without var.ref() or var.experimental_ref(). This patch is forward compatible with TF 2.2.0.

### DIFF
--- a/tensorflow_addons/optimizers/weight_decay_optimizers.py
+++ b/tensorflow_addons/optimizers/weight_decay_optimizers.py
@@ -26,7 +26,7 @@ def _ref(var):
     if release < "2.0.0":
         return var
     else:
-        return var.ref() if hasattr(var,"ref") else var.experimental_ref()
+        return var.ref() if hasattr(var, "ref") else var.experimental_ref()
 
 
 class DecoupledWeightDecayExtension:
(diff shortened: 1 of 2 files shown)


tensorflow_addons/optimizers/weight_decay_optimizers.py
```diff
@@ -21,6 +21,14 @@
 from typing import Union, Callable, Type
 
 
+def _ref(var):
```
PhilJd: tf.addons supports TF2 only anyway, maybe we can remove this string comparison and just return var.ref()/experimental_ref()?

gabrieldemarmiesse:
@theoldfather could you provide a test in your pull request to ensure that we don't re-introduce the bug in the future?

Ideally this test should fail without your patch.

tensorflow_addons/optimizers/weight_decay_optimizers.py
theoldfather: Done.

theoldfather:
> @theoldfather could you provide a test in your pull request to ensure that we don't re-introduce the bug in the future?
> 
> Ideally this test should fail without your patch.

Done.

gabrieldemarmiesse:
@theoldfather  thank you! I'll take a look.

theoldfather:
Thanks for reviewing!

PR #1520, opened 2020-04-01
Removed run_in_graph_and_eager_mode in npairs_test

autoih:
cc @gabrieldemarmiesse


tensorflow_addons/losses/npairs_test.py
```diff
@@ -16,139 +16,149 @@
...
 import pytest
 import tensorflow as tf
 from tensorflow_addons.losses import npairs
-from tensorflow_addons.utils import test_utils
 
 IS_WINDOWS = platform.system() == "Windows"
 
 
-@unittest.skipIf(
+@pytest.mark.skipif(
```
gabrieldemarmiesse, line 28: rather than using `pytest.mark.skipif` all the time, you can use `pytestmark = pytest.mark.skipif(...)` at the top of the file. See http://doc.pytest.org/en/latest/example/markers.html#marking-whole-classes-or-modules

autoih: Thanks for pointing this out, @gabrieldemarmiesse. It's very convenient.

gabrieldemarmiesse:
@googlebot I consent.

Issue #1316, opened 2020-03-15
Monitoring the speed of the tests.


gabrieldemarmiesse:
**Describe the feature and the current behavior/state.**

The CI should fail when the tests are not fast enough to prevent merging tests that are too time consuming. 

Technically, with pytest it's easy, there are even pugins to add timeouts. But we cannot take this implementation.

The problem is that test times are not deterministic. They vary from one run to another. So merging a pull request with the CI passing (and timeout implemented) does not garantee that the CI will continue to pass in the future.

Ideally, we should have the tests times of the first commit of the pull request branch and compare that to the new test times in the pull request, and make the CI fail if we notice that some tests have been changed/added and the running time for those tests increased too much. So we should put a limit on the difference of the running times, not the test times themselves. No idea how to implement that though.

**Relevant information**
- Are you willing to contribute it (yes/no): yes
- Are you willing to maintain it going forward? (yes/no): yes
- Is there a relevant academic paper? (if so, where): no
- Is there already an implementation in another framework? (if so, where): I wish I knew. Does anyone know an open source project where the test times are monitored?
- Was it part of tf.contrib? (if so, where): no

**Which API type would this fall under (layer, metric, optimizer, etc.)**

testing, no API

**Who will benefit with this feature?**

The reviewers, they wouldn't have to check the running times of the tests added by the contributors.

**Any other info.**

nope

PR #1524, opened 2020-04-01
Users don't need to declare the tests now in the image submodule.

gabrieldemarmiesse:
Follow up on #1499


gabrieldemarmiesse:
Pro tip: To avoid the bot spamming everyone, do the modifications in the second commit :p

tensorflow_addons/image/BUILD
```diff
@@ -36,169 +36,10 @@ filegroup(
 )
 
 py_test(
-    name = "color_ops_test",
+    name = "image_test",
     size = "small",
```
seanpmorgan, line 40: So placing all tests under a single "size" means that we'll restrict tests to < 1min of test time. Do we think that is sufficient for https://github.com/tensorflow/addons/issues/1316 or would we like tighter bounds?

PR #1525, opened 2020-04-01
Use a minimal docker image to build wheels. Fix the master branch build.

gabrieldemarmiesse:
@seanpmorgan 

That's still annoying because I wanted to add a `--mount=type=cache,dest=/root/.cache/pip` but It seems like it won't be possible.


gabrieldemarmiesse:
Well well well

gabrieldemarmiesse:
It seems that it won't work, even if we don't install the pytest dependencies (scikit-image and such).

seanpmorgan:
@googlebot I consent.

gabrieldemarmiesse:
Still missing #1527 to totally fix the master branch.

PR #1536, opened 2020-04-02
Pin pypi github action

seanpmorgan:
We've been getting a lot of intermittent failures from our nightly publishing:
https://github.com/tensorflow/addons/runs/553687744?check_suite_focus=true

I'm not sure this is the cause since we're getting a 502 error, but still a better practice to pin this action version.


gabrieldemarmiesse:
I believe that the internet is kind of broken these days. We've had issues with dockerhub and pypi downloads too.

Issue #838, opened 2020-01-06
Build npairs on windows


seanpmorgan:
Npairs can not run on windows due to no OpKernel registered `SparseMatMul`. This appears to have a related issue https://github.com/tensorflow/tensorflow/issues/22699 which has no resolution.

```
tensorflow.python.framework.errors_impl.InvalidArgumentError: No OpKernel was registered to support Op 'SparseMatMul' used by {{node MatMul}}with these attrs: [transpose_b=true, Ta=DT_FLOAT, Tb=DT_FLOAT, b_is_sparse=true, a_is_sparse=true, transpose_a=false]
Registered devices: [CPU]
Registered kernels:
  <no registered kernels>

         [[MatMul]]
         [[npairs_multilabel_loss/PartitionedCall]]

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\SeanM\Miniconda3\lib\site-packages\tensorflow_core\python\framework\test_util.py", line 1111, in decorated
    f(self, *args, **kwargs)
  File "\\?\C:\Users\SeanM\AppData\Local\Temp\Bazel.runfiles__8x3qnhl\runfiles\__main__\tensorflow_addons\losses\npairs_test.py", line 89, in test_single_label
    self.assertAllClose(loss, 0.253856)
  File "C:\Users\SeanM\Miniconda3\lib\site-packages\tensorflow_core\python\framework\test_util.py", line 1153, in decorated
    return f(*args, **kwds)
  File "C:\Users\SeanM\Miniconda3\lib\site-packages\tensorflow_core\python\framework\test_util.py", line 2495, in assertAllClose
    self._assertAllCloseRecursive(a, b, rtol=rtol, atol=atol, msg=msg)
  File "C:\Users\SeanM\Miniconda3\lib\site-packages\tensorflow_core\python\framework\test_util.py", line 2462, in _assertAllCloseRecursive
    (path_str, path_str, msg)))
  File "C:\Users\SeanM\Miniconda3\lib\site-packages\tensorflow_core\python\framework\test_util.py", line 2355, in _assertArrayLikeAllClose
    a = self._GetNdArray(a)
  File "C:\Users\SeanM\Miniconda3\lib\site-packages\tensorflow_core\python\framework\test_util.py", line 2349, in _GetNdArray
    a = self.evaluate(a)
[...]

PR #1539, opened 2020-04-02
[WIP] Check npairs

seanpmorgan:
Just want to confirm #838 is still a bug since moving away from graph mode testing. Also want to test it on tf2.2


seanpmorgan:
Looks like the error has been fixed in 2.2. Will re-open this once we've made the switch to >= 2.2 on master.

PR #1544, opened 2020-04-02
Add checks for code quality: usage of tf.compat.v1 and experimental APIs

### LINKED ISSUE #1279: Add tests to ensure that we don't use experimental/private/deprecated tensorflow APIs

gabrieldemarmiesse:
This should save us some work during code reviews. 

We might want to discuss for this policy about experimental APIs though. See https://github.com/tensorflow/community/pull/218
Should we be more permissive? Less permissive?

Closes  #1279

### DIFF
--- a/tools/testing/source_code_test.py
+++ b/tools/testing/source_code_test.py
@@ -65,32 +65,22 @@ def test_case_insensitive_filesystems():
             )
 
 
-def test_no_private_tf_api():
+def get_lines_of_source_code():
     source_dir = os.path.join(BASE_DIR, "tensorflow_addons")
     for path in glob.glob(source_dir + "/**/*.py", recursive=True):
-        if in_blacklist_private_api(path):
-            continue
         with open(path) as f:
-            for i, line in enumerate(f):
-                if (
-                    "import tensorflow.python" in line
-                    or "from tensorflow.python" in line
-                ):
-                    raise ImportError(
-                        "A private tensorflow API import was found in {} at line {}.\n"
-                        "tensorflow.python refers to TensorFlow's internal source "
-                        "code and private functions/classes.\n"
-                        "The use of those is forbidden in Addons for stability reasons."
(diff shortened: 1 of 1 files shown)


AakashKumarNain:
IMO we should definitely block anything that uses `tf.compat.v1`. If we allow this, we would be in somewhat similar scenario as `py2` -> `py3`. 

Regarding `experiemtal`, although most of the time things remain the same when they come out of experimental policy, IMO we should still defer its usage.

gabrieldemarmiesse:
The reason behing disallowing experimental in user-facing code is to be forward compatible. In a perfect world, since the public API of TF is backward compatible (with the exception of experimental APIs) if we use only the public API, the users should be able to use any version of TF above the recommended one and it will work. 

Experimental API are doomed to have their name changed as some points, which means that we they do, it forces users to update their TF addons version. Users should have to think as little as possible about compatibility between packages.

AakashKumarNain:
I am fine with the changes. @seanpmorgan thoughts?

PR #1545, opened 2020-04-02
Moved test out of run_in_graph_and_eager_mode in sparsemax_loss

autoih:
cc @gabrieldemarmiesse


tensorflow_addons/losses/sparsemax_loss_test.py
```diff
@@ -104,6 +104,33 @@ def test_gradient_against_estimate(dtype):
     test_utils.assert_allclose_according_to_type(jacob_sym, jacob_num)
 
 
+@pytest.mark.parametrize("dtype", ["float32", "float64"])
```
gabrieldemarmiesse, line 107: `_tf_sparsemax_loss` is not a test. As such, the pytest decorator has no effect.

autoih: Ops, my bad, updated.

PR #1546, opened 2020-04-02
Moved test out of run_in_graph_and_eager_mode in conditional_gradient

autoih:
cc @gabrieldemarmiesse

### DIFF
--- a/tensorflow_addons/optimizers/conditional_gradient_test.py
+++ b/tensorflow_addons/optimizers/conditional_gradient_test.py
@@ -14,41 +14,56 @@
 # ==============================================================================
 """Tests for Conditional Gradient."""
 
+import numpy as np
+import pytest
+import platform
+
 import tensorflow as tf
 from tensorflow_addons.utils import test_utils
-import numpy as np
 from tensorflow_addons.optimizers import conditional_gradient as cg_lib
-import platform
+
+
+WINDOWS_CPU = platform.system() == "Windows"
+
+pytestmark = pytest.mark.skipif(
+    WINDOWS_CPU,
+    reason="Doesn't work on Windows CPU version, tf.half is not registered for tf.linalg.svd function on Windows CPU version. So we have to remove tf.half when testing with Windows CPU version. See https://github.com/tensorflow/tensorflow/issues/36764",
+)
+
+
+def _dtypes_to_test(use_gpu):
+    if use_gpu:
+        return [tf.float32, tf.float64]
+    else:
+        return [tf.half, tf.float32, tf.float64]
+
+
+def _dtypes_with_checking_system(use_gpu, system):
+    if system == "Windows":
+        return [tf.float32, tf.float64]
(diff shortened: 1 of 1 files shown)


tensorflow_addons/optimizers/conditional_gradient_test.py
```diff
@@ -14,41 +14,56 @@
...
 from tensorflow_addons.optimizers import conditional_gradient as cg_lib
-import platform
 
 
-@test_utils.run_all_in_graph_and_eager_modes
-class ConditionalGradientTest(tf.test.TestCase):
+WINDOWS_CPU = platform.system() == "Windows"
+
+pytestmark = pytest.mark.skipif(
+    WINDOWS_CPU,
```
gabrieldemarmiesse, line 29: This will skip the whole file on windows. From what I understand, this is not what we want, right? We want to skip only one dtype, which you did in `_dtype_with_checking_system`. Am I missing something?

tensorflow_addons/optimizers/conditional_gradient_test.py
```diff
@@ -14,41 +14,56 @@
...
+
 
-    # Based on issue #347 in the following link,
-    #        "https://github.com/tensorflow/addons/issues/347"
-    # tf.half is not registered for 'ResourceScatterUpdate' OpKernel
-    # for 'GPU' devices.
-    # So we have to remove tf.half when testing with gpu.
-    # The function "_DtypesToTest" is from
-    #       "https://github.com/tensorflow/tensorflow/blob/5d4a6cee737a1dc6c20172a1dc1
-    #        5df10def2df72/tensorflow/python/kernel_tests/conv_ops_3d_test.py#L53-L62"
```
gabrieldemarmiesse, line 34: Can you keep those comments in the new version? I believe they're important.

tensorflow_addons/optimizers/conditional_gradient_test.py
```diff
@@ -14,41 +14,56 @@
...
-        if use_gpu:
-            return [tf.float32, tf.float64]
-        else:
-            return [tf.half, tf.float32, tf.float64]
-
-    # Based on issue #36764 in the following link,
-    #        "https://github.com/tensorflow/tensorflow/issues/36764"
-    # tf.half is not registered for tf.linalg.svd function on Windows
-    # CPU version.
-    # So we have to remove tf.half when testing with Windows CPU version.
```
gabrieldemarmiesse, line 45: Can you also keep this comment in the new version? It's handy to understand the code.

tensorflow_addons/optimizers/conditional_gradient_test.py
autoih: Yes, you are right, thanks for pointing out this, @gabrieldemarmiesse. Updated.

lokhande-vishnu:
The changes look good, thanks.

PR #1562, opened 2020-04-03
Update documentation regarding compatibility

### LINKED ISSUE #1317: Make a better UX around the issue of tensorflow ABI instability.

seanpmorgan:
Closes #1317 along with the warnings that have been merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,11 +63,33 @@ import tensorflow as tf
 import tensorflow_addons as tfa
 ```
 
+### Python Op Compatility
+TensorFlow Addons is actively working towards forward compatiblity with TF2.x. 
+However there are still a few private API uses within the respository so at the moment 
+we can only gurentee compatibility with the version of TF which it was built against. 
+Warnings will be emitted during TFA import if your TensorFlow version does not match 
+what it was built against.
+
+### C++ Op Compatibility
+TensorFlow C++ APIs are not stable and thus we can only guarentee compatibility with the 
+version TFA was built against. It is possible custom-ops will work with multiple 
+versions of TF, but there is also a chance for segfault or other problematic crash.
+Warnings will be emitted during TFA import if your TensorFlow version does not match 
+what it was built against.
+
+Additionally, registering custom-ops does not have an ABI stable interface so it is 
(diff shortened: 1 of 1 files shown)


seanpmorgan:
This needs a slight modification the the Python Op comaptibility. We'll support as many versions as we can but greater than whats in the warning will have no guarentees

README.md
```diff
@@ -63,11 +63,33 @@ import tensorflow as tf
 import tensorflow_addons as tfa
 ```
 
+### Python Op Compatility
+TensorFlow Addons is actively working towards forward compatiblity with TF2.x. 
+However there are still a few private API uses within the respository so at the moment 
+we can only gurentee compatibility with the version of TF which it was built against.
```
gabrieldemarmiesse, line 69:
```suggestion
we can only gurentee compatibility with the version of TF which it was tested against. 
```

We don't want to confuse users with the building of custom ops.

README.md
```diff
@@ -63,11 +63,33 @@ import tensorflow as tf
...
+However there are still a few private API uses within the respository so at the moment 
+we can only gurentee compatibility with the version of TF which it was built against. 
+Warnings will be emitted during TFA import if your TensorFlow version does not match 
+what it was built against.
+
+### C++ Op Compatibility
+TensorFlow C++ APIs are not stable and thus we can only guarentee compatibility with the 
+version TFA was built against. It is possible custom-ops will work with multiple 
+versions of TF, but there is also a chance for segfault or other problematic crash.
+Warnings will be emitted during TFA import if your TensorFlow version does not match
```
gabrieldemarmiesse, line 77:
```suggestion
Warnings will be emitted when loading a custom op if your TensorFlow version does not match 
```

Issue #1267, opened 2020-03-09
Missing argument in apply_gradients() in AdamW optimizer


eoinoreilly30:
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Google Colab
- TensorFlow version and how it was installed (source or binary): tf-nightly 2.2.0-dev20200309 (pip install)
- TensorFlow-Addons version and how it was installed (source or binary): 0.8.3 (pip install)
- Python version: 3.6
- Is GPU used? (yes/no): yes

**Describe the bug**

When running model.compile() with AdamW optimizer, a type error is thrown saying:
``apply_gradients() got an unexpected keyword argument 'all_reduce_sum_gradients'``

This can be fixed by adding in the argument to ``apply_gradients()`` in ``tensorflow_addons/optimizers/weight_decay_optimizers.py``

**Code to reproduce the issue**

https://colab.research.google.com/drive/1A6X8yYii5M8BDqwAFvoFglrTIqWvwQLm

**Other info / logs**

TypeError: in user code:

    /usr/local/lib/python3.6/dist-packages/tensorflow/python/keras/engine/training.py:503 train_function  *
        outputs = self.distribute_strategy.experimental_run_v2(
    /usr/local/lib/python3.6/dist-packages/tensorflow/python/distribute/distribute_lib.py:920 experimental_run_v2  **
        return self._extended.call_for_each_replica(fn, args=args, kwargs=kwargs)
    /usr/local/lib/python3.6/dist-packages/tensorflow/python/distribute/distribute_lib.py:2254 call_for_each_replica
        return self._call_for_each_replica(fn, args, kwargs)
    /usr/local/lib/python3.6/dist-packages/tensorflow/python/distribute/distribute_lib.py:2615 _call_for_each_replica
        return fn(*args, **kwargs)
    /usr/local/lib/python3.6/dist-packages/tensorflow/python/keras/engine/training.py:473 train_step  **
        _minimize(tape, self.optimizer, loss, self.trainable_variables)
    /usr/local/lib/python3.6/dist-packages/tensorflow/python/keras/engine/training.py:1737 _minimize
        all_reduce_sum_gradients=False)

    TypeError: apply_gradients() got an unexpected keyword argument 'all_reduce_sum_gradients'

gabrieldemarmiesse:
You code works with tensorflow==2.1.0 and tfa-nightly.

[...]

PR #1563, opened 2020-04-03
Update maximum version on master

seanpmorgan:
Unable to support TF2.1 and TF2.2 at the same time. See https://github.com/tensorflow/addons/issues/1267


PhilJd:
If this is only due to #1267, I've created PR #1566 to support TF2.1 and TF 2.2 :)

gabrieldemarmiesse:
@PhilJd good idea. 

Let's talk about it at the meeting to ensure we're all on the same page

seanpmorgan:
Yup going to go forward with supporting 2.1 + 2.2

PR #1564, opened 2020-04-03
Moved tests out of run_in_graph_and_eager_mode in sparsemax loss

autoih:
cc @gabrieldemarmiesse

### DIFF
--- a/tensorflow_addons/losses/sparsemax_loss_test.py
+++ b/tensorflow_addons/losses/sparsemax_loss_test.py
@@ -130,15 +130,17 @@ def test_sparsemax_loss_positive(dtype):
     assert np.zeros(test_obs).shape == tf_loss_op.shape
 
 
+@pytest.mark.parametrize("dtype", ["float32", "float64"])
+def _tf_sparsemax(z, dtype):
+    tf_sparsemax_op = sparsemax(z.astype(dtype))
+    tf_sparsemax_out = tf_sparsemax_op
+
+    return tf_sparsemax_op, tf_sparsemax_out
+
+
 @test_utils.run_all_with_types(["float32", "float64"])
 @test_utils.run_all_in_graph_and_eager_modes
 class SparsemaxTest(tf.test.TestCase):
-    def _tf_sparsemax(self, z, dtype):
-        tf_sparsemax_op = sparsemax(z.astype(dtype))
-        tf_sparsemax_out = self.evaluate(tf_sparsemax_op)
-
-        return tf_sparsemax_op, tf_sparsemax_out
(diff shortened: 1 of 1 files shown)


tensorflow_addons/losses/sparsemax_loss_test.py
```diff
@@ -130,15 +130,17 @@ def test_sparsemax_loss_positive(dtype):
     assert np.zeros(test_obs).shape == tf_loss_op.shape
 
 
+@pytest.mark.parametrize("dtype", ["float32", "float64"])
```
gabrieldemarmiesse, line 133: The decorator is not needed here because it's not a test.

tensorflow_addons/losses/sparsemax_loss_test.py
```diff
@@ -130,15 +130,17 @@ def test_sparsemax_loss_positive(dtype):
...
+def _tf_sparsemax(z, dtype):
+    tf_sparsemax_op = sparsemax(z.astype(dtype))
+    tf_sparsemax_out = tf_sparsemax_op
+
+    return tf_sparsemax_op, tf_sparsemax_out
```
gabrieldemarmiesse, line 138: Actually this function is so simple now with eager mode, maybe we could inline (remove) it?

autoih: Thanks @gabrieldemarmiesse. Updated.

tensorflow_addons/losses/sparsemax_loss_test.py
```diff
@@ -204,42 +198,48 @@ def test_sparsemax_loss_of_inf(self, dtype=None):
...
+
+    q = np.zeros((test_obs, 10))
+    q[:, 0] = 1
+
+    tf_loss_op, tf_loss_out = _tf_sparsemax_loss(z, q, dtype)
 
-        self.assertAllCloseAccordingToType(np.zeros(test_obs), tf_loss_out)
-        self.assertShapeEqual(np.zeros(test_obs), tf_loss_op)
+    tf_sparsemax_op = sparsemax(z.astype(dtype))
+    tf_sparsemax_out = tf_sparsemax_op
```
gabrieldemarmiesse, line 236: It's not needed to have two variable to represent the same thing. It made sense in graph mode where we had ops and tensors, but in eager mode it's not needed.

autoih: Got it, thanks @gabrieldemarmiesse

gabrieldemarmiesse:
Great thanks!

PR #1565, opened 2020-04-03
To be consistent with other tests in BUILD

autoih:
cc @gabrieldemarmiesse, please close this, if it's not necessary since it's only one small test here.


gabrieldemarmiesse:
Yeah good thing to do, this way it'll be easier to add more things into this directory.

PR #1571, opened 2020-04-05
Removing tests out of run_in_graph_and_eager_mode in conditional_gradient

autoih:
cc @gabrieldemarmiesse

### DIFF
--- a/tensorflow_addons/optimizers/conditional_gradient_test.py
+++ b/tensorflow_addons/optimizers/conditional_gradient_test.py
@@ -111,8 +111,10 @@ def test_minimize_sparse_resource_variable_frobenius():
         )
 
 
-def do_test_basic_frobenius(dtype, use_resource, use_callable_params=False):
-    # for i, dtype in enumerate([tf.half, tf.float32, tf.float64]):
+@pytest.mark.parametrize("dtype", [(tf.half, 0), (tf.float32, 1), (tf.float64, 2)])
+@pytest.mark.usefixtures("maybe_run_functions_eagerly")
+@pytest.mark.parametrize("use_resource", [True, False])
+def test_basic_frobenius(dtype, use_resource):
     if use_resource:
         var0 = tf.Variable([1.0, 2.0], dtype=dtype[0], name="var0_%d" % dtype[1])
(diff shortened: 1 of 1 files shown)


tensorflow_addons/optimizers/conditional_gradient_test.py
```diff
@@ -67,109 +67,141 @@ def test_like_dist_belief_nuclear_cg01():
...
+            var0.numpy(),
+        )
 
-            if not tf.executing_eagerly():
-                self.assertFalse(slot0 in tf.compat.v1.trainable_variables())
-                self.assertFalse(slot1 in tf.compat.v1.trainable_variables())
 
-            if not tf.executing_eagerly():
-                self.evaluate(cg_update)
+def do_test_basic_frobenius(dtype, use_resource, use_callable_params=False):
```
gabrieldemarmiesse, line 114: You can remove the `use_callable_params` because it's always `True`.

tensorflow_addons/optimizers/conditional_gradient_test.py
```diff
@@ -67,109 +67,141 @@ def test_like_dist_belief_nuclear_cg01():
...
-                        4.0 * 0.5 - (1 - 0.5) * 0.01 * 0.01 / norm1,
-                    ]
-                ),
-                self.evaluate(var1),
-            )
+@pytest.mark.parametrize("dtype", [(tf.half, 0), (tf.float32, 1), (tf.float64, 2)])
+@pytest.mark.usefixtures("maybe_run_functions_eagerly")
+@pytest.mark.parametrize("use_resource", [True, False])
+def test_basic_callable_params_frobenius(dtype, use_resource):
+    do_test_basic_frobenius(dtype, use_resource, use_callable_params=True)
```
gabrieldemarmiesse, line 200: There is no need for a second function `test_basic_callable_params_frobenius` here.

You can rename `do_test_basic_frobenius` to `test_basic_frobenius` and make it the main test.

tensorflow_addons/optimizers/conditional_gradient_test.py
autoih: That makes sense. Thanks @gabrieldemarmiesse.

PR #1573, opened 2020-04-05
Remove the need to declare source files in the BUILD of activations

gabrieldemarmiesse:
I used the feedback of the SIG meeting concerning moving the tests to another directory. I believe I found a nice middle ground, which is actually used by scikit-learn and numpy: 

https://github.com/scikit-learn/scikit-learn/tree/master/sklearn
https://github.com/numpy/numpy/tree/master/numpy

The tests of activations can be found now in `tensorflow_addons/activations/tests`.

This makes a clear distinction between source files and tests files which allows us in turn to use the glob feature of bazel to avoid declaring every python file that we use.

It also makes it easier for users who want to look at implementations to navigate the source code since they don't care about the tests.


AakashKumarNain:
Awesome. Looks perfect!

Issue #1203, opened 2020-03-02
cyclical_learning_rate_test.py is not run by bazel


gabrieldemarmiesse:
**Describe the bug**

Like the other ones, the 
```
if __name__ == "__main__":
    tf.test.main()
```

was missing. Since pytest doesn't need that, the tests failed with pytest.

**Other info / logs**

The error when actually running the tests:

```
proof_of_concept_pytest) /mnt/c/Users/gdemarmi/Desktop/projects/addons $ pytest -v tensorflow_addons/optimizers/cyclical_learning_rate_test.py

/home/gdemarmi/softwares/python/anaconda/lib/python3.7/site-packages/pep8.py:110: FutureWarning: Possible nested set at position 1
  EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')
============================================================= test session starts =============================================================
platform linux -- Python 3.7.5, pytest-5.0.1, py-1.8.0, pluggy-0.13.0 -- /home/gdemarmi/softwares/python/anaconda/bin/python
cachedir: .pytest_cache
rootdir: /mnt/c/Users/gdemarmi/Desktop/projects/addons
plugins: arraydiff-0.3, cov-2.7.1, doctestplus-0.2.0, flake8-1.0.4, forked-1.0.2, openfiles-0.4.0, pep8-1.0.6, remotedata-0.3.2, xdist-1.27.0, typeguard-2.7.1
collected 9 items

tensorflow_addons/optimizers/cyclical_learning_rate_test.py::CyclicalLearningRateTest::testCustomCyclicalLearningRateNotSerialized FAILED [ 11%]
tensorflow_addons/optimizers/cyclical_learning_rate_test.py::CyclicalLearningRateTest::testCustomCyclicalLearningRateSerialized FAILED  [ 22%]
tensorflow_addons/optimizers/cyclical_learning_rate_test.py::CyclicalLearningRateTest::testExponentialCyclicalLearningRateNotSerialized FAILED [ 33%]
tensorflow_addons/optimizers/cyclical_learning_rate_test.py::CyclicalLearningRateTest::testExponentialCyclicalLearningRateSerialized FAILED [ 44%]
tensorflow_addons/optimizers/cyclical_learning_rate_test.py::CyclicalLearningRateTest::testTriangular2CyclicalLearningRateNotSerialized FAILED [ 55%]
[...]

PR #1575, opened 2020-04-05
Moved test out of run_in_graph_and_eager_mode in cyclicallearningrate

autoih:
cc @gabrieldemarmiesse, since it skips all tests, do we need to wait until https://github.com/tensorflow/addons/issues/1203 solved?


gabrieldemarmiesse:
Yes, we'll likely need to fix the bugs before doing any changes to avoid changing the semantics of the tests.

gabrieldemarmiesse:
Closing as fixing the tests is already done in a case-by-case basis (e.g. #1623)

PR #1593, opened 2020-04-06
Remove the need to declare source files of tfa and utils in BUILD.

gabrieldemarmiesse:


tensorflow_addons/utils/BUILD
```diff
@@ -20,10 +13,8 @@ py_library(
 py_test(
     name = "keras_utils_test",
```
seanpmorgan, line 14:
```suggestion
    name = "utils_test",
```

PR #1600, opened 2020-04-06
Added recompute functionality for gradient checkpointing in eager mode

pidajay:
This PR implements gradient checkpointing in eager mode for TF keras sequential models. 
Gradient checkpointing allows for memory efficient training of 'large' models at the cost of increased compute time. 
Please check tutorial here - https://github.com/pidajay/addons/blob/grad_checkpointing_eager/docs/tutorials/training_gradient_checkpointing.ipynb
1. Currently only 'recompute' step is implemented. Results are highly dependent on type of model. My results can be found here
https://github.com/pidajay/tf2_gradient_checkpointing
2. Results Summary: 
a. For the sample CNN model defined in tutorial a memory savings of 6x was obtained at the cost of 3x performance degradation.
b. For the prebuilt keras application VGG16 model a memory savings of 3x was obtained at the cost of 6x performance degradation.


pidajay:
Note that there is recompute functionality in current TF core. But it does not seem to produce the desired memory savings. My implementation borrows from 
1. The original TF recompute implementation - https://github.com/tensorflow/tensorflow/blob/64f4a59d5e39b60d67047d5e0b82de0cbcc6c2df/tensorflow/python/ops/custom_gradient.py#L458
2. @davisyoshida implementation here - https://github.com/davisyoshida/tf2-gradient-checkpointing

gabrieldemarmiesse:
Thanks @pidajay for this nice pull request! It's a great starting point for people who want to use gradient checkpointing. It is a very new feature and we'll discuss with the other maintainers if we want to go forward with this or not. If not, we'll ask you to maintain it in one of your repositories instead. If we go forward, we'll help you go though the CI tests and make sure we don't duplicate functionalities with core TF.

We'll keep you updated about what we decide, thank you for your patience :)

pidajay:
Thanks @gabrieldemarmiesse. Just a heads up, I am working on implementing checkpointing on top of this  (This PR just does recompute. To my knowledge checkpointing does not exist anywhere in the TF ecosystem). Will wait for your feedback though before deciding on how to proceed. Thanks again!

pidajay:
The checkpointing step has been implemented on top of the recompute step. Results are below.
Sample CNN - https://github.com/pidajay/tf2_gradient_checkpointing/blob/master/Sample_CNN_memory_prof_results_with_ckpt.txt
VGG16 - https://github.com/pidajay/tf2_gradient_checkpointing/blob/master/VGG16_memory_prof_results_with_ckpt.txt

pidajay:
@gabrieldemarmiesse Any updates on the decision about how to proceed with this PR? Thanks.

gabrieldemarmiesse:
@pidajay sorry about that, we didn't have the time to discuss it yet. I'll ping the team.

davisyoshida:
@pidajay What's the solution implemented for things like dropout here?

seanpmorgan:
Hi @pidajay,

Firstly I want to apologize for not getting around to reviewing this with the other maintainers sooner, that is my fault. This implementation looks very nice and the tutorial is a great addition.

Unfortunately at this time we're not looking to add additional subpackages (e.g. tfa.training) since we're reaching the limit of subpackages that we can support given our bandwidth. To avoid this kind of uncessary work in a PR, we strongly recommend you file an issue prior to submitting a PR: 
https://github.com/tensorflow/addons/blob/master/CONTRIBUTING.md#pull-requests

Additionally this issue was raised in the past, and it appears that this ability is in core TensorFlow:
https://github.com/tensorflow/addons/issues/218
https://www.tensorflow.org/api_docs/python/tf/recompute_grad
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/custom_gradient.py#L459

Does this cover the functionality you're looking to provide in this PR?

pidajay:
@davisyoshida The current implementation does not address non-deterministic operations. I will need to look into that. Of course one possible way is to set a fixed random seed like you did in your implementation :). I am currently looking into adding support for non-sequential models. Will get back to non-deterministic support after that.

pidajay:
@seanpmorgan Thanks for the comments. Here are the answers to your questions.
1. I created an additional sub package because I had no idea where this functionality will slot in the existing sub packages. Sorry this is my first PR to the repo. Next time I will create an issue before proceeding to do things :).
2. As I mentioned in an earlier comment, recompute_grad exists in core TF but does not seem to work as advertised. Three main issues. One, no one seems to know how to use it. Two, even if one figures how to use, it does not seem to produce the desired memory savings. Three, it probably expects the user to explicitly partition the model before using the decorator on each of the individual partitions. With my solution the user just needs to add a decorator on top of the main model. That is it. No explicit partitioning required.

seanpmorgan:
> @seanpmorgan Thanks for the comments. Here are the answers to your questions.
> 
> 1. I created an additional sub package because I had no idea where this functionality will slot in the existing sub packages. Sorry this is my first PR to the repo. Next time I will create an issue before proceeding to do things :).
> 2. As I mentioned in an earlier comment, recompute_grad exists in core TF but does not seem to work as advertised. Three main issues. One, no one seems to know how to use it. Two, even if one figures how to use, it does not seem to produce the desired memory savings. Three, it probably expects the user to explicitly partition the model before using the decorator on each of the individual partitions. With my solution the user just needs to add a decorator on top of the main model. That is it. No explicit partitioning required.

Gotcha, so one thing that we really strive not to do is duplicate functionality between Addons and core (tf.contrib was notorious for re-implementing things with new API structure). Ultimately I think the `tf.recompute_grad` seems to be lacking in functionality so I would defer this PR to an issue in TF-Core to see what can be done to improve documentation and user experience. I understand that since it's a public API it's unlikely to get a ton of traction for changing its functionality, but perhaps stronger documentation could really help here (or if there is a bug in not saving memory that should be fixed.)

An alternative is to extend another repo (like @davisyoshida 's that you mentioned) to support this use case. Unfortunately, like I mentioned we have no bandwidth to support additional subpackages at the moment and as you mentioned this doesn't really fit into the ones we currently have. I'm going to close this issue since I don't think we'll be able to merge it into Addons.

Please continue to contribute to the repository though! Sorry to not have engaged in this discussion with you sooner. Please feel free to continue discussion in this thread if it is beneficial.

PR #1616, opened 2020-04-09
Select correct OS for python wheel uploads

seanpmorgan:
Since the runner is always ubuntu this only pulled down linux wheels. Likely the cause of the 502 errors from pypi.

https://test.pypi.org/manage/project/tensorflow-addons/release/0.9.0.dev2/
vs
https://test.pypi.org/manage/project/tensorflow-addons/release/0.9.0.dev0/


gabrieldemarmiesse:
Dang! I though I fixed that bug a while ago. I wondere when it's been introduced again.

gabrieldemarmiesse:
I see, it's been introduced here: https://github.com/tensorflow/addons/commit/10d29e9973317673f7a7ce7786b46c9eaa1de531

PR #1622, opened 2020-04-09
Fix the build badges on the central README

### LINKED ISSUE #1534: Update the CI badges on central README
Currently these badges point to workflows that we no longer run:
https://github.com/tensorflow/addons/actions?query=workflow%3Amanylinux-nightly

We can't update it to point toward our primary CI workflow though since this runs on PRs that are created as well as those that have been merged. Any good ideas jump out to anyone?

seanpmorgan:
Closes #1534


gabrieldemarmiesse:
It's a shame we cannot have one badge per os, but well, we'll have the CI passing all the time anyway so it's not useful :p

PR #1623, opened 2020-04-09
Fixed serialization and test for a learning rate scheduler.

gabrieldemarmiesse:
The gamma parameter was not saved. Causing the new object to behave differently from the original.


RaphaelMeudec:
@bot-of-gabrieldemarmiesse  @gabrieldemarmiesse LGTM!

PR #1625, opened 2020-04-09
Fix gpu build custom ops

gabrieldemarmiesse:


gabrieldemarmiesse:
Looking at the sizes of the wheels, it seems to work.

seanpmorgan:
> Looking at the sizes of the wheels, it seems to work.

Yeah, just wondering why koroko is hanging... probably not related to this, but there's no way to check.

gabrieldemarmiesse:
It's not the first time kokoro hangs sadly.

PR #1629, opened 2020-04-10
Specify CUDA information within docker for GPU tests

seanpmorgan:
Looks like the GPU tests were truly hanging:
https://source.cloud.google.com/results/invocations/67100f67-5a71-431b-952a-4efb61e8d1b8/targets/tensorflow_addons%2Fubuntu%2Fgpu%2Fpy3%2Fcontinuous/log

Interested to know why the release call to `build_and_release` somehow picked up default values:
https://github.com/tensorflow/addons/blob/master/tools/releases/release_linux.sh#L40
https://github.com/tensorflow/addons/runs/575305456?check_suite_focus=true#step:6:433

While the same call in the Dockerfile hangs waiting for a response:
https://github.com/tensorflow/addons/blob/master/tools/docker/gpu_tests.Dockerfile#L15

### DIFF
--- a/tools/docker/gpu_tests.Dockerfile
+++ b/tools/docker/gpu_tests.Dockerfile
@@ -1,5 +1,9 @@
 FROM tensorflow/tensorflow:2.1.0-custom-op-gpu-ubuntu16
 ENV TF_NEED_CUDA="1"
+ENV TF_CUDA_VERSION="10.1"
+ENV CUDA_TOOLKIT_PATH="/usr/local/cuda"
+ENV TF_CUDNN_VERSION="7"
+ENV CUDNN_INSTALL_PATH="/usr/lib/x86_64-linux-gnu"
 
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
> Interested to know why the release call to build_and_release somehow picked up default values

It seems that in github actions, (maybe because it's in a build step), user input is not available, as such, in this line: https://github.com/tensorflow/addons/blob/master/configure.py#L174 

answer is `""` (empty string).

In google cloud, for some reason, (maybe the `-t` argument in the docker run), python thinks that user input is available, as such it doesn't return en EOF in this line: https://github.com/tensorflow/addons/blob/master/configure.py#L51 .

I believe that we shouldn't ask for user input and always use the environement variables or default values. I'll make a PR for that.

gabrieldemarmiesse:
I think this can be closed because it's superseded by #1634

PR #1634, opened 2020-04-10
Fix configure.py (fixes the gpu build hanging).

gabrieldemarmiesse:
We shouldn't ask for user input, it's source of too many issues in the CI and hurts readability.


seanpmorgan:
@googlebot I consent.

PR #1639, opened 2020-04-10
Moved some functions out of run_all_in_graph_and_eager_mode in attention_wrapper_test.py.

gabrieldemarmiesse:


gabrieldemarmiesse:
> The indent seems to be quite weird in this file, are we using 4 space or 2 space?

I'm not sure I follow you, it's 4 spaces, what makes you think 2?


Also for the background, sorry, I should have been more explicit, see https://github.com/tensorflow/addons/issues/1328 

We want to get rid of the decorator that traces test functions (run_all_in_graph_and_eager_mode).

autoih:
> The indent seems to be quite weird in this file, are we using 4 space or 2 space?
> 
> Also, I haven't follow the change history closely, but currently there is a nest structure of test classes in this file, which is bit anti-pattern. Is that expected?

I think in TensorFlow using 2 space, but here is with `black`, which is 4 space.

PR #1642, opened 2020-04-10
Update compatibility matrix

seanpmorgan:
So wanted to discuss a couple of things here before merging:

1. Do we want to show all minor versions or just the ones that have changes to compatibility. For example 0.8.3 here is the same compatibility as 0.9.1 so it may be sufficient to just show 0.9.1. Cons I see is that documentation may have to be inferred, but otherwise this list may blow up in size through the years.

2. Regarding 3.8 compatibility on tf-nightly. We are compatible with code but don't publish a py38 package (until 2.2 released). Any thoughts on what to report there?

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,8 @@ what it was tested against.
 #### Python-Op Compatibility Matrix
 | TFA Version    | TensorFlow | Python  |
 |:----------------------- |:---|:---------- |
-| tfa-nightly | 2.1, 2.2 | 3.5, 3.6, 3.7, 3.8 | 
-| tensorflow-addons-0.8.3 | 2.1 |3.5, 3.6, 3.7 |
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
> Do we want to show all minor versions or just the ones that have changes to compatibility.

I believe we should display 0.6.x, 0.7.x, 0.8.x... so all of them, but let's not go back too far into the past. We could say that 4-5 versions are enough?

> Regarding 3.8 compatibility on tf-nightly. We are compatible with code but don't publish a py38 package (until 2.2 released). Any thoughts on what to report there?

Yup, in theory, we could publish to pypi only the source code, but that'd be a lot of work. Let's just add a note below to ask users who have 2.2.0 and py3.8 to do `pip install git+https://github.com/tensorflow/addons.git@r0.9`.

PR #1644, opened 2020-04-11
Update year and delete unnecessary tensorflow magic in _template.ipynb

shun-lin:
Change 2019 -> 2020, change small typo of tfo to tfa, remove unnecessary tensorflow version magic as TF 2 is now default.


shun-lin:
good catch @seanpmorgan , I have committed the suggestions! :)

shun-lin:
@seanpmorgan may you approve this if everything looks good? Thanks :)

PR #1649, opened 2020-04-12
Added TQDM Progressbar for evaluate()

### LINKED ISSUE #1620: Have TQDMProgressBar work in evaluate mode as well

Right now, the `TQDMProgressBar` callback can only be used when training a model.
When evaluating nothing happens. And even if I try to "cheat" by manually setting the `on_test_batch_(begin|end)` methods, it fails with an error about `mode`.

**Relevant information**
- Are you willing to contribute it (yes/no): don't know

callbacks

everyone using `tqdm` to evaluate in a notebook: I want something fancier than the default evaluate progressbar

**Any other info.**
It failed in both tf2.2 and tf2.1

You can find the colab gist [here](https://colab.research.google.com/drive/1eHI-6QhttMhdMYeH8VP0SwfyB0b46ugj).

ashutosh1919:
fixes #1620 .  
Although, I couldn't figure out the use of `batch` parameter in `on_test_batch_begin` method.
@zaccharieramzi and @gabrieldemarmiesse , please review and suggest changes.

### DIFF
--- a/tensorflow_addons/callbacks/tqdm_progress_bar.py
+++ b/tensorflow_addons/callbacks/tqdm_progress_bar.py
@@ -146,7 +146,10 @@ class TQDMProgressBar(Callback):
                 self.overall_progress_tqdm.close()
         elif hook == "test" or hook == "train_epoch":
             if self.show_epoch_progress:
-                metrics = self.format_metrics(logs, self.num_samples_seen)
+                if hook == "test":
+                    metrics = self.format_metrics(logs, self.num_samples_seen)
+                else:
+                    metrics = self.format_metrics(logs)
                 self.epoch_progress_tqdm.desc = metrics
                 # set miniters and mininterval to 0 so last update displays
                 self.epoch_progress_tqdm.miniters = 0
--- a/tensorflow_addons/callbacks/tqdm_progress_bar.py
+++ b/tensorflow_addons/callbacks/tqdm_progress_bar.py
@@ -98,35 +98,29 @@ class TQDMProgressBar(Callback):
         self.logs = None
         super().__init__()
 
-    def on_train_begin(self, logs=None):
-        self.is_training = True
+    def _initialize_progress_params(self):
+        self.num_samples_seen = 0
+        self.steps_to_update = 0
+        self.steps_so_far = 0
+        self.logs = defaultdict(float)
         self.num_epochs = self.params["epochs"]
-
-        if self.show_overall_progress:
-            self.overall_progress_tqdm = self.tqdm(
-                desc="Training",
-                total=self.num_epochs,
-                bar_format=self.overall_bar_format,
-                leave=self.leave_overall_progress,
-                dynamic_ncols=True,
-                unit="epochs",
-            )
-
-        # set counting mode
         self.mode = "steps"
         self.total_steps = self.params["steps"]
 
-    def on_train_end(self, logs={}):
-        self.is_training = False
-        if self.show_overall_progress:
-            self.overall_progress_tqdm.close()
-
-    def on_test_begin(self, logs={}):
-        if not self.is_training:
-            self.num_epochs = self.params["epochs"]
-            # set counting mode
-            self.mode = "steps"
-            self.total_steps = self.params["steps"]
+    def _on_call_begin(self, hook, data, logs=None):
+        if hook == "train":
+            self._initialize_progress_params()
+            if self.show_overall_progress:
+                self.overall_progress_tqdm = self.tqdm(
+                    desc="Training",
+                    total=self.num_epochs,
+                    bar_format=self.overall_bar_format,
+                    leave=self.leave_overall_progress,
+                    dynamic_ncols=True,
+                    unit="epochs",
+                )
+        elif hook == "test":
+            self._initialize_progress_params()
             if self.show_epoch_progress:
                 self.epoch_progress_tqdm = self.tqdm(
                     total=self.total_steps,
@@ -136,97 +130,41 @@ class TQDMProgressBar(Callback):
                     dynamic_ncols=True,
                     unit=self.mode,
                 )
-            self.num_samples_seen = 0
-            self.steps_to_update = 0
-            self.steps_so_far = 0
-            self.logs = defaultdict(float)
-
-    def on_test_batch_end(self, batch, logs={}):
-        if not self.is_training:
-            batch_size = 1
-
-            self.num_samples_seen += batch_size
-            self.steps_to_update += 1
-            self.steps_so_far += 1
-
-            if self.steps_so_far < self.total_steps:
-
-                for metric, value in logs.items():
-                    self.logs[metric] += value * batch_size
-
-                now = time.time()
-                time_diff = now - self.last_update_time
-
-                if self.show_epoch_progress and time_diff >= self.update_interval:
-
-                    # update the epoch progress bar
-                    metrics = self.format_metrics(self.logs, self.num_samples_seen)
-                    self.epoch_progress_tqdm.desc = metrics
-                    self.epoch_progress_tqdm.update(self.steps_to_update)
-
-                    # reset steps to update
-                    self.steps_to_update = 0
-
-                    # update timestamp for last update
-                    self.last_update_time = now
-            else:
-                self.logs = logs
-
-    def on_test_end(self, logs={}):
-        if not self.is_training:
-            metrics = self.format_metrics(self.logs)
-            self.epoch_progress_tqdm.desc = metrics
-
-            # set miniters and mininterval to 0 so last update displays
-            self.epoch_progress_tqdm.miniters = 0
-            self.epoch_progress_tqdm.mininterval = 0
-
-            # update the rest of the steps in epoch progress bar
-            self.epoch_progress_tqdm.update(
-                self.total_steps - self.epoch_progress_tqdm.n
+        elif hook == "epoch":
+            self._initialize_progress_params()
+            current_epoch_description = "Epoch {epoch}/{num_epochs}".format(
+                epoch=data + 1, num_epochs=self.num_epochs
             )
-            self.epoch_progress_tqdm.close()
+            if self.show_epoch_progress:
+                print(current_epoch_description)
+                self.epoch_progress_tqdm = self.tqdm(
+                    total=self.total_steps,
+                    bar_format=self.epoch_bar_format,
+                    leave=self.leave_epoch_progress,
+                    dynamic_ncols=True,
+                    unit=self.mode,
+                )
 
-    def on_epoch_begin(self, epoch, logs={}):
-        current_epoch_description = "Epoch {epoch}/{num_epochs}".format(
-            epoch=epoch + 1, num_epochs=self.num_epochs
-        )
+    def _on_call_end(self, hook, data, logs):
+        if hook == "train":
+            if self.show_overall_progress:
+                self.overall_progress_tqdm.close()
+        elif hook == "test" or hook == "epoch":
+            if self.show_epoch_progress:
+                metrics = self.format_metrics(logs)
+                self.epoch_progress_tqdm.desc = metrics
+                # set miniters and mininterval to 0 so last update displays
+                self.epoch_progress_tqdm.miniters = 0
+                self.epoch_progress_tqdm.mininterval = 0
+                # update the rest of the steps in epoch progress bar
+                self.epoch_progress_tqdm.update(
+                    self.total_steps - self.epoch_progress_tqdm.n
+                )
+                self.epoch_progress_tqdm.close()
+            if hook == "epoch" and self.show_overall_progress:
+                self.overall_progress_tqdm.update(1)
 
-        if self.show_epoch_progress:
-            print(current_epoch_description)
-            self.epoch_progress_tqdm = self.tqdm(
-                total=self.total_steps,
-                bar_format=self.epoch_bar_format,
-                leave=self.leave_epoch_progress,
-                dynamic_ncols=True,
-                unit=self.mode,
-            )
-
-        self.num_samples_seen = 0
-        self.steps_to_update = 0
-        self.steps_so_far = 0
-        self.logs = defaultdict(float)
-
-    def on_epoch_end(self, epoch, logs={}):
-
-        if self.show_epoch_progress:
-            metrics = self.format_metrics(logs)
-            self.epoch_progress_tqdm.desc = metrics
-
-            # set miniters and mininterval to 0 so last update displays
-            self.epoch_progress_tqdm.miniters = 0
-            self.epoch_progress_tqdm.mininterval = 0
-
-            # update the rest of the steps in epoch progress bar
-            self.epoch_progress_tqdm.update(
-                self.total_steps - self.epoch_progress_tqdm.n
-            )
-            self.epoch_progress_tqdm.close()
-
-        if self.show_overall_progress:
-            self.overall_progress_tqdm.update(1)
-
-    def on_batch_end(self, batch, logs={}):
+    def _on_call_batch_end(self, batch, logs):
         if self.mode == "samples":
             batch_size = logs["size"]
         else:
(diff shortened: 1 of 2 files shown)


shun-lin:
May you also update the test cases in https://github.com/tensorflow/addons/blob/master/tensorflow_addons/callbacks/tests/tqdm_progress_bar_test.py ? Thank you so much!

ashutosh1919:
@shun-lin , please suggest changes. I have got the problem. I think we need to add 4 methods : `on_test_begin()` (To initialize tqdm params) , `on_test_batch_begin()` (Probably to show batch num), `on_test_batch_end()` (To update progressbar with the info) and `on_test_end()` (to close progress bar)

shun-lin:
> @shun-lin , please suggest changes. I have got the problem. I think we need to add 4 methods : `on_test_begin()` (To initialize tqdm params) , `on_test_batch_begin()` (Probably to show batch num), `on_test_batch_end()` (To update progressbar with the info) and `on_test_end()` (to close progress bar)

Thanks! I agree with you that those are the methods that we need to add (and maybe for better readability we probably want to refactor the on_test|train_* code so it wouldn't have duplicate, but that can be done later). So because Model.evaluate() is equivalent to running **1 epoch** over the testing dataset (see screencast below), we can utilize how we show the progress within each epoch, see `self.epoch_progress_tqdm` and how we update it in `on_batch_end` method (maybe the evaluate will call `on_batch_end` if `on_test_batch_end` is absent? (double check this please thanks!, if this is not the case then just extract out the logic in `on_batch_end` and call that logic with `on_test_batch_end` too). Therefore `on_test_begin()` should look like `on_epoch_begin` and `on_test_end` should look like `on_epoch_end` (since there is no need for overall progress bar for evaluate).

Let me know if this is confusing to you, happy to explain it further! Thanks for the contribution!

Screen Shot 2020-04-11 at 11 20 15 PM

ashutosh1919:
@shun-lin ,  
**Small Help Required :**  

**The progress :**  
I was successfully able to produce the evaluate progress bar as shown below but the implementation was little bit modification o what you described in the previous comment. Below are the lessons learnt during implementation.
- We have to use exactly three methods to do what is required : `on_test_begin()` (to initialize the prograss bar and params for progress bar), `on_test_batch_end()` (to update progress bar after each batch is evaluated), `on_test_end()` (to update final metrices and close tqdm progress bar).  
- We can not just copy the code from `on_epoch_begin()` and `on_batch_end()` because, the `self.x` types of variables are shared across whole class. It won't get affected during call of `evaluate()` method, but it will produce wrong results when `fit()` is called and `validation_data` is provided. So, separate copies of every param is required for this new progress bar.
tqdm_pbar
Now, I have added code such that it will produce progress bar both in case of `evaluate()` method and also in case of `fit()` with `validation_data` defined.  

**The Problems :**  
But there are still couple of problems that I have noticed.  
- You can see in above image that there is separate list printed containing `loss` and `acc` below progress bar. But they must be defined in `tqdm.desc` (At the end of progress bar). I am searching, but still not able to locate the code which prints that list. Also, not able to find the reason why the `desc` is printing `loss` at the time when progress bar is filling but `desc` is not printing `loss` when progress bar is whole filled.
- Right now, the progress bar is using the steps taken from `self.params["steps"]` which is same as in training. Again, this want affect `evaluate()`, But in case of `fit()` with `validation_data`, `self.params["steps"]` will be shared which should not be. We need to create something to calculate total steps of `validation_data`.   

Sorry for long comment, the demo of current progress bar is [here](https://colab.research.google.com/drive/16p5Abagjzznxk6qUeYxYGlTU36DAuC-f). Please review it, and suggest changes to resolve problems. Thanks for help!

shun-lin:
@ashutosh1919 great progress and good observations! Couple of remarks.

1. The separate list of loss and accuracy is **expected**, this is the return of `model.evaluate()` and because notebook display the return of last line in a cell there is nothing  wrong with it. Here is the reference: https://www.tensorflow.org/api_docs/python/tf/keras/Model#evaluate

if you want the extra list to not print, just simply assign it to a variable, `metric_lst = model.evaluate(...)`.

2. I noticed that you have added `dynamic_col=True`, which should have solved the problem that the bar (the green part) is too long, but in your screenshot it's still too long, may you look into the version of your tqdm and check if updated version would help? 

3. In terms of printing progress bar for `train with validation_data`, I would suggest to follow how the original progress bar is doing. The original progress bar did not display a separate progress bar when computing metrics over data in `validation_data`, this behavior is different with `evaluate()`. Therefore we need to distinguish the case when the method is called from `evaluate()` vs `train with validation_data`. I think this helps with your 2nd problem. Let me know if this is not clear, thanks!

Sincerely,
Shun Lin

ashutosh1919:
@shun-lin , yes you are right. I was thinking the same that we should not display another progress bar in case of `train with validation_data`. But is there any parameter in `logs` which is passed from `evaluate()` or `fit()` which will help me distinguish these two cases?  

But I have very strange temporary idea which I don't think is good : 
```python
if(nb_epoch==1):
    # Came from evaluate()
else:
    # Came from fit()
```

shun-lin:
good question, I think you can find it in where TF Keras has the printing logic (they must distinguish that way). A quick but hacky way is just to set a variable like `self.is_training` at the beginning of training (and make sure to clean this variable up at the end).

ashutosh1919:
**Done from My Side**  

**Solutions of the problems described in [comment](https://github.com/tensorflow/addons/pull/1649#issuecomment-612608622) :**  
- The `desc` logs were not printing at the end of epoch because logs passed to `on_test_end()` method are `None` as of now. Solution was to assign `logs` to `self.test_logs` in last step in `on_test_batch_end()` method. And btw, if you are thinking that why this problem didn't come in case of train tqdm progress bar, because last logs of train progress bar is printed by `on_epoch_end()` method where `logs` param is not None.  
- Distinguishing `evaluate()` and `fit()` is done using class variable `is_training` which was set to `True` in `on_train_begin()` and set back to `False` in `on_train_end()`

**Also written test function for this PR.**
Below attached GIF will give you visualization of progress bar.
tqdm_pbar_demo - Colaboratory

ashutosh1919:
@shun-lin, please review.

tensorflow_addons/callbacks/tqdm_progress_bar.py
```diff
@@ -93,11 +93,17 @@ def __init__(
         self.last_update_time = time.time()
         self.overall_progress_tqdm = None
         self.epoch_progress_tqdm = None
+        self.is_training = False
+        self.test_last_update_time = time.time()
```
shun-lin, line 97: Are those `test_*` variables necessary anymore? Since we are not going to display the progress bar for `train with validation_data`, can we re-use the same variable for `train()` and `evaluate()`? This would make it easier to maintain, thanks!

shun-lin: Also since things should be closed properly in `on_*_end` and opend properly at `on_*_begin`, I think it would be possible to re-use those variables.

tensorflow_addons/callbacks/tqdm_progress_bar.py
```diff
@@ -115,9 +121,81 @@ def on_train_begin(self, logs=None):
...
+                    dynamic_ncols=True,
+                    unit=self.mode,
+                )
+            self.test_num_samples_seen = 0
+            self.test_steps_to_update = 0
+            self.test_steps_so_far = 0
+            self.test_logs = defaultdict(float)
+
+    def on_test_batch_end(self, batch, logs={}):
+        if self.is_training is False:
```
shun-lin, line 149: same nit from above.

tensorflow_addons/callbacks/tqdm_progress_bar.py
```diff
@@ -115,9 +121,81 @@ def on_train_begin(self, logs=None):
...
+            if self.mode == "samples":
```
shun-lin, line 150: will this ever be true, `self.mode == "sample"`? We have self.is_training is False and thus this will always be set to steps right?

tensorflow_addons/callbacks/tqdm_progress_bar.py
```diff
@@ -115,9 +121,81 @@ def on_train_begin(self, logs=None):
...
+                    # reset steps to update
+                    self.test_steps_to_update = 0
+
+                    # update timestamp for last update
+                    self.test_last_update_time = now
+            else:
+                self.test_logs = logs
+
+    def on_test_end(self, logs={}):
+        if self.is_training is False and self.show_epoch_progress:
```
shun-lin, line 185: Since how the `model.evaluate()` is just having 1 epoch, and because there will only be one progress bar for `evaluate()`, I think we can safely assume that the user would want to see it and thus we won't need to check if `self.show_epoch_progress` is set. This is my opinion.

tensorflow_addons/callbacks/tqdm_progress_bar.py
```diff
@@ -115,9 +121,81 @@ def on_train_begin(self, logs=None):
...
+                    leave=self.leave_epoch_progress,
+                    dynamic_ncols=True,
+                    unit=self.mode,
+                )
+            self.test_num_samples_seen = 0
+            self.test_steps_to_update = 0
+            self.test_steps_so_far = 0
+            self.test_logs = defaultdict(float)
+
+    def on_test_batch_end(self, batch, logs={}):
```
shun-lin, line 148: Follow up from the same comment for reusing variables instead of creating `test_*` variables, if we are able to re-use those variables, we are able to refactor `on_epoch_end|begin` method and extract the inner logic out and call those methods from `on_epoch_*` and `on_test_batch_*` as we can see that there are many duplicating code.

shun-lin:
> **Done from My Side**
> 
> **Solutions of the problems described in [comment](https://github.com/tensorflow/addons/pull/1649#issuecomment-612608622) :**
> 
> * The `desc` logs were not printing at the end of epoch because logs passed to `on_test_end()` method are `None` as of now. Solution was to assign `logs` to `self.test_logs` in last step in `on_test_batch_end()` method. And btw, if you are thinking that why this problem didn't come in case of train tqdm progress bar, because last logs of train progress bar is printed by `on_epoch_end()` method where `logs` param is not None.
> * Distinguishing `evaluate()` and `fit()` is done using class variable `is_training` which was set to `True` in `on_train_begin()` and set back to `False` in `on_train_end()`
> 
> **Also written test function for this PR.**
> Below attached GIF will give you visualization of progress bar.
> tqdm_pbar_demo - Colaboratory

Nice catch at the `on_test_batch_end`, maybe that's something we should file a bug for TF Keras if having `log == None` is not an expected behavior in the last step `on_test_batch_end`.

ashutosh1919:
@shun-lin , Done changes as per your last review comments.
Is below GIF able to convince you?
tqdm_pbar_demo - Colaboratory (2)

ashutosh1919:
**Future TODOs :**  
- Will raise PR soon to clean code and to make common functions for duplicate codes same as there are methods in tf repo like `_call_batch_hooks()` which are called from both `on_train_batch_*()` and `on_test_batch_*()`.
- Will add demo in the respective notebook to illustrate that progress bar also works with `evaluate()`.
- Will raise issue to tf repo about the issue with `on_test_end()` and `on_train_end()` that logs are currently empty. Instead, logs should be the logs from last call of `on_test_batch_end()` and `on_train_batch_end()` respectively.

shun-lin:
> @shun-lin , Done changes as per your last review comments.
> Is below GIF able to convince you?
> tqdm_pbar_demo - Colaboratory (2)

This looks amazing!

shun-lin:
> **Future TODOs :**
> 
> * Will raise PR soon to clean code and to make common functions for duplicate codes same as there are methods in tf repo like `_call_batch_hooks()` which are called from both `on_train_batch_*()` and `on_test_batch_*()`.
> * Will add demo in the respective notebook to illustrate that progress bar also works with `evaluate()`.
> * Will raise issue to tf repo about the issue with `on_test_end()` and `on_train_end()` that logs are currently empty. Instead, logs should be the logs from last call of `on_test_batch_end()` and `on_train_batch_end()` respectively.

Sounds like a good plan, are those for future PR (bullet point 1 and 2) or for finishing up this PR? (also adding @seanpmorgan and @gabrieldemarmiesse to the reviewer list)

ashutosh1919:
Those are for future because for that, this PR should be merged. I will do that within some days.

shun-lin:
@ashutosh1919 I would suggest to do the refactoring to remove duplicate code in this PR if you think it's not too much work to do so (I think it's easier to do it in this PR than in the future PR).

ashutosh1919:
> @ashutosh1919 I would suggest to do the refactoring to remove duplicate code in this PR if you think it's not too much work to do so (I think it's easier to do it in this PR than in the future PR).

Sure will try to do that today itself then.

ashutosh1919:
@shun-lin , I have refactored the existing code to remove duplicity. I have kept the function names temporarily. They can be changed if you say so. Please review it and suggest changes. I have tested the functionality in Google colab [notebook](https://colab.research.google.com/drive/16p5Abagjzznxk6qUeYxYGlTU36DAuC-f). It works fine.

tensorflow_addons/callbacks/tqdm_progress_bar.py
```diff
@@ -186,6 +193,37 @@ def on_batch_end(self, batch, logs={}):
 
                 # update timestamp for last update
                 self.last_update_time = now
+        else:
```
shun-lin, line 196: actually this else statement is not necessary and may be wrong after thinking about it more.
The reason why we have the logic of updating self.logs as we see examples is that (this is also how the progress bar is doing) we are displaying a `moving average` of metrics over time. If at the end we set `self.logs = log` this means that at the end we are displaying the metrics of last batch instead of the average across the entire epoch/evaluate. I am pretty sure that removing this else statement along with self.logs = logs will still work (since at `on_test_end` you are using `self.logs` and it's already been set.

tensorflow_addons/callbacks/tqdm_progress_bar.py
```diff
@@ -186,6 +193,37 @@ def on_batch_end(self, batch, logs={}):
...
+            self.logs = logs
+
+    def on_train_begin(self, logs=None):
```
shun-lin, line 199: Don't forget to change the internal method names here like `self._on_call_begin` to self._initialize_progbar` etc.

tensorflow_addons/callbacks/tqdm_progress_bar.py
```diff
@@ -93,71 +93,74 @@ def __init__(
...
+                    unit=self.mode,
+                )
+
+    def _clean_up_progbar(self, hook, logs):
+        if hook == "train_overall":
+            if self.show_overall_progress:
+                self.overall_progress_tqdm.close()
+        elif hook == "test" or hook == "train_epoch":
+            if self.show_epoch_progress:
+                if hook == "test":
```
shun-lin, line 149: a bit too much nest, you can bring this if/else statement outside this nest like

```
   else:
         if hook == "test":
               metrics = self.format_metrics(logs, self.num_samples_seen)
         else:
               metrics = self.format_metrics(logs)
         if self.show_epoch_progress:
               ......
```

ashutosh1919: @shun-lin , After last commit - I was able to generate progress bar in colab. But still gives error that desc is None in tests. Can you please help?

shun-lin: sure! which test cases?

ashutosh1919: [this one](https://github.com/tensorflow/addons/pull/1649/checks?check_run_id=584944056)

shun-lin: hmmmm I am thinking it may be `self.logs`'s values become None when we call `self._clean_up_progbar("test", self.logs)` at line 207, can you bring out metrics at line 151 like 

```
                 if hook == "test":  
                  metrics = self.format_metrics(logs, self.num_samples_seen)
                  print(metrics) <---- add this to see if metrics is not messed up.
                else:
                    metrics = self.format_metrics(logs)
```

shun-lin: ops I think I find a pre-existing problem, if you look at `_update_progbar` we are updating `self.logs`  in ` if self.steps_so_far < self.total_steps:` this is not too big of a problem when training because total_steps is normally greater than 1 but in eval there are many cases when total steps is 1 and thus this will never be called, may you try to either move `self.steps_so_far += 1` after the following `if` block or change `<` to `<=`? Thanks and I apologize for the bug.

ashutosh1919: great. I was aware of this small detail in the first commit of this PR. But forgot now. 😃

ashutosh1919: I think this should work now.

shun-lin: sounds good, thanks!

tensorflow_addons/callbacks/tqdm_progress_bar.py
```diff
@@ -168,10 +168,9 @@ def _update_progbar(self, logs):
 
         self.num_samples_seen += batch_size
         self.steps_to_update += 1
-        self.steps_so_far += 1
-
-        if self.steps_so_far < self.total_steps:
 
+        if self.steps_so_far <= self.total_steps:
```
shun-lin, line 172: Tiny NIT: actually we should either change `<` to `<=` OR putting `self.steps_so_far += 1` inside but not both, this will cause to take one extra step (this should not cause error anyway since `on_epoch_end` wouldn't be called the extra time for this to occur, but for readability may you move `self.steps_so_far += 1` back before, I think changing `<` to <=` alone should solve it, thanks!

ashutosh1919: @shun-lin , changed this tiny error. Thanks for noticing. Please approve it again.

ashutosh1919:
sorry, didn't see your comment on readability. Please sorry for this. Can you approve it again?

shun-lin:
> sorry, didn't see your comment on readability. Please sorry for this. Can you approve it again?

approved! Will merge after the tests finished running, thanks so much @ashutosh1919

ashutosh1919:
@shun-lin , all tests are passing. Merge it anytime if you are fine with it.

PR #1657, opened 2020-04-13
Fixing the flaky cell test.

gabrieldemarmiesse:
Depending on the orders of the tests, this one could or could not pass. Notably, 

```
pytest -v ./tensorflow_addons/rnn/tests/cell_test.py::test_config_nas
```

is enough to reproduce it because names of keras objects depends on which objects have already been created in the global context.

### DIFF
--- a/tensorflow_addons/conftest.py
+++ b/tensorflow_addons/conftest.py
@@ -1,6 +1,7 @@
 from tensorflow_addons.utils.test_utils import maybe_run_functions_eagerly  # noqa: F401
 from tensorflow_addons.utils.test_utils import cpu_and_gpu  # noqa: F401
 from tensorflow_addons.utils.test_utils import data_format  # noqa: F401
+from tensorflow_addons.utils.test_utils import set_seeds  # noqa: F401
 
 # fixtures present in this file will be available
 # when running tests and can be referenced with strings
--- a/tensorflow_addons/rnn/tests/cell_test.py
+++ b/tensorflow_addons/rnn/tests/cell_test.py
@@ -223,7 +223,7 @@ def test_keras_rnn():
 
 
 def test_config_nas():
-    cell = rnn_cell.NASCell(10, projection=5, use_bias=True)
+    cell = rnn_cell.NASCell(10, projection=5, use_bias=True, name="nas_cell_3")
 
     expected_config = {
         "dtype": "float32",
--- a/tensorflow_addons/utils/test_utils.py
+++ b/tensorflow_addons/utils/test_utils.py
@@ -17,6 +17,7 @@
 import contextlib
 import inspect
 import unittest
+import random
 
 import numpy as np
 import pytest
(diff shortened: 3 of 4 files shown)


qlzh727:
I guess this is a noticeable side effect of not using tf.test.TestCase, which ensure a proper setup and teardown of the test case, and it is missing for the current pytest setup.

The current fix is very fragile, which will break if any test is added or removed. We should find a better fix.

gabrieldemarmiesse:
> The current fix is very fragile, which will break if any test is added or removed. We should find a better fix.

Could you expand on that? / give an example?

qlzh727:
Before this change, the test case will fail randomly based on the execution order of the test. This change will fix the issue, but it seems to be hiding the real issue, which is the proper setup/teardown of the test case, and also run all tests in a global context.

gabrieldemarmiesse:
> it seems to be hiding the real issue, which is the proper setup/teardown of the test case, and also run all tests in a global context.

It's possible you see something that I'm missing here.

Could you detail what would be a proper setup/teardown? Also, many python open source projects use pytest to run all their tests in a global context (numpy, scipy, scikit-learn, pandas...), could you detail what would be different in our case which would cause us issues that they don't have?

qlzh727:
Sure. Please see https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/test_util.py#L1878.

The TensorflowTestCase has bunch of setup logic in setup and teardown, which will ensure each test case to be run in a fresh context, without being affected by other test cases.

gabrieldemarmiesse:
Notably, seeds are fixed in the `tf.test.TestCase` setup, but since we started the migration to pytest, we never fixed them automatically. Thanks for bringing it to my attention! I'll make the changes.

About the reset of the default graph, which happens in the teardown, we aim to use only eager mode + tf.functions to run the test, so we shouldn't rely on the default graph, and we won't modify global objects. We had no issues so far but if you see a potential one, feel free to give an example and we could work toward making sure we prevent it from happening.

qlzh727:
> Notably, seeds are fixed in the `tf.test.TestCase` setup, but since we started the migration to pytest, we never fixed them automatically. Thanks for bringing it to my attention! I'll make the changes.
> 
> About the reset of the default graph, which happens in the teardown, we aim to use only eager mode + tf.functions to run the test, so we shouldn't rely on the default graph, and we won't modify global objects. We had no issues so far but if you see a potential one, feel free to give an example and we could work toward making sure we prevent it from happening.
> 
> Edit: I added the automatic seeding to make sure our tests are more robusts. I added also a test to ensure the automatic seeding was working.

Also note that the test annotation also has some extra setup for eager and graph context. You might want to port some of them as well if you only test in eager.

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/test_util.py#L1063

AakashKumarNain:
> > Notably, seeds are fixed in the `tf.test.TestCase` setup, but since we started the migration to pytest, we never fixed them automatically. Thanks for bringing it to my attention! I'll make the changes.
> > About the reset of the default graph, which happens in the teardown, we aim to use only eager mode + tf.functions to run the test, so we shouldn't rely on the default graph, and we won't modify global objects. We had no issues so far but if you see a potential one, feel free to give an example and we could work toward making sure we prevent it from happening.
> > Edit: I added the automatic seeding to make sure our tests are more robusts. I added also a test to ensure the automatic seeding was working.
> 
> Also note that the test annotation also has some extra setup for eager and graph context. You might want to port some of them as well if you only test in eager.
> 
> https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/test_util.py#L1063

Thanks @qlzh727 for pointing it out. We will look into it. Thank you again.

PR #1661, opened 2020-04-13
Finishing the pytest migration.

gabrieldemarmiesse:
Finishing the pytest migration.


tensorflow_addons/seq2seq/tests/attention_wrapper_test.py
```diff
@@ -44,217 +43,241 @@ def setUp(self):
...
-        self.assertLen(score_val, 2)
-        self.assertEqual(score_val[0].shape, (self.batch, self.timestep))
-        self.assertEqual(score_val[1].shape, (self.batch, self.timestep))
-
-    @parameterized.named_parameters(
-        ("luong", wrapper.LuongAttention),
-        ("luong_monotonic", wrapper.LuongMonotonicAttention),
-        ("bahdanau", wrapper.BahdanauAttention),
-        ("bahdanau_monotonic", wrapper.BahdanauMonotonicAttention),
+@pytest.mark.parametrize(
```
qlzh727, line 47: Can we extra this annotation and reuse it? Having so many duplication seems to be a issue for future maintainence.

tensorflow_addons/seq2seq/tests/attention_wrapper_test.py
```diff
@@ -44,217 +41,193 @@ def setUp(self):
...
+    vocab = 20
+    embedding_dim = 6
+    num_batches = 5
+
+    dummy_data = DummyData()
+    model = MyModel(vocab, embedding_dim, dummy_data.memory_size, dummy_data.units)
+    if tf.executing_eagerly():
+        model.compile("rmsprop", "mse", run_eagerly=True)
+    else:
         model.compile("rmsprop", "mse")
```
seanpmorgan, line 216: Will this logic ever switch? TF would always be running eagerly right? The  `maybe_run_functions_eagerly` fixture wouldn't trip this so instead maybe we parameterize this as a boolean?

gabrieldemarmiesse: Good catch, fixed it!

tensorflow_addons/seq2seq/tests/attention_wrapper_test.py
```diff
@@ -297,683 +266,626 @@ def setUp(self):
...
+        attention_mechanisms.append(
+            creator(
+                units=depth,
+                memory=encoder_outputs,
+                memory_sequence_length=encoder_sequence_length,
+                **create_attention_kwargs,
             )
+        )
 
-        with self.cached_session(use_gpu=True):
```
seanpmorgan, line 437: Could we mark this as a TODO to enable `GPU` testing in #1713 for this

tensorflow_addons/seq2seq/tests/attention_wrapper_test.py
```diff
@@ -297,683 +266,626 @@ def setUp(self):
...
+            .dims[-1]
+            .value
+            for attention_layer in attention_layers
         )
+    else:
+        attention_depth = encoder_output_depth * len(create_attention_mechanisms)
 
-    def _testWithMaybeMultiAttention(
-        self,
-        is_multi,
```
seanpmorgan, line 360: So we remove all logic for handling `is_multi`. It was never used in the test suite, but I'm less familiar with this module and don't know if it'll ever be needed in testing going forward. Okay to remove IMO since it wasn't used and will be in git history.

tensorflow_addons/seq2seq/tests/attention_wrapper_test.py
gabrieldemarmiesse: We could, but I believe we don't need to test on gpu here. We don't depend on any custom op and we don't test in float16. I believe that this is tensorflow's job to do the right thing on gpu here, not ours. But I don't have a very strong opinion there, in the new setup, it's just going to be an additional decorator to add gpu testing.

PR #1665, opened 2020-04-14
Added code to call register_all() before loading model

### LINKED ISSUE #1567: Register keras objects in the function `register_all`, not at import time
Currently, it's possible to do 
```python
import tensorflow as tf

tf.keras.models.load_model("my_model.tf")
```

which is not readable and will throw an error with all main linting tools (import unused). 
Being explicit and readable come first in python, and similarely to how we do it with custom ops, `register_all()` should be called before loading a model.

See background #1151

ashutosh1919:
Fixes #1567 . 
@gabrieldemarmiesse , please review.

### DIFF
--- a/tensorflow_addons/utils/resource_loader.py
+++ b/tensorflow_addons/utils/resource_loader.py
@@ -19,7 +19,7 @@ import os
 import warnings
 
 import tensorflow as tf
-from tensorflow_addons import register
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
Closing as this is not exactly the right approach, I should have been more explicit about the problem. I gave some more guidance here: #1567

Issue #1655, opened 2020-04-13
Make our CI faster


gabrieldemarmiesse:
Currently our CI take ~30 min to run. After some analysis, here is the critical path and the big steps:

* Building the wheel on linux: 25m
  * pull the docker image: 3m
  * install TF and dependencies: 2m
  * build CPU and GPU ops: 8m (building CPU ops only takes 2m in another build, so we can assume 6m for building the GPU ops)
  * Run all tests: 6m, I should re-enable pytest-xdist here.
  * Rebuild all ops for the wheel: 5m (because the bazel cache was cleaned before)
* Test the release wheel on Windows: 4m10s
* Upload to pypi, but not really: 25s.

I'll work toward making this faster, help and ideas are welcome too of course.

bhack:
Can you point on the CI scripts sources?
Can we take any speed-up with https://github.com/actions/cache?

bhack:
See also this thread https://github.community/t5/GitHub-Actions/Cache-docker-images/td-p/44978

gabrieldemarmiesse:
Caching may help, but it may also make things slower. It's usually hard to know until we try.

bhack:
I've not built the linux wheel Dockerfile with the [progress option](https://docs.docker.com/develop/develop-images/build_enhancements/) but I suppose that in `25m` the main problem is that this is [`/root/.cache/bazel`](https://github.com/tensorflow/addons/blob/master/tools/docker/build_wheel.Dockerfile#L39) is an empty volume when you run the Action vs local build right?

bhack:
I've asked if we there is ware cache in action/cache the mount of buildkit `RUN --mount=type=cache` at https://github.com/actions/cache/issues/260.

bhack:
In the meantime I think that
```
--cache-from=type=registry,ref=$CACHE_TAG
--cache-to=type=registry,ref=$CACHE_TAG,mode=max
```
could work if we have a registry to push. See https://medium.com/titansoft-engineering/docker-build-cache-sharing-on-multi-hosts-with-buildkit-and-buildx-eb8f7005918e

gabrieldemarmiesse:
I did that a while ago, didn't seem very fast to me at the time. If someone wants to try again, a pull request is welcome: https://github.com/tensorflow/addons/pull/1043

seanpmorgan:
@gabrieldemarmiesse Can this be closed or are there pending actions we need to do?

gabrieldemarmiesse:
[...]

PR #1666, opened 2020-04-14
Fast failing test in ci

gabrieldemarmiesse:
Related to #1655 

So of course this doesn't make our CI any faster, but it provides fast feedback for contributors. With this new test, we catch most of the errors quickly since many changes are related to python-only code. The test runs in 4m and 20s, much faster than the 11m and 25s needed to run our fastest test (macos).

Some ressources: 

* [Documentation for command line options in pytest](https://docs.pytest.org/en/latest/example/simple.html#pass-different-values-to-a-test-function-depending-on-command-line-options)
* [Pandas' conftest.py](https://github.com/pandas-dev/pandas/blob/master/pandas/conftest.py)

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -84,6 +84,10 @@ jobs:
         with:
           python-version: 3.5
       - uses: actions/checkout@v2
+      - uses: actions/cache@v1
+          with:
+            path: ~/.cache/pip
+            key: ${{ runner.os }}-pip-${{ hashFiles('tools/install_deps/*.txt') }}
       - run: pip install -r tools/install_deps/tensorflow-cpu.txt -r tools/install_deps/pytest.txt
       - run: pip install -e ./
       - run: pytest -v -n auto --skip-custom-ops ./tensorflow_addons
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -84,10 +84,6 @@ jobs:
         with:
           python-version: 3.5
       - uses: actions/checkout@v2
-      - uses: actions/cache@v1
-        with:
-          path: ~/.cache/pip
-          key: ${{ runner.os }}-pip-${{ hashFiles('tools/install_deps/*.txt') }}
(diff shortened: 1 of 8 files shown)


.github/workflows/ci_test.yml
```diff
@@ -76,6 +76,21 @@ jobs:
...
+    steps:
+      - uses: actions/setup-python@v1
+        with:
+          python-version: 3.5
+      - uses: actions/checkout@v2
+      - uses: actions/cache@v1
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('tools/install_deps/*.txt') }}
+      - run: pip install -r tools/install_deps/tensorflow-cpu.txt -r tools/install_deps/pytest.txt
```
gabrieldemarmiesse, line 91:
```suggestion
      - uses: actions/checkout@v2
      - run: pip install -r tools/install_deps/tensorflow-cpu.txt -r tools/install_deps/pytest.txt
```

.github/workflows/ci_test.yml
```diff
@@ -76,6 +76,17 @@ jobs:
...
       - run: bash tools/run_build.sh test_editable_mode
+  test_python_code_only:
+    name: Fast build to run python-only tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-python@v1
+        with:
+          python-version: 3.5
+      - uses: actions/checkout@v2
+      - run: pip install -r tools/install_deps/tensorflow-cpu.txt -r tools/install_deps/pytest.txt
```
bhack, line 87: Any speedup caching pip packages? https://github.com/actions/cache/blob/master/examples.md#simple-example

gabrieldemarmiesse: I tried, but I removed it in https://github.com/tensorflow/addons/pull/1666/commits/00832394a4d6a948eb76b139b4cfe984c9d3f092 because we didn't get any speed benefits.

bhack: Have you tried another build After the cache build?

gabrieldemarmiesse: Yes, you can take a look for yourself since the build logs are public for each commit.

PR #1667, opened 2020-04-14
Added tests for AverageModelCheckpoint() callback

### LINKED ISSUE #1502: AverageModelCheckpoint is untested
We have no way of knowing if it's working or not. Some tests would ensure no bugs are introduced.

ashutosh1919:
Fixes #1502 .  
@gabrieldemarmiesse, @shun-lin and @Squadrick , Please review.

### DIFF
--- a/tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
+++ b/tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
@@ -238,7 +238,7 @@ def _run_load_weights_on_restart_test_common_iterations():
     return model, train_ds, filepath, weights_after_one_more_epoch
 
 
-def get_checkpoint_load_weights_on_restart_true_test():
+def test_checkpoint_load_weights():
     (
         model,
         train_ds,
@@ -258,6 +258,6 @@ def get_checkpoint_load_weights_on_restart_true_test():
     np.testing.assert_almost_equal(
         weights_after_one_more_epoch, weights_after_model_restoring_and_one_more_epoch
     )
-    not np.testing.assert_almost_equal(
+    np.testing.assert_almost_equal(
         weights_after_one_more_epoch, weights_with_one_final_extra_epoch
     )
--- a/tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
+++ b/tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
@@ -1,6 +1,5 @@
 import os
 import pytest
-import numpy as np
 import tensorflow as tf
 from tensorflow import keras
 from tensorflow_addons.callbacks import AverageModelCheckpoint
@@ -197,71 +196,4 @@ def test_invalid_save_freq(tmp_path):
     with pytest.raises(ValueError, match="Unrecognized save_freq"):
         AverageModelCheckpoint(
             update_weights=True, filepath=test_model_filepath, save_freq=save_freq
-        )
-
-
-def _get_dummy_resource_for_checkpoint_testing(tmp_path):
-    def get_input_datasets():
-        # Simple training input.
-        train_input = [[1.0]] * 16
-        train_label = [[0.0]] * 16
-        ds = tf.data.Dataset.from_tensor_slices((train_input, train_label))
-        return ds.batch(8, drop_remainder=True)
-
-    # Very simple model to eliminate randomness.
-    optimizer = MovingAverage(
-        tf.keras.optimizers.SGD(lr=2.0), sequential_update=True, average_decay=0.5
-    )
-    model = keras.Sequential()
-    model.add(keras.layers.Dense(units=1, input_shape=(1,)))
-    model.compile(loss="mae", optimizer=optimizer, metrics=["mae"])
-    train_ds = get_input_datasets()
-    filepath = str(tmp_path / "test_model.{epoch:02d}.h5")
-    callback = AverageModelCheckpoint(
-        update_weights=True, filepath=filepath, save_weights_only=True
-    )
-    return model, train_ds, callback, filepath
-
-
-def _run_load_weights_on_restart_test_common_iterations(tmp_path):
-    (model, train_ds, callback, filepath) = _get_dummy_resource_for_checkpoint_testing(
-        tmp_path
-    )
-    initial_epochs = 3
-    model.fit(train_ds, epochs=initial_epochs, callbacks=[callback])
-    # The files should exist after fitting with callback.
-    for epoch in range(initial_epochs):
-        assert os.path.exists(filepath.format(epoch=epoch + 1))
-    model.fit(train_ds, epochs=1)
-    weights_after_one_more_epoch = model.get_weights()
-    # The filepath should continue to exist after fitting without callback.
-    for epoch in range(initial_epochs):
-        assert os.path.exists(filepath.format(epoch=epoch + 1))
-    return model, train_ds, filepath, weights_after_one_more_epoch
-
-
-def test_checkpoint_load_weights(tmp_path):
-    (
-        model,
-        train_ds,
-        filepath,
-        weights_after_one_more_epoch,
-    ) = _run_load_weights_on_restart_test_common_iterations(tmp_path)
-    callback = AverageModelCheckpoint(
-        update_weights=True,
-        filepath=filepath,
-        save_weights_only=True,
-        load_weights_on_restart=True,
-    )
-    model.fit(train_ds, epochs=1, callbacks=[callback])
-    weights_after_model_restoring_and_one_more_epoch = model.get_weights()
-    model.fit(train_ds, epochs=1, callbacks=[callback])
-    weights_with_one_final_extra_epoch = model.get_weights()
-    with np.testing.assert_raises(AssertionError):
-        np.testing.assert_almost_equal(
-            weights_after_one_more_epoch,
-            weights_after_model_restoring_and_one_more_epoch,
-        )
-        np.testing.assert_almost_equal(
-            weights_after_one_more_epoch, weights_with_one_final_extra_epoch
         )
--- a/tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
+++ b/tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
@@ -218,7 +218,7 @@ def _get_dummy_resource_for_checkpoint_testing(tmp_path):
     train_ds = get_input_datasets()
     filepath = str(tmp_path / "test_model.{epoch:02d}.h5")
     callback = AverageModelCheckpoint(
-        update_weights=False, filepath=filepath, save_weights_only=True
+        update_weights=True, filepath=filepath, save_weights_only=True
     )
     return model, train_ds, callback, filepath
 
@@ -248,7 +248,7 @@ def test_checkpoint_load_weights(tmp_path):
         weights_after_one_more_epoch,
     ) = _run_load_weights_on_restart_test_common_iterations(tmp_path)
     callback = AverageModelCheckpoint(
-        update_weights=False,
+        update_weights=True,
         filepath=filepath,
         save_weights_only=True,
         load_weights_on_restart=True,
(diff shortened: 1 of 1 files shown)


ashutosh1919:
@shun-lin , @Squadrick - The build is cancelled. I don't understand. Please help

gabrieldemarmiesse:
That behavior from github actions is very unexpected. Either the tensorflow org has run out of build minutes, either there is a problem with github itself. I'll try to restart the CI.

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
```diff
@@ -0,0 +1,206 @@
...
+import tensorflow as tf
+from tensorflow import keras
+from tensorflow_addons.callbacks import AverageModelCheckpoint
+from tensorflow_addons.optimizers import MovingAverage
+import os
+
+try:
+    import h5py
+except ImportError:
+    h5py = None
```
AakashKumarNain, line 11: Please add a message in the exception

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
```diff
@@ -0,0 +1,206 @@
+import pytest
+import tensorflow as tf
+from tensorflow import keras
+from tensorflow_addons.callbacks import AverageModelCheckpoint
+from tensorflow_addons.optimizers import MovingAverage
+import os
```
AakashKumarNain, line 6: `os` should be imported at the top. Please refer to [PEP8](https://www.python.org/dev/peps/pep-0008/#imports) for more information

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
gabrieldemarmiesse: Actually I believe we can remove this import as it's not used in the file.

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
```diff
@@ -0,0 +1,206 @@
...
+NUM_CLASSES = 2
+INPUT_DIM = 3
+NUM_HIDDEN = 5
+BATCH_SIZE = 5
+EPOCHS = 5
+test_model_filepath = "test_model.h5"
+
+moving_avg = MovingAverage(
+    tf.keras.optimizers.SGD(lr=2.0), sequential_update=True, average_decay=0.5,
+)
```
gabrieldemarmiesse, line 23: Please move the optimizer in the `get_data_and_model`, tests should not interfer with each other, and as such, they should not modify any global object (in this case an optimizer). The optimizer should be re-created in each test.

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
```diff
@@ -0,0 +1,206 @@
...
+
+def test_compatibility_with_some_opts_only():
+    x, y, model = get_data_and_model(optimizer="rmsprop")
+    avg_model_ckpt = AverageModelCheckpoint(
+        update_weights=True, filepath=test_model_filepath
+    )
+    with pytest.raises(
+        TypeError,
+        match="AverageModelCheckpoint is only used when trainingwith"
+        + " MovingAverage or StochasticAverage",
```
gabrieldemarmiesse, line 47:
```suggestion
        " MovingAverage or StochasticAverage",
```

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
```diff
@@ -0,0 +1,206 @@
...
+    ):
+        assert model.fit(
```
gabrieldemarmiesse, line 49:
```suggestion
        model.fit(
```

That would try to convert the keras history into a boolean object. Tests don't necesarily need assert statements.

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
```diff
@@ -0,0 +1,206 @@
...
+        validation_data=(x, y),
+        callbacks=[avg_model_ckpt],
+    )
+    assert os.path.exists(test_model_filepath)
+    os.remove(test_model_filepath)
+
+
+def test_save_best_only():
+    x, y, model = get_data_and_model()
+    save_best_only = True
```
gabrieldemarmiesse, line 138: Here we test that the callback runs, which is good, but we could also check for correctness (the callback does what we want it to do). 

To do that, grab the keras history in the fit, load the saved model, do an evaluation pass and ensure the metric is equal to the best one found in the keras history.


I don't know if it's possible to do that for other tests, but if it's possible, we should aim for that.

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
```diff
@@ -0,0 +1,206 @@
...
+except ImportError:
+    h5py = None
+
+TRAIN_SAMPLES = 10
+NUM_CLASSES = 2
+INPUT_DIM = 3
+NUM_HIDDEN = 5
+BATCH_SIZE = 5
+EPOCHS = 5
+test_model_filepath = "test_model.h5"
```
gabrieldemarmiesse, line 19: Using the same path for all the tests won't work if the tests are all run concurrently (with pytest-xdist for example). We also don't want to pollute the current working directory if one test fails and the file is not removed. The clean solution is to use this fixture: https://docs.pytest.org/en/latest/tmpdir.html#the-tmp-path-fixture , we also won't have to do `os.remove`.

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
```diff
@@ -0,0 +1,206 @@
...
+moving_avg = MovingAverage(
+    tf.keras.optimizers.SGD(lr=2.0), sequential_update=True, average_decay=0.5,
+)
+
+
+def get_data_and_model(optimizer=moving_avg):
+    x = tf.random.normal([TRAIN_SAMPLES, INPUT_DIM])
+    y = tf.random.normal([TRAIN_SAMPLES, NUM_CLASSES])
+    inputs = keras.layers.Input(INPUT_DIM)
+    layer1 = keras.layers.Dense(NUM_HIDDEN, input_dim=INPUT_DIM, activation="relu")(
```
shun-lin, line 30: NIT: since you only have 1 layer maybe this variable to `hidden_layer` is better for readability.

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
ashutosh1919: done

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
ashutosh1919: @gabrieldemarmiesse , please refer to [this comment](https://github.com/keras-team/keras/issues/6977#issuecomment-400207482). I don't think it is a good idea to test `fit()` results with `evaluate()`

ashutosh1919:
@gabrieldemarmiesse , @AakashKumarNain , @shun-lin - Resolved all the reviews except one and modified the code accordingly. The review which is puzzling is to test the functionality of the model. Please give some ideas. I will also try to experiment.

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
gabrieldemarmiesse: You're right, you'll have problems if you compare on the train data. On the validation data, the keras history should give you the same result as a standalone evaluate call since the model does't change during an evaluation pass.

ashutosh1919: FYI, all types of data here are same. And I have tested it and after testing it, I can't strongly suggest that method because the results I was getting are not good.

ashutosh1919: Maybe We should use any particular simple data which doesn't deviate the model.

AakashKumarNain: @ashutosh1919 can you paste your code here which you have tested?

gabrieldemarmiesse: If it doesn't work, it may be that there is a bug in the implementation of the callback, in which case we should decorate the test with `@pytest.xfail` and open an issue to track the bug. @ashutosh1919 we won't ask you to fix the bug, it's outside the scope of this pull request.

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
```diff
@@ -0,0 +1,199 @@
...
+    hidden_layer = keras.layers.Dense(
+        NUM_HIDDEN, input_dim=INPUT_DIM, activation="relu"
+    )(inputs)
+    outputs = keras.layers.Dense(NUM_CLASSES, activation="softmax")(hidden_layer)
+    model = keras.Model(inputs=inputs, outputs=outputs)
+    model.compile(loss="categorical_crossentropy", optimizer=optimizer, metrics=["acc"])
+    return x, y, model
+
+
+def test_compatibility_with_some_opts_only(tmpdir):
```
gabrieldemarmiesse, line 34: Please use `tmp_path` instead of `tmpdir`, the reason is that it uses pathlib, which is a modern library to handle paths. Going forward, we're going to try to remove progressively the usage of `os.path` in addons to make the codebase more readable and bug-free.

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
ashutosh1919: @gabrieldemarmiesse , I think I got the idea of testing this. We need to test the model weights. Got the idea from some tests in tf repo.

AakashKumarNain: I would have expected that weird behavior in earlier versions because of how batchnorm used to operate but that has been resolved long ago. So, IMO, `evaluate` would give you the same result as during the `fit`

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
AakashKumarNain: Yes, we should rely on `pathlib`

ashutosh1919:
@gabrieldemarmiesse , Changed `tmpdir` to `tmp_path`. Also, added last function to test functionality of model by asserting model weights. Please review now.

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
```diff
@@ -0,0 +1,263 @@
...
+        assert os.path.exists(filepath.format(epoch=epoch + 1))
+    model.fit(train_ds, epochs=1)
+    weights_after_one_more_epoch = model.get_weights()
+    # The filepath should continue to exist after fitting without callback.
+    for epoch in range(initial_epochs):
+        assert os.path.exists(filepath.format(epoch=epoch + 1))
+    return model, train_ds, filepath, weights_after_one_more_epoch
+
+
+def get_checkpoint_load_weights_on_restart_true_test():
```
gabrieldemarmiesse, line 241: test functions should have this form: `test_*`. Otherwise they're not run.

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
```diff
@@ -0,0 +1,263 @@
...
+        load_weights_on_restart=True,
+    )
+    model.fit(train_ds, epochs=1, callbacks=[callback])
+    weights_after_model_restoring_and_one_more_epoch = model.get_weights()
+    model.fit(train_ds, epochs=1, callbacks=[callback])
+    weights_with_one_final_extra_epoch = model.get_weights()
+    np.testing.assert_almost_equal(
+        weights_after_one_more_epoch, weights_after_model_restoring_and_one_more_epoch
+    )
+    not np.testing.assert_almost_equal(
```
gabrieldemarmiesse, line 261:
```suggestion
    np.testing.assert_almost_equal(
```

The not has no effect here. Overall those tests show that the model doesn't change during the last epochs, I'm not sure this is what the callback does. I may be mistaken.

ashutosh1919: not is put because the weights will be different both times. Let me see another fn.

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
ashutosh1919: Sorry, didn't notice while naming.

gabrieldemarmiesse:
@googlebot I consent.

tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
```diff
@@ -0,0 +1,267 @@
...
+        assert os.path.exists(filepath.format(epoch=epoch + 1))
+    model.fit(train_ds, epochs=1)
+    weights_after_one_more_epoch = model.get_weights()
+    # The filepath should continue to exist after fitting without callback.
+    for epoch in range(initial_epochs):
+        assert os.path.exists(filepath.format(epoch=epoch + 1))
+    return model, train_ds, filepath, weights_after_one_more_epoch
+
+
+def test_checkpoint_load_weights(tmp_path):
```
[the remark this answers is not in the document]
ashutosh1919: @gabrieldemarmiesse , You are right. We need to think in a better direction.

PR #1670, opened 2020-04-14
Moved run_all_in_graph_and_eager_mode in basic decoder

autoih:
cc @gabrieldemarmiesse. Dividing into several parts since it's large, and not sure if removing `use_gpu` is a good idea. The reason is using gpu seems not helping a lot in LSTM, only if it's self-attention(transformer) based. May need your suggestions.

### DIFF
--- a/tensorflow_addons/seq2seq/tests/basic_decoder_test.py
+++ b/tensorflow_addons/seq2seq/tests/basic_decoder_test.py
@@ -25,7 +25,8 @@ from tensorflow_addons.seq2seq import basic_decoder
 from tensorflow_addons.seq2seq import sampler as sampler_py
 
 
-def test_step_with_training_helper_output_layer(use_output_layer=True):
+@pytest.mark.parametrize("use_output_layer", [True, False])
+def test_step_with_training_helper_output_layer(use_output_layer):
     sequence_length = [3, 4, 3, 1, 0]
     batch_size = 5
     max_time = 8
--- a/tensorflow_addons/seq2seq/tests/basic_decoder_test.py
+++ b/tensorflow_addons/seq2seq/tests/basic_decoder_test.py
@@ -25,7 +25,6 @@ from tensorflow_addons.seq2seq import basic_decoder
 from tensorflow_addons.seq2seq import sampler as sampler_py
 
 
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
About the use of gpu for testing going forward, unless we're testing some code with a custom op, we need a strong reason to run in both cpu and gpu mode.

tensorflow_addons/seq2seq/tests/basic_decoder_test.py
```diff
@@ -24,372 +25,301 @@
 from tensorflow_addons.seq2seq import sampler as sampler_py
 
 
-class BasicDecoderTest(tf.test.TestCase, parameterized.TestCase):
-    """Unit test for basic_decoder.BasicDecoder."""
-
-    @parameterized.named_parameters(
-        ("use_output_layer", True), ("without_output_layer", False)
+def test_step_with_training_helper_output_layer(use_output_layer=True):
```
gabrieldemarmiesse, line 28:
```suggestion
@pytest.mark.parametrize("use_output_layer", [True, False])
def test_step_with_training_helper_output_layer(use_output_layer):
```

tensorflow_addons/seq2seq/tests/basic_decoder_test.py
```diff
@@ -24,372 +25,301 @@
...
-            output_size = my_decoder.output_size
-            output_dtype = my_decoder.output_dtype
-            self.assertEqual(
-                basic_decoder.BasicDecoderOutput(cell_depth, tf.TensorShape([])),
-                output_size,
-            )
-            self.assertEqual(
-                basic_decoder.BasicDecoderOutput(tf.float32, tf.int32), output_dtype
-            )
+@pytest.mark.parametrize("use_mask", [True, False])
```
gabrieldemarmiesse, line 108:
```suggestion
@pytest.mark.parametrize("use_mask", [True, False, None])
```

tensorflow_addons/seq2seq/tests/basic_decoder_test.py
autoih: Oh, yeah, it also tests none use_output_layer. Thanks @gabrieldemarmiesse

PR #1671, opened 2020-04-14
Test the release wheels in a docker build step.

gabrieldemarmiesse:
Related to #1655

Removing dotnet gives us 18GB of extra space, but the `rm -rf ...` takes 1m to complete. To avoid wasting time, the rm is executed in the background.

So with this we don't test the wheels for windows and macos in a fresh environement, but that should be ok because the making of the wheels is similar. I don't expect an issue to happen when creating a wheel only in macos for example.

Instead of having 25m for the linux wheel + 4m10s for testing the windows wheel, we're now at 25m total, while still testing the linux wheel.

We can speed things up further down the road afterwards.

### DIFF
--- a/.github/workflows/make_wheel_Linux.sh
+++ b/.github/workflows/make_wheel_Linux.sh
@@ -1,4 +1,9 @@
 set -e -x
+
+df -h
+
+du -h --max-depth=2 /
+
 DOCKER_BUILDKIT=1 docker build \
     -f tools/docker/build_wheel.Dockerfile \
     --output type=local,dest=wheelhouse \
--- a/.github/workflows/make_wheel_Linux.sh
+++ b/.github/workflows/make_wheel_Linux.sh
@@ -1,13 +1,8 @@
 set -e -x
 
 df -h
-
 docker info
-
-du -h --max-depth=1 /usr/share
-du -h --max-depth=1 /usr/local
-du -h --max-depth=2 /
-
+rm -rf /usr/share/dotnet
 
 DOCKER_BUILDKIT=1 docker build \
     -f tools/docker/build_wheel.Dockerfile \
--- a/.github/workflows/make_wheel_Linux.sh
+++ b/.github/workflows/make_wheel_Linux.sh
@@ -2,9 +2,10 @@ set -e -x
 
 df -h
 
+docker info
+
 du -h --max-depth=2 /
 
-docker info
 
 DOCKER_BUILDKIT=1 docker build \
     -f tools/docker/build_wheel.Dockerfile \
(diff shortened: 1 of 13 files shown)


.github/workflows/make_wheel_Linux.sh
```diff
@@ -1,4 +1,10 @@
 set -e -x
+
+df -h
+docker info
+rm -rf /usr/share/dotnet &
+df -h
```
gabrieldemarmiesse, line 6:
```suggestion
rm -rf /usr/share/dotnet &
```

tools/docker/build_wheel.Dockerfile
```diff
@@ -54,7 +54,18 @@ RUN bash tools/releases/tf_auditwheel_patch.sh
 RUN auditwheel repair --plat manylinux2010_x86_64 artifacts/*.whl
 RUN ls -al wheelhouse/
 
+# -------------------------------------------------------------------
+FROM python:3.5 as test_wheel_in_fresh_environement
```
gabrieldemarmiesse, line 58:
```suggestion
# -------------------------------------------------------------------
ARG PY_VERSION
FROM python:$PY_VERSION as test_wheel_in_fresh_environement
```

tools/docker/build_wheel.Dockerfile
```diff
@@ -61,7 +62,19 @@ RUN bash tools/releases/tf_auditwheel_patch.sh
 RUN auditwheel repair --plat manylinux2010_x86_64 artifacts/*.whl
 RUN ls -al wheelhouse/
 
+# -------------------------------------------------------------------
+
+FROM python:$PY_VERSION as test_wheel_in_fresh_environement
```
seanpmorgan, line 67:
```suggestion
FROM python:$PY_VERSION as test_wheel_in_fresh_environment
```

tools/docker/build_wheel.Dockerfile
```diff
@@ -61,7 +62,19 @@ RUN bash tools/releases/tf_auditwheel_patch.sh
...
+COPY --from=make_wheel /addons/wheelhouse/ /addons/wheelhouse/
+RUN pip install /addons/wheelhouse/*.whl
+
+RUN python -c "import tensorflow_addons as tfa; print(tfa.register_all())"
+
 # -------------------------------------------------------------------
 FROM scratch as output
 
-COPY --from=make_wheel /addons/wheelhouse/ .
+COPY --from=test_wheel_in_fresh_environement /addons/wheelhouse/ .
```
seanpmorgan, line 80:
```suggestion
COPY --from=test_wheel_in_fresh_environment /addons/wheelhouse/ .
```

PR #1674, opened 2020-04-15
Added demo of TQDMProgressBar() for evaluate()

ashutosh1919:
Added demo as per [#1649 (comment)](https://github.com/tensorflow/addons/pull/1649#pullrequestreview-391989472).  
@shun-lin, @gabrieldemarmiesse  - Please review.

### DIFF
--- a/docs/tutorials/tqdm_progress_bar.ipynb
+++ b/docs/tutorials/tqdm_progress_bar.ipynb
@@ -278,6 +278,16 @@
         "**Below is the expected output when you run the cell above**\n",
         "![TQDM Progress Bar Figure](https://raw.githubusercontent.com/tensorflow/addons/59961669a0e21eb4c045d4ad38d008a529d566c2/docs/tutorials/assets/tqdm_progress_bar_demo.png)"
       ]
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "# TQDMProgressBar() also works with evaluate()\n",
+        "model.evaluate(x_test, y_test, batch_size=64, callbacks=[tqdm_callback])"
+      ]
     }
   ]
 }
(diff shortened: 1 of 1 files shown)


docs/tutorials/tqdm_progress_bar.ipynb
```diff
@@ -278,6 +278,16 @@
         "**Below is the expected output when you run the cell above**\n",
```
shun-lin, line 278: <p>you need to set verbose=0, otherwise there will be 2 progress bars.</p><p> </p><p>also if you have time, may you take a screenshot like the one above and link it in this colab? (because when we display the colab on tensorflow.org, the output of the cell won't be saved as display as desired, <a href="https://www.tensorflow.org/addons/tutorials/tqdm_progress_bar" target="_blank">https://www.tensorflow.org/addons/tutorials/tqdm_progress_bar</a>) this is optional, thanks!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/addons/pull/1674/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

ashutosh1919: @shun-lin, yes I tried to do that. But the markdown of the above image has path somewhere from rawgithubcontent. There is no asset folder in the repo. Where should I keep screenshot. Please help me. I don’t know

shun-lin: https://gist.github.com/vinkla/dca76249ba6b73c5dd66a4e986df4c8d

It is kind of weird, but just take a screenshot, drag it into a comment in an issue (or anywhere like here should work) wait a few second until the url comes up. Then you can copy that URL into the colab).

ashutosh1919: evaluatepbar

ashutosh1919:
@shun-lin , changed as per your comment.

PR #1675, opened 2020-04-15
Added code to get the slowest test functions, test files and sum of all durations.

gabrieldemarmiesse:
Related to #1655 and #1316

This should give us useful information. 

the --durations option gives us the slowest tests, but it's not useful because one function may result in multiple tests. With this, it's easier to target what we need to optimize.

The question is do we want that code here or should we make a separate repo that will act as a pytest plugin? I don't have a strong opinion on that.

Note that the code is not clean, I should add some command line arguments to control that. It's just a POC.


gabrieldemarmiesse:
I think I'll make a separate plugin, it's a pretty general feature.

PR #1678, opened 2020-04-16
Fix two testing functions and move out tf.test

autoih:
cc @gabrieldemarmiesse.

### DIFF
--- a/tensorflow_addons/seq2seq/tests/basic_decoder_test.py
+++ b/tensorflow_addons/seq2seq/tests/basic_decoder_test.py
@@ -319,274 +319,243 @@ def test_step_with_sample_embedding_helper():
     np.testing.assert_equal(expected_step_next_inputs, eval_result["step_next_inputs"])
 
 
+def test_step_with_scheduled_embedding_training_helper():
+    sequence_length = [3, 4, 3, 1, 0]
+    batch_size = 5
+    max_time = 8
+    input_depth = 7
+    vocabulary_size = 10
+
+    inputs = np.random.randn(batch_size, max_time, input_depth).astype(np.float32)
+    input_t = tf.constant(inputs)
+    embeddings = np.random.randn(vocabulary_size, input_depth).astype(np.float32)
+    half = tf.constant(0.5)
+    cell = tf.keras.layers.LSTMCell(vocabulary_size)
+    sampler = sampler_py.ScheduledEmbeddingTrainingSampler(
+        sampling_probability=half, time_major=False
+    )
+    initial_state = cell.get_initial_state(batch_size=batch_size, dtype=tf.float32)
+    my_decoder = basic_decoder.BasicDecoder(cell=cell, sampler=sampler)
+    (first_finished, first_inputs, first_state) = my_decoder.initialize(
+        input_t,
+        sequence_length=sequence_length,
+        embedding=embeddings,
+        initial_state=initial_state,
+    )
+    output_size = my_decoder.output_size
+    output_dtype = my_decoder.output_dtype
+    assert (
+        basic_decoder.BasicDecoderOutput(vocabulary_size, tf.TensorShape([]))
+        == output_size
+    )
+
+    assert basic_decoder.BasicDecoderOutput(tf.float32, tf.int32) == output_dtype
+
+    (step_outputs, step_state, step_next_inputs, step_finished,) = my_decoder.step(
+        tf.constant(0), first_inputs, first_state
+    )
+    batch_size_t = my_decoder.batch_size
+
+    assert len(first_state) == 2
+    assert len(step_state) == 2
+    assert isinstance(step_outputs, basic_decoder.BasicDecoderOutput)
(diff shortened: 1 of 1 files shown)


tensorflow_addons/seq2seq/tests/basic_decoder_test.py
```diff
@@ -319,274 +319,243 @@ def test_step_with_sample_embedding_helper():
...
+        eval_result["first_finished"].numpy(),
+    )
+    np.testing.assert_equal(
+        np.asanyarray([False, False, False, True, True]),
+        eval_result["step_finished"].numpy(),
+    )
+    sample_ids = eval_result["step_outputs"].sample_id
+    assert output_dtype.sample_id == sample_ids.dtype
+    batch_where_not_sampling = tf.cast(np.where(sample_ids == -1), dtype=tf.int32)
+    batch_where_sampling = tf.cast(np.where(sample_ids > -1), dtype=tf.int32)
```
gabrieldemarmiesse, line 392: Could you explain a bit where the tf.cast come from? Are you fixing a bug or something like that?

tensorflow_addons/seq2seq/tests/basic_decoder_test.py
```diff
@@ -319,274 +319,243 @@ def test_step_with_sample_embedding_helper():
...
+            tf.gather(eval_result["step_next_inputs"], batch_where_sampling)
+        ).numpy(),
+        tf.squeeze(tf.gather(embeddings[sample_ids, :], batch_where_sampling)).numpy(),
+    )
+    np.testing.assert_equal(
+        tf.squeeze(
+            tf.gather(eval_result["step_next_inputs"], batch_where_not_sampling)
+        ),
+        np.squeeze(inputs[batch_where_not_sampling, 1], axis=0),
+    )
```
gabrieldemarmiesse, line 405: Here could you detail a bit more why you did the changes? It's not clear to me.

tensorflow_addons/seq2seq/tests/basic_decoder_test.py
```diff
@@ -319,274 +319,243 @@ def test_step_with_sample_embedding_helper():
...
-        )
+    np.testing.assert_equal(
+        tf.squeeze(
+            tf.gather(eval_result["step_next_inputs"], batch_where_not_sampling)
+        ),
+        np.concatenate(
+            (
+                np.squeeze(inputs[batch_where_not_sampling, 1], axis=0),
+                tf.squeeze(
+                    tf.gather(auxiliary_inputs_to_concat, batch_where_not_sampling)
```
gabrieldemarmiesse, line 549: Same question here.

tensorflow_addons/seq2seq/tests/basic_decoder_test.py
autoih: The original slice method can't pass the testing until the changes, i.e., dimension inconsistent.

tensorflow_addons/seq2seq/tests/basic_decoder_test.py
autoih: Same reason.

tensorflow_addons/seq2seq/tests/basic_decoder_test.py
gabrieldemarmiesse: Why did the tests passed before then? Were the test functions not run? Or is it because it's in eager mode now?

tensorflow_addons/seq2seq/tests/basic_decoder_test.py
autoih: Thanks @gabrieldemarmiesse. This part may not be necessary, it just uses tensor for slicing. Using original slicing can't pass the testing, so come up with this way; not sure any easy way to fix it.

gabrieldemarmiesse:
Ok I found the reason. When calling evaluate on the dictionary to get `eval_result`, all the values in the dictionary become numpy arrays (not TF tensors). Since here we're not calling evaluate, we end up with tensorflow tensors in the dictionary. As such, the same functions cannot be called. A cleaner change would be to call `.numpy()` when creating the `eval_result` dictionary, then we wouldn't need to do that cast, gather and such.

tensorflow_addons/seq2seq/tests/basic_decoder_test.py
AakashKumarNain: Can we just quickly check the `dtype` of `sample_ids` before casting? I am asking because it shouldn't have passed the test cases earlier as well if casting was required

gabrieldemarmiesse: @AakashKumarNain , tests were passing before because all tensors were numpy tensors or tuples. See https://github.com/tensorflow/addons/pull/1678#issuecomment-614757184

autoih:
Thanks @gabrieldemarmiesse. It looks cleaner instead of simply conversion.

PR #1680, opened 2020-04-16
Remove the cleaning of the cache to speed up the build.

gabrieldemarmiesse:
See #1655

We cleaned the cache to make sure that the options for building the wheels were taken into account, but one could argue that it's bazel's job to invalidate the cache, not ours, so we can let bazel decide what to do.

When building the wheel, we went from 5m to 1s.


bhack:
@gabrieldemarmiesse I still see `https://github.com/tensorflow/addons/runs/591598847` 21m 20s. Is this just a local effect?

PR #1685, opened 2020-04-17
Update to rc3 with py38 for all OS

seanpmorgan:


.github/workflows/release.yml
```diff
@@ -58,9 +58,7 @@ jobs:
           python-version: ${{ matrix.py-version }}
       - name: Build wheels
         if: |
-          (matrix.os != 'macos-latest' || matrix.py-version != '3.8')
-          && (matrix.py-version != '3.8' || matrix.tf-version != '2.1.0')
-          && (matrix.os != 'ubuntu-18.04' || matrix.tf-version != '2.2.0rc2')
```
seanpmorgan, line 63: @gabrieldemarmiesse I wasn't sure for the logic of this 3rd check. Only mac/windows build for 2.2 , is it safe to remove this now or was there a need for that logic check. Only think I could think of was that the build breaks in the 2.1 docker container... but we should fix that (in a separate PR is this one fails)

seanpmorgan: Wheel appears to have built successfully

gabrieldemarmiesse: Ho dang, i just though about it, just 10 seconds after the merge. It won't work for linux py3.8 because py3. 8 is not in the docker image

PR #1688, opened 2020-04-17
Added LossFunctionWrapper for losses and simplified code

### LINKED ISSUE #1324: Make LossFunctionWrapper public

ashutosh1919:
Fixes #1324 .  
The ideas of the fix are taken from PR #1329 .  
@gabrieldemarmiesse , please review.


ashutosh1919:
@AakashKumarNain , @gabrieldemarmiesse and @WindQAQ - If the code added for `LossFunctionWrapper` in this PR is correct, then it is exposing some of the wrong test cases written for some loss in my opinion. Please give your comments.

ashutosh1919:
@WindQAQ , I have undone changes for npairs and sparsemax_loss. I think it is good to create separate issue to correct logical errors in these two losses and then we can include wrapper to it.

ashutosh1919:
@WindQAQ , let me also dive deeper into `npairs` and `sparsemax_loss`. Let's change that too.

tools/testing/source_code_test.py
```diff
@@ -91,6 +91,11 @@ def test_no_private_tf_api():
         "tensorflow_addons/optimizers/moving_average.py",
         "tensorflow_addons/metrics/r_square.py",
         "tensorflow_addons/utils/test_utils.py",
+        "tensorflow_addons/losses/contrastive.py",
+        "tensorflow_addons/losses/focal_loss.py",
+        "tensorflow_addons/losses/lifted.py",
+        "tensorflow_addons/losses/quantiles.py",
+        "tensorflow_addons/losses/triplet.py",
```
gabrieldemarmiesse, line 98: Friendly reminder that those rules are here to enforce the usage of the public TF API only, by contributors in the future, so that we can be forward compatible with tensorflow. The blacklist is here to keep track of the things we have not migrated yet, and so should not grow :) Maybe I should have written a clearer message in the comment above.

PR #1691, opened 2020-04-17
Removed run_all_in_graph_and_eager_mode in cohens_kappa.

gabrieldemarmiesse:


AakashKumarNain:
Apart from one query, everything is looking good.

tensorflow_addons/metrics/tests/cohens_kappa_test.py
```diff
@@ -18,180 +18,175 @@
...
 
-        obj = CohenKappa(num_classes=5, sparse_labels=True)
-        self.evaluate(tf.compat.v1.variables_initializer(obj.variables))
+def check_results(objs, values):
+    obj1, obj2, obj3 = objs
+    val1, val2, val3 = values
 
-        self.evaluate(obj.update_state(y_true, y_pred))
-        self.assertAllClose(0.19999999, obj.result())
+    np.testing.assert_allclose(val1, obj1.result(), atol=1e-5)
```
AakashKumarNain, line 61: Shouldn't we check for a higher tolerance value, like 1e-6 or 1e-7?

tensorflow_addons/metrics/tests/cohens_kappa_test.py
```diff
@@ -18,180 +18,175 @@
...
+    np.testing.assert_allclose(val2, obj2.result(), atol=1e-5)
```
AakashKumarNain, line 62: Same question

tensorflow_addons/metrics/tests/cohens_kappa_test.py
gabrieldemarmiesse: Possibly, I kept the tolerances used previously. Maybe that can be done in another pull request?

AakashKumarNain: Sure. Let's check for higher in the next one

AakashKumarNain:
Thanks for the refactor.

PR #1699, opened 2020-04-17
Silence memset warning

WindQAQ:
For gcc >= 8.0 (our docker uses 8.3), using `memset` to clearing non-trivial object will throw the following warning.

image

To silence it, we can explicitly cast the first argument with type (T*) to (void*).


WindQAQ:
Oops, it's my fault. GCC 8.3 is the version used to compile python instead of the one of toolchain. Thanks for pointing out!

PR #1702, opened 2020-04-18
Fixed average_model_checkpoint to have integer save_freq

### LINKED ISSUE #1701: save_freq is not accepting integer value in AverageModelCheckpoint
It is observed that `AverageModelCheckpoint` is derived from `ModelCheckpoint`. `ModelCheckpoint` accepts `save_freq` param as both `epoch` string and integer. But it is not the case with `AverageModelCheckpoint`. The problem is because of the typing annotations. `save_freq` is given only `str` annotation, which should be `Union[str, int]`.

ashutosh1919:
Fixes #1701 .  
Also added test cases to verify that the changes done works perfectly.  
@gabrieldemarmiesse and @shun-lin - Please review.

### DIFF
--- a/tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
+++ b/tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
@@ -190,6 +190,31 @@ def test_save_freq(tmp_path):
     assert os.path.exists(test_filepath.format(epoch=5))
 
 
+def test_save_freq_integer(tmp_path):
+    test_filepath = str(tmp_path / "test_model.{epoch:02d}.h5")
+    x, y, model = get_data_and_model()
+    save_freq = 2
+    avg_model_ckpt = AverageModelCheckpoint(
+        update_weights=False,
+        save_best_only=False,
+        filepath=test_filepath,
+        save_freq=save_freq,
+    )
+    model.fit(
+        x,
+        y,
+        epochs=EPOCHS,
+        batch_size=BATCH_SIZE,
+        validation_data=(x, y),
+        callbacks=[avg_model_ckpt],
+    )
+    assert os.path.exists(test_filepath.format(epoch=1))
+    assert os.path.exists(test_filepath.format(epoch=2))
+    assert os.path.exists(test_filepath.format(epoch=3))
+    assert os.path.exists(test_filepath.format(epoch=4))
+    assert os.path.exists(test_filepath.format(epoch=5))
+
+
 def test_invalid_save_freq(tmp_path):
     test_model_filepath = str(tmp_path / "test_model.h5")
     save_freq = "invalid_save_freq"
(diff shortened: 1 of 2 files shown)


tensorflow_addons/callbacks/tests/avg_model_checkpoint_test.py
```diff
@@ -190,6 +190,31 @@ def test_save_freq(tmp_path):
...
+        epochs=EPOCHS,
+        batch_size=BATCH_SIZE,
+        validation_data=(x, y),
+        callbacks=[avg_model_ckpt],
+    )
+    assert os.path.exists(test_filepath.format(epoch=1))
+    assert os.path.exists(test_filepath.format(epoch=2))
+    assert os.path.exists(test_filepath.format(epoch=3))
+    assert os.path.exists(test_filepath.format(epoch=4))
+    assert os.path.exists(test_filepath.format(epoch=5))
```
gabrieldemarmiesse, line 215: I'm not sure this is the best thing to check as we'd get a similar result if we had `save_freq="epoch"`. Since there are 5 epochs with 10 samples eachs, we then see 50 samples.  One good way to test would be to put `save_freq=25` and then check that only 2 files are present in the directory.

ashutosh1919: @gabrieldemarmiesse , Thanks for noticing this. Now, I have run sample code for this on `ModelCheckpoint` and it is not working. If I put `file.batch{batch:02d}epoch{epoch:02d}.h5` in filepath name, it is giving error like below:
```python
KeyError                                  Traceback (most recent call last)
/usr/local/lib/python3.6/dist-packages/tensorflow/python/keras/callbacks.py in _get_file_path(self, epoch, logs)
   1243         # placeholders can cause formatting to fail.
-> 1244         return self.filepath.format(epoch=epoch + 1, **logs)
   1245       except KeyError as e:

KeyError: 'batch'

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
8 frames
/usr/local/lib/python3.6/dist-packages/tensorflow/python/keras/callbacks.py in _get_file_path(self, epoch, logs)
   1245       except KeyError as e:
   1246         raise KeyError('Failed to format this callback filepath: "{}". '
-> 1247                        'Reason: {}'.format(self.filepath, e))
   1248     else:
   1249       # If this is multi-worker training, and this worker should not

KeyError: 'Failed to format this callback filepath: "test_model.batch{batch:02d}epoch{epoch:02d}.hdf5". Reason: \'batch\''
```

And when searching the error in tf code, I found out that there is no support for batches in tf code even if they have this option of having integer `save_freq`. I can say that from [this line](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/callbacks.py#L1312). Please suggest me what should I do?

ashutosh1919:
@gabrieldemarmiesse , I have raised issue [#38668](https://github.com/tensorflow/tensorflow/issues/38668) and PR [#38669](https://github.com/tensorflow/tensorflow/pull/38669) on tf repo.  
But should we keep this PR open or should we close this?

shun-lin:
> #38669

Thanks for the PR!

bhack:
I will close this. We will track this issue at https://github.com/tensorflow/tensorflow/issues/38668

PR #1706, opened 2020-04-18
Fix Ubuntu py38 build

seanpmorgan:
For some reason the python3.8 build installs the pytest.txt prior to tensorflow:
https://github.com/tensorflow/addons/runs/598441627

This caused issue since numpy was not available for scikit-image installation. I couldn't find documentation about why this order would be changed.

This is a more explicit order regardless.

Marking WIP -- confirming locally since CI won't test this until it merges

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,7 +16,8 @@ on:
 
 env:
   BAZEL_VERSION: 1.1.0
-  OLDEST_PY_VERSION: '3.5'
+  MIN_PY_VERSION: '3.5'
+  MAX_PY_VERSION: '3.8'
 
 jobs:
   test-with-bazel:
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse:
Maybe for pull requests, we could tests against python 3.5 and 3.8, and skip 3.6 and 3.7? That would help because it's not the first time we get problems when inserting a new python version. That would help with forward compatibility too.

seanpmorgan:
> Maybe for pull requests, we could tests against python 3.5 and 3.8, and skip 3.6 and 3.7? That would help because it's not the first time we get problems when inserting a new python version. That would help with forward compatibility too.

I changed it so all python builds are tested, these are done independently now so we shouldn't see any test time increase.

gabrieldemarmiesse:
Since the pool of concurrent builds is shared with the rest of the TF org, if possible, it'd be best to skip builds that have a very low probability of failure. Once more and more repos are using github actions, we'll see the benefits.

gabrieldemarmiesse:
https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits

I think the tf org has 180 concurrent builds total. And another limit is 50 concurrent macos jobs.

gabrieldemarmiesse:
Great thanks!

Issue #643, opened 2019-09-10
Expected Center Loss


KiryanovKD:
**System information**

TensorFlow version (you are using): 2 rc
Are you willing to contribute it (Yes/No): No

**Describe the feature and the current behavior/state.**
Expected Center Loss can help in learning discriminative features in seq training without framewising

**Will this change the current api? How?**
No

**Who will benefit with this feature?**
Anyone who wants to use this loss function in ASR or OCR

**Any Other info.**
https://arxiv.org/pdf/1905.06907.pdf

seanpmorgan:
TensorFlow Addons is transitioning to a minimal maintenance and release mode. New features will not be added to this repository. For more information, please see our public messaging on this decision:
[TensorFlow Addons Wind Down](https://github.com/tensorflow/addons/issues/2807)

Please consider sending feature requests / contributions to other repositories in the TF community with a similar charters to TFA:
[Keras](https://github.com/keras-team/keras)
[Keras-CV](https://github.com/keras-team/keras-cv)
[Keras-NLP](https://github.com/keras-team/keras-nlp)

PR #1709, opened 2020-04-19
Added Center loss

ashutosh1919:
Fixes [#643](https://github.com/tensorflow/addons/issues/643).  
I observed that last PR was closed and there was no activity since many days. That's why I am raising fresh PR on the issue. Modified some code for data type consistency.  
@gabrieldemarmiesse and @shun-lin  - Please review.


gabrieldemarmiesse:
I'm just wondering, did you have any issue with running tests locally? Pusing a lot of commits id not a problem, but maybe our contributing guide needs improvements if some people are not able to run tests locally.

ashutosh1919:
> I'm just wondering, did you have any issue with running tests locally? Pusing a lot of commits id not a problem, but maybe our contributing guide needs improvements if some people are not able to run tests locally.

@gabrieldemarmiesse , sorry for so many commits. But I am not able to setup the environment properly for running tests locally. Let me try on collab. Sorry for these many commits.

gabrieldemarmiesse:
The many commits are not a problem. We squash them at the end anyway. I'm just trying to understand what are the issues that users are facing so that we can improve the contributing guide. As a side note, I really don't recommend using colab as it's hard to edit files. Could you explain what are the steps where you encounter issies in the contributing guide?

ashutosh1919:
@gabrieldemarmiesse , yes sure.  
I am using windows 10.
doing `pytest` causes import error of tf even if I have tf 2.0 installed in my system.
I am only able to test `flake8` and able to do `black` locally.

gabrieldemarmiesse:
It's likely that you have multiple interpreters installed locally..when it happens, the good solution is to use `python -m pytest` or `python -m pip install` ot ensures that it uses the pip and pytest linked to the "python" you are currently using. "python -m black" and "python -m flake8" should work too. Could you try ans report if it works?

ashutosh1919:
@gabrieldemarmiesse , Still don't work.
I tried doing it on fresh docker container with python 3.6, tensorflow-2.1.0. 
when I do `pip show tensorflow`, it shows the correct installed version. But whenever I am doing `pytest ...` or `import tensorflow as tf` in python3, it is showing `ModuleNotFound : tensorflow`. I tried doing same in venv. But still don't work.

gabrieldemarmiesse:
Could you give me the output of:

* `python -V`
* `python -m pip freeze`
* `python -c "import tensorflow"`
* `python -c "import pytest"`

ashutosh1919:
- `python3 -V`:
`Python 3.6.9`
- `python3 -m pip freeze`:
```python
absl-py==0.9.0
apipkg==1.5
appdirs==1.4.3
asn1crypto==0.24.0
astor==0.8.1
attrs==19.3.0
cachetools==4.1.0
certifi==2020.4.5.1
chardet==3.0.4
cryptography==2.1.4
cycler==0.10.0
decorator==4.4.2
distlib==0.3.0
execnet==1.7.1
filelock==3.0.12
gast==0.2.2
google-auth==1.14.0
google-auth-oauthlib==0.4.1
google-pasta==0.2.0
grpcio==1.28.1
h5py==2.10.0
idna==2.6
imageio==2.8.0
importlib-metadata==1.6.0
importlib-resources==1.4.0
joblib==0.14.1
Keras-Applications==1.0.8
Keras-Preprocessing==1.1.0
keyring==10.6.0
keyrings.alt==3.0
kiwisolver==1.2.0
Markdown==3.2.1
matplotlib==3.2.1
more-itertools==8.2.0
networkx==2.4
numpy==1.18.2
oauthlib==3.1.0
opt-einsum==3.2.1
packaging==20.3
Pillow==7.0.0
pluggy==0.13.1
protobuf==3.11.3
py==1.8.1
pyasn1==0.4.8
pyasn1-modules==0.2.8
pycrypto==2.6.1
pygobject==3.26.1
pyparsing==2.4.7
pytest==5.4.1
pytest-forked==1.1.3
pytest-xdist==1.31.0
python-apt==1.6.5+ubuntu0.2
python-dateutil==2.8.1
PyWavelets==1.1.1
pyxdg==0.25
requests==2.23.0
requests-oauthlib==1.3.0
rsa==4.0
scikit-image==0.15.0
scikit-learn==0.22.2.post1
scipy==1.4.1
SecretStorage==2.3.1
six==1.14.0
tensorboard==2.1.1
-e git+https://github.com/ashutosh1919/addons.git@863c08e685cb368863b74c5ef42fd1bace5ceaec#egg=tensorflow_addons
tensorflow-estimator==2.1.0
termcolor==1.1.0
tqdm==4.45.0
typeguard==2.7.1
unattended-upgrades==0.1
urllib3==1.25.9
virtualenv==20.0.18
wcwidth==0.1.9
Werkzeug==1.0.1
wrapt==1.12.1
zipp==3.1.0
```
- `python3 -c "import tensorflow"`
```python
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'tensorflow'
```
- `python3 -c "import pytest"`:
Doesn't output anything.

gabrieldemarmiesse:
So it seems tensorflow is not installed in your python3 (it must be installed in another environment.

You should do `python3 -m pip install tensorflow-cpu` and then `python3 -c "import tensorflow"` should work. In this case you should be able to use `python3 -m pytest ./path/to/test/file.py` without any issues.

ashutosh1919:
@gabrieldemarmiesse , Thank you so much. Today's whole day work payed off. I am now able to run tests for both tf and tfa locally in same docker container but in different virtualenv. It's great. Thank you so much for valuable help.  
One small thing that I want you to notice is that for pytest to run locally, user needs to have tf version >=2.1.0 but <=2.3.0. This thing is not mentioned in the doc. I think we should mention that.

gabrieldemarmiesse:
I believe we already mention it in the [readme](https://github.com/tensorflow/addons#python-op-compatibility-matrix), but we definitly should mention it in the CONTRIBUTING.md to make it clear, it's true that it's not obvious. Thanks for the advice! I'll make a PR.

ashutosh1919:
Can anyone help me here? How to use `tf.Variable` inside `tf.function`. This one is really tricky. Not able to do it.

ashutosh1919:
@WindQAQ , after doing this, I am getting following error : 
```python
ERROR    tensorflow:test_util.py:1644 Container localhost does not exist. (Could not find resource: localhost/center_loss/centers)
         [[node center_loss/Gather (defined at /src/addons/tensorflow_addons/losses/center_loss.py:102) ]]
```
Can you tell me how this error occurs?

gabrieldemarmiesse:
The solution is to avoid using tf.test.TestCase and to use plain pytest functions

ashutosh1919:
**Credits :** 
- Thanks @gabrieldemarmiesse for resolving issues on local testing of code and pytesting of Loss.
- Thanks @WindQAQ for hinting me with stateful objects and their non-compatibility with Loss.  
Please review, ready for changes required.

tensorflow_addons/losses/center_loss.py
```diff
@@ -0,0 +1,86 @@
...
+    CNNs to have a better result.
+    We expect labels `y_pred` must be the output of the last fully connected
+    layer, which is 2-D float `Tensor` of l2 normalized embedding vectors and
+    `y_true` to be provided as 1-D integer `Tensor` with shape [batch_size]
+    of multi-class integer labels.
+
+    Args:
+      name: Optional name for the op.
+      reduction: (Optional) Type of `tf.keras.losses.Reduction` to apply to
+        loss. Default value is `SUM_OVER_BATCH_SIZE`.
```
WindQAQ, line 41: The doc does not match actual arguments.

tensorflow_addons/losses/center_loss.py
```diff
@@ -0,0 +1,86 @@
...
+    def loss_fn(self, labels, feature, alpha):
+        labels = tf.convert_to_tensor(labels)
+        feature = tf.convert_to_tensor(feature)
+        len_features = feature.shape[-1]
+        num_classes = labels.shape[-1]
+        centers = tf.Variable(
+            tf.constant(0.0, shape=[num_classes, len_features]),
+            name="centers",
+            dtype=tf.float32,
+        )
```
WindQAQ, line 66: Maybe it is better to move variable creation into `build`. Also please use `add_weight` to create variables when subclassing `Layer`

https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer#add_weight

tensorflow_addons/losses/center_loss.py
```diff
@@ -0,0 +1,86 @@
...
+        labels = tf.reshape(labels, [-1])
+        centers_batch = tf.gather(centers, labels)
+        diff = centers_batch - feature
+        unique_label, unique_idx, unique_count = tf.unique_with_counts(labels)
+        appear_times = tf.gather(unique_count, unique_idx)
+        appear_times = tf.reshape(appear_times, [-1, 1])
+        diff = diff / tf.cast((1 + appear_times), tf.float32)
+        diff = alpha * diff
+        # update centers
+        centers_update_op = tf.compat.v1.scatter_sub(centers, labels, diff)
```
WindQAQ, line 76: Is it possible to use `tf.tensor_scatter_nd_sub` here?

https://www.tensorflow.org/api_docs/python/tf/tensor_scatter_nd_sub

ashutosh1919:
@WindQAQ , In my opinion, we should put this into layers. Reason - TF repo has something similar like this for PRelu which is derived from `Layer` and they have put it in `advanced_activations.py` in layers.

tensorflow_addons/losses/center_loss.py
ashutosh1919: @WindQAQ, Here we are using two shapes `labels.shape[-1]` and `feature.shape[-1]` to define `centers` variable. But in `build` method, we can only get one of them, explicitly saying only `labels.shape`. So, to change that, I have to pass `num_classes` explicitly in `__init__`. So, should we do that or should we keep this?

tensorflow_addons/losses/center_loss.py
ashutosh1919: Done

WindQAQ:
So in keras, the `call` for `Layer` is designed to accept **a Tensor** or **List/Tuple of Tensors**.

In this way, you might pass args like `loss = center_loss_obj([y_true, y_pred])`. Then, you can get the shape of `y_true` by `input_shape[0]` and the one of `y_pred` by `input_shape[1]`.

You can see the similar practice in `multihead_attention` and its test.
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/layers/multihead_attention.py
https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer#call

However, if you do this, you cannot compile this loss as what we do normally because the loss  is supposed to be a callable object with signature `call(y_true, y_pred)` instead of a list of tensors.

This is why I would like to have more suggestions on this PR before merging it because it is a little bit weird than the normal one LOL.

- If subclassing `Layer` and follow the convention of its `call` signature, it cannot be compiled with `model.compile(loss=CenterLoss())`.
- If subclassing `Loss`, it breaks the rule that anything stateful subclasses `Layer`.

IMO, the better (might not be the best) solution is to subclass `Layer` but not to follow the signature `call(inputs, **kargs)`. This is exactly what you do now. But let's just leave the design alone until we receive more suggestions :-)

ashutosh1919:
@gabrieldemarmiesse , Can you take a look at @WindQAQ's last comment and update me? Thanks!

gabrieldemarmiesse:
Would it be possible for it to be a simple layer in the model? And then we do `self.add_loss` in the layer.call? I didn't look too much into the pull request, but from what you're saying @WindQAQ this looks like the API we should use.

tensorflow_addons/losses/center_loss.py
```diff
@@ -0,0 +1,86 @@
...
+        centers_batch = tf.gather(centers, labels)
+        diff = centers_batch - feature
+        unique_label, unique_idx, unique_count = tf.unique_with_counts(labels)
+        appear_times = tf.gather(unique_count, unique_idx)
+        appear_times = tf.reshape(appear_times, [-1, 1])
+        diff = diff / tf.cast((1 + appear_times), tf.float32)
+        diff = alpha * diff
+        labels = tf.reshape(labels, [-1, 1])
+        # update centers
+        centers_update_op = tf.tensor_scatter_nd_sub(centers, labels, diff)
```
gabrieldemarmiesse, line 76: Does the variable needs to be trainable? Also does the variable needs to be re-created at each forward pass?

ashutosh1919: Yes that variable is trainable IMO. Also, with what I understood in the loss, I think the centers are shifting. So, it will be recalculated at each forward pass.

seanpmorgan:

Issue #1682, opened 2020-04-16
Enable multiprocessing when testing with GPU and support distributed strategies in the tests.


gabrieldemarmiesse:
**Describe the feature and the current behavior/state.**

Here I'm not going to discuss the bazel case as it's much more complicated to handle, and we currently advertize using pytest anyway to run the tests. We can of course make sure everything is compatible though.

This revamping of gpu testing has multiple objectives:

* The tests should behave the same weither the contributor has a gpu or not. Meaning we shouldn't run all the tests on a gpu just because a gpu is available, otherwise it hurts reproducibility.
* The test suite should be able to run with multiple workers in kokoro or when a user has multiple gpus. Pytest should use all gpus visible by the main process.
* We need to support testing with distributed strategies. Currently it doesn't work. A fix has been started in #1209 but we need to update it for pytest. 
* Making the whole thing simple to use and to maintain. Notably, we would get rid of this file: https://github.com/tensorflow/addons/blob/master/tools/testing/parallel_gpu_execute.sh which is quite hard to work on.

To do all that, here is my proposal:

Stuff to know: 
* Pytest-xdist uses multiprocessing to run the tests, not multithreading. 
* 2 environement variables are available in each of the workers to identify them. https://github.com/pytest-dev/pytest-xdist#identifying-the-worker-process-during-a-test

### Test workers
Suppose we have a machine with 10CPUs and 4 GPUs, 10 processes will start to run the test suite. Workers 0 to 3 will have ownership of one GPU each (we can use CUDA_VISIBLE_DEVICES to enforce that, but I'm not even sure that's needed with the proposed implementation). Workers 4 to 9 will have no gpu available.

### Virtual devices
Each of those processes, when starting, will split their physical device into 2 virtual device. Tests that just need to run on gpu will use the first of those virtual devices. Processes which need to test distributed strategies will use the two of them. We assume here that 2 virtual devices are enough to test distributed strategies.

### Impact on the contributors:
For this whole machinery to work, we need to know which test needs to run on CPU, GPU, or in distributed strategies. To do that we'll use pytest markers: `@pytest.mark.....`

* By default, if no marker is found, the test will run on CPU: `with device("CPU:0")`. It's equivalent to 
`@pytest.mark.run_on(["cpu"])`.
* To run with gpu only: `@pytest.mark.run_on(["gpu"])`.
* To run on the cpu and gpu: `@pytest.mark.run_on(["cpu", "gpu"])` (test runs twice)
[...]

PR #1713, opened 2020-04-21
Device selection with pytest.

gabrieldemarmiesse:
Fixes part of #1682 

Note that I left out the idea of using many gpus for now as it's not necessary. One is more than enough. 

This focuses on enabling multiprocessing with the gpu available.
This also have another benefit: it doesn't matter if your system has a gpu or not anymore, the same tests will get exectuted on the same devices.

We can now use many workers, even when only one GPU is available to run the tests.

It's compatible with calling manually `with tf.device("GPU:0")`, but I believe it's cleaner to use this built-in solution so we should replace the leftovers later on.

Also in the TODO, instead of using one virtual device per process, we can make multiple ones, and then do something like `@pytest.mark.with_device(["mirror_strategy"])`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -295,6 +295,10 @@ Run selected tests:
 python3 -m pytest path/to/file/or/directory/to/test
 ```
 
+Run the cpu only tests with `pytest -m 'not needs_gpu' ./tensorflow_addons`.
+Run the gpu only tests with `pytest -m needs_gpu ./tensorflow_addons`.
+
+
 #### Testing with Bazel
 
 Testing with Bazel is still supported but not recommended unless you have prior experience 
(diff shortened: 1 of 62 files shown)


CONTRIBUTING.md
```diff
@@ -295,6 +295,10 @@ Run selected tests:
 python3 -m pytest path/to/file/or/directory/to/test
 ```
 
+Run the cpu only tests with `pytest -m 'not needs_gpu' ./tensorflow_addons`.
+Run the gpu only tests with `pytest -m needs_gpu ./tensorflow_addons`.
```
seanpmorgan, line 299:
```suggestion
Run the gpu only tests with `pytest -m needs_gpu ./tensorflow_addons`.
Run the cpu only tests with `pytest -m 'not needs_gpu' ./tensorflow_addons`.
```

Very minor, but think it's a bit more intuive in the doc to read the positive and then the negative afterward. Similar to how it's done on pytest markers docs. This pytest marker methodology is also not my usage pattern but it's very useful

gabrieldemarmiesse:
Yeah, I totally forgot to remove the logging of the devices. It's done now.

PR #1714, opened 2020-04-21
In scheduled samplers, don't convert sampling_probability if it's a tf.Variable

ageron:
To change the `sampling_probability` during training, the simplest option is to make it a `tf.Variable` (alternatively, it could be a placeholder). However, the current constructors of the scheduled samplers call `tf.convert_to_tensor(sampling_probability)`, which converts variables into static tensors. This forces the user to write:

```python
sampler = tfa.seq2seq.sampler.ScheduledEmbeddingTrainingSampler(
    sampling_probability=0.,
    embedding_fn=decoder_embedding_layer)
sampler.sampling_probability = tf.Variable(0.)
```

This is not too hard, but it's not immediately obvious (and the documentation does not cover this), and it really feels like a hack. It would be more natural to just keep the `sampling_probability` as a variable in the constructor, so we could just write:

```
sampling_probability = tf.Variable(0.)
sampler = tfa.seq2seq.sampler.ScheduledEmbeddingTrainingSampler(
    sampling_probability=sampling_probability,
    embedding_fn=decoder_embedding_layer)
```

Note: I also fixed a typo `_sampling_probability` => `sampling_probability`

### DIFF
--- a/tensorflow_addons/seq2seq/sampler.py
+++ b/tensorflow_addons/seq2seq/sampler.py
@@ -336,9 +336,12 @@ class ScheduledEmbeddingTrainingSampler(TrainingSampler):
             raise ValueError(
                 "embedding_fn is expected to be callable, got %s" % type(embedding_fn)
             )
-        self.sampling_probability = tf.convert_to_tensor(
-            sampling_probability, name="sampling_probability"
-        )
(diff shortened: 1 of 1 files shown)


guillaumekln:
Thanks!

ageron:
Thanks for merging.

PR #1715, opened 2020-04-22
Added layer_test in addons and refactoring the code

ashutosh1919:
We need to modify code in many files to resolve issue #1567 .  
This PR adds `layer_test` in `testing_utils` of addons.
@gabrieldemarmiesse - I have removed all codes in the function corresponding to model checking and `set|get_weights()`. Currently, tests will fail because the method doesn't return anything (previously it was returning something like `actual_output` which was part of Sequential API tests part.). Please review the code and guide me to the right direction.


gabrieldemarmiesse:
* We don't need a model for the tests but we still need to do a foward pass to ensure that no exception is raised. To get the output values just do `output_tensor = layer(input_tensor).numpy()`.  So we can still return the result without using a keras model.

* Another thing to check is that the layer output the same numbers after being serialized and deserialized.

ashutosh1919:
@gabrieldemarmiesse , I have added codes as per you comment. But still many tests fails. Tests fails because of lib `NotFoundError`. Please can you take a look and give me a hint?

tensorflow_addons/utils/test_utils.py
```diff
@@ -25,8 +25,90 @@
...
+    # Test adapt, if data was passed.
+    if adapt_data is not None:
+        layer.adapt(adapt_data)
+
+    x = tf.keras.Input(shape=input_shape[1:], dtype=input_dtype)
+    y = layer(x)
+    model = tf.keras.Model(x, y)
+    actual_output = model.predict(input_data)
+    model_config = model.get_config()
+    recovered_model = tf.keras.models.model_from_config(model_config, custom_objects)
```
gabrieldemarmiesse, line 99:
```suggestion
    actual_output = layer(input_data)
    layer_config= tf.keras.layers.serialize(layer)
    recovered_layer = tf.keras.layers.deserialize(layer)
```

ashutosh1919: done changes

ashutosh1919:
cc @bhack

seanpmorgan:

Issue #1318, opened 2020-03-16
tfa.image.rotate Different Fill Modes


srilman:
**Describe the feature and the current behavior/state.**
Currently, when an image is rotated using the `tfa.image.rotate()` function, the outer regions of the image are filled with 0's, resulting in an image with black corners usually. This is usually not preferred, especially when the user is trying to create new images via data augmentation. Images with black corners are usually not a good representation of the intended creation.

`tf.keras.preprocessing.image.random_rotation` has a really good option where you can control how exactly the blank space is filled. The following pictures shows some of the common methods and how the resulting image looks closer to the original.

![](https://cdn-images-1.medium.com/max/720/1*rG4YJyVdu28LZwkwkwC55w.jpeg)

It would be nice to see the same option added to tfa.image.rotate.

**Relevant information**
- Are you willing to contribute it (yes/no): Yes!
- Are you willing to maintain it going forward? (yes/no): Yes!
- Is there a relevant academic paper? (if so, where): No
- Is there already an implementation in another framework? (if so, where): Keras (obviously)
- Was it part of tf.contrib? (if so, where): Not that I am aware of

**Which API type would this fall under (layer, metric, optimizer, etc.)**
`tfa.image`

**Who will benefit with this feature?**
Anyone who is currently using `tfa.image.rotate` for data / image augmentation purposes and want different methods to fill the image boundaries to create better augmented images for training.

**Any other info.**
While `tf.keras.preprocessing.image.random_rotation` (assuming you are performing rotations randomly) is capable of doing the same thing, it is generally not convenient to use it in a data pipeline. For one reason, it's not really designed for tensors in the form of `(num_pictures, num_rows, num_cols, num_channels)`, the channels last format which is whats normally expected. Secondly, it expects inputs as Numpy arrays. Finally it is hard to use directly with @tf.function. Thus, I think a native implementation would be preferred.

gabrieldemarmiesse:
It's widely used, already present in keras with numpy, I just hope it won't be too hard to implement. Contributions welcome!

vinayvr11:
Can shall we use the same method used in keras.

gabrieldemarmiesse:
I don't have a preference on the method as long as it's fast and the gradients and the results are correct.

vinayvr11:
Ok so we need to make it from scratch.

srilman:
I currently have a solution built in Python using high level Tensorflow ops for the following extensions:

* Extending the closest pixel outwards (the second picture)
[...]

PR #1719, opened 2020-04-23
Extend Image Function

srilman:
Related to #1318.

Pads an image by extending the borders of the image by some predefined size. I separated it out to a new function because it seemed useful for more situations than just rotation (for example, I used it before calling `tf.image.random_crop` to get a randomly shifted image). I currently named it `tfa.image.extend` but it might make more sense to use something like `tfa.image.pad____` to fit in line with the other pad functions (`tf.image.resize_and_pad`, `tf.image.pad_to_bounding_box`, etc.).

### DIFF
--- a/tensorflow_addons/image/transform_ops.py
+++ b/tensorflow_addons/image/transform_ops.py
@@ -434,6 +434,3 @@ def shear_y(image: TensorLike, level: float, replace: int) -> TensorLike:
     #  level  1].
     image = transform(wrap(image), [1.0, 0.0, 0.0, level, 1.0, 0.0, 0.0, 0.0])
     return unwrap(image, replace)
-
-
-
--- a/tensorflow_addons/image/transform_ops.py
+++ b/tensorflow_addons/image/transform_ops.py
@@ -20,7 +20,7 @@ from tensorflow_addons.utils.resource_loader import LazySO
 from tensorflow_addons.utils.types import TensorLike
 from tensorflow_addons.image.utils import wrap, unwrap
 
-from typing import Optional
+from typing import Optional, Tuple, Union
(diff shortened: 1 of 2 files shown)


tensorflow_addons/image/transform_ops.py
```diff
@@ -309,6 +309,53 @@ def _image_projective_transform_grad(op, grad):
...
+    Args:
+      images: A tensor of shape
+        (num_images, num_rows, num_columns, num_channels) (NHWC),
+        (num_rows, num_columns, num_channels) (HWC), 
+        or (num_rows, num_columns) (HW).
+      length: One of the following
+        A scalar length to extend the image by on each side. 
+        A tuple of pairs (LH, LW) to extend the image by in each dimension.
+        A tuple of tuple pairs ((LT, LB), (LL, LR)) to extend the image by for
+        each individual side
```
WindQAQ, line 327: This is not implemented yet, right?

srilman: Oops! I intended it to add it before but it slipped my mind.

tanzhenyu:
> /cc @tanzhenyu What do you think about this?

Can't this be done through tf.image.pad_xxx?

bhack:
It seems to me could be covered by https://www.tensorflow.org/api_docs/python/tf/image/pad_to_bounding_box.

tanzhenyu:
> It seems to me could be covered by https://www.tensorflow.org/api_docs/python/tf/image/pad_to_bounding_box.

Yeah

bhack:
I am closing this. Please reopen if you think there there are some others use cases that could be covered by this new API.

PR #1721, opened 2020-04-23
Support extended mode on projective transform

### LINKED ISSUE #1318: tfa.image.rotate Different Fill Modes

WindQAQ:
Closed #1318. Support extend mode on projective transform. Mainly follow the [scipy's implementation](https://github.com/scipy/scipy/blob/master/scipy/ndimage/src/ni_interpolation.c).

From top to bottom: original, reflect, constant, mirror, nearest, wrap. Addons is on the left, scipy is on the right.

test

Testing script:
```python3
import numpy as np

import tensorflow as tf

from scipy import ndimage, misc

image = misc.face()

tf.io.write_file("test.jpeg", tf.io.encode_jpeg(image))

for mode in ["CONSTANT", "REFLECT", "WRAP", "MIRROR", "NEAREST"]:
    t = tfa.image.rotate(image, np.pi / 4, extend=mode)
    tf.io.write_file(mode.lower() + ".jpeg", tf.io.encode_jpeg(t))
    st = ndimage.rotate(image, 45, mode=mode.lower(), reshape=False)
    tf.io.write_file("scipy-" + mode.lower() + ".jpeg", tf.io.encode_jpeg(st))
```

### DIFF
--- a/tensorflow_addons/custom_ops/image/cc/kernels/image_projective_transform_op.cc
+++ b/tensorflow_addons/custom_ops/image/cc/kernels/image_projective_transform_op.cc
@@ -43,32 +43,70 @@ template struct FillProjectiveTransform<CPUDevice, double>;
 
 }  // end namespace functor
 
-typedef Eigen::ThreadPoolDevice CPUDevice;
+using CPUDevice = Eigen::ThreadPoolDevice;
 
 using functor::FillProjectiveTransform;
+using generator::Extend;
 using generator::Interpolation;
-using generator::INTERPOLATION_BILINEAR;
-using generator::INTERPOLATION_NEAREST;
 using generator::ProjectiveGenerator;
+
+bool InterpolationFromString(const string& interpolation_str,
+                             Interpolation* interpolation) {
+  if (interpolation_str == "NEAREST") {
+    *interpolation = Interpolation::INTERPOLATION_NEAREST;
+  } else if (interpolation_str == "BILINEAR") {
+    *interpolation = Interpolation::INTERPOLATION_BILINEAR;
+  } else {
+    return false;
+  }
+  return true;
+}
+
+bool ExtendFromString(const string& extend_str, Extend* extend) {
+  if (extend_str == "REFLECT") {
+    *extend = Extend::EXTEND_REFLECT;
+  } else if (extend_str == "CONSTANT") {
+    *extend = Extend::EXTEND_CONSTANT;
+  } else if (extend_str == "NEAREST") {
+    *extend = Extend::EXTEND_NEAREST;
+  } else if (extend_str == "MIRROR") {
+    *extend = Extend::EXTEND_MIRROR;
+  } else if (extend_str == "WRAP") {
+    *extend = Extend::EXTEND_WRAP;
+  } else {
+    return false;
+  }
+  return true;
+}
 
 template <typename Device, typename T>
 class ImageProjectiveTransformV2 : public OpKernel {
  private:
   Interpolation interpolation_;
+  Extend extend_;
+  T constant_values_;
 
  public:
   explicit ImageProjectiveTransformV2(OpKernelConstruction* ctx)
       : OpKernel(ctx) {
     string interpolation_str;
     OP_REQUIRES_OK(ctx, ctx->GetAttr("interpolation", &interpolation_str));
-    if (interpolation_str == "NEAREST") {
-      interpolation_ = INTERPOLATION_NEAREST;
-    } else if (interpolation_str == "BILINEAR") {
-      interpolation_ = INTERPOLATION_BILINEAR;
-    } else {
-      LOG(FATAL) << "Invalid interpolation " << interpolation_str
-                 << ". Supported types: NEAREST, BILINEAR";
-    }
+    OP_REQUIRES(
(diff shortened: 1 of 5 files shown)


AakashKumarNain:
@WindQAQ overall looking good but I have one query. Please check the comments

tensorflow_addons/custom_ops/image/cc/kernels/image_projective_transform_op.cc
```diff
@@ -43,32 +43,70 @@ template struct FillProjectiveTransform<CPUDevice, double>;
 
 }  // end namespace functor
 
-typedef Eigen::ThreadPoolDevice CPUDevice;
+using CPUDevice = Eigen::ThreadPoolDevice;
```
AakashKumarNain, line 46: Are we forcing to run all these ops on CPU only?

WindQAQ: Nope, GPUDevice is also registered [here](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/custom_ops/image/cc/kernels/image_projective_transform_op.cc#L148-L160).

WindQAQ:
Just figure out that there is another PR #1719. @srilman what do you think about this :-) Thank you.

tensorflow_addons/custom_ops/image/cc/kernels/image_projective_transform_op.cc
AakashKumarNain: Cool. Thanks for the clarification

srilman:
@WindQAQ Looks great to me! Definitely preferable to #1719. Only thing is that it would be nice if the different extended modes could be used separately (outside of projective transforms) for more flexibility.

For example, I mentioned that I liked to extend the image before performing some image ops like `tf.image.random_crop`. In general, extend can be used in lieu of `tf.image.resize_with_pad` in nearly every situation.

WindQAQ:
> @WindQAQ Looks great to me! Definitely preferable to #1719. Only thing is that it would be nice if the different extended modes could be used separately (outside of projective transforms) for more flexibility.
> 
> For example, I mentioned that I liked to extend the image before performing some image ops like `tf.image.random_crop`. In general, extend can be used in lieu of `tf.image.resize_with_pad` in nearly every situation.

We can have both of them actually if they are both useful :-)

bhack:
I have a question but I don't know if it is strictly related. Do you know why we have our own  custom `ImageProjectiveTransformV2`? What is https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/image_ops.cc#L52?

WindQAQ:
@bhack It seems that it is migrated to core TF in Jan. I have no idea about this. Thanks for pointing out!

https://github.com/tensorflow/tensorflow/commit/26992815bf7ee136c49f658596a9fbde7881d425

bhack:
@WindQAQ I know that I am blamed as the cross-linking guy :stuck_out_tongue_winking_eye:  but I suppose it was part of a more general effort we already commented at https://github.com/tensorflow/addons/issues/1226#issuecomment-608594527 and some follow-up messages in that thread mentioning Keras.

WindQAQ:
> @WindQAQ I know that I am blamed as the cross-linking guy 

Never mind :smile: I don't even care about it LOL. Your links are super helpful for me.

> but I suppose it was part of a more general effort that we already commented at [#1226 (comment)](https://github.com/tensorflow/addons/issues/1226#issuecomment-608594527) and in some follow-up messages in that thread mentioning Keras.

Umm, it seems to be the miscommunication between us and many places. A related issue has been posted https://github.com/tensorflow/addons/issues/1779. Thanks again for pointing out it :-)

bhack:
@WindQAQ Users are looking for some of these filling parameters at https://github.com/tensorflow/tensorflow/issues/39177

PR #1723, opened 2020-04-24
Refactored code to include common parent classes for adaptive 1D, 2D and 3D pooling

Susmit-A:
This is to add future scope for implementing multiple Adaptive Pooling Layers without redundant code.


Susmit-A:
Okay, I'll work on this. As for the tests, I'll have to perform them on a specific type of pooling. I'll add separate tests for Average and Max pooling, along with the layers themselves with future commits. Or should I do that in a separate PR?

WindQAQ:
> Okay, I'll work on this. As for the tests, I'll have to perform them on a specific type of pooling. I'll add separate tests for Average and Max pooling, along with the layers themselves with future commits. Or should I do that in a separate PR?

Including in this PR would be better. Normally, we add a feature along with its test :-)

tools/testing/source_code_test.py
```diff
@@ -91,6 +91,7 @@ def test_no_private_tf_api():
         "tensorflow_addons/optimizers/moving_average.py",
         "tensorflow_addons/metrics/r_square.py",
         "tensorflow_addons/utils/test_utils.py",
+        "tensorflow_addons/layers/adaptive_pooling.py",
```
gabrieldemarmiesse, line 94: Please find an alternative. You can copy the code in tensorflow addons if you wish. Those blacklist are supposed to get smaller with the time. Rules are meant to be respected :)

WindQAQ: @Susmit-A you can directly copy codes into `tensorflow_addons/utils/keras_utils.py` directly :-)

Susmit-A: I've made the requested changes.

tensorflow_addons/utils/keras_utils.py
```diff
@@ -14,6 +14,20 @@
 # ==============================================================================
 """Utilities for tf.keras."""
 
+from tensorflow.keras import backend
```
WindQAQ, line 17: Access through `tf.keras.backend.image_data_format()` so we don't need to access private API.

tensorflow_addons/layers/adaptive_pooling.py
```diff
@@ -15,12 +15,185 @@
...
 
 import tensorflow as tf
+import tensorflow_addons.utils.keras_utils as conv_utils
+
 from typeguard import typechecked
-from typing import Union, List, Tuple
+from typing import Union, Callable, Iterable
+
+
+class AdaptivePooling1D(tf.keras.layers.Layer):
```
WindQAQ, line 24: My thought is to put `get_config` under `AdaptivePooling{1,2,3}D` so that we do not need to override it every time. What do you think? @Susmit-A

tensorflow_addons/utils/keras_utils.py
Susmit-A: Thst's called in the `normalize_data_format()` method. Should I still add that to the classes?

tensorflow_addons/layers/adaptive_pooling.py
Susmit-A: Sounds good. I'll move it.

Issue #1294, opened 2020-03-12
Add a public testing module


gabrieldemarmiesse:
**Describe the feature and the current behavior/state.**

Currently, tensorflow offers some low level functions and classes for testing: https://www.tensorflow.org/api_docs/python/tf/test 

But overall, we show in the test suite that we lack tests that we could apply to any keras objects as seen in #992

My proposal is to take #992 and that we expose publicly some of the functions created. I can be a codeowner. Some good candidates would be convergence tests for optimizers, a copy of `layer_test` for layers and serialization tests for all keras objects.

We could then use those functions in our test suite and other projects could use them too.

**Relevant information**
- Are you willing to contribute it (yes/no): yes
- Are you willing to maintain it going forward? (yes/no): yes
- Is there a relevant academic paper? (if so, where): no
- Is there already an implementation in another framework? (if so, where): keras-team/keras tf.keras
- Was it part of tf.contrib? (if so, where): no

**Which API type would this fall under (layer, metric, optimizer, etc.)**

`tfa.testing`

**Who will benefit with this feature?**

Everyone who make new keras objects and want to test them.

**Any other info.**

#992

seanpmorgan:
TensorFlow Addons is transitioning to a minimal maintenance and release mode. New features will not be added to this repository. For more information, please see our public messaging on this decision:
[TensorFlow Addons Wind Down](https://github.com/tensorflow/addons/issues/2807)

Please consider sending feature requests / contributions to other repositories in the TF community with a similar charters to TFA:
[Keras](https://github.com/keras-team/keras)
[Keras-CV](https://github.com/keras-team/keras-cv)
[Keras-NLP](https://github.com/keras-team/keras-nlp)

PR #1728, opened 2020-04-25
Added serializability tests in all optimizers and fixed serializability bug in yogi

ashutosh1919:
Related to #1294 .

@gabrieldemarmiesse - Please review this one.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,7 +41,7 @@ jobs:
       matrix:
         os: ['macos-latest', 'windows-latest', 'ubuntu-18.04']
         py-version: ['3.5', '3.6', '3.7', '3.8']
-        tf-version: ['2.1.0', '2.2.0rc3']
+        tf-version: ['2.2.0rc3']
       fail-fast: false
     steps:
       - uses: actions/github-script@0.3.0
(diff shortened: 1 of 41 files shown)


ashutosh1919:
@seanpmorgan , I think the method you suggested is too constrained.  
I am checking this `serialization.check_config(new_optimizer.get_config(), optimizer, False)` and it is giving error as `AttributeError: 'ConditionalGradient' object has no attribute '__name__'`. What do it mean by that?

ashutosh1919:
@seanpmorgan ,Please take a look at latest commit. With the use of `check_config()` in `serialization.py`, I have modified only one file and its tests - `conditional_gradient.py` and `conditional_graident_test.py` files. I have also slightly modified the function `check_config()`. Please review these three files and guide me whether I am doing right or not?

seanpmorgan:
Hi @ashutosh1919 so the bazel tests are failing because you have to explicitly link python imports that are not in the same directory. See https://github.com/tensorflow/addons/pull/1740 as a fix or we can merge master into this branch after it merges.

seanpmorgan:
@googlebot I consent.

seanpmorgan:
> @seanpmorgan , I think the method you suggested is too constrained.
> I am checking this `serialization.check_config(new_optimizer.get_config(), optimizer, False)` and it is giving error as `AttributeError: 'ConditionalGradient' object has no attribute '__name__'`. What do it mean by that?

Ah, this is a bit too restrictive. Could you revert your commits using the shared serialization check please. I'll create an issue to standardize all optimizer serialization checks, but want to get the bugfix in, using your first serialization check.

ashutosh1919:
@seanpmorgan - I have reverted the changes. And please create a separate issue to add serailizability check for optimizers in `serialization.py`. I definitely want to work on it.

PR #1730, opened 2020-04-25
Added angular distance option to triplet loss

henry-eigen:
Added an option to base pairwise distances on cosine similarity in addition to euclidean distance as suggested by several papers.

Formula for distance taken from  [Metric Learning for Adversarial Robustness](https://arxiv.org/pdf/1909.00900.pdf)

### DIFF
--- a/tensorflow_addons/losses/metric_learning.py
+++ b/tensorflow_addons/losses/metric_learning.py
@@ -65,6 +65,7 @@ def pairwise_distance(feature, squared=False):
     pairwise_distances = tf.math.multiply(pairwise_distances, mask_offdiagonals)
     return pairwise_distances
 
+
 @tf.function
 def angular_distance(feature):
     """Computes the angular distance matrix
@@ -78,12 +79,12 @@ def angular_distance(feature):
       angular_distances: 2-D Tensor of size [number of data, number of data].
     """
     # normalize input
-    feature = tf.math.l2_normalize(feature, axis = 1)
+    feature = tf.math.l2_normalize(feature, axis=1)
 
     # create adjaceny matrix of cosine similarity
     angular_distances = 1 - tf.matmul(feature, tf.transpose(feature))
 
     # ensure all distances > 1e-16
     angular_distances = tf.maximum(angular_distances, 1e-16)
-    
+
     return angular_distances
--- a/tensorflow_addons/losses/metric_learning.py
+++ b/tensorflow_addons/losses/metric_learning.py
@@ -68,7 +68,7 @@ def pairwise_distance(feature, squared=False):
 
 @tf.function
 def angular_distance(feature):
-    """Computes the angular distance matrix
+    """Computes the angular distance matrix.
 
     output[i, j] = 1 - cosine_similarity(feature[i, :], feature[j, :])
 
@@ -82,9 +82,9 @@ def angular_distance(feature):
     feature = tf.math.l2_normalize(feature, axis=1)
 
     # create adjaceny matrix of cosine similarity
-    angular_distances = 1 - tf.matmul(feature, tf.transpose(feature))
+    angular_distances = 1 - tf.matmul(feature, feature, transpose_b=True)
 
     # ensure all distances > 1e-16
-    angular_distances = tf.maximum(angular_distances, 1e-16)
+    angular_distances = tf.maximum(angular_distances, 0.0)
 
     return angular_distances
--- a/tensorflow_addons/losses/metric_learning.py
+++ b/tensorflow_addons/losses/metric_learning.py
@@ -70,7 +70,7 @@ def pairwise_distance(feature, squared=False):
 def angular_distance(feature):
     """Computes the angular distance matrix
 
-    output[i, j] = cosine_similarity(feature[i, :], feature[j, :])
+    output[i, j] = 1 - cosine_similarity(feature[i, :], feature[j, :])
 
     Args:
       feature: 2-D Tensor of size [number of data, feature dimension].
--- a/tensorflow_addons/losses/metric_learning.py
+++ b/tensorflow_addons/losses/metric_learning.py
@@ -64,3 +64,26 @@ def pairwise_distance(feature, squared=False):
     )
     pairwise_distances = tf.math.multiply(pairwise_distances, mask_offdiagonals)
     return pairwise_distances
+
+@tf.function
+def angular_distance(feature):
+    """Computes the angular distance matrix
+
+    output[i, j] = cosine_similarity(feature[i, :], feature[j, :])
+
+    Args:
+      feature: 2-D Tensor of size [number of data, feature dimension].
+
+    Returns:
+      angular_distances: 2-D Tensor of size [number of data, number of data].
+    """
+    # normalize input
+    feature = tf.math.l2_normalize(feature, axis = 1)
+
+    # create adjaceny matrix of cosine similarity
+    angular_distances = 1 - tf.matmul(feature, tf.transpose(feature))
+
+    # ensure all distances > 1e-16
+    angular_distances = tf.maximum(angular_distances, 1e-16)
+    
+    return angular_distances
--- a/tensorflow_addons/losses/metric_learning.py
+++ b/tensorflow_addons/losses/metric_learning.py
@@ -15,10 +15,11 @@
 """Functions of metric learning."""
 
 import tensorflow as tf
+from tensorflow_addons.utils.types import TensorLike
 
 
 @tf.function
-def pairwise_distance(feature, squared=False):
+def pairwise_distance(feature: TensorLike, squared: bool = False):
     """Computes the pairwise distance matrix with numerical stability.
 
     output[i, j] = || feature[i, :] - feature[j, :] ||_2
@@ -67,7 +68,7 @@ def pairwise_distance(feature, squared=False):
 
 
 @tf.function
-def angular_distance(feature):
+def angular_distance(feature: TensorLike):
     """Computes the angular distance matrix.
 
     output[i, j] = 1 - cosine_similarity(feature[i, :], feature[j, :])
--- a/tensorflow_addons/losses/triplet.py
+++ b/tensorflow_addons/losses/triplet.py
@@ -69,7 +69,10 @@ def _masked_minimum(data, mask, dim=1):
 @tf.keras.utils.register_keras_serializable(package="Addons")
 @tf.function
 def triplet_semihard_loss(
-    y_true: TensorLike, y_pred: TensorLike, margin: FloatTensorLike = 1.0, angular: bool = False
+    y_true: TensorLike,
+    y_pred: TensorLike,
+    margin: FloatTensorLike = 1.0,
+    angular: bool = False,
 ) -> tf.Tensor:
     """Computes the triplet loss with semi-hard negative mining.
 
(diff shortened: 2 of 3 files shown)


henry-eigen:
@googlebot I signed it!

tensorflow_addons/losses/metric_learning.py
```diff
@@ -64,3 +64,27 @@ def pairwise_distance(feature, squared=False):
     )
     pairwise_distances = tf.math.multiply(pairwise_distances, mask_offdiagonals)
     return pairwise_distances
+
+
+@tf.function
+def angular_distance(feature):
```
WindQAQ, line 70: Could we annotate the type of `feature`?

tensorflow_addons/losses/metric_learning.py
```diff
@@ -64,3 +64,27 @@ def pairwise_distance(feature, squared=False):
...
+    """Computes the angular distance matrix
```
WindQAQ, line 71: Missing period at the end of sentence.

tensorflow_addons/losses/metric_learning.py
```diff
@@ -64,3 +64,27 @@ def pairwise_distance(feature, squared=False):
...
+      feature: 2-D Tensor of size [number of data, feature dimension].
+
+    Returns:
+      angular_distances: 2-D Tensor of size [number of data, number of data].
+    """
+    # normalize input
+    feature = tf.math.l2_normalize(feature, axis=1)
+
+    # create adjaceny matrix of cosine similarity
+    angular_distances = 1 - tf.matmul(feature, tf.transpose(feature))
```
WindQAQ, line 85: `tf.matmul` has the option to transpose matrix (`transpose_b`) so you do not need to manually transpose it.

https://www.tensorflow.org/api_docs/python/tf/linalg/matmul#expandable-1-label

tensorflow_addons/losses/metric_learning.py
```diff
@@ -64,3 +64,27 @@ def pairwise_distance(feature, squared=False):
...
+
+    # ensure all distances > 1e-16
+    angular_distances = tf.maximum(angular_distances, 1e-16)
```
WindQAQ, line 88: What is this magic number 1e-16? Shouldn't it be zero as distance is always positive or zero?

tensorflow_addons/losses/triplet.py
```diff
@@ -168,6 +178,7 @@ def triplet_hard_loss(
     y_true: TensorLike,
     y_pred: TensorLike,
     margin: FloatTensorLike = 1.0,
+    angular: bool = False,
```
WindQAQ, line 181: please move this to the last arg.

tensorflow_addons/losses/metric_learning.py
henry-eigen: I was trying to adhere to the style of the above function, but I could certainly annotate the argument in the "feature: TensorLike" style if that is what you mean

henry-eigen:
@WindQAQ If you are asking me to add automated tests, I am afraid I am not sure how to do that. I have manually tested the results produced by the angular_distances function against a few matrices I generated with sklearn.metrics.pairwise.cosine_similarity

tensorflow_addons/losses/metric_learning.py
WindQAQ: Please do :-) The type hint is not yet finish in our repo.

WindQAQ:
> @WindQAQ If you are asking me to add automated tests, I am afraid I am not sure how to do that. I have manually tested the results produced by the angular_distances function against a few matrices I generated with sklearn.metrics.pairwise.cosine_similarity. Once that matrix is computed, performance of the triplet_loss function should be unchanged aside from the different distance values.

So you can put the testing function, say `test_semihard_tripled_loss_angular`, `test_hard_tripled_loss_angular` or somthing like that and  into [this file](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/losses/tests/triplet_test.py). You might also want to verify the results from numpy/scipy implementation so feel free to modify the functions in the file.

henry-eigen:
@WindQAQ I added the tests you suggested as well as one for the angular distance matrix. I also added the angular argument to the loss wrappers which I initially forgot to do.

tensorflow_addons/losses/triplet.py
```diff
@@ -97,7 +100,14 @@ def triplet_semihard_loss(
     labels = tf.reshape(labels, [lshape[0], 1])
 
     # Build pairwise squared distance matrix.
-    pdist_matrix = metric_learning.pairwise_distance(precise_embeddings, squared=True)
+
+    if angular:
```
shun-lin, line 104: Maybe other people can join in on this, should we add support for custom distance method? (probably easy to implement, 1 more arg and another `if` statement here)

tensorflow_addons/losses/triplet.py
```diff
@@ -69,7 +69,10 @@ def _masked_minimum(data, mask, dim=1):
 @tf.keras.utils.register_keras_serializable(package="Addons")
 @tf.function
 def triplet_semihard_loss(
-    y_true: TensorLike, y_pred: TensorLike, margin: FloatTensorLike = 1.0
+    y_true: TensorLike,
+    y_pred: TensorLike,
+    margin: FloatTensorLike = 1.0,
+    angular: bool = False,
```
shun-lin, line 75: just wondering, why isn't "squared" a parameter originally? (for l1 vs l2 distance)?

henry-eigen: I can only speak to what I have added, but I agree it is odd that there is not an option especially give that the pairwise_dist function accepts squared as an option. I can certainly add this parameter.

tensorflow_addons/losses/triplet.py
henry-eigen: I agree that would be helpful, but how do you suggest the custom distance method is passed to the function. Do you mean that the function itself would be passed as an argument?

tensorflow_addons/losses/triplet.py
henry-eigen: Or maybe not. Adding the squared parameter is causing a single build release wheel check to fail so maybe I should just revert that commit.

henry-eigen:
@shun-lin  I have relatively little experience with contributing to large projects such as this one, and I am hoping you can help me with something. This PR passed all of the checks initially, but after making a minor revision, it is now failing on a single test (Build release for ubuntu 18.04) even though it passes all of the other builds. How would I go about figuring out the source of the issue?

shun-lin:
> @shun-lin I have relatively little experience with contributing to large projects such as this one, and I am hoping you can point me in the right direction. An earlier version of PR passed all of the checks, but after making a minor revision, it is now failing on a single test (Build release for ubuntu 18.04) even though it passes all of the other builds. How would I go about figuring out the source of the issue?

@henry-eigen thanks for letting us know the issue, sometimes a test failed (especially when only 1 failed for a specific env) may not due to your PR. (maybe other maintainer can explain better than I do).

tensorflow_addons/losses/triplet.py
shun-lin: We can do this similar to TensorFlow API that this parameter can take in either a string (eud/angular like you mentioned) and a function. For example when you build the model you can either have loss = some string or some class, so I agree with you here.

henry-eigen:
@shun-lin I fixed the issue of the missing squared argument in the triplet loss. What do you think of my response to your other comment about the custom distance metrics?

shun-lin:
I have added comments on that thread :)

henry-eigen:
@shun-lin Whoops, didn't see that. I removed the angular argument and added a "dist_metric" parameter which behaves as indicated above. Tests on my side show that it performs the same now if, for example, you were to set dist_metric="L_2" or dist_metric=metric_learning.pairwise_distance

tensorflow_addons/losses/triplet.py
```diff
@@ -79,6 +83,11 @@ def triplet_semihard_loss(
       y_pred: 2-D float `Tensor` of embedding vectors. Embeddings should
         be l2 normalized.
       margin: Float, margin term in the loss definition.
+      squared: bool, determines whether L-1 or L-2 norm is used
+      dist_metric: str or function, determines distance metric:
+                   "L_2" for euclidean distance
```
shun-lin, line 88: hmmm having bot h squared and "L_2" is confusing, i.e. if people want "L_1" they need to set "L_2" and squared = False, I think adding an "L_1" option to dist_metric would be better than having `squared` argument if we are using dist_metric this way :). What do you think ?

NIT: May you also (quickly) check with TensorFlow repo how they name dist_metric? Do they use `distance_metric` or short-handed `dist_metric` for similar argument? This is for consistency :)

henry-eigen: I added the option for L-1 distance and removed the squared argument. I changed dist_metric to distance_metric, because even though I wasn't able to find that exact name anywhere else in the repo, the long version seemed more in line with the rest of tensorflow's code. I have also completely overhauled the tests to work with the restructured function.

henry-eigen:
@shun-lin I added the option for L-1 distance and removed the squared argument. I changed dist_metric to distance_metric, because even though I wasn't able to find that exact name anywhere else in the repo, the long version seemed more in line with the rest of tensorflow's code. I have also completely overhauled the tests to work with the restructured function.

tensorflow_addons/losses/triplet.py
```diff
@@ -79,6 +82,11 @@ def triplet_semihard_loss(
       y_pred: 2-D float `Tensor` of embedding vectors. Embeddings should
         be l2 normalized.
       margin: Float, margin term in the loss definition.
+      distance_metric: str or function, determines distance metric:
+                       "L_1" for l1-norm distance
```
shun-lin, line 86: NIT: would "L1" instead of "L_1" be more consistent with rest of the repo / TF?

tensorflow_addons/losses/triplet.py
```diff
@@ -79,6 +82,11 @@ def triplet_semihard_loss(
...
+                       "L_2" for l2-norm distance
+                       "angular" for cosine similarity
+                        function for custom metric
```
shun-lin, line 89: You probably want to give an example of custom function,

You can see the example here:

https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/LearningRateScheduler

tensorflow_addons/losses/triplet.py
```diff
@@ -96,8 +104,24 @@ def triplet_semihard_loss(
     lshape = tf.shape(labels)
     labels = tf.reshape(labels, [lshape[0], 1])
 
-    # Build pairwise squared distance matrix.
-    pdist_matrix = metric_learning.pairwise_distance(precise_embeddings, squared=True)
+    # Build pairwise squared distance matrix
+
+    if distance_metric == "L_1":
```
shun-lin, line 109: Same as above, "L1" may be better.

shun-lin:
@henry-eigen thanks so much, will wait for @WindQAQ to take another look :)

henry-eigen:
@shun-lin @WindQAQ I made the suggested changes to notation and documentation. Unless there are any other major issue, I think this should be good to go.

henry-eigen:
@WindQAQ Any thoughts?

tensorflow_addons/losses/metric_learning.py
```diff
@@ -64,3 +65,27 @@ def pairwise_distance(feature, squared=False):
...
+      angular_distances: 2-D Tensor of size [number of data, number of data].
+    """
+    # normalize input
+    feature = tf.math.l2_normalize(feature, axis=1)
+
+    # create adjaceny matrix of cosine similarity
+    angular_distances = 1 - tf.matmul(feature, feature, transpose_b=True)
+
+    # ensure all distances > 1e-16
+    angular_distances = tf.maximum(angular_distances, 0.0)
```
lc0, line 89: isn't it supposed to be
`tf.maximum(angular_distances, 1e-16)`

in order to 
` # ensure all distances > 1e-16`

henry-eigen: I can take another look, but if I remember, it wasn’t necessary here since there is no sqrt and therefore no issue of the gradient of sqrt(0). I guess I forgot to update the comment

PR #1738, opened 2020-04-26
Drop support for TF 2.1

### LINKED ISSUE #1734: Dropping support for tensorflow < 2.2.
We have three pull requests that need TF 2.2 and above: #1733 #1726 #905 

They are blocked until we drop support, since Addons 0.9 was released with support for both tf 2.1 and 2.2, I don't think it's going to be a very big inconvenience for users if we drop support for 2.1 now. They have a good fallback.

gabrieldemarmiesse:
Fixes #1734


gabrieldemarmiesse:
I updated the 0.9.x branch as it's compatible with tf 2.2.

PR #1739, opened 2020-04-26
DOC: Correct example of new experimental argument

seanpmorgan:
The new argument was actually `experimental_aggregate_gradients`

Reference: https://github.com/tensorflow/models/commit/9fc4fd082968c0c959c1d7bc938a9fcbf0f9e50d


PhilJd:
Thanks for the fix, sorry about that. I might have looked at an outdated version there.

PR #1746, opened 2020-04-27
Migrated some files from TF 1.0 to 2.0

### LINKED ISSUE #1735: Remove all elements from allowlists in source_code_test.py
We enforce some rules in [source_code_test.py](https://github.com/tensorflow/addons/blob/master/tools/testing/source_code_test.py) about what API we can't use in addons. (no private api, no tf.compat.v1, and no experimental API in user-facing code).

We didn't enforce it in some files because they didn't respect the rules before the check was put into place. Replacing the problematic code takes some time, so this is why there are blacklists now. To track the progress. Replacing some APIs by public ones can be very easy or very hard depending on what the code is about. We'll gladly take pull request removing some elements from the blacklists. When we remove all elements, we can remove the blacklists entirely.

It's possible to copy the code needed into addons, or to find another similar API as long as the effect is the same.

ashutosh1919:
Fixes part of #1735 .
@gabrieldemarmiesse - Please review.

### DIFF
--- a/tensorflow_addons/seq2seq/tests/attention_wrapper_test.py
+++ b/tensorflow_addons/seq2seq/tests/attention_wrapper_test.py
@@ -396,9 +396,7 @@ def _test_with_attention(
     )
     if cell._attention_layers is not None:
         for layer in cell._attention_layers:
-            layer.kernel_initializer = tf.keras.initializers.GlorotUniform(
-                seed=1337
-            )
+            layer.kernel_initializer = tf.keras.initializers.GlorotUniform(seed=1337)
 
     sampler = sampler_py.TrainingSampler()
     my_decoder = basic_decoder.BasicDecoder(cell=cell, sampler=sampler)
--- a/tensorflow_addons/seq2seq/tests/attention_wrapper_test.py
+++ b/tensorflow_addons/seq2seq/tests/attention_wrapper_test.py
@@ -396,7 +396,7 @@ def _test_with_attention(
     )
     if cell._attention_layers is not None:
         for layer in cell._attention_layers:
-            layer.kernel_initializer = tf.compat.v1.keras.initializers.glorot_uniform(
+            layer.kernel_initializer = tf.keras.initializers.GlorotUniform(
                 seed=1337
             )
 
--- a/tensorflow_addons/text/skip_gram_ops.py
+++ b/tensorflow_addons/text/skip_gram_ops.py
@@ -15,7 +15,6 @@
 """Skip-gram sampling ops from https://arxiv.org/abs/1301.3781."""
 
 import csv
-import numpy as np
 import tensorflow as tf
 
 from tensorflow_addons.utils.resource_loader import LazySO
@@ -176,9 +175,7 @@ def skip_gram_sample(
             seed=seed,
         )
 
-        seed1 = np.random.get_state()[1][0]
-        seed2 = np.random.get_state()[1][1]
-        # seed1, seed2 = tf.compat.v1.get_seed(seed)
+        seed1, seed2 = tf.compat.v1.get_seed(seed)
         tokens, labels = _skip_gram_so.ops.addons_skip_gram_generate_candidates(
             input_tensor=input_tensor,
             min_skips=min_skips,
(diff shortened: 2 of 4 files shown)


tensorflow_addons/seq2seq/tests/attention_wrapper_test.py
```diff
@@ -396,9 +396,7 @@ def _test_with_attention(
     )
     if cell._attention_layers is not None:
         for layer in cell._attention_layers:
-            layer.kernel_initializer = tf.compat.v1.keras.initializers.glorot_uniform(
-                seed=1337
-            )
+            layer.kernel_initializer = tf.keras.initializers.GlorotUniform(seed=1337)
```
guillaumekln, line 399: Unfortunately it is not as simple as that. V1 and V2 initializers don't seem to produce the same random number sequence. So this change would require to update all numerical checks. Do you want to do that?

ashutosh1919: @guillaumekln , Is there any other solution to upgrade to v2 but other things should remain same? If there is a way then we need to pursue that and if not then anyhow, we will have to upgrade it in future. So, why not now?

gabrieldemarmiesse: I belive that going forward, we should avoid testing by hardcoding the numerical results. Those tests don't ensure that the feature works as expected, they ensure that the feature doesn't change. They also rely too much on the random seed, which leads to a situation like this one (or like [this other one](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/seq2seq/tests/attention_wrapper_test.py#L538))

Obviously for complicated features, it's also much easier to hardcode the numerical results, and that works okay, but as we get more and more users, we need to have stronger and stronger tests because bugs make more damage.

AakashKumarNain: I agree that we need to stop hardcoding the results for testing until unless it is the only way to test the feature

seanpmorgan: > I believe that going forward, we should avoid testing by hardcoding the numerical results. Those tests don't ensure that the feature works as expected, they ensure that the feature doesn't change. They also rely too much on the random seed, which leads to a situation like this one (or like [this other one](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/seq2seq/tests/attention_wrapper_test.py#L538))
> 
> Obviously for complicated features, it's also much easier to hardcode the numerical results, and that works okay, but as we get more and more users, we need to have stronger and stronger tests because bugs make more damage.

So we've had this discussion a few times (failing to find them at the moment.) I think there are a number of different times when we validate values and I'd like to discuss all of them to make a guide for choosing how to test them. There are pros and cons to all the methods

1. The correct values can be computed by hand. For example many metrics. Hand calculating values means we can validate not only the result doesn't change but that it is correct in it's calculation. The con is if it was hand-calculated incorrectly or someone just ran the function to get a value. It's a maintence burden when reviewing this.

2. The correct values can be computed from another well-tested library (e.g. numpy, scikit-learn). To me this is preferable to hand-calculated since we can be reasonably sure the values are correct and also that the tested function does not change. As an example f1 score.

3. The correct values can't be computed from another library. As an example the seq2seq tests that you referenced @gabrieldemarmiesse . They're troublesome when we need to refactor tests (i.e. moving from TF1 to TF2 or if the random seed gets changed). But they have value in making sure the function output does not change. I guess I'm missing how removing these makes a more robust test suite?

ashutosh1919: I am definitely agree with @seanpmorgan , I have seen the same thing in many losses tests too. We can build the numpy version of the same function in the test and we can compare the results of tf function and numpy function for tests. That will work. So, if you all agree, I will start working on that in case of `attention_wrapper_test.py`.

guillaumekln: Do you mean rewriting the attention layers in terms on Numpy functions? This sounds like a lot of code duplication.

@gabrieldemarmiesse suggests a pytest tool to help testing numerical outputs https://github.com/tensorflow/addons/issues/1793. @ashutosh1919 Maybe do you want to look into that?

seanpmorgan:

PR #1751, opened 2020-04-28
Accept passing maximum_iterations as a Tensor

### LINKED ISSUE #1749: Incorrect type of maximum_iterations argument in tfa.seq2seq.decoder causes runtime error.
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Linux Debian Stable
- TensorFlow version and how it was installed (source or binary): 2.1 binary
- TensorFlow-Addons version and how it was installed (source or binary): 0.8.1 binary
- Python version: 3.7
- Is GPU used? (yes/no): no

**Describe the bug**

Calling `tfa.seq2seq.decoder.BaseDecoder` with a `tf.Tensor` for `maximum_iterations`, which worked before, now results in

```
TypeError: type of argument "maximum_iterations" must be one of (float, int, float16, float32, int8, int16, int32, int64, uint8, uint16, uint32, uint64, NoneType); got tensorflow.python.framework.ops.Tensor instead
```

**Other info / logs**

The problem is caused by a faulty type annotation
```python
  maximum_iterations: Optional[Number] = None,
```

Note that this annotation is also used in `tfa.seq2seq.decoder.dynamic_decode`.

guillaumekln:
Also add `@typechecked` to `dynamic_decode` for consistency.

Fixes #1749.

### DIFF
--- a/tensorflow_addons/seq2seq/decoder.py
+++ b/tensorflow_addons/seq2seq/decoder.py
@@ -17,7 +17,7 @@
 import abc
 
 import tensorflow as tf
-from tensorflow_addons.utils.types import Number
+from tensorflow_addons.utils.types import TensorLike
 from typeguard import typechecked
 from typing import Any, Optional, Tuple, Union
 
@@ -144,7 +144,7 @@ class BaseDecoder(tf.keras.layers.Layer):
         self,
         output_time_major: bool = False,
         impute_finished: bool = False,
-        maximum_iterations: Optional[Number] = None,
+        maximum_iterations: Optional[TensorLike] = None,
         parallel_iterations: int = 32,
         swap_memory: bool = False,
         **kwargs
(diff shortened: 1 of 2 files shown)


tensorflow_addons/seq2seq/decoder.py
```diff
@@ -299,14 +300,8 @@ def dynamic_decode(
       `(final_outputs, final_state, final_sequence_lengths)`.
 
     Raises:
-      TypeError: if `decoder` is not an instance of `Decoder`.
       ValueError: if `maximum_iterations` is provided but is not a scalar.
     """
-    if not isinstance(decoder, (Decoder, BaseDecoder)):
```
autoih, line 305: Thanks @guillaumekln, LGTM. May I know why do you remove this?

guillaumekln: I added the `typechecked` annotation instead.

autoih: Yeah. Should we also modify `ValueError: ... not a scalar` as `scalar or tensor`?

PR #1759, opened 2020-04-30
Bump dependency to tf2.2rc4

seanpmorgan:
Also want to troubleshoot what is going on with broken master branch

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -92,7 +92,7 @@ jobs:
       - uses: actions/download-artifact@v1
         if: github.event_name != 'pull_request' || matrix.py-version == env.MIN_PY_VERSION  || matrix.py-version == env.MAX_PY_VERSION
         with:
-          name: ${{ matrix.os }}-${{ matrix.py-version }}-tf2.2.0rc3-wheel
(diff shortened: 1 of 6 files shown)


seanpmorgan:
Blocked by missing macos py38 whl for ttf2.2rc4

gabrieldemarmiesse:
@bhack we can support multiple versions simultaneously. Here the oldest and more recent versions are the same, so it's trivial with a Ctrl+R. Even in the cases where we have multiple tf version, I'm not sure it's worth the effort to write a bash script.

bhack:
It was just that we need to manually track 5 files every RC..

seanpmorgan:
> @bhack we can support multiple versions simultaneously. Here the oldest and more recent versions are the same, so it's trivial with a Ctrl+R. Even in the cases where we have multiple tf version, I'm not sure it's worth the effort to write a bash script.

> It was just that we need to manually track 6 files every RC..

I do want to remove the version number from the contribution docs if possible so we don't need to update them constantly. 

Ctrl+R is pretty straightforward, but I agree there are ways to streamline this. Open to PRs, but not high priority

PR #1760, opened 2020-04-30
Fix master: Keep pip version constant

seanpmorgan:
No need to upgrade these two packages in our build environment.. this can introduce possible errors we don't control

### DIFF
--- a/.github/workflows/make_wheel_macOS.sh
+++ b/.github/workflows/make_wheel_macOS.sh
@@ -5,7 +5,7 @@ export TF_NEED_CUDA=0
 python --version
 python -m pip install delocate wheel setuptools tensorflow==$TF_VERSION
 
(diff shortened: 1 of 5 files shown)


seanpmorgan:
So the macos wheels are failing for 403 on github api requesting bazel releases. I've ran into something similar for another project and I _think_ this is related to all macos VMs running the same IP address:
https://github.com/actions/virtual-environments/issues/602#issuecomment-602472951

If there's anyway we can authenticate the API request that would solve the issue but seeing as it's done inside of bazel build this may be tricky to do.

gabrieldemarmiesse:
We're merging this because it's urgent, but I believe the setup bazel step should be skipped on linux and windows, we already have bazel installed in docker and on windows it's downloaded in the bash script.

PR #1767, opened 2020-05-01
Use the mark for the gpu tests in conditional_gradient_test.py

gabrieldemarmiesse:
Made conditional_gradient_test.py use the pytest mark as it's more clean.

I also had to expose a way to grab the device used to run the test function. This can now be done by adding a parameter `device` to the test function. I adapted the CONTRIBUTING to reflect that.


gabrieldemarmiesse:
Nice, this will unblock some good stuff :)

PR #1770, opened 2020-05-02
Enable distributed strategies in tests.

### LINKED ISSUE #1682: Enable multiprocessing when testing with GPU and support distributed strategies in the tests.

gabrieldemarmiesse:
Needs #1767 to be merged first this will make the diff smaller.

Closes #1682 

tf.test.is_gpu_available() is deprecated and was causing issues so I replaced it.

Maybe we should replace the name `device` for the fixture? Maybe it's strange that there is a device being a strategy.

I think we'll rename `with_device` -> `with_scope` and `device` -> `scope`. But the renaming would make a big diff, so it's a TODO for a next pull request.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -428,6 +428,8 @@ the test twice, on CPU and on GPU, or only on GPU. Here is how to do it.
 
 ```python
 import pytest
+import tensorflow as tf
+from tensorflow_addons.utils import test_utils
 
 @pytest.mark.with_device(["cpu", "gpu"])
 def test_something():
@@ -443,6 +445,23 @@ def test_something2(device):
         print("do something.")
     if "gpu" in device:
         print("do something else.")
+
+
+@pytest.mark.with_device(["cpu", "gpu", tf.distribute.MirroredStrategy])
+def test_something3(device):
+    # the code here will run three times, once on gpu, once on cpu and once with 
+    # a mirror distributed strategy.
+    # device will be "cpu:0" or "gpu:0" or "gpu:1" or "gpu:2" or the strategy.
+    # with the MirroredStrategy, it's equivalent to:
+    # strategy = tf.distribute.MirroredStrategy(...)
+    # with strategy.scope():
+    #     test_function(strategy)
+    if "cpu" in device:
+        print("do something.")
+    if "gpu" in device:
+        print("do something else.")
+    if isinstance(device, tf.distribute.Strategy):
+        device.run(...)
 
 
 @pytest.mark.with_device(["gpu"])
@@ -461,6 +480,14 @@ def test_something_more():
 def test_something_more2():
     # When running the function, there will be no `with tf.device` wrapper.
     # You are free to do whatever you wish with the devices in there.
+    # Make sure to use only the cpu, or only gpus available to the current process with
+    # test_utils.gpu_for_testing() , otherwise, it might not play nice with 
+    # pytest's multiprocessing.
+    # for example
+    ...
+    strategy = tf.distribute.MirroredStrategy(test_utils.gpu_for_testing())
+    with strategy.scope():
+        print("I'm doing whatever I want.") 
     ...
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -452,7 +452,7 @@ def test_something2(device):
 def test_something3(device):
     # the code here will run three times, once on gpu, once on cpu and once with 
     # a mirror distributed strategy.
-    # device will be "cpu:0" or "gpu:0" or "gpu:1" or "gpu:2" or the strategy.
+    # device will be "cpu:0" or "gpu:0" or the strategy.
     # with the MirroredStrategy, it's equivalent to:
     # strategy = tf.distribute.MirroredStrategy(...)
     # with strategy.scope():
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -477,12 +477,15 @@ def test_something_more():
 
 
 @pytest.mark.with_device(["no_device"])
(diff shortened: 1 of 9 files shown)


Squadrick:
@gabrieldemarmiesse What's blocking this PR?

gabrieldemarmiesse:
#1767

guillaumekln:
On this subject, I've been using this trick to test distribution strategies locally:

https://github.com/OpenNMT/OpenNMT-tf/blob/v2.10.0/opennmt/tests/test_util.py#L41-L61

It's a decorator that resets the TensorFlow context and creates 2 logical CPU devices for the test being run.

CONTRIBUTING.md
```diff
@@ -445,6 +447,24 @@ def test_something2(device):
...
+
+@pytest.mark.with_device(["cpu", "gpu", tf.distribute.MirroredStrategy])
+def test_something3(device):
+    # the code here will run three times, once on gpu, once on cpu and once with 
+    # a mirror distributed strategy.
+    # device will be "cpu:0" or "gpu:0" or "gpu:1" or "gpu:2" or the strategy.
+    # with the MirroredStrategy, it's equivalent to:
+    # strategy = tf.distribute.MirroredStrategy(...)
+    # with strategy.scope():
+    #     test_function(strategy)
```
seanpmorgan, line 459:
```suggestion
    # the code here will run three times, once on gpu, once on cpu and once with 
    # a mirrored distribution strategy on two virtual gpu devices.
    # `device` will be either "cpu:0", "gpu:0", or  "gpu:0" / "gpu:1" respectively.
    # With the MirroredStrategy, it's equivalent to:
    # strategy = tf.distribute.MirroredStrategy(...)
    # with strategy.scope():
    #     test_function(strategy)
```
Slight re-wording. Per the `gpus_for_testing` function I'm not sure how we'd get a gpu:1 and gpu:2:
```
RMSprop/RMSprop/update_1/update_1/sub/x: (Const): /job:localhost/replica:0/task:0/device:GPU:1
RMSprop/RMSprop/update_0/Const: (Const): /job:localhost/replica:0/task:0/device:GPU:0
```

gabrieldemarmiesse: Yeah, actually device is going to be cpu:0 or gpu:0 or the strategy. Let me reword that. Device can be the strategy too.

PR #1771, opened 2020-05-02
Make gaussian filter meet scipy's implementation

WindQAQ:
- Support different sigma for x and y direction.
- Minor docs.
- Depthwise conv2d on floating type and retain the original dtype if it is not floating type.
- Simplify tests, including 2d, 3d and 4d image, three different padding modes and different dtype.
- The radius of gaussian kernel is computed as `ceil(sigma * truncate) * 2 + 1`. So for sigma=1.0, truncate=4.0, the kernel size should be 9 instead 3, which is the value we used before.
https://github.com/scipy/scipy/blob/v1.4.1/scipy/ndimage/filters.py#L216
- Use `scipy.ndimage.filters.gaussian_filter` instead of the one in `skimage` to test 4D image directly.

### DIFF
--- a/tensorflow_addons/image/filters.py
+++ b/tensorflow_addons/image/filters.py
@@ -206,10 +206,10 @@ def median_filter2d(
         return output
 
 
-def _get_gaussian_kernel(sigma, filter_shape_1d):
-    "This function creates a kernel of size [filter_shape]."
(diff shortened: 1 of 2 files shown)


AakashKumarNain:
LGTM! Thanks @WindQAQ

Issue #1353, opened 2020-03-20
[2/2] Add Image Ops for Data Augmentation


vinayvr11:
Part 1: https://github.com/tensorflow/addons/issues/1333

As part of #1226, relevant image ops need to be added to `tfa.image`. List of ops tracked in this issue (tick the op only once the corresponding PR is *merged*):
- [x] to grayscale
- [ ] color jitter
- [ ] color jitter nonrand
- [ ] color jitter rand
- [ ] compute crop shape
- [ ] center crop
- [ ] distorted bounding box crop
- [ ] crop and resize
- [x] gaussian blur
- [ ] random crop with resize
- [ ] random color jitter
- [ ] random blur
- [ ] batch random blur

Reference: https://github.com/google-research/simclr/blob/master/data_util.py

vinayvr11:
@gabrieldemarmiesse, @dynamicwebpaige does this issue is meaningfull for Tensorlfow.

Squadrick:
@vinayvr11 Closing #1342, create separate PRs for each op.

vinayvr11:
Sure @Squadrick i will make seperate PR for all operations.

ghosalsattam:
@gabrieldemarmiesse The list does not contain Canny Edge detection. But it is useful functionality an is contained in almost all Standard image processing libraries. Should it be added?

bhack:
> @gabrieldemarmiesse The list does not contain Canny Edge detection. But it is useful functionality and is contained in almost all Standard image processing libraries. Should it be added?

About image processing I think that we need to think about priorities:

- Is the op useful for image (auto)augmentation/preprocessing?
- Is the op useful for postprocessing the network output?
- if are not in the first two cases what is the use case of that specific op VS a network that learn that task (e.g. in the edge detector case see https://arxiv.org/abs/2003.08235 just to mention a recent work)? Are we planning to contribute a differentiable version of the ops to learn that task? 

In the first two case, as an example, see [how many Opencv calculators ](https://github.com/google/mediapipe/tree/master/mediapipe/calculators/image) (search for `cv::`)  mediapipe still need to implement in its demos other then tensorflow inference.

seanpmorgan:
Per https://github.com/tensorflow/addons/issues/1226#issuecomment-634755922 the autoaugment/randaugment features will be a part of https://github.com/keras-team/keras-cv

Apologies for not being able to present this information sooner to the contributors.

PR #1772, opened 2020-05-02
RGB to Grayscale

ghosalsattam:
See #1333 #1353 #1226 
This PR is responsible for RGB to Grayscale Conversion of image(s).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -64,6 +64,8 @@
 /tensorflow_addons/image/tests/interpolate_spline_test.py
 /tensorflow_addons/image/resampler_ops.py @autoih
 /tensorflow_addons/image/tests/resampler_ops_test.py @autoih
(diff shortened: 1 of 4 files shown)


ghosalsattam:
@bhack @gabrieldemarmiesse @seanpmorgan @WindQAQ Please provide a review.

gabrieldemarmiesse:
I agree with @WindQAQ , there is not enough added value here compared to what tensorflow offers, we can use directly tf.image.rgb_to_grayscale  . I think we need to edit https://github.com/tensorflow/addons/issues/1353#issue-585039702 . I can understand why someone might think that we need to copy the function here. I'll correct the message.

Thank you @ghosalsattam for the pull request, and sorry for not looking too much into the issue. I'll check the box for rgb to grayscale.

bhack:
I agree cause in our cause just for this single line call we need +137 lines of diff (overhead, tests etc.) also if the original wrapper was just 5 lines.

PR #1774, opened 2020-05-02
Add a small script to change version

bhack:
In the meantime that we centralize/refactor the versioning this script could be helpful.

`./update_release_version.sh 2.2.0rc5`

### DIFF
--- a/update_release_version.sh
+++ b/update_release_version.sh
@@ -1,0 +1,7 @@
+sed -ri "s/(TF_VERSION=|tensorflow(-cpu)*(~|=)=|tf-version: \[')[0-9]+[a-zA-Z0-9_.-]+/\1$1/g" \
+	.github/workflows/release.yml \
+	CONTRIBUTING.md \
+	tools/docker/cpu_tests.Dockerfile \
+	tools/install_deps/tensorflow-cpu.txt \
+	tools/install_deps/tensorflow.txt \
+	tools/run_gpu_tests.sh
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -85,6 +85,7 @@ jobs:
       matrix:
         os: ['macOS', 'Windows', 'Linux']
         py-version: ['3.5', '3.6', '3.7', '3.8']
+        tf-version: ['2.2.0rc4']
       fail-fast: false
     steps:
       - uses: actions/download-artifact@v1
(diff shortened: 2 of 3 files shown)


gabrieldemarmiesse:
I'm not sure I understand, what's the benefit of running 
```
./update_release_version.sh 2.2.0rc5
```
compared to opening Pycharm, VSCode, vim... and do 
```
Ctrl+Shift+R (or equivalent) 2.2.0rc4 -> 2.2.0rc5
```

bhack:
Until we need to track multiple files to pilot the version I think it is easier to have single script in the root without opening an ide, find and replace, saving all the file in the editor, etc.

gabrieldemarmiesse:
I'm not sure the complexity of this script and regex is worth the saving of the few seconds needed to open and IDE/text editor. I'm not a fan of maintaining a functionality present already in other tools. But maybe it's just me. If another maintainer is interested, we can keep this script.

seanpmorgan:
> I'm not sure I understand, what's the benefit of running
> 
> ```
> ./update_release_version.sh 2.2.0rc5
> ```
> 
> compared to opening Pycharm, VSCode, vim... and do
> 
> ```
> Ctrl+Shift+R (or equivalent) 2.2.0rc4 -> 2.2.0rc5
> ```

So I think the replace strategy is safe to do for specific versions like `2.2.0rc4` becoming `2.2.0rc5`. However, for stable release updates, running CTRL+SHIFT+R is susceptible to make replacements in docs strings and other places where versioning is being discussed but shouldn't be updated.

update_release_version.sh
```diff
@@ -0,0 +1,7 @@
+sed -ri "s/(TF_VERSION=|tensorflow(-cpu)*(~|=)=|tf-version: \[')[0-9]+[a-zA-Z0-9_.-]+/\1$1/g" \
```
seanpmorgan, line 1: Please add copyright info at the top, and relocate to the `tools/` directory to keep project root clean.

bhack:
@seanpmorgan As you can see the list of files is constrained but please double check the safety of the regular expression.

update_release_version.sh
bhack: I moved it to `tools/` but I don't see copyright info in bash scripts. Do you have an example?

seanpmorgan: https://github.com/tensorflow/addons/blob/master/tools/install_deps/buildifier.sh as an example. You're right this is missing from most of our bash scripts... but should still add it I suppose.

PR #1784, opened 2020-05-05
Don't hide utils from the API docs generator

yashk2810:
Utils contains types/ and since the API generator now supports type annotation with links, it will be good to link off to the type pages.

### DIFF
--- a/docs/build_docs.py
+++ b/docs/build_docs.py
@@ -73,7 +73,7 @@ def main(argv):
         root_title=PROJECT_FULL_NAME,
         py_modules=[(PROJECT_SHORT_NAME, tfa)],
         code_url_prefix=code_url_prefix,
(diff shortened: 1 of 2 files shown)


MarkDaoust:
Would it make sense to import that `utils.types` in the root __init__.py so you don't document the rest of `utils` too?

Or switch to that explicit package thing and add `import tensorflow_addons.utils.types` to utils?

Can you post a diff of the output TOC either way?

gabrieldemarmiesse:
Maybe you'd need to bump the version for the CI to pass? See https://github.com/tensorflow/addons/blob/master/tools/install_deps/doc_requirements.txt

facaiy:
I think `utils` module is hidden intentionally, because we don't want to guarantee its backward compatibility.

WindQAQ:
Should be done in https://github.com/tensorflow/addons/pull/2162

Issue #1788, opened 2020-05-07
Unable to save model using WeightNormalization in TF2.1 / load saved model in TF2.2rc4


SNeugberYoti:
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Linux Ubuntu 18.04
- TensorFlow version and how it was installed (source or binary): 2.1 / 2.2rc4, both installed from pip
- TensorFlow-Addons version and how it was installed (source or binary): 0.9.1, installed from pip
- Python version: 3.7.5
- Is GPU used? (yes/no): no

**Describe the bug**

Building a model with a WeightNormalization layer and saving it as a tensorflow saved model throws this error under tensorflow 2.1:

`ValueError: Attempted to save a function b'__inference_model_layer_call_and_return_conditional_losses_692' which references a symbolic Tensor Tensor("model/weight_normalization_1/compute_weights/mul:0", shape=(3, 3, 3, 128), dtype=float32) that is not a simple constant. This is not supported.`

Saving works under tensorflow 2.2rc4, but subsequently trying to load the saved model then throws this error:

`KeyError: '__inference_model_layer_call_and_return_conditional_losses_516'`

**Code to reproduce the issue**

```
import tensorflow as tf
import tensorflow_addons as tfa

# some basic model
inp = tf.keras.layers.Input((112,112,3))
x = tfa.layers.WeightNormalization(tf.keras.layers.Conv2D(128, kernel_size=3))(inp)
x = tf.keras.layers.GlobalAveragePooling2D()(x)
x = tf.keras.layers.Dense(2)(x)

model = tf.keras.models.Model(inputs=[inp], outputs=[x])

tf.saved_model.save(model, "./test_export") # throws in TF2.1
restored_model = tf.saved_model.load("./test_export") # throws in TF2.2rc4
```

SNeugberYoti:
Maybe this helps narrow it down further: I also just tried loading the model in TF2.2rc4 using

`restored_model  = tf.keras.models.load_model("./test_export")`

the error message is now:

`tensorflow.python.saved_model.nested_structure_coder.NotEncodableError: No encoder for object [] of type [<class 'tensorflow.core.protobuf.struct_pb2.StructuredValue'>].`

bhack:
Do you have the same error with `model.save('model',save_format='tf')`?

SNeugberYoti:
Yes

SNeugberYoti:
[...]

PR #1789, opened 2020-05-07
Test WeightNormalization layer can be saved in SavedModel format

SNeugberYoti:
Failing test to highlight issue #1788


SNeugberYoti:
@googlebot I signed it!

gabrieldemarmiesse:
Thank you @bhack  and @SNeugberYoti for the pull request with the test, that's very helpful. I'll use your branch and try to fix it.

gabrieldemarmiesse:
@googlebot I consent.

gabrieldemarmiesse:
So in the fix, we need to force the use of autograph with `@tf.function` so that we don't have to use any `tf.control_dependencies` after the assign.

gabrieldemarmiesse:
I checked and my fix is wrong. `g` has no gradients. I don't think there is an easy way to fix this, to have something that works correctly we might deprecate this and give utility functions to plug into the layer instead. I don't believe that the WeightNormalization layer in this current form is a good idea. We'll have a regular flux of users complaining about bugs.

SNeugberYoti:
Hi @gabrieldemarmiesse 

Looks like `g` has no gradients in your fix because the gradients of `assign` are undefined?

Maybe using `tf.custom_gradient` would work here? Similar to the solution here: https://github.com/tensorflow/tensorflow/issues/17735#issuecomment-437621701

gabrieldemarmiesse:
@SNeugberYoti  thanks for the idea. The bigger issue here is the true value of the weights, and the gradients. I don't think the assignment is good at the semantic level, and it will confuse users who expect the kernel and `g` to be two different things. Now they're mixed together because of the assignment, which isn't correct, and will likely give use other problems moving forward. The right solution is a restructuring of the WeightNormalization layer. I lack time on my side. It would be very nice if a contributor could pick this up. I can give some guidance, but doing the pull request myself requires too much time.

SNeugberYoti:
@gabrieldemarmiesse fair enough. I'd generally be happy to give this a go, if you could outline what you had in mind? I struggle to see a way around changing the kernel of the wrapped layer while also applying the weight normalisation at the right time.

SNeugberYoti:
Regarding your comment [here](https://github.com/tensorflow/addons/pull/1797/files#r422347435), I'm not sure subclassing is the way forward, the wrapper class does seem like the correct approach to me.

> Now they're mixed together because of the assignment, which isn't correct, and will likely give use other problems moving forward. 

Technically this is correct though as this method fundamentally redefines the kernel. Going by the definition of the paper:

>  [...] We do so by expressing the weight vectors in terms of the new parameters using w = g v / ||v||


Going with the subclass approach, wouldn't this require the user to re-implement the "call" method of the base class and change the kernel there, which is basically what the wrapper is doing currently? The current approach seems more robust to me in that case.

I may equally just not know enough of the details though. Do you have an example of where you see the current approach running in to problems in the future?

Since @MokkeMeguru has been working on the layer for the other issue, maybe worth involving here as well.

gabrieldemarmiesse:
For example, one huge problem we have with the implementation that I made in this pull requests is that the weights will change when doing a "model.predict()", because of the assign()

SNeugberYoti:
Right, makes sense.

I have just subclassed Conv2D and added the weightnorm code in. It's quite messy and I can't see an easy way to pull the relevant parts out into utility functions, but it works. Gradients are fine and saving/loading the model is also fine. Especially the data-dependent initialisation made this tricky, having to run the layer once without the normalization functionality.

I wonder why just updating the kernel as in the original implementation (`self.layer.kernel = ...`) breaks model loading though, as I'm not doing much else in my subclassed implementation either.

MokkeMeguru:
(I have one alternative idea. That is to create a wrapper class of **tf.keras.Layer** , and ask someone wants to use WeightNormalization to create a class such as WeightNorm-Conv2D inherit from its class. When I have a similar problem in [my Flow-based Model's Library](https://github.com/MokkeMeguru/TFGENZOO/blob/master/TFGENZOO/flows/flowbase.py#L10-L57), I used this solution. But this may be a misguided idea sry)

e.g.
```
class WeightNormLayer(tf.keras.layers.Layer):
    ...

class WeightNormedConv2D(WeightNormLayer):
    def build(input_shape):
        self.w = self.add_weightnormed_weight(name="weight_w", shape=self.filter_sizes)
```

dathudeptrai:
what is a different betweent tf_addons's weightnorm and tfp's weightnorm here (https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/layers/weight_norm.py) ?

bhack:
The reference paper Is the same https://arxiv.org/abs/1602.07868 Tim Salimans, Diederik P. Kingma (2016)

In TF probability it was introduced Aug 28, 2019 so It Is a more recent adoption then Addons but I've not check impl details.

SNeugberYoti:
Implementation looks relatively similar, and suffers from what appears to be the same issue. Saving a model with their implementation works fine, but loading it again breaks.

bhack:
I think It could confuse a little bit users cause probability It Is not a SIG and it is something ufficially supported by the Google Tensorflow team /cc @seanpmorgan

seanpmorgan:

Issue #1775, opened 2020-05-02
Roadmap for dropping activation custom ops


gabrieldemarmiesse:
Open for discussion, here is a step-by-step timeline to drop custom ops for activations:

1) We add a warning when the user is loading an activation custom op through the python interface, for example, here: https://github.com/tensorflow/addons/blob/master/tensorflow_addons/activations/hardshrink.py#L56 
we recommend switching the `TF_ADDONS_PY_OPS` to 1 or 0 to freeze the behavior. We also recommend in the docs to try `tf.config.optimizer.set_jit(True)` with the pure python implementation to see if there is a speedup.

2) `TF_ADDONS_PY_OPS` 's value is 1 by default. The warning is still there to tell the user that activation custom ops will go away at some point.

3) We remove the python entrypoint for activation custom ops, but we keep the SO file. So even if `TF_ADDONS_PY_OPS` is set to 0, the user will get the pure python version. It's still possible for models to get loaded because by calling `tfa.register_all()`, the SO file gets loaded. 
We should have a warning in the c++ code to tell the custom op for activation are deprecated and will go away in release xxxxx.

4) We remove the custom op completely. Users who haven't re-created their old models need to pin the Tensorflow Addons version for their model to keep working.

bhack:
Do we really need `3`? As we are warning users in both stages `1` and `2` probably it is fine to skip 3 if we give them enough time.

gabrieldemarmiesse:
Maybe for people who don't read warnings? I don't have a strong opinion on this. I don't mind if we skip step 3.

bhack:
We could put a `sleep` in `2` after the second run with `TF_ADDONS_PY_OPS=0` :smile:

AakashKumarNain:
I am fine with `1` and `2`

gabrieldemarmiesse:
I think we should add the warning of 1), and then backport the warning to the 0.9.x branch and make a 0.9.2 release. The would ensure more people get the warning. If that's not too much work @seanpmorgan for you to draft the 0.9.2 release of course.

seanpmorgan:
> I think we should add the warning of 1), and then backport the warning to the 0.9.x branch and make a 0.9.2 release. The would ensure more people get the warning. If that's not too much work @seanpmorgan for you to draft the 0.9.2 release of course.

Good idea... no issue creating a 0.9.2 release.

bhack:
[...]

PR #1791, opened 2020-05-07
Added the DepreciationWarning for the custom op version of activations functions

gabrieldemarmiesse:
Step 1 of #1775


bhack:
Can we have few words in the string so that we advice the user/dev that is a global change for all the activations?
So if you use a specific single activation in your model you will be advised for all early.

gabrieldemarmiesse:
Sure, why not.

gabrieldemarmiesse:
Thanks for the review!

PR #1792, opened 2020-05-07
Bump the tensorflow version to 2.2.0 stable.

gabrieldemarmiesse:


gabrieldemarmiesse:
Oups, the release is only in github at the moment, we need to wait until it's on pypi. Let's just restart the CI once it's done.

PR #1797, opened 2020-05-08
Apply Multi-GPU/TPU Normalization

MokkeMeguru:
In Tensorflow 2.2.0, they implement multi-gpu/tpu's batch normalization layer, So I implement Weight Normalization Layer by reference to it.
Sources:
https://github.com/tensorflow/tensorflow/blob/v2.2.0/tensorflow/python/keras/layers/normalization_v2.py#L165-L204

### DIFF
--- a/tensorflow_addons/layers/wrappers.py
+++ b/tensorflow_addons/layers/wrappers.py
@@ -54,7 +54,7 @@ class WeightNormalization(tf.keras.layers.Wrapper):
     """
 
     @typechecked
-    def __init__(self, layer: tf.keras.layers.Layer, data_init: bool = True, **kwargs):
+    def __init__(self, layer: tf.keras.layers, data_init: bool = True, **kwargs):
         super().__init__(layer, **kwargs)
         self.data_init = data_init
         self._track_trackable(layer, name="layer")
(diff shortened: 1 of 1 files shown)


MokkeMeguru:
Sorry, I am suffering from your Formatter...

tensorflow_addons/layers/wrappers.py
```diff
@@ -182,13 +180,56 @@ def _init_norm(self):
             g_tensor = self.g.assign(tf.reshape(v_norm, (self.layer_depth,)))
             return [g_tensor]
 
+    def _calculate_moments(self, x, axes, keep_dims=True):
```
bhack, line 183: Can we ask to expose this in Tensorflow directly?

MokkeMeguru: I think Tensorflow has a too complex test system. (In my other pull request, I had a very hard time.)
And also, its framework always doesn't have backward compatibility in particular with Multi-GPU/TPU.

So I have no enough power for sending the pull request for TensorFlow...

bhack: I don't know if you could push more effort here cause https://github.com/tensorflow/addons/pull/1789#issuecomment-625876871. /cc @gabrieldemarmiesse

gabrieldemarmiesse: Yeah, we need to rethink the implementation of `WeightNormalization`. To the best of my knowledge it's impossible to hook ourselves because the use of a weight in a layer without using hacky stuff that will come back to bite us later.

We'll likely have to deprecate the wrapper layer and provide utility functions instead, with a good tutorial to explain how the use need to do the subclassing for it to work. We could maybe provide a new Dense layer with the weight normalization implemented inside. That would cover many use cases and provide an example of how to do it for other layers.

It would be nice if somebody could work on this and provide a proposal/possibly a new API to fix all those issues.

seanpmorgan:

Issue #1794, opened 2020-05-07
Remove the tf.cond and tf.control_dependencies in the codebase.


gabrieldemarmiesse:
**Describe the feature and the current behavior/state.**

With tf.function, and more precisely autograph, tf.cond and tf.control_dependencies are not needed anymore. One can take a look at the the [RFC](https://github.com/tensorflow/community/blob/master/rfcs/20180918-functions-not-sessions-20.md) for reference. TensorFlow Addons should aim to be an example for contributors who want to know how to make custom keras objects in TF2 in a clean way. It will also help us by making the code easier to read.

**Relevant information**
- Are you willing to contribute it (yes/no): Some of it
- Are you willing to maintain it going forward? (yes/no): Some of it
- Is there a relevant academic paper? (if so, where): the [RFC](https://github.com/tensorflow/community/blob/master/rfcs/20180918-functions-not-sessions-20.md) but it's not academic :p 
- Is there already an implementation in another framework? (if so, where): no
- Was it part of tf.contrib? (if so, where): no

**Which API type would this fall under (layer, metric, optimizer, etc.)**

All

**Who will benefit with this feature?**

devs who read the code, so some users, contributors and maintainers.

**Any other info.**

I left `tf.while_loop` out of this proposal because there are some extra arguments that we might need to give to the `while_loop` function, and that's not possible with autograph.

Feel free to comment if you think it's a good/bad idea.

bhack:
> I left tf.while_loop out of this proposal because there are some extra arguments that we might need to give to the while_loop function, and that's not possible with autograph.

Is it only in the [seq2seq decoder](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/seq2seq/decoder.py)?

gabrieldemarmiesse:
Yes, it's the only place where there is a tf.cond.

guillaumekln:
I think `tf.control_dependencies` are still required for assert ops. For example the documentation of [`tf.debugging.assert_equal`](https://www.tensorflow.org/api_docs/python/tf/debugging/assert_equal) states:

> **Returns:**
>
> Op that raises `InvalidArgumentError` if `x == y` is False. This can be used with `tf.control_dependencies` inside of `tf.functions` to block followup computation until the check has executed.

bhack:
Are they missing this case in the [main doc reference](https://www.tensorflow.org/api_docs/python/tf/control_dependencies?version=nightly)?

[...]

PR #1799, opened 2020-05-08
Prohibit the use of tf.control_dependencies.

gabrieldemarmiesse:
See #1794  for background


bhack:
I don't know if we could or it is safe enough to check `sys.gettrace` if we are really under a debugger and to enable only that case.

gabrieldemarmiesse:
the `__debug__` directive is very simple to use and specifically targeted for our use case. We'll use that and it's easy to turn off.

Issue #1417, opened 2020-03-26
Discussion: When to do checks of user inputs?


gabrieldemarmiesse:
**Describe the feature and the current behavior/state.**

Currently, we don't have a policy specifying when to do checks on numerical values (user input checks). Sometimes, it's done in eager mode only, sometimes in eager and graph mode...

For example: https://github.com/tensorflow/addons/pull/1338#discussion_r398463749 

To have maximum speed, it would be better if the checks are not running, but we need a way of debugging too. But eager mode is not a garantee of debug mode. Some people us eager mode for flexibility because they can't use graph mode.

To have a strong policy would allow us to do very expensive checks and would improve the UX by allowing users to have checks on their numerical values if they want to do a dry run.

I propose the we do the following:

```python
if tf.executing_eagerly() and __debug__:
    # do very expensive checks of user inputs here, like comparing values, raising exceptions and such.
```

the `__debug__` directive is `True` by default in python (yes it's strange). If python is run in optimized mode with `python -O`, the `__debug__` variable is set to `False`.

In the end, user have several choices:

* graph mode and maximal performance.
* eager mode and maximal performance: `python -O my_script.py`
* eager mode and debug mode to monitor strange values: `python my_script.py`.

Of course this doesn't concern checks of user inputs done in `__init__` functions because those parts are never critial for performance.

**Relevant information**
- Are you willing to contribute it (yes/no): yes
- Are you willing to maintain it going forward? (yes/no): yes
- Is there a relevant academic paper? (if so, where): no
- Is there already an implementation in another framework? (if so, where): not that I'm aware of. If people know other libraries doing similar things, I'd like to know about it.
- Was it part of tf.contrib? (if so, where): no

**Which API type would this fall under (layer, metric, optimizer, etc.)**

All APIs

**Who will benefit with this feature?**

Users as they would benefits from many checks while having maximal performance in graph mode as all checks would have been removed.

**Any other info.**

fsx950223:
It sounds good.

bhack:
I don't know if we could or it is safe enough to check only `sys.gettrace` to see if we are really under a debugger and to enable only the checks in that case.
It seems to me not so much UX friendly to require to use `python -O my_script.py` for having performance in eager_mode if we want to be `eager friendly` by default.
I have not read many other docs about explicit `-O` in other libraries so my point is just related to the user experience point of view and habits.

gabrieldemarmiesse:
[...]

PR #1800, opened 2020-05-08
Enfore the runtime checks in specific conditions.

gabrieldemarmiesse:
Fixes part of https://github.com/tensorflow/addons/issues/1417 . We need to remove elements from the blacklist afterwards.


gabrieldemarmiesse:
Needs more discussion

PR #1805, opened 2020-05-09
Remove tf.control_dependencies from SWA

Squadrick:
Also: clean up the code a bit by replacing tf conds with regular python conditions.

### DIFF
--- a/tensorflow_addons/optimizers/stochastic_weight_averaging.py
+++ b/tensorflow_addons/optimizers/stochastic_weight_averaging.py
@@ -120,28 +120,29 @@ class SWA(AveragedOptimizerWrapper):
     def average_op(self, var, average_var):
         average_period = self._get_hyper("average_period", tf.dtypes.int64)
         start_averaging = self._get_hyper("start_averaging", tf.dtypes.int64)
-        # check if the correct number of iterations has taken place to start
-        # averaging.
-        thresold_cond = tf.greater_equal(self.iterations, start_averaging)
         # number of times snapshots of weights have been taken (using max to
         # avoid negative values of num_snapshots).
         num_snapshots = tf.math.maximum(
             tf.cast(0, tf.int64),
             tf.math.floordiv(self.iterations - start_averaging, average_period),
         )
-        # checks if the iteration is one in which a snapshot should be taken.
-        sync_cond = tf.equal(
-            start_averaging + num_snapshots * average_period, self.iterations
+
+        # The average update should happen iff two conditions are met:
+        # 1. A min number of iterations (start_averaging) have taken place.
+        # 2. Iteration is one in which snapshot should be taken.
+        update_cond = tf.reduce_all(
+            [
+                self.iterations >= start_averaging,
+                self.iterations == start_averaging + num_snapshots * average_period,
+            ]
         )
+
         num_snapshots = tf.cast(num_snapshots, tf.float32)
         average_value = (average_var * num_snapshots + var) / (num_snapshots + 1.0)
-        average_cond = tf.reduce_all([thresold_cond, sync_cond])
-        with tf.control_dependencies([average_value]):
-            average_update = average_var.assign(
-                tf.where(average_cond, average_value, average_var,),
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse:
Thanks! I think we need to force autograph if we're removing tf.control_dependencies. Could you decorate the method with `@tf.function`?

Squadrick:
@gabrieldemarmiesse Added `tf.function`, should be good to go.

tensorflow_addons/optimizers/stochastic_weight_averaging.py
```diff
@@ -117,31 +117,33 @@ def __init__(
...
         average_value = (average_var * num_snapshots + var) / (num_snapshots + 1.0)
-        average_cond = tf.reduce_all([thresold_cond, sync_cond])
-        with tf.control_dependencies([average_value]):
-            average_update = average_var.assign(
-                tf.where(average_cond, average_value, average_var,),
-                use_locking=self._use_locking,
-            )
-        return average_update
+        return average_var.assign(
+            tf.where(update_cond, average_value, average_var),
```
gabrieldemarmiesse, line 144: If I'm not mistaken, we work with scalar tensors here. Would it be possible to use `if` instead of tf.where and `and` instead of tf.reduce_all? That would be easier to read.

Squadrick: Updated to use python `if`.

Squadrick:
cc @shreyashpatodia Could you review this again? Replaced `tf.where` with python `if`.

tensorflow_addons/optimizers/stochastic_weight_averaging.py
```diff
@@ -117,31 +117,27 @@ def __init__(
...
-        average_value = (average_var * num_snapshots + var) / (num_snapshots + 1.0)
-        average_cond = tf.reduce_all([thresold_cond, sync_cond])
-        with tf.control_dependencies([average_value]):
-            average_update = average_var.assign(
-                tf.where(average_cond, average_value, average_var,),
-                use_locking=self._use_locking,
-            )
-        return average_update
+
+        # The average update should happen iff two conditions are met:
```
seanpmorgan, line 131:
```suggestion
        # The average update should happen if two conditions are met:
```

Squadrick: I was using the mathematical notion: `iff` meaning [if and only if](https://en.wikipedia.org/wiki/If_and_only_if).

seanpmorgan: :D should look more carefully

tensorflow_addons/optimizers/stochastic_weight_averaging.py
```diff
@@ -117,31 +117,27 @@ def __init__(
...
+        # 1. A min number of iterations (start_averaging) have taken place.
+        # 2. Iteration is one in which snapshot should be taken.
+        checkpoint = start_averaging + num_snapshots * average_period
+        if self.iterations >= start_averaging and self.iterations == checkpoint:
+            num_snapshots = tf.cast(num_snapshots, tf.float32)
+            average_value = (average_var * num_snapshots + var) / (num_snapshots + 1.0)
+            return average_var.assign(average_value)
```
facaiy, line 138:
```suggestion
            return average_var.assign(average_value, use_locking=self._use_locking)
```

facaiy:
@seanpmorgan Sean, is it necessary to add a test for graph mode?

seanpmorgan:
> @seanpmorgan Sean, is it necessary to add a test for graph mode?

Sorry for the delay. We've switched to only supporting TF2 paradigms and so decorating with `@tf.function` is the only graph mode we support. This has made UX for testing addons significantly easier and aligns with the recommended TF2 patterns.

Squadrick:
@seanpmorgan Thanks for pointing that out. Removed from exceptions. Should be good to go.

PR #1806, opened 2020-05-09
Create new type for Optimizer

Squadrick:


tensorflow_addons/optimizers/average_wrapper.py
```diff
@@ -16,15 +16,16 @@
...
 from typeguard import typechecked
-from typing import Union
 
 
 class AveragedOptimizerWrapper(tf.keras.optimizers.Optimizer, metaclass=abc.ABCMeta):
     @typechecked
     def __init__(
         self,
-        optimizer: Union[tf.keras.optimizers.Optimizer, str],
+        optimizer: Optimizer,
```
gabrieldemarmiesse, line 28:
```suggestion
        optimizer: types.Optimizer,
```

It may confuse readers who are not looking at the import. They may think it's the keras Optimizer class.

Squadrick: That's a good catch, looking back at it it does look a lot like keras's Optimizer class.

Squadrick:
@gabrieldemarmiesse Changed to `types.Optimizer`.

PR #1807, opened 2020-05-09
Remove `sequential_update` from AverageWrapper

Squadrick:
In TF2.0, sequential_update is redundant. This allows the removal of `tf.control_dependencies` from average_wrapper and its subclasses: moving_average and stochastic_weight_averaging.

### DIFF
--- a/tensorflow_addons/optimizers/average_wrapper.py
+++ b/tensorflow_addons/optimizers/average_wrapper.py
@@ -48,11 +48,11 @@ class AveragedOptimizerWrapper(tf.keras.optimizers.Optimizer, metaclass=abc.ABCM
         self._optimizer = optimizer
 
         if sequential_update is not None:
-          warnings.warn(
-              "The parameter `sequential_update` is redundant due to AutoGraph. "
-              "This behavior is deprecated and in Addons 0.12, this will raise an error. ",
-              DeprecationWarning,
-          )
+            warnings.warn(
+                "The parameter `sequential_update` is redundant due to AutoGraph. "
+                "This behavior is deprecated and in Addons 0.12, this will raise an error. ",
+                DeprecationWarning,
+            )
 
     def _create_slots(self, var_list):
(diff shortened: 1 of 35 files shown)


gabrieldemarmiesse:
Thanks for the pull request! A nice cleanup of the API! To ensure we're consistent, we should wrap the method with a `@tf.function` with autograph enabled, and since it's a public API we can't remove an argument without warnings. Can you remove the tf.control_dependencies while keeping the sequential update? With a DeprecationWarning saying that it will go away in TFA 0.12.0?

tensorflow_addons/optimizers/average_wrapper.py
```diff
@@ -45,6 +47,12 @@ def __init__(
         self._optimizer = optimizer
         self._sequential_update = sequential_update
 
+        warnings.warn(
```
gabrieldemarmiesse, line 50: Here all the users will see the warning, even if they've never used it. To target exactly the users who haved changed the parameter, set the default value to None. Then do:

```python
if sequential_update is not None:
    warnings.warn(...)
```

Squadrick: Unfortunately, the default value for sequential_update was set to `True` when it was added, and that has remained.

So testing for not None is not effective. Since the type of sequential_update is strictly Bool.

gabrieldemarmiesse: We can relax the bool requirement by typing it as Optional[bool], then we can set it to None by default and the warning will trigger only if the user had been setting the value manually. Do you think it makes sense?

Squadrick: Yeah, made it `Optional[bool]`. I removed `sequential_update` from the optimizer's config as well. Loading an older model with `sequential_update` will show the `DeperecationWarning`.

gabrieldemarmiesse:
Thanks a lot! It makes the code a lot simpler!

PR #1812, opened 2020-05-11
remove internal modules

fsx950223:
cc @gabrieldemarmiesse

### DIFF
--- a/tensorflow_addons/utils/keras_utils.py
+++ b/tensorflow_addons/utils/keras_utils.py
@@ -59,9 +59,6 @@ class LossFunctionWrapper(tf.keras.losses.Loss):
     Returns:
       Loss values per sample.
     """
-        # if tf.is_tensor(y_pred) and tf.is_tensor(y_true):
-        #   y_pred, y_true = tf_losses_util.squeeze_or_expand_dimensions(
-        #       y_pred, y_true)
         return self.fn(y_true, y_pred, **self._fn_kwargs)
 
     def get_config(self):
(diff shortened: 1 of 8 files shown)


tensorflow_addons/utils/keras_utils.py
```diff
@@ -15,6 +15,58 @@
 """Utilities for tf.keras."""
 
 import tensorflow as tf
+import six
```
WindQAQ, line 18: We might not need to use `six` as we have dropped the support of python2

tensorflow_addons/utils/keras_utils.py
```diff
@@ -15,6 +15,58 @@
...
+        this defaults to `SUM_OVER_BATCH_SIZE`. When used with
+        `tf.distribute.Strategy`, outside of built-in training loops such as
+        `tf.keras` `compile` and `fit`, using `AUTO` or `SUM_OVER_BATCH_SIZE`
+        will raise an error. Please see this custom training [tutorial](
+          https://www.tensorflow.org/tutorials/distribute/custom_training)
+        for more details.
+      name: (Optional) name for the loss.
+      **kwargs: The keyword arguments that are passed on to `fn`.
+    """
+        super(LossFunctionWrapper, self).__init__(reduction=reduction, name=name)
```
gabrieldemarmiesse, line 48: No need for arguments in super in py3

fsx950223:
cc @WindQAQ @gabrieldemarmiesse

tensorflow_addons/utils/keras_utils.py
```diff
@@ -17,6 +17,57 @@
...
+      Loss values per sample.
+    """
+        return self.fn(y_true, y_pred, **self._fn_kwargs)
+
+    def get_config(self):
+        config = {}
+        for k, v in iter(self._fn_kwargs.items()):
+            config[k] = tf.keras.backend.eval(v) if is_tensor_or_variable(v) else v
+        base_config = super(LossFunctionWrapper, self).get_config()
+        return dict(list(base_config.items()) + list(config.items()))
```
WindQAQ, line 68: No need to feed args into `super`, and we can use literal instead of constructor for dict :-)

https://github.com/tensorflow/addons/blob/master/tensorflow_addons/layers/adaptive_pooling.py#L388-L389

PR #1813, opened 2020-05-11
Linear transformation

ghosalsattam:
This PR is aimed at Linear Transformation of images.


ghosalsattam:
@WindQAQ I can modify this function if required. 
The negative operation and identity operation(#1333 #1353 ) are special cases of the linear transformation. Besides this PR with slight modification can address the shifting of grayscale value from one range to other.

WindQAQ:
> @WindQAQ I can modify this function if required.
> The negative operation and identity operation(#1333 #1353 ) are special cases of the linear transformation. Besides this PR with slight modification can address the shifting of grayscale value from one range to other.

So maybe we should discuss more about this kind of single-lined wrapper. Personally, I discourage the single-lined wrapper because it adds more testing codes than "real" source codes.

@gabrieldemarmiesse @seanpmorgan @Squadrick @bhack  Hi, thoughts on this?

> And what can be acceptable tollerance value?

Just leave the argument default.

However, for example, if you are trying to test against scipy's implementation, we might want to make tolerance larger because the underlying implementation of scipy uses double aka float64 to compute, which produces more precise results. In this case, we can adjust the tolerance to acceptable value say 1e-6 or 1e-5 etc.

bhack:
For `output = a * image + b` we need to maintain 107 lines of code in the repo.

gabrieldemarmiesse:
We really need to clean up #1333 and #1353, those issues were made without inspecting too much the list of said transformations or how it was phrased, and now contributors are being confused about why we need those and what form it should take. We are wasting contributors' time and that's not good. I'll try to take a look at it tomorrow and cleanup / clarify those issues. Apologies @ghosalsattam we should have been more careful about the issues and what was said in them. We need those transformations for randaugment and autoaugment, but we don't need to expose all of them with a public API.

Issue #1826, opened 2020-05-13
tfa.image.sharpness does not expose in __init__.py


WindQAQ:
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): colab
- TensorFlow version and how it was installed (source or binary): binary
- TensorFlow-Addons version and how it was installed (source or binary): nightly
- Python version: 3.6
- Is GPU used? (yes/no): no

**Describe the bug**
should expose `sharpness` under [\_\_init\_\_.py](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/__init__.py).

**Code to reproduce the issue**

https://colab.research.google.com/drive/1jSa4B06CopeVfgExeSoO6roPR5eTdSYn?usp=sharing

WindQAQ:
closed via #1827

PR #1827, opened 2020-05-13
Expose sharpness

bhack:


bhack:
Fix https://github.com/tensorflow/addons/issues/1826

Issue #1829, opened 2020-05-13
tfa.image.equalize is not compatible with tf.data.Dataset


WindQAQ:
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): colab
- TensorFlow version and how it was installed (source or binary): 2.2.0
- TensorFlow-Addons version and how it was installed (source or binary): tfa-nightly
- Python version: 3.6
- Is GPU used? (yes/no): no

**Describe the bug**
The error raises if using `tfa.image.equalize` with `tf.data.Dataset`. The error is either
```
OperatorNotAllowedInGraphError: using a `tf.Tensor` as a Python `bool` is not allowed in Graph execution
```
or
```
OperatorNotAllowedInGraphError: using a `tf.Tensor` as a Python `bool` is not allowed: AutoGraph did not convert this function.
```
depending on whether `tf.function` is used.

Seems that there are something mysterious about the subgraph in `tf.data.Dataset`. The direct fix is to use `tf.cond` instead of python control flow, but it might violate TF 2.x style.

**Code to reproduce the issue**

https://colab.research.google.com/drive/1B2Bl0Dwmto7sN8rqRodrtXEtZdp07e0w?usp=sharing

bhack:
https://github.com/tensorflow/addons/pull/1830

WindQAQ:
For tracking, this is related to tensorflow/tensorflow#26091 and tensorflow/tensorflow#26602.

bhack:
@WindQAQ It was `map_fn`. My PR was merged in Tensorflow https://github.com/tensorflow/tensorflow/pull/39541

PR #1830, opened 2020-05-13
Fix condition tracing in scale_channel

bhack:
Try to fix https://github.com/tensorflow/addons/issues/1829

### DIFF
--- a/tensorflow_addons/image/color_ops.py
+++ b/tensorflow_addons/image/color_ops.py
@@ -30,6 +30,7 @@ from functools import partial
 def equalize_image(image: TensorLike, data_format: str = "channels_last") -> tf.Tensor:
     """Implements Equalize function from PIL using TF ops."""
 
+    @tf.function
     def scale_channel(image, channel):
         """Scale the data in the channel to implement equalize."""
         image_dtype = image.dtype


seanpmorgan:
Thanks! Tracing the encapsulated function appears to work:
https://colab.research.google.com/drive/18QvTLJbS-21QD1tzNyP1G3gjq2SY4RjQ?usp=sharing

Would very much like the standardize the testing for tf.data.Dataset on tfa.image with #992

bhack:
@seanpmorgan I don't know if in the original colab example, the one with `tf.function`,  there is a know tracing issue with `partial`.

bhack:
I think that we have a more general issue here about `partial` with `@tf.function`.
Instead about `tf.data.Dataset` tests what is the plan with https://github.com/tensorflow/addons/issues/992?

WindQAQ:
> I think that we have a more general issue here about `partial` with `@tf.function`.
> Instead about `tf.data.Dataset` tests what is the plan with #992?

Could you give me the link to the issue about `partial` and `tf.function`? Thank you!

bhack:
@WindQAQ 

`fn = tf.function(partial(equalize_image, data_format=data_format))` 
it works.

```
@tf_function
def equalize(
```
doesn't trace `equalize_image` in partial

```
@tf_function
def equalize_image(
```
it works

bhack:
We had some previous issues with `partial`:
https://github.com/tensorflow/tensorflow/issues/26091
https://github.com/tensorflow/tensorflow/issues/26602

But seems that they have not tested entering in `partial` with `@tf.function` like in your original colab.

bhack:
If you are interested I am tying to check if `partial` inside a decorated function is a bug/limit of the tracer at https://github.com/tensorflow/tensorflow/pull/39541

PR #1836, opened 2020-05-15
Speedup sharpness

WindQAQ:
`tf.nn.depthwise_conv2d` can work on 4D image directly so we do not need to use `tf.map_fn` to operate image-by-image. Achieve about 1.5X speed on large batch of images.
https://colab.research.google.com/drive/1yqqBp6-aXXgUbmN-c-DMOl0RgSaRYX4j?usp=sharing


bhack:
> we do not need to use tf.map_fn

For the general knowledge the original issue releted `tf.map_fn` is going to be solved merging https://github.com/tensorflow/tensorflow/pull/39541

PR #1848, opened 2020-05-16
Remove squadrick from subpackage CODEOWNERS

Squadrick:
`metrics` is currently without any owners.


Squadrick:
Thanks, Sean. It's been an absolute pleasure working with you!

PR #1849, opened 2020-05-16
Decorate _update_multi_class_model

bhack:


AakashKumarNain:
@bhack why not `binary` as well?

bhack:
> @bhack why not `binary` as well?

Is it needed there? I don't see python control flow in that function (I could be wrong cause I am checking on a mobile device so not the best usability experience :smile:)

AakashKumarNain:
Yeah, it's not there but we can anyways decorate that as well. Let's keep that for the next PR. Thanks for the fix.

bhack:
> Yeah, it's not there but we can anyways decorate that as well. Let's keep that for the next PR. Thanks for the fix.

Yes I asked to @seanpmorgan some days ago how we could handle the decoration more in general also in relation to the experimental_compile parameter if/when we need to enable test this on demand so that we have a coherent guideline. 
E.g. these are private methods so It Is not clear if we want to use the decoration directly up to call chain in the public method and to let It hynerit  or we want leave this decoration atomicity on single functions so that we could pilot single function compilation.

PR #1852, opened 2020-05-17
Enable half and double for resampler GPU ops

WindQAQ:
Actually, I don't know how we passed tests previously...
We don't register half and double backward kernel, and even don't specialize half forward kernel. But we indeed test them.

https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/tests/resampler_ops_test.py#L95-L124
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/tests/resampler_ops_test.py#L185-L213
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/util/gpu_device_functions.h#L670-L691


seanpmorgan:
@alextp No urgency, but whenever time allows could you or someone you could direct us to take a look at why this was working before. Would like to improve our understanding of the registering mechanism.

PR #1855, opened 2020-05-17
Enable some compiler optimalization tricks for correlation cost GPU ops

WindQAQ:
- use `ldg` to read data
- add `__restrict__` qualifier to pointer

### DIFF
--- a/tensorflow_addons/custom_ops/layers/cc/kernels/correlation_cost_op_gpu.cu.cc
+++ b/tensorflow_addons/custom_ops/layers/cc/kernels/correlation_cost_op_gpu.cu.cc
@@ -20,6 +20,7 @@ limitations under the License.
 #include "gpu/cub/device/device_reduce.cuh"
 #include "tensorflow/core/framework/tensor.h"
 #include "tensorflow/core/framework/tensor_shape.h"
+#include "tensorflow/core/util/gpu_device_functions.h"
 #include "tensorflow/core/util/gpu_kernel_helper.h"
 #include "tensorflow/core/util/tensor_format.h"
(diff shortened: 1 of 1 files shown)


tensorflow_addons/custom_ops/layers/cc/kernels/correlation_cost_op_gpu.cu.cc
```diff
@@ -69,19 +71,20 @@ __global__ void pad_and_no_transpose(const float *input, float *output, int C,
...
   }
 }
 
 template <unsigned int THREADS_PER_BLOCK>
-__global__ void Correlation_forward(float *output, int Cout, int Hout, int Wout,
-                                    float *pInput1, int Cin, int Hin, int Win,
-                                    float *pInput2, int pad, int kernel_size,
-                                    int max_displacement, int stride1,
-                                    int stride2) {
+__global__ void Correlation_forward(float *__restrict__ output, int Cout,
```
bhack, line 81: We are not supporting Visual c++ right?

WindQAQ: @bhack Sorry for the late reply. We currently only build GPU ops linux wheel. https://github.com/tensorflow/addons/issues/784

bhack: OK cause in TF lite we had guards in multiple places lke https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/kernels/internal/optimized/sse_tensor_utils_impl.h#L22

PR #1877, opened 2020-05-25
Zoneout LSTM Cell

### LINKED ISSUE #1867: want zoneout lstm supported

no zoneout lstm found in addons

**Relevant information**
- Is there a relevant academic paper? (if so, where): https://arxiv.org/abs/1606.01305
- Is there already an implementation in another framework? (if so, where): https://github.com/LaurentMazare/deep-models/blob/c54a0f5227d1b03018bbbf19075454250266bf90/rhn/lstm_zoneout.py

tfa.layers

people implementing Tacotron2 with tf.keras API

**Any other info.**

failure-to-thrive:
Closes #1867

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -153,8 +153,8 @@
 /tensorflow_addons/optimizers/yogi.py @manzilz
 /tensorflow_addons/optimizers/tests/yogi_test.py @manzilz
 
-/tensorflow_addons/rnn/cell.py @qlzh727 @pedrolarben
-/tensorflow_addons/rnn/tests/cell_test.py @qlzh727 @pedrolarben
+/tensorflow_addons/rnn/cell.py @qlzh727 @pedrolarben @failure-to-thrive
+/tensorflow_addons/rnn/tests/cell_test.py @qlzh727 @pedrolarben @failure-to-thrive
 
 /tensorflow_addons/seq2seq/ @qlzh727 @guillaumekln
 
--- a/tensorflow_addons/rnn/cell.py
+++ b/tensorflow_addons/rnn/cell.py
@@ -799,3 +799,50 @@ class ESNCell(keras.layers.AbstractRNNCell):
         }
         base_config = super().get_config()
         return {**base_config, **config}
+
+
+@tf.keras.utils.register_keras_serializable(package="Addons")
+class ZoneoutLSTMCell(keras.layers.LSTMCell):
+    """LSTM cell with recurrent zoneout.
+
+    https://arxiv.org/abs/1606.01305
+    """
+
+    @typechecked
+    def __init__(
+        self,
+        units: int,
+        zoneout_h: float = 0,
+        zoneout_c: float = 0,
+        seed: int = None,
+        **kwargs
+    ):
+        """
+        """
+        super().__init__(units, **kwargs)
+        self.zoneout_h = zoneout_h
+        self.zoneout_c = zoneout_c
+        self.seed = seed
+
+    def _zoneout(self, t, tm1, rate, training):
+        dt = tf.cast(
+            tf.random.uniform(t.shape, seed=self.seed) >= rate * training, t.dtype
+        )
+        return dt * t + (1 - dt) * tm1
+
+    def call(self, inputs, states, training=None):
+        if training is None:
+            training = keras.backend.learning_phase()
+        output, new_states = super().call(inputs, states, training)
+        h = self._zoneout(new_states[0], states[0], self.zoneout_h, training)
+        c = self._zoneout(new_states[1], states[1], self.zoneout_c, training)
+        return h, [h, c]
+
+    def get_config(self):
+        config = {
+            "zoneout_h": self.zoneout_h,
+            "zoneout_c": self.zoneout_c,
+            "seed": self.seed,
+        }
+        base_config = super().get_config()
+        return {**base_config, **config}
--- a/tensorflow_addons/rnn/tests/cell_test.py
+++ b/tensorflow_addons/rnn/tests/cell_test.py
@@ -558,3 +558,64 @@ def test_esn_config():
     restored_cell = rnn_cell.ESNCell.from_config(config)
     restored_config = restored_cell.get_config()
     assert config == restored_config
+
+
+def test_zoneout():
+    units = 3
+    inputs = tf.constant([[[1.0, 2, 3, 4, 5]]])
+    initial_state = [tf.constant([[11.0, 12, 13]]), tf.constant([[21.0, 22, 23]])]
+
+    const_initializer = tf.constant_initializer(0.5)
+
+    # No zoneout. Testing against vanila LSTM.
+    lstm_layer = keras.layers.LSTM(
+        units,
+        kernel_initializer=const_initializer,
+        recurrent_initializer=const_initializer,
+        return_state=True,
+    )
+    lstm_output = lstm_layer(inputs, training=True)
+
+    zoneout_cell = rnn_cell.ZoneoutLSTMCell(
+        units,
+        kernel_initializer=const_initializer,
+        recurrent_initializer=const_initializer,
+    )
+    zoneout_layer = keras.layers.RNN(zoneout_cell, return_state=True)
+    zoneout_output = zoneout_layer(inputs, training=True)
+
+    np.testing.assert_allclose(lstm_output, zoneout_output)
+
+    # Total zoneout. Testing against initial state.
+    zoneout_cell = rnn_cell.ZoneoutLSTMCell(
+        units,
+        zoneout_h=1,
+        zoneout_c=1,
+        kernel_initializer=const_initializer,
+        recurrent_initializer=const_initializer,
+    )
+    zoneout_layer = keras.layers.RNN(zoneout_cell, return_state=True)
+    zoneout_output = zoneout_layer(inputs, initial_state=initial_state, training=True)
+
+    np.testing.assert_allclose(
+        [initial_state[0], initial_state[0], initial_state[1]], zoneout_output
+    )
+
+    # Inference. Expected to be the same as LSTM.
+    zoneout_output = zoneout_layer(inputs)
+
+    np.testing.assert_allclose(lstm_output, zoneout_output)
+
+
+def test_zoneout_config():
+    cell = rnn_cell.ZoneoutLSTMCell(3, zoneout_c=0.5)
+    config = cell.get_config()
+
+    our_config = {
+        "units": 3,  # just to ensure that we do get the base's class config
+        "zoneout_h": 0,
+        "zoneout_c": 0.5,
+        "seed": None,
+    }
+
+    assert our_config == {key: config[key] for key in our_config.keys() & config.keys()}
--- a/tensorflow_addons/rnn/__init__.py
+++ b/tensorflow_addons/rnn/__init__.py
@@ -18,3 +18,4 @@ from tensorflow_addons.rnn.cell import LayerNormLSTMCell
 from tensorflow_addons.rnn.cell import NASCell
 from tensorflow_addons.rnn.cell import LayerNormSimpleRNNCell
 from tensorflow_addons.rnn.cell import ESNCell
+from tensorflow_addons.rnn.cell import ZoneoutLSTMCell


pedrolarben:
Sure, I will do it asap

tensorflow_addons/rnn/cell.py
```diff
@@ -799,3 +799,50 @@ def get_config(self):
...
         base_config = super().get_config()
         return {**base_config, **config}
+
+
+@tf.keras.utils.register_keras_serializable(package="Addons")
+class ZoneoutLSTMCell(keras.layers.LSTMCell):
+    """LSTM cell with recurrent zoneout.
+
+    https://arxiv.org/abs/1606.01305
+    """
```
pedrolarben, line 809: Please, add the description of the arguments of the class (units, zoneout_h, zoneout_c, seed) in the docstring

tensorflow_addons/rnn/cell.py
```diff
@@ -799,3 +799,50 @@ def get_config(self):
...
+    def __init__(
+        self,
+        units: int,
+        zoneout_h: float = 0,
+        zoneout_c: float = 0,
+        seed: int = None,
+        **kwargs
+    ):
+        """
+        """
```
pedrolarben, line 821: Remove this empty docstring

tensorflow_addons/rnn/cell.py
```diff
@@ -799,3 +799,50 @@ def get_config(self):
...
+        self.zoneout_c = zoneout_c
+        self.seed = seed
+
+    def _zoneout(self, t, tm1, rate, training):
+        dt = tf.cast(
+            tf.random.uniform(t.shape, seed=self.seed) >= rate * training, t.dtype
+        )
+        return dt * t + (1 - dt) * tm1
+
+    def call(self, inputs, states, training=None):
```
pedrolarben, line 833: In order to get the `training` value when it's None, try first with  `base_layer_utils.call_context().training`,  which is a more specific approach. Then, if not found, use keras backend as you were doing. 
You can use [this](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/backend.py#L4364) as a reference.

tensorflow_addons/rnn/cell.py
```diff
@@ -799,3 +799,50 @@ def get_config(self):
...
+        """
+        """
+        super().__init__(units, **kwargs)
+        self.zoneout_h = zoneout_h
+        self.zoneout_c = zoneout_c
+        self.seed = seed
+
+    def _zoneout(self, t, tm1, rate, training):
+        dt = tf.cast(
+            tf.random.uniform(t.shape, seed=self.seed) >= rate * training, t.dtype
```
pedrolarben, line 829: Instead of multiplying by `training`, it could be surrounded by and if/else statement so that we avoid unnecessary computation (the random matrix)

tensorflow_addons/rnn/tests/cell_test.py
```diff
@@ -558,3 +558,64 @@ def test_esn_config():
     restored_cell = rnn_cell.ESNCell.from_config(config)
     restored_config = restored_cell.get_config()
     assert config == restored_config
+
+
+def test_zoneout():
```
pedrolarben, line 563: Please, add another test case for an e2e test with keras model and compile/fit. You can take `test_esn_keras_rnn_e2e` as an example.

tensorflow_addons/rnn/tests/cell_test.py
```diff
@@ -558,3 +558,64 @@ def test_esn_config():
...
+        [initial_state[0], initial_state[0], initial_state[1]], zoneout_output
+    )
+
+    # Inference. Expected to be the same as LSTM.
+    zoneout_output = zoneout_layer(inputs)
+
+    np.testing.assert_allclose(lstm_output, zoneout_output)
+
+
+def test_zoneout_config():
```
pedrolarben, line 610: You should test if you can create the cell using `from_config` function.
It could be something like this:
` 
restored_cell = rnn_cell.ZoneoutLSTMCell.from_config(config)
`
`
restored_config = restored_cell.get_config()
`
`
assert config == restored_config
`

tensorflow_addons/rnn/tests/cell_test.py
```diff
@@ -558,3 +558,64 @@ def test_esn_config():
...
+    cell = rnn_cell.ZoneoutLSTMCell(3, zoneout_c=0.5)
+    config = cell.get_config()
+
+    our_config = {
+        "units": 3,  # just to ensure that we do get the base's class config
```
pedrolarben, line 615: I don't see the point of this comment. If you are referring to the value of `units`, I don't think you should write it. If you are referring to the method in general, please add it in a new line before defining `our_config`.

tensorflow_addons/rnn/cell.py
failure-to-thrive: How does your approach outperform the current one? It's just a fallback. Your fallback still requires that fallback. :) I'm even not sure that we need it here but just peeked it from Tensorflow's Dropout implementation.

tensorflow_addons/rnn/cell.py
failure-to-thrive: If I understand things correctly, 99% of the execution time is a training. No harm at all but looks much nicer.

tensorflow_addons/rnn/tests/cell_test.py
failure-to-thrive: We do not override `from_config()` nor any of its helpers, why do we have to test it? When it could fail and how can we help if it fails? IMHO it's not a child class authors responsibility.

tensorflow_addons/rnn/tests/cell_test.py
failure-to-thrive: Let me explain. As opposed to the common approach, I do not test base's class config. Only our child class' values. We are not responsible for parent class' values. However, since we override `get_config()` we still have to be sure that parent class' values got back too. Testing for `units` should be enough.

tensorflow_addons/rnn/tests/cell_test.py
pedrolarben: I don't think the inherited method `from_config()` is going to fail, it just runs `cls(**config)`, where `cls` is `ZoneoutLSTMCell`, so it is calling the constructor. In this case, I'm sure it will not fail but it is better to have good tests implemented so that we avoid future mistakes while modifying this class (adding new arguments or changes in the `get_config`). 

A possible case when testing the `from_config` method can fail is if an argument from the constructor is not well defined in the `get_config` method.

tensorflow_addons/rnn/tests/cell_test.py
pedrolarben: Now it's clear for me, thanks! :)

tensorflow_addons/rnn/cell.py
pedrolarben: the `call_context` function seemed to be a more specific function. The approach I suggested is the one used by default, it is implemented here: [in_train_phase](https://github.com/tensorflow/tensorflow/blob/a3aec8d37ec6dcd7e82a68067162f492be00944f/tensorflow/python/keras/backend.py#L4346).
However, I see your reference from Dropout [implementation](https://github.com/tensorflow/tensorflow/blob/2b96f3662bd776e277f86997659e61046b56c315/tensorflow/python/keras/layers/core.py#L198). Seems ok for me, thanks again :).

tensorflow_addons/rnn/cell.py
pedrolarben: I totally agree with you that it looks much nicer with the multiplications. But I am concerned that it could harm prediction performance. Sometimes we forget about the importance of predicting time as most of the trained models don't end in productions, but the main goal of training models is to use it in productions and therefore prediction time matters.

failure-to-thrive: Performance of production code is a much more broader topic. If someone report that that redundant random dramatically degrades their model inference to an unacceptable level, I will definitely try to find the best possible solution. Otherwise, I see no reasons to change it now and break something accidentally.
Also, just for note, branching is a bad thing for performance. Especially in the GPU world.

tensorflow_addons/rnn/tests/cell_test.py
failure-to-thrive: Remember, the only thing that may happen to a public class is an addition of an optional argument. Any other updates break backward compatibility and must be prohibited in the first place. At the end of the all, it's all about maintaining appropriate tests. Sorry, I personally see no reasons why to complicate the things.

tensorflow_addons/rnn/cell.py
pedrolarben: It's ok, we can leave it in this more elegant way and take into account if someone reports any performance issue.

tensorflow_addons/rnn/tests/cell_test.py
pedrolarben: I was not talking about modifications of the LSTM layer but the child ZoneoutLSTM. 
I personally see no complication on adding those three lines that assure the complete process of initializing the recurrent cell from the config generated works fine. 
Apart from its utility, which I really believe in, It has been tested for every other recurrent cell in tensorflow_addons, so you can take consistency as another reason.

tensorflow_addons/rnn/cell.py
qlzh727: Please note that training can be a tensor, which means if check will yield error. In the case you want to disable the zone out in the inference mode, please use tf.cond. Please check the keras.backend.dropout as an example.

tensorflow_addons/rnn/cell.py
```diff
@@ -799,3 +799,50 @@ def get_config(self):
...
+        )
+        return dt * t + (1 - dt) * tm1
+
+    def call(self, inputs, states, training=None):
+        if training is None:
+            training = keras.backend.learning_phase()
+        output, new_states = super().call(inputs, states, training)
```
qlzh727, line 836: Since the output is not used here, probably replace it with _.

.github/CODEOWNERS
```diff
@@ -153,8 +153,8 @@
 /tensorflow_addons/optimizers/yogi.py @manzilz
 /tensorflow_addons/optimizers/tests/yogi_test.py @manzilz
 
-/tensorflow_addons/rnn/cell.py @qlzh727 @pedrolarben
-/tensorflow_addons/rnn/tests/cell_test.py @qlzh727 @pedrolarben
+/tensorflow_addons/rnn/cell.py @qlzh727 @pedrolarben @failure-to-thrive
```
qlzh727, line 156: I think we probably want to split the file in future, so that the PR/issue will be forward to the correct person.

pedrolarben: Yes, I think it would be better

tensorflow_addons/rnn/cell.py
failure-to-thrive: `tf.cond` is phasing out now: https://github.com/tensorflow/addons/issues/1794

.github/CODEOWNERS
failure-to-thrive: > I think we probably want to split the file in future, so that the PR/issue will be forward to the correct person.

@gabrieldemarmiesse, could you please take a look?

gabrieldemarmiesse: I agree, not all cell classes should be in the same file. We can move the code of zoneoutLSTM in another file.

failure-to-thrive: To simplify the process, let's merge that PR and split the `cell.py` and `cell_test.py` thereafter. Those files contains lots of different cells already.

Mushoz:
Any reason why this was never merged?

pedrolarben:
> Any reason why this was never merged?

There are some comments that need to be addressed and this is now outdated, as cell.py has been refactored and split into different files: https://github.com/tensorflow/addons/tree/master/tensorflow_addons/rnn

Mushoz:
@failure-to-thrive Any chance this could be picked up again? Or do you reckon the interest of the community at large is insufficient for this to be a worthwhile pursuit? I, for one, would love to play around with it, to see if I can find improvements over the vanilla LSTM implementation.

failure-to-thrive:
> Or do you reckon the interest of the community at large is insufficient for this to be a worthwhile pursuit? 

Yes, while it's technically possible to remaster, there is an inclusion criteria outlined at the link above.

seanpmorgan:

PR #1878, opened 2020-05-25
fix docstring in focal loss

### LINKED ISSUE #1859: Documentation: Focal loss example misleading
In the example of the focal loss (`tfa.losses.SigmoidFocalCrossEntropy`) the label and the prediction data seems to be swapped. 
This is misleading as `sigmoid_focal_crossentropy()` expects `y_true, y_pred`, whereas the example presents `y_pred, y_true`.

https://github.com/tensorflow/addons/blob/efa0cd9c86578f53a2eb9d80849e186bffa4d64a/tensorflow_addons/losses/focal_loss.py#L42-L47

https://github.com/tensorflow/addons/blob/efa0cd9c86578f53a2eb9d80849e186bffa4d64a/tensorflow_addons/losses/focal_loss.py#L90-L92

autoih:
Closes #1859

### DIFF
--- a/tensorflow_addons/losses/focal_loss.py
+++ b/tensorflow_addons/losses/focal_loss.py
@@ -40,11 +40,11 @@ class SigmoidFocalCrossEntropy(LossFunctionWrapper):
     ```python
     fl = tfa.losses.SigmoidFocalCrossEntropy()
     loss = fl(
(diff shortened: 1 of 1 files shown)


AakashKumarNain:
Thanks @autoih

PR #1884, opened 2020-05-25
Update tutorials with TF2 Colab default

seanpmorgan:
Diffs might be a bit ugly. Hopefully clear in ReviewNB


seanpmorgan:
Ugh, diffs are brutal. Will break these up using GitHub

PR #1888, opened 2020-05-26
Publish dev container

### LINKED ISSUE #1414: Publish developer docker containers on dockerhub
See #1309

seanpmorgan:
Closes #1414

Discussion Points:
1. This adds a bit of computation time to the action on commit push. I don't think this is a big deal as this doesn't slow down feedback to contributors
2. There is no addons repository included in the docker image. I believe this is the correct pattern since we want users to load their own repository within the container
3. Because of 2 there is no pre-compiled bazel cache. We could add it but then the calls to bazel in the repository would need to be pointed to that cache. Given that our build only takes a couple of minutes I'd prefer to do this in another PR if we want to.

I tested the image in VScode using a dev-container and it worked as expected.


.github/workflows/build_dev_container.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -x -e
+
+df -h
+docker info
+# to get more disk space
+rm -rf /usr/share/dotnet &
```
bhack, line 8: Could this be a problem if users test the dev_container image builds process locally (i.e. PRs over dev_container recepits itself)?

bhack:
I am looking at the compressed size of devel image TF vs TFA:

https://hub.docker.com/r/tensorflow/tensorflow/tags
https://hub.docker.com/r/tfaddons/tensorflow/tags

The TF CPU devel is 1/3 of our devel image and we have 1GB of compressed overhead over tensorflow GPU devel (uncompressed our is 9.2GB).

bhack:
I think it could be useful to have a default small size 1GB image to lower the download overhead but I think that is still not possible to handle multiple configuration (GPU/CPU) in the dev-containers https://github.com/microsoft/vscode-remote-release/issues/2143#issuecomment-600087021.

The new devcontainer support PR is at https://github.com/microsoft/vscode-dev-containers/pull/347

bhack:
Could you add
```
sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-8 100
sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-8 100
```

bhack:
If you could set the alias also for `clang-format` it could be useful

seanpmorgan:
> I am looking at the compressed size of devel image TF vs TFA:
> 
> https://hub.docker.com/r/tensorflow/tensorflow/tags
> https://hub.docker.com/r/tfaddons/tensorflow/tags
> 
> The TF CPU devel is 1/3 of our devel image and we have 1GB of compressed overhead over tensorflow GPU devel (uncompressed our is 9.2GB).



> I think it could be useful to have a default small size 1GB image to lower the download overhead but I think that is still not possible to handle multiple configuration (GPU/CPU) in the dev-containers [microsoft/vscode-remote-release#2143 (comment)](https://github.com/microsoft/vscode-remote-release/issues/2143#issuecomment-600087021).

As discussed offline this is a consequence of https://github.com/tensorflow/tensorflow/issues/38352 . There is not much we can do without deviating from the custom-op recommended images.

.github/workflows/build_dev_container.sh
seanpmorgan: Yeah which is why I moved it into `.github/workflows` it is not meant to be built locally. We could make it so it calls a script in `tools/` but then the user would need download the base image which is almost as large so I don't see a lot of benefit

seanpmorgan:
> Could you add
> 
> ```
> sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-8 100
> sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-8 100
> ```

Sure, but could you expand on why its needed. All of our documentation uses gcc and nvcc to compile. I know there was work on getting TF to compile using cuda-clang but not sure if we want to encourage that?

bhack:
I've not checked in detail where we use what but I was thinking as a general rule to not going to ovveride to much settings properties in vsocde devcontainer about specific strings or to monitor if  we change clang version cause there we are on another repository. 
I prefer that we have the control in the image we are generating so we have constant alias.
If we don't care about clang probably we could do It only for clang-format?

bhack:
I don't know if was clear. But I prefer to not set all these properties  e.g. `"C_Cpp.clang_format_path"` to the specific `/usr/local/bin/clang-format-9`. 
If we use the default alias it is easier to maintain cause we don't need to set all the properties.

CONTRIBUTING.md
```diff
@@ -267,28 +267,22 @@ to install any additional tools.
...
 Install in editable mode
 ```
-python3 -m pip install -e .
-python3 -m pip install -r tools/install_deps/pytest.txt
+python -m pip install -e .
 ```
 
 Compile the custom ops
 ```
+export TF_NEED_CUDA=1 # If GPU is to be used
```
bhack, line 285: This seems that is set by default to 1 in the image/container. 
So if I follow CPU custom_op documentation it build cuda files also without exporting this env.

bhack: As we are using the `custom_ops` GPU image version as our baseline it is inherited by https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/Dockerfile.custom_op_gpu#L21

bhack:
> As discussed offline this is a consequence of [tensorflow/tensorflow#38352](https://github.com/tensorflow/tensorflow/issues/38352) . There is not much we can do without deviating from the custom-op recommended images.

I've changed `tools/docker/build_wheel.Dockerfile` using `FROM tensorflow/tensorflow:devel-gpu as base_install` and the `bazel build --enable_runfiles build_pip_pkg` TF addons compile command seems to run fine.

bhack:
I think that we need to update a little bit https://github.com/tensorflow/addons/blob/master/CONTRIBUTING.md after merging this. Can you open a ticket as a reminder?

bhack:
Need we still to run `bash tools/install_so_files.sh`?

CONTRIBUTING.md
seanpmorgan: Resolved since using the cpu custom-op image as base.

tools/docker/dev_container.Dockerfile
```diff
@@ -1,5 +1,12 @@
 #syntax=docker/dockerfile:1.1.5-experimental
-FROM addons_base:latest as build_dev_container
+FROM tensorflow/tensorflow:2.1.0-custom-op-ubuntu16 as dev_container_cpu
+ARG TF_PACKAGE
+ARG TF_VERSION
+
+# Temporary until custom-op container is updated
+RUN ln -sf /usr/bin/python3 /usr/bin/python
+RUN ln -sf /usr/local/bin/pip3 /usr/local/bin/pip
+RUN pip install --default-timeout=1000 $TF_PACKAGE==$TF_VERSION
```
bhack, line 9: With this we are duplicating the tensorflow install layer.  Do you think to remove it after https://github.com/tensorflow/tensorflow/pull/40189 or need we to duplicate it for other goals?

tools/install_deps/clang-format.sh
```diff
@@ -16,4 +16,5 @@
 
 
 wget -O /usr/local/bin/clang-format-9 https://github.com/DoozyX/clang-format-lint-action/raw/master/clang-format/clang-format9
-chmod +x /usr/local/bin/clang-format-9
\ No newline at end of file
+chmod +x /usr/local/bin/clang-format-9
+ln -s /usr/local/bin/clang-format-9 /usr/local/bin/clang-format
```
bhack, line 20: As you want. Finally I don't think that vscode intellisense is using the executable.

tools/docker/dev_container.Dockerfile
seanpmorgan: Yeah remove it after that is merged and a new container is published

tools/docker/dev_container.Dockerfile
```diff
@@ -0,0 +1,25 @@
+#syntax=docker/dockerfile:1.1.5-experimental
+FROM tensorflow/tensorflow:2.1.0-custom-op-ubuntu16 as dev_container_cpu
```
bhack, line 2: Here you lost the build ARGS to prepare the switch between cpu and gpu that was in https://github.com/seanpmorgan/addons/pull/14/files#diff-76fe3c36c889ccfcf62f6e35439d030fR3-R5

tools/build_dev_container.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -x -e
+
+DOCKER_BUILDKIT=1 docker build \
```
bhack, line 5: We could duplicate this command with the build_args to publish the GPU image. See the ARGS comment below.

.github/workflows/release.yml
```diff
@@ -97,7 +98,17 @@ jobs:
...
+    name: Upload dev container to DockerHub
+    needs: [release-wheel, test-with-bazel]
+    runs-on: ubuntu-18.04
+    if: (github.event_name == 'push' && github.ref == 'refs/heads/master')
+    steps:
+      - run: |
+          set -e -x
+          docker login --username ${{ secrets.DOCKER_USER }} --password ${{ secrets.DOCKER_PW }}
+          bash bash .github/workflows/github_build_dev_container.sh
+          docker push tfaddons/dev_container:latest-cpu
```
bhack, line 114: We could duplicate this command to push the `latest` without suffixs (the is the GPU image)

bhack:
As a reminder if you merge https://github.com/tensorflow/addons/pull/1808/ you need to add the requirements here.

CONTRIBUTING.md
```diff
@@ -289,6 +289,7 @@ python3 -m pip install -r tools/install_deps/pytest.txt
...
```
bhack, line 292: I think that we need to set this only in the GPU image

bhack:
`custom-ops` images are not produced for RC https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/release/ubuntu_16/custom_op/release.sh#L39

CONTRIBUTING.md
seanpmorgan: Yes, but also if it's built from source locally.

seanpmorgan:
@bhack @tensorflow/sig-addons-maintainers so there are still a few sharp edges here, but it aligns with the recent merge of SIG-IO:
https://github.com/tensorflow/io/issues/1087

I would prefer that we merge this semi-hardcoded cpu devel build and then I'll open an issue and start work on a seperate GPU build. This got a bit stale and it'd be easier to get this merged and then upgrade it IMO. Also waiting for clarity from SIG-Build about custom-op images and their future. Does that seem okay?

tools/docker/dev_container.Dockerfile
bhack: Ok I think that this for cpu could be removed now

bhack (approved):

PR #1899, opened 2020-05-29
typo

feyn-aman:
changed hindi to german


feyn-aman:
@googlebot I signed it!

PR #1901, opened 2020-05-31
fix - in multhead_attention, correct the location where key & value shapes are checked

ksachdeva:
at present this check is done only if mask is not None whereas this should be checked irrespective of the msk

### DIFF
--- a/tensorflow_addons/layers/multihead_attention.py
+++ b/tensorflow_addons/layers/multihead_attention.py
@@ -189,6 +189,11 @@ class MultiHeadAttention(tf.keras.layers.Layer):
         key = inputs[1]
         value = inputs[2] if len(inputs) > 2 else key
 
+        if key.shape[-2] != value.shape[-2]:
+            raise ValueError(
+                "the number of elements in 'key' must be equal to the same as the number of elements in 'value'"
+            )
+
         # verify shapes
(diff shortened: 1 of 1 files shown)


tensorflow_addons/layers/multihead_attention.py
```diff
@@ -189,6 +189,11 @@ def call(self, inputs, training=None, mask=None):
         key = inputs[1]
         value = inputs[2] if len(inputs) > 2 else key
 
+        if key.shape[-2] != value.shape[-2]:
+            raise ValueError(
+                "the number of elements in 'key' must be equal to the same as the number of elements in 'value'"
+            )
+
         # verify shapes
```
WindQAQ, line 197: we might also want to move this comment upwards

ksachdeva: done

ksachdeva:
Hi,
I did run the checks as recommended and even had the hoot installed as per the instructions. Not sure what is wrong here.

WindQAQ:
> Hi,
> I did run the checks as recommended and even had the hoot installed as per the instructions. Not sure what is wrong here.

No worries. GitHub CI sometimes fails due to network issue. Let me try to re-trigger it again.

PR #1909, opened 2020-06-03
Handle the so loading in dev mode

bhack:


tools/testing/build_and_run_tests.sh
```diff
@@ -34,5 +34,5 @@ if ! [ -x "$(command -v nvidia-smi)" ]; then
   EXTRA_ARGS="-n auto"
 fi
 
-
+bazel clean
```
bhack, line 37: @seanpmorgan I want to remove this but if I don't remove this it find in the cache some so that I don't undertands:

```
tensorflow.python.framework.errors_impl.NotFoundError: /addons/bazel-bin/tensorflow_addons/custom_ops/layers/lib_correlation_cost_ops_gpu.so: undefined symbol: _ZNK10tensorflow6Tensor21CheckTypeAndIsAlignedENS_8DataTypeE
```

In the local build I don't find any `lib_correlation_cost_ops_gpu.so`

bhack:
Just as a reminder note:
I don't know how this PR works in `windows` and `osx` build release wheels. 
But probably if it works there we could try to totally remove `https://github.com/tensorflow/addons/blob/master/tools/install_so_files.sh`

tools/run_gpu_tests.sh
```diff
@@ -9,4 +9,4 @@ docker build \
        --build-arg TF_VERSION=2.2.0 \
        --build-arg PY_VERSION=3.5 \
        -t tfa_gpu_tests ./
-docker run --rm -t -v cache_bazel:/root/.cache/bazel --gpus=all tfa_gpu_tests
+docker run --rm -t --gpus=all tfa_gpu_tests
```
bhack, line 12: It is working so I don't know if anybody remember why the bazel cache volume was required here but we could revert this change.

gabrieldemarmiesse:
Could you give a bit of background to help us understand what you want to enable with this pull request?

bhack:
@seanpmorgan knows a little bit more the context of this cause it is partially related to https://github.com/tensorflow/addons/pull/1888 but it doesn't depend on that.

When you develop in a container env generally and/or you are integrating in the IDO you don't want to have to much manual steps on bash scripts or have in you source tree files produced in the container environment like when you mount in the container your `tf_addons` checkout volume on the host and you need to run internally `install_so_files.sh` in the container.

With this we are trying to detect if we are in "developer mode"  so that we don't need to continuously manually run `install_so_files.sh` when we are developing on custom_op cause it will enable so loading from the bazel-bin. With that you will also not have `.so` file in the source tree on the host when you exit from the container.

seanpmorgan (approved):
LGTM Thanks. Apologies for the very slow review. I was debating if we could avoid users paying this dev tool directory lookup, but it's minimal and done only when a custom op is called for the first time.

PR #1914, opened 2020-06-07
use raw ops

### LINKED ISSUE #1779: Deprecate ImageProjectiveTransformV2
`ImageProjectiveTransformV2` seems to be integrated into core TF at https://github.com/tensorflow/tensorflow/commit/26992815bf7ee136c49f658596a9fbde7881d425 in Jan. We have to deprecate the use of it.

https://github.com/tensorflow/addons/blob/master/tensorflow_addons/custom_ops/image/cc/kernels/image_projective_transform_op.cc

Also, it was unclear to me that `ImageProjectiveTransformV2` has been integrated into TF many months ago. Maybe we should check new stuff in TF regularly. Thanks @bhack for pointing out!

fsx950223:
Fix #1779.
cc @WindQAQ 
Related https://github.com/google/automl/issues/480

### DIFF
--- a/tensorflow_addons/image/transform_ops.py
+++ b/tensorflow_addons/image/transform_ops.py
@@ -16,13 +16,11 @@
 
 import tensorflow as tf
 from tensorflow_addons.image import utils as img_utils
-from tensorflow_addons.utils.resource_loader import LazySO
 from tensorflow_addons.utils.types import TensorLike
 from tensorflow_addons.image.utils import wrap, unwrap
 
 from typing import Optional
 
-_image_so = LazySO("custom_ops/image/_image_ops.so")
 
 _IMAGE_DTYPES = {
     tf.dtypes.uint8,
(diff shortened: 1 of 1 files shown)


fsx950223:
cc @gabrieldemarmiesse.

seanpmorgan:
Related: https://github.com/tensorflow/addons/issues/1779

fsx950223:
> Thank you @fsx950223! I believe this is correct, but the serialization/deserialization might break the saved models (for whom also save the input pipeline) due to different registration name.
> 
> Hi @seanpmorgan @gabrieldemarmiesse, should we throw some warning to tell users the implementation has changed? Thank you!

It won't influence saved model. Since saved model needs to load custom kernel separately.

gabrieldemarmiesse:
We have a function to load all SO, so as long as we keep the SO somewhere, we won't break existing models. This PR doesn't break existing saved models but it will remove the gradient for the op though

fsx950223:
Because the PR fix AA on TPUs, I believe the above issues shouldn't block it.

PR #1918, opened 2020-06-08
super_serial: automate saving and restoring tfrecords 

markemus:
This module saves Datasets as TFRecords files alongside a .header file containing the metadata for reconstructing the Dataset. Users only need to call `save(tfrecordpath, headerpath)` and `load(tfrecordpath, headerpath)`. It really is that easy.

Currently using TFRecords in Tensorflow has a steep learning curve, it can be difficult to write code for complex datasets and it requires you to keep the code needed to read the TFRecord back into memory. Super_serial eliminates these headaches entirely. 

Includes a test to demonstrate how it works.

### DIFF
--- a/tensorflow_addons/super_serial/super_serial.py
+++ b/tensorflow_addons/super_serial/super_serial.py
@@ -1,0 +1,177 @@
+"""Easily save tf.data.Datasets as tfrecord files, and restore tfrecords as Datasets.
+
+The goal of this module is to create a SIMPLE api to tfrecords that can be used without
+learning all of the underlying mechanics.
+
+Users only need to deal with 2 functions:
+save(dataset)
+dataset = load(tfrecord, header)
+
+To make this work, we create a .header file for each tfrecord which encodes metadata
+needed to reconstruct the original dataset.
+
+Saving must be done in eager mode, but loading is compatible with both eager and
+graph execution modes.
+
+GOTCHAS:
+- This module is only compatible with "dictionary-style" datasets {key: val, key2:val2,..., keyN: valN}.
+- The restored dataset will have the TFRecord dtypes {float32, int64, string} instead of the original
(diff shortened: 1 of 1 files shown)


bhack:
Is this related to https://github.com/tensorflow/tensorflow/issues/38483?

markemus:
Yes, this is a solution to that problem.

bhack:
My impression is that this could have a better fit in [Tensorflow IO](https://github.com/tensorflow/io) /cc @yongtang @jsimsa

seanpmorgan:
> My impression is that this could have a better fit in [Tensorflow IO](https://github.com/tensorflow/io) /cc @yongtang @jsimsa

Agree. Thanks for the PR @markemus but this seems to scope better in tf/io. You may want to open an issue there to see if they'd like to add it to their repository!

markemus:
Will do, thanks for the quick feedback everyone.

jsimsa:
I am actually working on a PR that will provide support for `save` and `load` of datasets. It should be available later this month (and before TF 2.3).

markemus:
Good to hear! This functionality really should have been built into the TFRecord API to begin with. They're a nightmare to work with directly atm.

bhack:
@markemus The origin of this was quite old see tensorflow/tensorflow#16926. 
Thanks @jsimsa for disclosing your roadmap we care about third party contributor time it is so precious :wink:  (@martinwicke)

markemus:
Two years later there's still no good way to save and restore TFRecords in Tensorflow. This PR was accepted and has been included in tf-io for over a year now (and also improved, with tests and additional support added). 

https://github.com/tensorflow/io/pull/1280

Any chance you folks have changed your mind about adding it to tensorflow proper? I have used this in multiple real projects and so have some of my coworkers. It's easy and it works well, and the header is stored separately so old reading approaches will still work. I'm also happy to add more features if required.

PR #1919, opened 2020-06-09
Update context checks in dynamic_decode

guillaumekln:
There are several things here:

- The XLA check is updated to match what is done in `tf.keras.backend.rnn` and also mentionned in https://github.com/tensorflow/tensorflow/commit/e67e29e3c8a150be847e17bbfe4f5d399e7b08ea

- As far as I know, setting a caching device on the variable scope only impacts variables created with `tf.compat.v1.get_variable` which is no longer used in V2. The current check was wrong in V2 anyway (see https://github.com/tensorflow/tensorflow/commit/e67e29e3c8a150be847e17bbfe4f5d399e7b08ea)

- If we don't need to configure the variable scope, we can change it to a name scope which has the same effect in V2.

### DIFF
--- a/tensorflow_addons/seq2seq/decoder.py
+++ b/tensorflow_addons/seq2seq/decoder.py
@@ -291,7 +291,7 @@ def dynamic_decode(
       training: Python boolean. Indicates whether the layer should behave
           in training  mode or in inference mode. Only relevant
           when `dropout` or `recurrent_dropout` is used.
-      scope: Optional variable scope to use.
+      scope: Optional name scope to use.
       **kwargs: dict, other keyword arguments for dynamic_decode. It might
         contain arguments for `BaseDecoder` to initialize, which takes all
         tensor inputs during call().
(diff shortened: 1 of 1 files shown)


tensorflow_addons/seq2seq/decoder.py
```diff
@@ -302,26 +302,19 @@ def dynamic_decode(
...
-        # Don't set a caching device when running in a loop, since it is
-        # possible that train steps could be wrapped in a tf.while_loop. In that
-        # scenario caching prevents forward computations in loop iterations from
-        # re-reading the updated weights.
-        if not tf.executing_eagerly() and not in_while_loop:
-            if varscope.caching_device is None:
-                varscope.set_caching_device(lambda op: op.device)
+    with tf.name_scope(scope or "decoder"):
+        is_xla = not tf.executing_eagerly() and control_flow_util.GraphOrParentsInXlaContext(
+            tf.compat.v1.get_default_graph()
```
qlzh727, line 307: I think you are missing _get_control_flow_context() for get_default_graph()

tensorflow_addons/seq2seq/decoder.py
```diff
@@ -302,26 +302,19 @@ def dynamic_decode(
...
-        ctxt = tf.compat.v1.get_default_graph()._get_control_flow_context()
-        is_xla = control_flow_util.GetContainingXLAContext(ctxt) is not None
-        in_while_loop = control_flow_util.GetContainingWhileContext(ctxt) is not None
-        # Properly cache variable values inside the while_loop.
-        # Don't set a caching device when running in a loop, since it is
-        # possible that train steps could be wrapped in a tf.while_loop. In that
-        # scenario caching prevents forward computations in loop iterations from
-        # re-reading the updated weights.
-        if not tf.executing_eagerly() and not in_while_loop:
-            if varscope.caching_device is None:
```
qlzh727, line 316: The caching device will have a big performance boost if the code is run in a remote parameter server and worker setting, so that the variable is not read for every iteration of while loop. Is there any reason to remove it?

qlzh727: Ah, you are right. The code is only expect to work in tf v1. Removing it for v2 should be fine.

tensorflow_addons/seq2seq/decoder.py
guillaumekln: `control_flow_util.GraphOrParentsInXlaContext` expects a graph. See an example usage here:

https://github.com/tensorflow/tensorflow/blob/v2.2.0/tensorflow/python/keras/backend.py#L4106

PR #1921, opened 2020-06-09
Use a tf.function with an input signature instead of tf.placeholder

guillaumekln:
This is more in line with how these functions would actually be called.


AakashKumarNain:
Thanks

Issue #1920, opened 2020-06-09
LookAhead + RAdam + mixed_fp16 = apply_gradients() got an unexpected keyword argument 'experimental_aggregate_gradients'


shkarupa-alex:
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Ubuntu 20.04
- TensorFlow version and how it was installed (source or binary): source
- TensorFlow-Addons version and how it was installed (source or binary): source
- Python version: 3.8.2
- Is GPU used? (yes/no): yes

**Describe the bug**
When i set global mixed precision policy `tf.keras.mixed_precision.experimental.set_policy('mixed_float16')` meta-optimizer LookAhead + RAdam raises error

```python
Epoch 1/5
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-19-841302a83165> in <module>
----> 1 model.fit(
      2     train_ds,
      3     epochs=1 if lr_finder else 5,
      4     callbacks=callbacks,
      5     steps_per_epoch=findlr_steps if lr_finder else None,

~/.local/lib/python3.8/site-packages/tensorflow/python/keras/engine/training.py in _method_wrapper(self, *args, **kwargs)
     64   def _method_wrapper(self, *args, **kwargs):
     65     if not self._in_multi_worker_mode():  # pylint: disable=protected-access
---> 66       return method(self, *args, **kwargs)
     67 
     68     # Running inside `run_distribute_coordinator` already.

~/.local/lib/python3.8/site-packages/tensorflow/python/keras/engine/training.py in fit(self, x, y, batch_size, epochs, verbose, callbacks, validation_split, validation_data, shuffle, class_weight, sample_weight, initial_epoch, steps_per_epoch, validation_steps, validation_batch_size, validation_freq, max_queue_size, workers, use_multiprocessing)
    846                 batch_size=batch_size):
    847               callbacks.on_train_batch_begin(step)
--> 848               tmp_logs = train_function(iterator)
    849               # Catch OutOfRangeError for Datasets of unknown size.
    850               # This blocks until the batch has finished executing.

~/.local/lib/python3.8/site-packages/tensorflow/python/eager/def_function.py in __call__(self, *args, **kwds)
    578         xla_context.Exit()
    579     else:
--> 580       result = self._call(*args, **kwds)
    581 
    582     if tracing_count == self._get_tracing_count():

[...]

PR #1924, opened 2020-06-12
Add experimental_aggregate_gradients support

bhack:
Try to fix https://github.com/tensorflow/addons/issues/1920

### DIFF
--- a/tensorflow_addons/optimizers/tests/lookahead_test.py
+++ b/tensorflow_addons/optimizers/tests/lookahead_test.py
@@ -18,7 +18,7 @@ import numpy as np
 import pytest
 import tensorflow as tf
 
-from tensorflow_addons.optimizers import Lookahead
+from tensorflow_addons.optimizers import Lookahead, RectifiedAdam
 
 
 def run_dense_sample(iterations, optimizer, seed=0x2019):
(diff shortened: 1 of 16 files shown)


bhack:
Let me know if it makes sense to add this test:

```
def test_fit_simple_linear_model_mixed_precision():
    np.random.seed(0x2019)
    tf.random.set_seed(0x2019)

    x = np.random.standard_normal((10000, 3))
    w = np.random.standard_normal((3, 1))
    y = np.dot(x, w) + np.random.standard_normal((10000, 1)) * 1e-4

    tf.keras.mixed_precision.experimental.set_policy("mixed_float16")
    model.add(tf.keras.layers.Dense(input_shape=(3,), units=1))
    model.compile(Lookahead("sgd"), loss="mse")

    model.fit(x, y, epochs=3)

    x = np.random.standard_normal((100, 3))
    y = np.dot(x, w)
    predicted = model.predict(x)

    max_abs_diff = np.max(np.abs(predicted - y))
    assert max_abs_diff < 2.3e-3
```

tensorflow_addons/optimizers/tests/lookahead_test.py
```diff
@@ -19,6 +19,7 @@
 import tensorflow as tf
 
 from tensorflow_addons.optimizers import Lookahead
+from tensorflow.python.keras.mixed_precision.experimental import policy
```
bhack, line 22: One of the check is failing cause our policy doesn't let us to import this namespace but `policy_scope` is not exposed and in many parts there are `thread_local`  todos like in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/mixed_precision/experimental/policy.py#L484

bhack: See also https://github.com/tensorflow/tensorflow/pull/40475

seanpmorgan: Seeing as we have no public API to make this thread local, I'm in favor of including this test case and adding the file to the list of allowed private usage:
https://github.com/tensorflow/addons/blob/master/tools/testing/source_code_test.py#L86

This has been confirmed by the user to solve the problem, and the private usage is not called in the optimizer source code, only the test case. Seems counterproductive to not include the test. If https://github.com/tensorflow/tensorflow/pull/40475 or some variant merges we can remove it. @gabrieldemarmiesse thoughts?

tensorflow_addons/optimizers/tests/lookahead_test.py
```diff
@@ -120,6 +121,30 @@ def test_fit_simple_linear_model():
...
+
+    model.fit(x, y, epochs=3)
+
+    x = np.random.standard_normal((100, 3))
+    y = np.dot(x, w)
+    predicted = model.predict(x)
+
+    max_abs_diff = np.max(np.abs(predicted - y))
+    assert max_abs_diff < 2.3e-3
+    assert max_abs_diff >= 1e-3
```
seanpmorgan, line 145: Why assert a greater than difference here? Mixed precision variation?

bhack: It Is what I think. There Is a default dynamic loss scaling to avoid numerical precision underrun https://www.tensorflow.org/api_docs/python/tf/keras/mixed_precision/experimental/Policy?version=nightly

PR #1925, opened 2020-06-12
Add a Python alternative to seq2seq.gather_tree

guillaumekln:
This PR adds a pure Python TensorFlow implementation of `tfa.seq2seq.gather_tree`, which can be enabled with the global flag [`TF_ADDONS_PY_OP`](https://github.com/tensorflow/addons#gpu-and-cpu-custom-ops-1). This is useful when Addons custom ops are not readily available (e.g. TensorFlow Serving).

I tested the performance on a real world application: beam search decoding of a neural machine translation model. I'm using a [custom beam search implementation](https://github.com/OpenNMT/OpenNMT-tf/blob/master/opennmt/utils/decoding.py) but it is close to the `BeamSearchDecoder` included in Addons.

I did not find significant performance impact.  And I think this makes sense: the function is only used at the very end of the decoding and does not involve very complex ops.

### DIFF
--- a/tensorflow_addons/seq2seq/BUILD
+++ b/tensorflow_addons/seq2seq/BUILD
@@ -6,8 +6,8 @@ py_library(
     name = "seq2seq",
     srcs = glob(["*.py"]),
     data = [
-        "//tensorflow_addons/custom_ops/seq2seq:_beam_search_ops.so",
         "//tensorflow_addons:options.py",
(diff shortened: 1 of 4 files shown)


guillaumekln:
Marking this PR as draft. The fixture from #1929 is needed to run tests for both the custom op and the py op.

guillaumekln:
Tests are now running for this new implementation.

PR #1926, opened 2020-06-12
Enable Python alternatives when running tests with --skip-custom-ops

guillaumekln:


guillaumekln:
Any idea why `from tensorflow_addons import options` is failing in Bazel tests? I also got that in my other PR.

tensorflow_addons/utils/test_utils.py
```diff
@@ -21,6 +21,7 @@
 import pytest
 import tensorflow as tf
 
+from tensorflow_addons import options
```
bhack, line 24: if you use https://github.com/tensorflow/addons/blob/master/tensorflow_addons/options.py#L36-L40 seems it works.

bhack:
To import `options` in bazel you need to add `"//tensorflow_addons:options.py",` in https://github.com/tensorflow/addons/blob/master/tensorflow_addons/utils/BUILD#L8

guillaumekln:
Ah thanks @bhack, I forgot about that.

I'm going to close this PR though, we might want to do it differently.

PR #1928, opened 2020-06-14
Fix typo in docstring of CohenKappa __init__

xvr-hlt:
As above.


xvr-hlt:
@googlebot I signed it!

AakashKumarNain:
Thank you for the fix

PR #1941, opened 2020-06-19
Fix docstring with incorrect API usage

### LINKED ISSUE #1936: Usage with tf.keras API
https://github.com/tensorflow/addons/blob/5f618fdb92d9737da059de2a33fa606e97505398/tensorflow_addons/losses/focal_loss.py#L52-L53

The usage in `tf.keras` API example is incorrect. It should be replaced with:

```python
model = tf.keras.Model(inputs, outputs)
model.compile('sgd', loss=tfa.losses.SigmoidFocalCrossEntropy())
```

Squadrick:
Fix #1936


AakashKumarNain:
Thanks @Squadrick for the fix. The build is failing but it is not related to this. Thinking if I should wait for that to be fixed first

Issue #1942, opened 2020-06-19
Failed GPU optimizer test with mixed precision


WindQAQ:
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): CI
- TensorFlow version and how it was installed (source or binary): CI
- TensorFlow-Addons version and how it was installed (source or binary): CI
- Python version: CI
- Is GPU used? (yes/no): yes

**Describe the bug**
GPU test is broken after https://github.com/tensorflow/addons/commit/bd5bbfc66e750cba94360236e7f85481662ba4de.

https://source.cloud.google.com/results/invocations/147c0642-7973-429b-87b1-b8f1b1d2cb14/log

WindQAQ:
cc @bhack

bhack:
https://github.com/tensorflow/tensorflow/issues/39775

bhack:
In the maintime check if we could bypass with https://github.com/tensorflow/addons/pull/1943

bhack:
We can close this and just add https://github.com/tensorflow/addons/pull/1951#issuecomment-651004300

seanpmorgan:
> We can close this and just add [#1951 (comment)](https://github.com/tensorflow/addons/pull/1951#issuecomment-651004300)

Agree, this appears to be solved in upstream for 2.3

PR #1943, opened 2020-06-19
temp bypass gpu test mixed_precision

bhack:


bhack:
Workaround Fixes https://github.com/tensorflow/addons/issues/1942

PR #1951, opened 2020-06-28
Add TF 2.3.0rc to testing

seanpmorgan:


bhack:
Can you add a version check on this line https://github.com/tensorflow/addons/blob/master/tensorflow_addons/optimizers/tests/lookahead_test.py#L125?

seanpmorgan:
Looks like there's an error when compiling Windows + 2.3rc0

bhack:
Can we have more details on the bulid?

seanpmorgan:
Not really since this is appears to be a bash segfault. Trying to replicate it on my local windows box and I as of now I believe this to be the issue: 
https://github.com/bazelbuild/bazel/issues/3636

Will dig into how we can avoid this.

seanpmorgan:
Welp... spent way too much time working through that for a relatively straight forward fix. @tensorflow/sig-addons-maintainers ready for review.

build_deps/tf_dependency/tf_configure.bzl
```diff
@@ -185,7 +185,7 @@ def _symlink_genrule_for_dir(
 
     genrule = _genrule(
         genrule_name,
-        " && ".join(command),
+        ";\n".join(command),
```
seanpmorgan, line 188: Bash on windows and macos have line length limits (4964 and 7937 respectively). TF2.3 added more include files so we finally were pushed over the cliff and our bash shell segfaulted with a very uninformative error.

Quite lucky this was [investigated by the Bazel team a long time ago](https://github.com/bazelbuild/bazel/issues/3636), since it's really not related to Bazel (other than the way genrules expect each copied element to be specified).

WindQAQ:
BTW, can we also upgradte bazel version in this (or next) PR? We are quite lag behind TF, which uses 3.1.0 in TF 2.3.0rc* and 2.0.0 in TF 2.2.*

seanpmorgan:
> BTW, can we also upgradte bazel version in this (or next) PR? We are quite lag behind TF, which uses 3.1.0 in TF 2.3.0rc* and 2.0.0 in TF 2.2.*

Should be no issue with 3.1.0 bazel and RC1-- good catch. Will see how the tests run

WindQAQ:
Seems good :-) But there are two files staying at version 1.1.0 haha.

https://github.com/tensorflow/addons/blob/master/tools/install_deps/bazel_linux.sh#L6
https://github.com/tensorflow/addons/blob/master/.github/workflows/ci_test.yml#L14

Issue #1962, opened 2020-07-02
CohenKappa raise exception for batch_size=1


brunodoamaral:
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Centos 7
- TensorFlow version and how it was installed (source or binary): 2.2.0
- TensorFlow-Addons version and how it was installed (source or binary): 0.10.0
- Python version: 3.6.9
- Is GPU used? (yes/no): no

**Describe the bug**

`tfa.metrics.CohenKappa` raise exception when batch_size=1.

**Code to reproduce the issue**

```python
import tensorflow as tf
import tensorflow_addons as tfa

kappa = tfa.metrics.CohenKappa(5, regression=True, sparse_labels=True)

# Batch-size > 1: this will not raise an exception
kappa.update_state(tf.ones(2), tf.ones(2))

# Batch-size = 1: this will raise an exception
kappa.update_state(tf.ones(1), tf.ones(1))
```

**Other info / logs**

The issue is due to `tf.squeeze` being use in `_update_confusion_matrix`:
https://github.com/tensorflow/addons/blob/5f618fdb92d9737da059de2a33fa606e97505398/tensorflow_addons/metrics/cohens_kappa.py#L170-L171

When the tensor has max shape equals to 1, like in `(1,)`, all dims are removed and a single scalar is returned by `tf.squeeze`. A subsequent call to `tf.math.confusion_matrix` will raise the exception while trying to stack tensors over an invalid axis.

```
ValueError                                Traceback (most recent call last)
<ipython-input-12-7ea8164c36b0> in <module>
      9
     10 # Batch-size = 1: this will raise an exception due to tf.squeeze
---> 11 kappa.update_state(tf.ones(1), tf.ones(1))

/usr/local/lib/python3.6/dist-packages/tensorflow/python/keras/utils/metrics_utils.py in decorated(metric_obj, *args, **kwargs)
     88
     89     with tf_utils.graph_context_for_symbolic_tensors(*args, **kwargs):
---> 90       update_op = update_state_fn(*args, **kwargs)
     91     if update_op is not None:  # update_op will be None in eager execution.
     92       metric_obj.add_update(update_op)

/usr/local/lib/python3.6/dist-packages/tensorflow_addons/metrics/cohens_kappa.py in update_state(self, y_true, y_pred, sample_weight)
    137           Update op.
    138         """
--> 139         return self._update(y_true, y_pred, sample_weight)
    140
[...]

PR #1963, opened 2020-07-02
Fixes exception when batch_size=1

brunodoamaral:
Here is a proposal to fix issue #1962

### DIFF
--- a/tensorflow_addons/metrics/cohens_kappa.py
+++ b/tensorflow_addons/metrics/cohens_kappa.py
@@ -171,6 +171,12 @@ class CohenKappa(Metric):
         y_true = tf.squeeze(y_true)
         y_pred = tf.squeeze(y_pred)
 
+        if tf.rank(y_true) == 0:
+            y_true = tf.expand_dims(y_true, 0)
+
+        if tf.rank(y_pred) == 0:
+            y_pred = tf.expand_dims(y_pred, 0)
+        
         new_conf_mtx = tf.math.confusion_matrix(
             labels=y_true,
             predictions=y_pred,
--- a/tensorflow_addons/metrics/cohens_kappa.py
+++ b/tensorflow_addons/metrics/cohens_kappa.py
@@ -171,11 +171,8 @@ class CohenKappa(Metric):
         y_true = tf.squeeze(y_true)
         y_pred = tf.squeeze(y_pred)
 
-        if tf.rank(y_true) == 0:
-            y_true = tf.expand_dims(y_true, 0)
-
-        if tf.rank(y_pred) == 0:
-            y_pred = tf.expand_dims(y_pred, 0)
+        y_true = tf.cond(tf.rank(y_true) == 0, lambda: tf.expand_dims(y_true, 0), lambda: y_true)
+        y_pred = tf.cond(tf.rank(y_pred) == 0, lambda: tf.expand_dims(y_pred, 0), lambda: y_pred)
         
         new_conf_mtx = tf.math.confusion_matrix(
             labels=y_true,
--- a/tensorflow_addons/metrics/cohens_kappa.py
+++ b/tensorflow_addons/metrics/cohens_kappa.py
@@ -167,16 +167,19 @@ class CohenKappa(Metric):
             y_pred = tf.cast(y_pred, dtype=tf.int64)
         return y_pred
 
-    def _update_confusion_matrix(self, y_true, y_pred, sample_weight):
-        y_true = tf.squeeze(y_true)
-        y_pred = tf.squeeze(y_pred)
+    @tf.function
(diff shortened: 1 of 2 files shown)


brunodoamaral:
@WindQAQ thanks for making clear the pre-commit process. I just included a test case for issue #1962.

brunodoamaral:
@googlebot  I fixed it.

tensorflow_addons/metrics/tests/cohens_kappa_test.py
```diff
@@ -237,3 +237,13 @@ def test_cohen_kappa_serialization():
...
+
+def test_cohen_kappa_single_batch():
+    # Test for issue #1962
+    obj = CohenKappa(5, regression=True, sparse_labels=True)
+
+    try:
+        # Test single batch update
+        obj.update_state(tf.ones(1), tf.ones(1))
+    except:
+        pytest.fail("Unexpected exception")
```
WindQAQ, line 249: Why we have to use `try` and `except` here? It should work after the fix. So it might be something like
```python
def test_cohen_kappa_single_batch():
    obj = CohenKappa(5, regression=True, sparse_labels=True)
    obj.update_state(tf.ones(1), tf.ones(1))
    np.testing.assert_allclose(xxx, obj.result())
```
where `xxx` is the numeric result.

tensorflow_addons/metrics/cohens_kappa.py
```diff
@@ -171,6 +171,9 @@ def _update_confusion_matrix(self, y_true, y_pred, sample_weight):
         y_true = tf.squeeze(y_true)
         y_pred = tf.squeeze(y_pred)
 
+        y_true = tf.cond(tf.rank(y_true) == 0, lambda: tf.expand_dims(y_true, 0), lambda: y_true)
```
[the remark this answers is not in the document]
WindQAQ: Thanks for pointing out! I am lacking of the context for that discussion.

tensorflow_addons/metrics/cohens_kappa.py
```diff
@@ -171,6 +171,13 @@ def _update_confusion_matrix(self, y_true, y_pred, sample_weight):
         y_true = tf.squeeze(y_true)
         y_pred = tf.squeeze(y_pred)
 
+        y_true = tf.cond(
+            tf.rank(y_true) == 0, lambda: tf.expand_dims(y_true, 0), lambda: y_true
+        )
+        y_pred = tf.cond(
+            tf.rank(y_pred) == 0, lambda: tf.expand_dims(y_pred, 0), lambda: y_pred
+        )
```
WindQAQ, line 179: Please use
```python
if tf.rank(y_true) == 0:
    y_true = tf.expand_dims(y_true, 0)
if tf.rank(y_pred) == 0:
    y_pred = tf.expand_dims(y_pred, 0)
```
instead.

tensorflow_addons/metrics/tests/cohens_kappa_test.py
brunodoamaral: Ok. I'll take this approach. I added the assertion value check in my last commit, but for a "perfect" Kappa, it return `nan` (I double-checked if this was an issue, but even sklearn implementation has the same result, due to denominator being equal to zero).
The solution was to call `obj.update_state(tf.ones(1), tf.zeros(1))` instead, which results in zero kappa value.

tensorflow_addons/metrics/cohens_kappa.py
brunodoamaral: Actually this was my first version, but it gave me the error:

`OperatorNotAllowedInGraphError: using a tf.Tensor as a Python bool is not allowed: AutoGraph did not convert this function. Try decorating it directly with @tf.function.`

That's why I changed to `tf.cond` implementation, which is has been prohibited.

My last commit address both requirements, using a `@tf.function`:
```python
    @tf.function
    def _safe_squeeze(self, y):
        y = tf.squeeze(y)

        # Check for scalar result
        if tf.rank(y) == 0:
            y = tf.expand_dims(y, 0)

        return y

    def _update_confusion_matrix(self, y_true, y_pred, sample_weight):
        y_true = self._safe_squeeze(y_true)
        y_pred = self._safe_squeeze(y_pred)

        # ...
```

AakashKumarNain:
LGTM but let @bhack and @WindQAQ take a final look at it before we merge.

bhack:
I've not followed the original issue. Was it simlar to https://github.com/tensorflow/models/pull/7478?

WindQAQ:
> I've not followed the original issue. Was it simlar to [tensorflow/models#7478](https://github.com/tensorflow/models/pull/7478)?

Yes it is.

PR #1964, opened 2020-07-02
Make `dynamic_decode()` TFLite-convertible

jaeyoo:
This PR adds some logics to make `dynamic_decoder()` TFLite-convertible.

Hi. I am Jae from TensorFlow Lite team.
Because TFLite doesn't support dynamic shape of `TensorArray` yet, it changes the `TensorArray`'s into static shape.
Also, 2-D `TensorArray` is not supported, so some reshaping at the input & output tensors is required.
Because TFLite assumes `batch_size = 1` at the inference, we can simply implement the above.

Thank you.

### DIFF
--- a/tensorflow_addons/seq2seq/decoder.py
+++ b/tensorflow_addons/seq2seq/decoder.py
@@ -340,15 +340,18 @@ def dynamic_decode(
             # Assume the batch_size = 1 for inference.
             # So we can change 2-D TensorArray into 1-D by reshaping it.
             zero_outputs = tf.nest.map_structure(
-                lambda shape, dtype: tf.reshape(tf.zeros(
-                    _prepend_batch(decoder.batch_size, shape), dtype=dtype), [-1]),
+                lambda shape, dtype: tf.reshape(
+                    tf.zeros(_prepend_batch(decoder.batch_size, shape), dtype=dtype),
+                    [-1],
+                ),
                 decoder.output_size,
                 decoder.output_dtype,
             )
         else:
             zero_outputs = tf.nest.map_structure(
                 lambda shape, dtype: tf.zeros(
-                    _prepend_batch(decoder.batch_size, shape), dtype=dtype),
+                    _prepend_batch(decoder.batch_size, shape), dtype=dtype
+                ),
                 decoder.output_size,
                 decoder.output_dtype,
             )
@@ -534,6 +537,7 @@ def dynamic_decode(
                 # Reshape the output to the original shape.
                 def _restore_batch(x):
                     return tf.expand_dims(x, [1])
+
                 final_outputs = tf.nest.map_structure(_restore_batch, final_outputs)
 
             final_outputs = tf.nest.map_structure(_transpose_batch_time, final_outputs)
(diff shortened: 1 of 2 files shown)


bhack:
Is this documented somewhere in tflite?

jaeyoo:
> Is this documented somewhere in tflite?

The documentation is in google internal issue tracker. We are preparing for public documentation as a TF blog post. Please stay tuned. Thank you for your interest.

guillaumekln:
Thanks for the change. Is there a way to add a unit test running the TF Lite converter on this function?

jaeyoo:
> Thanks for the change. Is there a way to add a unit test running the TF Lite converter on this function?

I think the unit tests can be added after the new tf is released. Currently, this function is only working with tf-nightly. In this case, what should I do? May I open a new PR to push a colab notebook to show how to convert a model with dynamic_decode() into a TFLite file?

bhack:
You can add an unit test like https://github.com/seanpmorgan/addons/pull/15 then you can share a colab.

tensorflow_addons/seq2seq/decoder.py
```diff
@@ -352,9 +368,15 @@ def _shape(batch_size, from_shape):
                 batch_size = tf.get_static_value(
                     tf.convert_to_tensor(batch_size, name="batch_size")
                 )
+                if enable_tflite_convertible:
+                    # Since we can't use 2-D TensoArray and assume `batch_size` = 1,
```
lc0, line 372:
```suggestion
                    # Since we can't use 2-D TensorArray and assume `batch_size` = 1,
```

tensorflow_addons/seq2seq/decoder.py
```diff
@@ -352,9 +368,15 @@ def _shape(batch_size, from_shape):
...
+                    # we use `from_shape` dimension only.
+                    return from_shape
                 return tf.TensorShape([batch_size]).concatenate(from_shape)
 
         dynamic_size = maximum_iterations is None or not is_xla
+        # The dynamic shape `TensoArray` is not allowed in TFLite yet.
```
lc0, line 378:
```suggestion
        # The dynamic shape `TensorArray` is not allowed in TFLite yet.
```

guillaumekln:
@qlzh727 Do you have any comment on this one?

qlzh727:
> @qlzh727 Do you have any comment on this one?

Sorry for the late reply. Functional wise, I think it is fine to add this feature. On the other hand, the additional if-else branch makes code less readable, but I don't have a good alternative, unless we want to breakdown this function and/or create another decode function just for tflite.

jaeyoo:
Sorry, due to the limited bandwidth, preparing unittest is begin delayed now. May I push this code & example colab first, and then put the unittest later? In fact, the colab will be the modified version of this:

https://github.com/TensorSpeech/TensorflowTTS/blob/master/notebooks/TensorFlowTTS_Tacotron2_with_TFLite.ipynb

And I am about to remove TensorflowTTS dependency, and use our TensorFlow addons only.

bhack:
Can it work with the condition on `decoder.batch_size`?

guillaumekln:
@jaeyoo To move forward on this, I added a test that runs the conversion on your branch. Can you take a look and comment?

jaeyoo:
> @jaeyoo To move forward on this, I added a test that runs the conversion on your branch. Can you take a look and comment?

Yes it looks good to me. Thank you Guillaume!

guillaumekln:
@lc0 Your code suggestions were applied and generated a co-authored commit. We need you to confirm that you are ok with that as explained in https://github.com/tensorflow/addons/pull/1964#issuecomment-664847840.

PR #1966, opened 2020-07-03
Fixes up nomenclature on L1 and L2 norms in triplet losses.

gugarosa:
I had a hard time figuring out why my siamese model was not being able to converge, i.e., it was always converging to the pre-defined margin. After a while, I found out that the nomenclature used along the triplet losses (`L1` and `L2`) might be wrong according to their mathematical formulation.

Let's start with the package `losses/metric_learning` and function `pairwise_distance()`:

- At L34, the euclidean distances between all samples are calculated;
- According to the euclidean distance definition, it is initially calculated by summing up the square of the difference between every dimension;
- From L47 to L53, it is optionally possible to reduce the distance with a square root. Hence, we have the square root of the previously calculated distance, thus creating the L2-distance.
- On the other hand, if we decide not to square root the previously calculated distance, we are not returning the L1-distance, as we have already calculated the sum of the square of the differences. Instead, we are returning a squared version of the L2-distance, usually called `squared-L2`.

This might lead users in thinking that using the `L2` argument they are using the euclidean distance, but they are using its squared version in which, regarding some problems and architectures, will not convergence.

Therefore, this PR proposes to tackle such an issue and rename every `L1` to `L2` on the triplet losses, as well as every `L2` to `squared-L2`. Please, correct if I am wrong, but this definitely fixed my convergence problem.

Best regards,
Gustavo.

### DIFF
--- a/tensorflow_addons/losses/triplet.py
+++ b/tensorflow_addons/losses/triplet.py
@@ -83,8 +83,8 @@ def triplet_semihard_loss(
         be l2 normalized.
       margin: Float, margin term in the loss definition.
       distance_metric: str or function, determines distance metric:
-                       "L1" for l1-norm distance
                        "L2" for l2-norm distance
+                       "squared-L2" for squared l2-norm distance
                        "angular" for cosine similarity
                         A custom function returning a 2d adjacency
                           matrix of a chosen distance metric can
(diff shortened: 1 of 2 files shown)


gugarosa:
No problem, it's fine! Sorry for taking so long, just passed down the `pre-commit` script and updated this PR. Thanks again for handling this out!!

lc0:
Thank you for contribution @gugarosa! It looks like the existing tutorials **do not** use `L1`, so should be fine!

tensorflow_addons/losses/triplet.py
```diff
@@ -118,12 +118,12 @@ def custom_distance(batch):
 
     # Build pairwise squared distance matrix
```
lc0, line 119: Not sure how relevant this comment anymore. It's also hanging with an empty line, next to the line with `squared=False`

PR #1967, opened 2020-07-03
Snake layer and activation to learn periodic functions

### LINKED ISSUE #1939: feature request for Snake activation function
**Capable of learning a periodic function from the data better than current standard activation functions such as ReLU and tanh. Also performs as well as ReLU/leaky-ReLU on image task (refer paper)**

**Relevant information**
- Is there a relevant academic paper? (if so, where): https://arxiv.org/pdf/2006.08195.pdf

Implement under: **tfa.activation for base case, possibly tfa.layer for learnable param 'a' (frequency - refer paper)**

Who it benefits: **researches and practitioners from the time-series related space**

**Any other info.**

failure-to-thrive:
Closes #1939

### DIFF
--- a/tensorflow_addons/activations/__init__.py
+++ b/tensorflow_addons/activations/__init__.py
@@ -20,6 +20,6 @@ from tensorflow_addons.activations.lisht import lisht
 from tensorflow_addons.activations.mish import mish
 from tensorflow_addons.activations.softshrink import softshrink
 from tensorflow_addons.activations.rrelu import rrelu
+from tensorflow_addons.activations.snake import snake
 from tensorflow_addons.activations.sparsemax import sparsemax
 from tensorflow_addons.activations.tanhshrink import tanhshrink
-from tensorflow_addons.activations.snake import snake
--- a/tensorflow_addons/activations/snake.py
+++ b/tensorflow_addons/activations/snake.py
@@ -18,7 +18,6 @@ import tensorflow as tf
 from tensorflow_addons.utils import types
 
 
-@tf.function
 @tf.keras.utils.register_keras_serializable(package="Addons")
 def snake(inputs: types.TensorLike, freq: types.Number = 1) -> tf.Tensor:
     """Snake activation to learn periodic functions.
--- a/tensorflow_addons/activations/snake.py
+++ b/tensorflow_addons/activations/snake.py
@@ -1,4 +1,4 @@
-# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+# Copyright 2020 The TensorFlow Authors. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,14 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
-"""Additional activation functions."""
 
-from tensorflow_addons.activations.gelu import gelu
-from tensorflow_addons.activations.hardshrink import hardshrink
-from tensorflow_addons.activations.lisht import lisht
-from tensorflow_addons.activations.mish import mish
-from tensorflow_addons.activations.softshrink import softshrink
-from tensorflow_addons.activations.rrelu import rrelu
-from tensorflow_addons.activations.sparsemax import sparsemax
-from tensorflow_addons.activations.tanhshrink import tanhshrink
-from tensorflow_addons.activations.snake import snake
+import tensorflow as tf
+
+from tensorflow_addons.utils import types
+
+
+@tf.function
+@tf.keras.utils.register_keras_serializable(package="Addons")
+def snake(inputs: types.TensorLike, freq: types.Number = 1) -> tf.Tensor:
+    """Snake activation to learn periodic functions.
+
+    https://arxiv.org/abs/2006.08195
+    """
+
+    inputs = tf.convert_to_tensor(inputs)
+    freq = tf.cast(freq, inputs.dtype)
+
+    return inputs + (1 - tf.cos(2 * freq * inputs)) / (2 * freq)
--- a/tensorflow_addons/activations/snake.py
+++ b/tensorflow_addons/activations/snake.py
@@ -19,13 +19,19 @@ from tensorflow_addons.utils import types
 
 
 @tf.keras.utils.register_keras_serializable(package="Addons")
-def snake(inputs: types.TensorLike, freq: types.Number = 1) -> tf.Tensor:
+def snake(logits: types.TensorLike, frequency: types.Number = 1) -> tf.Tensor:
     """Snake activation to learn periodic functions.
 
     https://arxiv.org/abs/2006.08195
+
+    Args:
+        logits: Input tensor.
+        frequency: A scalar, frequency of the periodic part.
+    Returns:
+        Tensor of the same type and shape as `logits`.
     """
 
-    inputs = tf.convert_to_tensor(inputs)
-    freq = tf.cast(freq, inputs.dtype)
+    logits = tf.convert_to_tensor(logits)
+    frequency = tf.cast(frequency, logits.dtype)
 
-    return inputs + (1 - tf.cos(2 * freq * inputs)) / (2 * freq)
+    return logits + (1 - tf.cos(2 * frequency * logits)) / (2 * frequency)
--- a/tensorflow_addons/activations/tests/snake_test.py
+++ b/tensorflow_addons/activations/tests/snake_test.py
@@ -1,4 +1,4 @@
-# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+# Copyright 2020 The TensorFlow Authors. All Rights Reserved.
 #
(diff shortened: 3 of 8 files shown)


tensorflow_addons/activations/snake.py
```diff
@@ -0,0 +1,31 @@
...
+# limitations under the License.
+# ==============================================================================
+
+import tensorflow as tf
+
+from tensorflow_addons.utils import types
+
+
+@tf.keras.utils.register_keras_serializable(package="Addons")
+def snake(inputs: types.TensorLike, freq: types.Number = 1) -> tf.Tensor:
```
WindQAQ, line 22: can we use the full name `frequency` here?

tensorflow_addons/layers/snake.py
```diff
@@ -0,0 +1,47 @@
...
+
+@tf.keras.utils.register_keras_serializable(package="Addons")
+class Snake(tf.keras.layers.Layer):
+    """Snake layer to learn periodic functions with the 'frequency' param trainable.
+
+    https://arxiv.org/abs/2006.08195
+    """
+
+    @typechecked
+    def __init__(self, freq_initializer: types.Initializer = "ones", **kwargs):
```
WindQAQ, line 31: ditto

tensorflow_addons/layers/snake.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        freq_initializer: Initializer for the 'frequency' param.
+        """
+        super().__init__(**kwargs)
+        self.freq_initializer = tf.keras.initializers.get(freq_initializer)
+        self.freq = self.add_weight(initializer=freq_initializer, trainable=True)
+
+    def call(self, inputs):
+        return inputs + (1 - tf.cos(2 * self.freq * inputs)) / (2 * self.freq)
```
WindQAQ, line 40: can we directly call `tensorflow_addons.activations.snake` here?

tensorflow_addons/activations/tests/snake_test.py
```diff
@@ -0,0 +1,31 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+import pytest
+
+import numpy as np
+
+from tensorflow_addons.activations import snake
+
```
WindQAQ, line 21: nit: remove this empty line

tensorflow_addons/layers/tests/snake_test.py
```diff
@@ -0,0 +1,39 @@
...
+"""Tests for Snake layer."""
+
+import pytest
+
+import numpy as np
+import tensorflow as tf
+
+from tensorflow_addons.layers.snake import Snake
+from tensorflow_addons.activations.snake import snake
+
```
WindQAQ, line 24: ditto

tensorflow_addons/activations/__init__.py
```diff
@@ -22,3 +22,4 @@
 from tensorflow_addons.activations.rrelu import rrelu
 from tensorflow_addons.activations.sparsemax import sparsemax
 from tensorflow_addons.activations.tanhshrink import tanhshrink
+from tensorflow_addons.activations.snake import snake
```
WindQAQ, line 25: place in alphabetical order

tensorflow_addons/layers/__init__.py
```diff
@@ -36,3 +36,4 @@
 from tensorflow_addons.layers.tlu import TLU
 from tensorflow_addons.layers.wrappers import WeightNormalization
 from tensorflow_addons.layers.esn import ESN
+from tensorflow_addons.layers.snake import Snake
```
WindQAQ, line 39: ditto

tensorflow_addons/activations/__init__.py
bhack: @WindQAQ @seanpmorgan Just as a reminder can we automate with https://pypi.org/project/flake8-isort/?

tensorflow_addons/layers/snake.py
failure-to-thrive: It imposes additional overhead. That approach used in test degraded performance by 30%.

tensorflow_addons/activations/snake.py
```diff
@@ -0,0 +1,31 @@
...
+    """Snake activation to learn periodic functions.
+
+    https://arxiv.org/abs/2006.08195
+    """
```
WindQAQ, line 26: could we add the arguments doc here?

tensorflow_addons/layers/snake.py
```diff
@@ -0,0 +1,47 @@
...
+
+from tensorflow_addons.utils import types
+
+
+@tf.keras.utils.register_keras_serializable(package="Addons")
+class Snake(tf.keras.layers.Layer):
+    """Snake layer to learn periodic functions with the 'frequency' param trainable.
+
+    https://arxiv.org/abs/2006.08195
+    """
```
WindQAQ, line 28: could we add the arguments doc here?

tensorflow_addons/layers/snake.py
WindQAQ: that's weird. I don't think one `convert_to_tensor` and `cast` have such impact on performance. But anyway, because that's how core TF deals with activation and layer subclass, it'd be better to use `tensorflow_addons.activations.snake` IMO.

tensorflow_addons/activations/__init__.py
WindQAQ: Looks good to me. Could you open and issue and PR for it? I'm fine to automate it.

tensorflow_addons/activations/tests/snake_test.py
failure-to-thrive: Sure. The reason I was doing so is I was trying to somehow group similar imports since there are no strict guidelines on how to order imports, nor other layers and activations adhere some consistent pattern.

tensorflow_addons/activations/snake.py
failure-to-thrive: I was trying, but the original academic paper is quite obscure about explanations. So it could has end-up with words like 'frequency is a frequency' and other obvious things such as 'input is an input'. If it is OK anyway, no problems, I'll add.

tensorflow_addons/layers/snake.py
failure-to-thrive: I put them near the __init__ for an IDE autocomplete.

tensorflow_addons/activations/snake.py
WindQAQ: Got it. Maybe we can have something like

```
      inputs: Input tensor.
      frequency: A scalar, frequency of the periodic part.
```

tensorflow_addons/layers/snake.py
failure-to-thrive: OK. They know the best.

tensorflow_addons/layers/snake.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @typechecked
+    def __init__(self, frequency_initializer: types.Initializer = "ones", **kwargs):
+        """
+        frequency_initializer: Initializer for the 'frequency' param.
+        """
```
WindQAQ, line 36: Sorry for the confusion, after a rough look at our docs, it would be better to move this near `class Snake`.

```python
class Snake(tf.keras.layers.Layer):
    """Snake layer to learn periodic functions with the trainable `frequency` scalar.

    It follows:
    ```
      f(x) = x + (1 - cos(2 * frequency * x)) / (2 * frequency)
    ```

    where `frequency` is a learned scalar.

    See [Neural Networks Fail to Learn Periodic Functions and How to Fix It](https://arxiv.org/abs/2006.08195) for more details.

    Arguments:
        frequency_initializer: Initializer for the `frequency` scalar.
        frequency_regularizer: Regularizer for the `frequency` scalar.
        frequency_constraint: Constraint for the `frequency` scalar.
    """
```

I'm following doc of the parametric activation like [PReLU](https://www.tensorflow.org/api_docs/python/tf/keras/layers/PReLU) here. Feel free to elaborate more.

failure-to-thrive: > IMO, it suggests `frequency` to be with the same shape as the input (like PReLU). What do you think?

As I said earlier, the paper is quite obscure in many aspects. The same statement could have been made about the activation. Guesswork is a Sisyphus work. So I propose to leave it as it is. We can always get back to it later.

> Also, following the layers with learnable parameter in core TF, we should also support some arguments like `frequency_regularizer` and `frequency_constraint`.

Same here. I thought of it but the paper does not give a glue on how these 'regularizers and co' should be applied. If they are relevant here at all.

WindQAQ: Got it :-) Then just move the doc near `class Snake`.

failure-to-thrive: I had to eliminate the math formula because it looks very awkward in the plain text. In the end of the all, there is the academic paper an one must be in good knowledge before utilizing the layer.

WindQAQ: ok

PR #1971, opened 2020-07-06
Remove unnecessary eval_result container in tests

guillaumekln:
This was used for V1 session.run.


AakashKumarNain:
Thanks for the fix

PR #1978, opened 2020-07-08
fix typos and examples for test cases

AakashKumarNain:
The syntax for the test cases wasn't correct. Added more information regarding the implementation as well.

### DIFF
--- a/tensorflow_addons/metrics/README.md
+++ b/tensorflow_addons/metrics/README.md
@@ -12,7 +12,13 @@ must:
 
 #### Metric Requirements
 
-Any PR which adds a new metric must ensure that the following three cases are well tested and supported:
+Any PR which adds a new metric must ensure that:
+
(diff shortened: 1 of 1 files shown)


bhack:
Was this derived from https://twitter.com/fchollet/status/1280730971717820416?

AakashKumarNain:
Thanks for the review @WindQAQ . In the earlier version. I kept it that way only and I agree that it is better to use that way. But if you look at the docs for `metrics` in TF core, for example [categorical accuracy](https://www.tensorflow.org/versions/r2.3/api_docs/python/tf/keras/metrics/CategoricalAccuracy), they explicitly show `update_state()`. I thought it would be better to mention it here that way as well so that there is no confusion at all but we can take a call on which version we want to go with

AakashKumarNain:
@bhack nah, not from that.

bhack:
> @bhack nah, not from that.

Ok cause point 2. is that list.

PR #1980, opened 2020-07-09
Drop data_format argument

### LINKED ISSUE #1825: Inconsistent APIs within tfa.image and between tf.image and tfa.image
Currently, some functions in `tfa.image` that support `data_format` args like [equalize](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/color_ops.py#L79), [cutout](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/cutout_ops.py#L97), and [random_cutout](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/cutout_ops.py#L46), while most of them do not.  Moreover, all functions in [tf.image](https://www.tensorflow.org/api_docs/python/tf/image) support only channel-last images.

Therefore, I wonder if we can drop the support for `data_format` argument and channel-first image in `tfa.image` for some reasons:

- Be consistent with `tf.image`: Usually, we compose multiple functions in `tf.image` and `tfa.image`. So if one of the function does not support channel-first image, there must exist at least one `tf.transpose` operation to change the data format. In this case, the support of channel-first image seems vague and useless as we have to explicitly change it into channel-last format in order for input pipeline.
- Unlike Conv2D or CorrelationCost that has specific custom ops to handle different format, the support for channel first images in `tfa.image` just selects different axis to unpack, compute and pack.
- Shape casting: we do `to_4D_image` and `from_4D_image` to make most operations compatible with `tf.data.Dataset`, while they operate only on channel last images. We can indeed extend those functions for channel first images, but I am also confused if it's worthwhile.

**Relevant information**
- Is there already an implementation in another framework? (if so, where): addons

image

developers who use tfa.image

**Any other info.**

WindQAQ:
Closes #1825. Should we have some warning on deprecated arguments?

### DIFF
--- a/tensorflow_addons/image/color_ops.py
+++ b/tensorflow_addons/image/color_ops.py
@@ -24,10 +24,9 @@ from tensorflow_addons.image.utils import to_4D_image, from_4D_image
 from tensorflow_addons.image.compose_ops import blend
 
 from typing import Optional
-from functools import partial
 
 
-def equalize_image(image: TensorLike, data_format: str = "channels_last") -> tf.Tensor:
+def equalize_image(image: TensorLike) -> tf.Tensor:
     """Implements Equalize function from PIL using TF ops."""
 
     @tf.function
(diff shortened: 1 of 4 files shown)


AakashKumarNain:
LGTM! Thanks

facaiy:
> Should we have some warning on deprecated arguments?

I think it's a good idea to warn user that the argument is not valid any more. And we can safely remove those warnings after 2~3 releases :-)

WindQAQ:
> > Should we have some warning on deprecated arguments?
> 
> I think it's a good idea to warn user that the argument is not valid any more. And we can safely remove those warnings after 2~3 releases :-)

Sounds good to me. Should we revert the change and add the warning?

facaiy:
Great, it's not necessary to revert the change, let's just add the warning :-)

facaiy:
which package to use, Tzu-wei ? eg: deprecation, etc

WindQAQ:
> which package to add warninig, Tzu-wei ? eg: deprecation, etc

[This](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/optimizers/average_wrapper.py#L50-L55) looks like what we deal with deprecation.

facaiy:
Thanks, I'm fine with it

PR #1985, opened 2020-07-11
Use .ref() instead of .experimental_ref()

WindQAQ:
Per https://github.com/tensorflow/tensorflow/blob/master/RELEASE.md#bug-fixes-and-other-changes-5 and https://www.tensorflow.org/api_docs/python/tf/Tensor#experimental_ref.


AakashKumarNain:
LGTM! Thanks.

PR #1990, opened 2020-07-12
Add deprecation warning

WindQAQ:
Per https://github.com/tensorflow/addons/pull/1980

### DIFF
--- a/tensorflow_addons/image/color_ops.py
+++ b/tensorflow_addons/image/color_ops.py
@@ -17,6 +17,8 @@
     sharpness: Sharpen image
 """
 
+import warnings
+
 import tensorflow as tf
 
 from tensorflow_addons.utils.types import TensorLike, Number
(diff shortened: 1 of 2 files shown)


facaiy:
> Sounds good to me. Should we revert the change and add the warning?

Tzu-wei, sorry, I think you're right. Let's revert #1980 and remove the argument after next release, what do you think?

WindQAQ:
> > Sounds good to me. Should we revert the change and add the warning?
> 
> Tzu-wei, sorry, I think you're right. Let's revert #1980 and remove the argument after next release, what do you think?

Sure. Let's do it.

AakashKumarNain:
LGTM! Thanks. I didn't weighed in #1980 because IMO end user would always use core image ops as a ref, hence for the end user that argument shouldn't have been in there in the first place. Sorry for overlooking it though.

AakashKumarNain:
@WindQAQ I think we can merge this now or are there any more changes that need to be made?

WindQAQ:
Sure let's merge it :-)

PR #1992, opened 2020-07-12
Duplicate descriptions also to functions

lc0:
Currently [documentation](https://www.tensorflow.org/addons/api_docs/python/tfa/losses#functions) looks slightly broken, for some losses

image

Not a big fan of repeating docs in two places, but it looks like a consistent behaviour we do in other places as well.

Also, not sure if it also makes sense to adjust usage examples and bring to the added docs

### DIFF
--- a/tensorflow_addons/losses/focal_loss.py
+++ b/tensorflow_addons/losses/focal_loss.py
@@ -94,7 +94,17 @@ def sigmoid_focal_crossentropy(
     gamma: FloatTensorLike = 2.0,
     from_logits: bool = False,
 ) -> tf.Tensor:
-    """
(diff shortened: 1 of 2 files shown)


lc0:
> > Also, not sure if it also makes sense to adjust usage examples and bring to the added docs
> 
> It makes sense to me. It would be great if we can have a PR to unify the usage example for the same module. Let's do it in the future. Thank you!

@WindQAQ, documentation is generated from Docstring, so in this case we would have two different usage examples.
On other hand, these two different approaches one is function based and another one is class based ex: `sigmoid_focal_crossentropy` vs `SigmoidFocalCrossEntropy`

PR #1993, opened 2020-07-13
Set shape for dense image warp

### LINKED ISSUE #1969: (None, None, None, None)-shaped variable returned by tfa.image.dense_image_warp
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Linux 20.04
- TensorFlow version and how it was installed (source or binary): 2.1.0, source
- TensorFlow-Addons version and how it was installed (source or binary): latest, source
- Python version: 3.7
- Is GPU used? (yes/no): Yes

**Describe the bug**

When passing a "None"-batch size image with flow, such as a (None, 7, 7, 192) image with a valid flow, the returned value will have shape (None, None, None, None) 

```
import tensorflow as tf

tf.config.experimental_run_functions_eagerly(True)

image = tf.keras.layers.Input(shape=(7, 7, 192))
flow = tf.ones((1, 7, 7, 2))
warped = tfa.image.dense_image_warp(image, flow)
```

**Other info / logs**
```
>>> warped.shape
TensorShape([None, None, None, None])
```

Include any logs or source code that would be helpful to diagnose the problem. If including tracebacks, please include the full traceback. Large logs and files should be attached.

WindQAQ:
Closes #1969. Because feeding keras symbolic tensor into `tf.function` only supports when `tf.config.experimental_run_functions_eagerly(True)`, I create a new fixture.

### DIFF
--- a/tensorflow_addons/conftest.py
+++ b/tensorflow_addons/conftest.py
@@ -1,5 +1,6 @@
 from tensorflow_addons.utils.test_utils import (  # noqa: F401
     maybe_run_functions_eagerly,
+    only_run_functions_eagerly,
     run_custom_and_py_ops,
     run_with_mixed_precision_policy,
     pytest_make_parametrize_id,
--- a/tensorflow_addons/image/dense_image_warp.py
+++ b/tensorflow_addons/image/dense_image_warp.py
@@ -249,4 +249,5 @@ def dense_image_warp(
         # image grid.
         interpolated = interpolate_bilinear(image, query_points_flattened)
         interpolated = tf.reshape(interpolated, [batch_size, height, width, channels])
+        interpolated.set_shape(image.shape)
         return interpolated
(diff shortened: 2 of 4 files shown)


facaiy:
@WindQAQ Tzu-Wei, is it possible to create a private helper function which attempts to get static shape ?

```python
       def _get_shape(x):
           static_shape = x.shape
           dynamic_shape = tf.shape(x)
           shape = []
           for idx, dim in zip(static_shape):
               if idx is None:
                    shape.append(dynamic_shape[idx])
               else:
                    shape.append(dim)
          return shape
```

WindQAQ:
> @WindQAQ Tzu-Wei, is it possible to create a private helper function which attempts to get static shape ?
> 
> ```python
>        def _get_shape(x):
>            static_shape = x.shape
>            dynamic_shape = tf.shape(x)
>            shape = []
>            for idx, dim in zip(static_shape):
>                if idx is None:
>                     shape.append(dynamic_shape[idx])
>                else:
>                     shape.append(dim)
>           return shape
> ```

Sure. It makes sense to me. Updated! Thank you, Facai.

WindQAQ:
Seems that it will break our tests with unknown shapes. In that case, `static_shape` will be rank unknown, and can not be iterated.

facaiy:
Hi, Tzu-Wei, sorry for misleading you. I mean to use the helper function to replace tf.shape here. Does it work?
https://github.com/tensorflow/addons/blob/9cdc1855af5dcbfa61584203328775a6b733ceff/tensorflow_addons/image/dense_image_warp.py#L244-L248

WindQAQ:
Oh, it's my mistake haha. Seems that it works, lets see how CI says :-)

PR #1994, opened 2020-07-13
Clean up .assign in optmizers

WindQAQ:
- Use `.assign` instead of raw ops.
- Apply `use_locking` when doing assignment
- Remove trivial `.op`

Cause they are all related to `.assign`, I put them into one PR.

### DIFF
--- a/tensorflow_addons/optimizers/conditional_gradient.py
+++ b/tensorflow_addons/optimizers/conditional_gradient.py
@@ -178,9 +178,8 @@ class ConditionalGradient(tf.keras.optimizers.Optimizer):
             )
             s = top_singular_vector
 
-        return var.assign(
-            var * lr - (1 - lr) * lambda_ * s, use_locking=self._use_locking
-        )
+        var_update = tf.math.multiply(var, lr) - (1 - lr) * lambda_ * s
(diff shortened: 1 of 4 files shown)


tensorflow_addons/optimizers/conditional_gradient.py
```diff
@@ -182,13 +178,9 @@ def _resource_apply_dense(self, grad, var, apply_state=None):
...
 
-        var_update_tensor = tf.math.multiply(var, lr) - (1 - lr) * lambda_ * s
-        var_update_kwargs = {
-            "resource": var.handle,
-            "value": var_update_tensor,
-        }
-        var_update_op = tf.raw_ops.AssignVariableOp(**var_update_kwargs)
-        return tf.group(var_update_op)
+        return var.assign(
+            var * lr - (1 - lr) * lambda_ * s, use_locking=self._use_locking
```
facaiy, line 182: How about creating a temp variable for the expression? 
```python
var_update = xxxx
var.assign(xxx)
```

WindQAQ: Done :-)

Issue #1996, opened 2020-07-13
Make lookahead, weight decay, and stochastic weight averaging incompatible.


ben-arnao:
I was playing around with different architectures and training techniques and i noticed some odd behavior.

It seems like when we create a SGDW optimizer, and then try to wrap it with either lookahead or SWA, no error is thrown, the weigh_decay is simply truncated from the optimizer.

If you try to wrap a SWA optimizer with lookahead or visa versa, you'll get an explicit error.

In retrospect it makes sense that these 3 optimizer are incompatibility as they all operate on weights, and combining them doesn't make sense.

However i believe it might be a good idea to explicitly throw an error that let's us know, weight decay optimizer is not compatible with lookahead or SWA wrapper, instead of just silently removing the weight decay attribute.

WindQAQ:
Hi @ben-arnao, it's more than welcome to open an PR for better documentation and error handling! Also, can you provide the minimal runnable code snippet to reproduce what you say? Thank you for the input.

ben-arnao:
@WindQAQ I think this should work.

```
from tensorflow_addons.optimizers import Lookahead, SGDW
opt = SGDW(learning_rate=1e-1,
           weight_decay=1e-4)
print(hasattr(opt, 'weight_decay'))
opt = Lookahead(opt)
print(hasattr(opt, 'weight_decay'))
```

WindQAQ:
Sorry that I rushed to review this yesterday. What do you mean that the `weigh_decay` is simply truncated from the optimizer?

ben-arnao:
Lookahead (and SWA) will silently remove the attribute `weight_decay` from the optimizer as snippet above shows. This makes sense i guess, because they all operate on weights and we can't do weight decay while also doing lookahead/SWA.

The issue i had was that there is no sort of warning or error thrown if you try to wrap an optimizer that has weight decay (SGDW for example), with lookahead/SWA. It's not necessarily a functionality issue but it would be nice if TF gave some feedback.

Also, not sure if there is something deeper in the structure of these optimizers i'm not aware of, but if we try to extend an optimizer with Lookback, and then we try to extend with SWA, we'll get an explicit error thrown that SWA can't extend an optimizer of type Lookback or something of that nature.  However if we extend SGD with the decoupled weight decay extension to get SGDW, the base class is still SGD i guess so we don't get the same error.

WindQAQ:
[...]

PR #2000, opened 2020-07-14
Address #1996

ben-arnao:
Let users know that weight decay optimizers are not compatible with lookahead

### DIFF
--- a/tensorflow_addons/optimizers/lookahead.py
+++ b/tensorflow_addons/optimizers/lookahead.py
@@ -75,6 +75,10 @@ class Lookahead(tf.keras.optimizers.Optimizer):
             raise TypeError(
                 "optimizer is not an object of tf.keras.optimizers.Optimizer"
             )
+        if hasattr(optimizer, 'weight_decay'):
+            raise TypeError(
+                "Lookahead is not compatible with weight decay optimizers"
+            )
 
         self._optimizer = optimizer
         self._set_hyper("sync_period", sync_period)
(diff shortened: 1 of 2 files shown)


ben-arnao:
@WindQAQ I have uploaded a test. I do apologize if it is not good, i'm not too familiar with creating tests.

WindQAQ:
LGTM! We can use `pytest.raises` to catch the exception, such as
https://github.com/tensorflow/addons/blob/993f2d8587d20fc319dde4e2eb46b0f565d42fb7/tensorflow_addons/utils/tests/keras_utils_test.py#L29-L30

WindQAQ:
Just figure out than one common use is to wrap RectifiedAdam with Lookahead, and it seems to work probably.

ben-arnao:
Yea, actually sees like RAdam has weight_decay... which causes the tests to fail

```
from tensorflow_addons.optimizers import RectifiedAdam
from tensorflow.keras.optimizers import Adam

print(hasattr(Adam(), "weight_decay"))
print(hasattr(RectifiedAdam(), "weight_decay"))
```

Not sure if RAdam should have weight decay... the vanilla Adam does not.

WindQAQ:
The implementation from author also includes weight decay.

https://github.com/LiyuanLucasLiu/RAdam/blob/master/radam/radam.py

seanpmorgan:
See the colab in the #1996:
https://colab.research.google.com/drive/1b-8JSEMx0Ruqfue9L092sCQxq7-KJaZW?usp=sharing

PR #2007, opened 2020-07-16
Add PR template

### LINKED ISSUE #1815: Pull request template?
Sometime we have direct PR with new features without an issue so we don't know many important elements that are required in the ISSUE template like who is the maintainer or backup maintainer when we have new contributors (or need we to assume that the merger is the backup maintainer?). 

So, do we want to have a similar template for PR or we just want to enforce users to open an issue if they have opened a direct PR with new features?

seanpmorgan:
Closes #1815 

It would be great to break these out into several templates that are more specific to the contribution. However, from what I gather that would require users to navigate to the URL of the specific PR template which seems unlikely to happen. 

LMK if there are other things we want covered here or a better way to present the contribution guidelines (It may be a bit unwieldy when editing the markdown instead of viewing the rendered file.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,50 @@
+# Description
+
+Please include a summary of the change and which issue is fixed. 
+
+Fixes # (issue)
+
+## Type of change
+
+
+- [ ] Bug fix
+- [ ] New Tutorial
+- [ ] Updated or additional documentation
+- [ ] Additional Testing
+- [ ] New Activation (Requires creating an issue first)
+    - [ ] The changes conform to the [contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/activations/README.md#contribution-guidelines)
+- [ ] New Callback (Requires creating an issue first)
+    - [ ] The changes conform to the [contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/callbacks/README.md#contribution-guidelines)
+- [ ] New Image addition (Requires creating an issue first)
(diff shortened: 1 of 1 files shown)


bhack:
I would limit the list only to the mandatory checklist and not to alternative check-boxes so that we could apply something like https://github.com/clarkbw/task-list-zero-bot (or any other similar solution).

.github/pull_request_template.md
```diff
@@ -0,0 +1,50 @@
+# Description
+
+Please include a summary of the change and which issue is fixed. 
+
+Fixes # (issue)
+
+## Type of change
```
bhack, line 7: Type of change could be removed here, clustered in 5 classes and used as new templates. So that we could do also do some extra auto-labeling.

bhack: Oh is that for PR github hasn't an UX like for issue templates: https://github.community/t/multiple-pull-request-templates/1850/5

seanpmorgan: Yeah that would be the preferred method, but we're waiting on GitHub to support that

seanpmorgan:
> I would limit the list only to the mandatory checklist and not to alternative check-boxes so that we could apply something like:
> 
> * https://github.com/clarkbw/task-list-zero-bot
> * https://github.com/marketplace/task-list-completed/
> * or any other similar solution

I like this idea a lot, but IMO it requires additional support from GitHub so that we can have different PR templates for different contribution types.

seanpmorgan:
Going to merge this as is. But we can certainly iterate. I've also subscribed to the ticket for separate PR types.

PR #2008, opened 2020-07-16
Use tf.keras.backend.epsilon() as dtype

### LINKED ISSUE #2006: Keras model save using WeightedKappaLoss errors, not json serializable
**Describe the bug**

Keras model compiled with WeightedKappaLoss errors when saving, "TypeError: ('Not JSON Serializable:', tf.float32)"

```
model = Sequential()

model._set_inputs(tf.keras.Input((256,256,3)))
model.add(layers.Dense(6, activation='softmax'))

model.compile(Adam(lr=1e-3), tfa.losses.WeightedKappaLoss(num_classes=6, weightage='quadratic'))
model.save('test')
```

WindQAQ:
Fixes #2006.

There is no way to serialize `tf.dtypes.DType` and in `tf.keras.losses.*`, the computation dtype depends on either input (`y_true`, `y_pred`) or `tf.keras.backend.floatx()`. It terms out the most reasonable way for me is to cast values to `tf.keras.backend.floatx()` for computation because it involves something like log and div.

https://github.com/tensorflow/tensorflow/blob/v2.2.0/tensorflow/python/keras/losses.py

### DIFF
--- a/tensorflow_addons/losses/kappa_loss.py
+++ b/tensorflow_addons/losses/kappa_loss.py
@@ -113,7 +113,7 @@ class WeightedKappaLoss(tf.keras.losses.Loss):
 
     def call(self, y_true, y_pred):
         y_true = tf.cast(y_true, dtype=self.col_label_vec.dtype)
-        y_pred = tf.cast(y_pred, dtype=self.weight_mat)
+        y_pred = tf.cast(y_pred, dtype=self.weight_mat.dtype)
         batch_size = tf.shape(y_true)[0]
         cat_labels = tf.matmul(y_true, self.col_label_vec)
         cat_label_mat = tf.tile(cat_labels, [1, self.num_classes])
--- a/tensorflow_addons/losses/kappa_loss.py
+++ b/tensorflow_addons/losses/kappa_loss.py
@@ -14,7 +14,6 @@
 # ==============================================================================
 """Implements Weighted kappa loss."""
 
-import warnings
 from typing import Optional
 
 import tensorflow as tf
@@ -88,20 +87,14 @@ class WeightedKappaLoss(tf.keras.losses.Loss):
 
         super().__init__(name=name, reduction=reduction)
 
-        warnings.warn(
-            "The data type for `WeightedKappaLoss` defaults to "
-            "`tf.keras.backend.floatx()`."
-            "The argument `dtype` will be removed in Addons `0.12`.",
-            DeprecationWarning,
-        )
-
         if weightage not in ("linear", "quadratic"):
             raise ValueError("Unknown kappa weighting type.")
 
         self.weightage = weightage
         self.num_classes = num_classes
         self.epsilon = epsilon or tf.keras.backend.epsilon()
-        label_vec = tf.range(num_classes, dtype=tf.keras.backend.floatx())
+        self.dtype = dtype
+        label_vec = tf.range(num_classes, dtype=self.dtype)
         self.row_label_vec = tf.reshape(label_vec, [1, num_classes])
         self.col_label_vec = tf.reshape(label_vec, [num_classes, 1])
         col_mat = tf.tile(self.col_label_vec, [1, num_classes])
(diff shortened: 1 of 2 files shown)


tensorflow_addons/losses/kappa_loss.py
```diff
@@ -103,7 +112,7 @@ def __init__(
             self.weight_mat = (col_mat - row_mat) ** 2
 
     def call(self, y_true, y_pred):
-        y_true = tf.cast(y_true, dtype=self.dtype)
+        y_true = tf.cast(y_true, dtype=tf.keras.backend.floatx())
```
bhack, line 115: Is this required?

tensorflow_addons/losses/tests/kappa_loss_test.py
```diff
@@ -90,3 +90,14 @@ def test_config():
 def test_serialization():
     loss = WeightedKappaLoss(num_classes=3)
     tf.keras.losses.deserialize(tf.keras.losses.serialize(loss))
+
+
+def test_save_model(tmpdir):
```
bhack, line 95: Where `tmpdir` is coming for?

WindQAQ: it's purest fixture

https://docs.pytest.org/en/stable/tmpdir.html

tensorflow_addons/losses/kappa_loss.py
WindQAQ: Test, because there is a `tf.matmul(y_true, self.col_label_vec)` down below, where `self.col_label_vec` is of type `tf.keras.backend.floatx()`.

bhack: Ok so the direct constrain is  `dtype=self.col_label_vec.dtype`?

facaiy:
Hi, Tzu-Wei, I know there is one way to serialize tf.dtypes.DType:

```python
In [21]: tf.float32.as_datatype_enum
Out[21]: 1

In [22]: tf.as_dtype(tf.float32.as_datatype_enum)
Out[22]: tf.float32
```

WindQAQ:
> Hi, Tzu-Wei, I know there is one way to serialize tf.dtypes.DType:
> 
> ```python
> In [21]: tf.float32.as_datatype_enum
> Out[21]: 1
> 
> In [22]: tf.as_dtype(tf.float32.as_datatype_enum)
> Out[22]: tf.float32
> ```

Awesome! Let me try it.

tensorflow_addons/losses/tests/kappa_loss_test.py
```diff
@@ -90,3 +90,22 @@ def test_config():
...
+            tf.keras.layers.Input((256, 256, 3)),
+            tf.keras.layers.Dense(6, activation="softmax"),
+        ]
+    )
+    model.compile(optimizer="adam", loss=WeightedKappaLoss(num_classes=6))
+    model.save(str(tmpdir / "test.h5"))
+
+
+def test_dtype_config():
+    wkl = WeightedKappaLoss(num_classes=4, dtype=tf.float32)
```
bhack, line 107: Is this the only loss in the repo where we handle `dtype` in the loss class?

WindQAQ: yes it is :-)

bhack: is there a particular reason?

WindQAQ: Because it is very weird to pass `dtype` into losses. In core TF, the dtype of losses is determined either from inputs or from `backend.floatx()`. Consider the loss is going to pass into optimizers, which (mostly) creates the slots with the same dtype of variables, I think loss will be better to have either dtype `floatx()` or the one same with `y_pred/y_true`, so that the any computation in backward pass will not throw some errors like incompatible dtype.

bhack: I meant Is there any particular reason why we are using it in this case?

WindQAQ: ohoh, sorry for misunderstanding :-(

Actually, I think no reason we do that. I would say `backend.floatx()` is the better solution. What do you think?

bhack: If we don't break serialized users models I think It Is better

AakashKumarNain: Similar thoughts. That argument doesn't make sense as this should be inferred from the input dtype

WindQAQ: Ok, I revert the commit and use `tf.keras.backend.floatx()` instead.

PR #2010, opened 2020-07-17
CLEAN: Remove Duplicated reset value

marload:
Hello, TensorFlow Addons Community!

I removed a duplicate initial value generation call from `reset_state`

Thanks!

### DIFF
--- a/tensorflow_addons/metrics/f_scores.py
+++ b/tensorflow_addons/metrics/f_scores.py
@@ -15,6 +15,7 @@
 """Implements F scores."""
 
 import tensorflow as tf
+from tensorflow.python.keras import backend as K
 from typeguard import typechecked
 
 from tensorflow_addons.utils.types import AcceptableDTypes, FloatTensorLike
@@ -185,10 +186,7 @@ class FBetaScore(tf.keras.metrics.Metric):
 
     def reset_states(self):
         reset_value = tf.zeros(self.init_shape, dtype=self.dtype)
-        self.true_positives.assign(reset_value)
-        self.false_positives.assign(reset_value)
-        self.false_negatives.assign(reset_value)
-        self.weights_intermediate.assign(reset_value)
+        K.batch_set_value([(v, reset_value) for v in self.variables])
 
 
 @tf.keras.utils.register_keras_serializable(package="Addons")
--- a/tensorflow_addons/metrics/f_scores.py
+++ b/tensorflow_addons/metrics/f_scores.py
@@ -184,10 +184,11 @@ class FBetaScore(tf.keras.metrics.Metric):
         return {**base_config, **config}
 
     def reset_states(self):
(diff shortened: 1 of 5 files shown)


tensorflow_addons/metrics/r_square.py
```diff
@@ -139,7 +139,8 @@ def result(self) -> tf.Tensor:
...
         # The state of the metric will be reset at the start of each epoch.
-        self.squared_sum.assign(0)
-        self.sum.assign(0)
-        self.res.assign(0)
-        self.count.assign(0)
+        reset_value = 0
+        self.squared_sum.assign(reset_value)
+        self.sum.assign(reset_value)
+        self.res.assign(reset_value)
+        self.count.assign(reset_value)
```
WindQAQ, line 146: using `0` directly will not have duplicate initial value generation call :-)

bhack: Yes but it could simplify semantic readibility if all the values need to be the same

tensorflow_addons/metrics/f_scores.py
```diff
@@ -15,6 +15,7 @@
 """Implements F scores."""
 
 import tensorflow as tf
+from tensorflow.keras import backend as K
```
WindQAQ, line 18: remove this

tensorflow_addons/metrics/f_scores.py
```diff
@@ -184,10 +185,8 @@ def get_config(self):
         return {**base_config, **config}
 
     def reset_states(self):
-        self.true_positives.assign(tf.zeros(self.init_shape, self.dtype))
-        self.false_positives.assign(tf.zeros(self.init_shape, self.dtype))
-        self.false_negatives.assign(tf.zeros(self.init_shape, self.dtype))
-        self.weights_intermediate.assign(tf.zeros(self.init_shape, self.dtype))
+        reset_value = tf.zeros(self.init_shape, dtype=self.dtype)
+        K.batch_set_value([(v, reset_value) for v in self.variables])
```
WindQAQ, line 189:
```python
tf.keras.backend.batch_set_value([(v, reset_value) for v in self.variables])
```

WindQAQ: Just for the consistent style in the repo :-)

bhack: Yes probably it is more popular/consistent but not uniform. :wink: 
See https://github.com/tensorflow/addons/search?q=%22backend+as+K%22&unscoped_q=%22backend+as+K%22

marload: Thanks!

Looking at the other codes in tesnorflow_addons, its seem to be used inconsistently. 

If you choose one style, I will modify it to be consistent. :)

PR #2011, opened 2020-07-17
EHN: Using Language Tag in CONTRIBUTING.md

marload:
Hello!

I improved the `CONTRIBUTING.md` file using Language Tag (I don't know if this work is right)

Thanks!


marload:
Thank you for your efforts.

> I've not investigated if it is also auto-detected without language identifier (e.g. in the python case). But in any case it could be ok.
> 
> More info at:
> https://docs.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks#syntax-highlighting

Thank you for your efforts!

PR #2013, opened 2020-07-18
REFACOTR: Function Extraction _zero_wt_init(name)

marload:
Thank You!

### DIFF
--- a/tensorflow_addons/metrics/f_scores.py
+++ b/tensorflow_addons/metrics/f_scores.py
@@ -19,6 +19,7 @@ from typeguard import typechecked
 
 from tensorflow_addons.utils.types import AcceptableDTypes, FloatTensorLike
 from typing import Optional
+from utils import zero_wt_init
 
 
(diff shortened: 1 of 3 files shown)


marload:
https://github.com/tensorflow/addons/blob/f03f26f8c16097cfbc36739135ea3a38b6287896/tensorflow_addons/metrics/f_scores.py#L114

This refactoring is already in use by F-Score. In my view, I think this is useful. :smile:

bhack:
I meant if this pattern It Is used/useful in multiple place or just in multipleetrics could this function go in `utils.py` instead of redefine It multiple time?

marload:
Oh, I see. It seems like a better way.

bhack:
Have you seen https://github.com/tensorflow/addons/pull/2013#pullrequestreview-451060161?

bhack:
I meant only to see https://github.com/tensorflow/addons/blob/master/tensorflow_addons/metrics/multilabel_confusion_matrix.py#L84

PR #2014, opened 2020-07-18
REFACTOR: Early Returnn Pattern (if elif else -> if if return)

marload:
In Early Return Pattern, `if if return` is more recommended than `if elif else`.

Thank You! :)

### DIFF
--- a/tensorflow_addons/metrics/r_square.py
+++ b/tensorflow_addons/metrics/r_square.py
@@ -126,16 +126,15 @@ class RSquare(Metric):
 
         if self.multioutput == "raw_values":
             return raw_scores
-        elif self.multioutput == "uniform_average":
+        if self.multioutput == "uniform_average":
             return tf.reduce_mean(raw_scores)
-        elif self.multioutput == "variance_weighted":
+        if self.multioutput == "variance_weighted":
             return _reduce_average(raw_scores, weights=total)
-        else:
-            raise RuntimeError(
-                "The multioutput attribute must be one of {}, but was: {}".format(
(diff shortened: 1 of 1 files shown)


marload:
> I don't know if we could enforce this cause It seems that this rule is supported in pylint but we are using other tools:
> 
> https://pheanex.github.io/pylint/#r1705-no-else-return

Thank you for your review!

We do not need to enforce this style. However, I think this style is more useful in many cases, including maintenance.

Thank You 😄

bhack:
Let's see if the other maintainers want to introduce https://pypi.org/project/flake8-return/ in https://github.com/tensorflow/addons/blob/master/tools/install_deps/flake8.txt.

bhack:
I think that It Is a quite open topic.
Mozilla c++ codestyle (with some controversy):
https://blog.mozilla.org/nnethercote/2009/08/31/no-else-after-return-considered-harmful/

In they python ecosystem It was introduced in `pylint 1.7.1` as a warning but seems to me no option/plugin in the linter that we are using.

See also for Python https://dmerej.info/blog/post/else-after-return-yea-or-nay/

bhack:
But I don't think that Tensorflow cares about this:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/utils/conv_utils.py#L28:L48

bhack:
As you se It is hard to enforce what you cannot lint:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/cloud/gcs_file_system.cc#L151:L158

PR #2020, opened 2020-07-20
TYPO: Add missing spaces in some multi-line messages

csachs:
Getting an error message, I've noticed that two words were joined because a space had been forgotten in the message composed of continued multi-line strings.
I've added it, looked around for similar issues and found two more. (I don't vouch to have found all, tho'.)

Furthermore, it seems that multi-line messages are assembled incoherently within the project, sometimes spaces are at the end of lines to be continued, sometimes at the beginnings, furthermore sometimes `\n` stands alone, sometimes with a preceding space. I would suggest to have a standardization to follow here. (My suggestion would be space at the end of an individual line, and no spaces preceding `\n`).


seanpmorgan:
Our formatter black doesn't have a preference on how lines are broken so for simplicity neither will we. But they should be correct! Thanks @csachs!

PR #2023, opened 2020-07-20
Use updated method for launching GPU containers

seanpmorgan:
Per https://github.com/NVIDIA/nvidia-docker#usage

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -272,7 +272,7 @@ docker run --rm -it -v ${PWD}:/addons -w /addons tensorflow/tensorflow:2.1.0-cus
 
 GPU Docker: 
 ```bash
-docker run --runtime=nvidia --rm -it -v ${PWD}:/addons -w /addons tensorflow/tensorflow:2.1.0-custom-op-gpu-ubuntu16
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -272,7 +272,7 @@ docker run --rm -it -v ${PWD}:/addons -w /addons tensorflow/tensorflow:2.1.0-cus
 
 GPU Docker: 
 ```bash
-docker run --runtime=nvidia --rm -it -v ${PWD}:/addons -w /addons tensorflow/tensorflow:2.1.0-custom-op-gpu-ubuntu16
+docker run --gpus all --rm -it -v ${PWD}:/addons -w /addons tensorflow/tensorflow:2.1.0-custom-op-gpu-ubuntu16
```
bhack, line 275: 2.1.0?

bhack: I am trying to explore what happens with an aligned version at https://github.com/tensorflow/addons/pull/2027

seanpmorgan: As mentioned in the other PR. There's no reason to bump this docker container since we install our own version of TF in CI. The devtoolset and CUDA versions have remain unchanged. A bit confusing I agree but until we clean up that docker image it's not worth updating.

PR #2026, opened 2020-07-21
REFACTOR: Remove Duplicated Logit

marload:
I removed duplication and improved readability by refactoring the code.

Thank You!

### DIFF
--- a/tensorflow_addons/layers/adaptive_pooling.py
+++ b/tensorflow_addons/layers/adaptive_pooling.py
@@ -53,11 +53,12 @@ class AdaptivePooling1D(tf.keras.layers.Layer):
         super().__init__(**kwargs)
 
     def call(self, inputs, *args):
+        bins = self.output_size[0]
         if self.data_format == "channels_last":
             s_ax, v_ax = 1, 2
         else:
             s_ax, v_ax = 2, 3
-        splits = tf.split(inputs, self.output_size[0], axis=s_ax)
+        splits = tf.split(inputs, bins, axis=s_ax)
         splits = tf.stack(splits, axis=s_ax)
         return self.reduce_function(splits, axis=v_ax)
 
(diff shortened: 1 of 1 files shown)


tensorflow_addons/layers/adaptive_pooling.py
```diff
@@ -53,16 +53,13 @@ def __init__(
         super().__init__(**kwargs)
 
     def call(self, inputs, *args):
-        bins = self.output_size[0]
```
bhack, line 56: Here bins probably has more semantic info.

marload: I agree.
Thank You!

Susmit-A:
I believe it makes the logic sligtly less obvious to readers. Also, pooling in higher dimensions can't be simplified as neatly as in case of 1D pooling, so this would break the trend of code applied in all layers.

bhack:
@WindQAQ are we deprecating data_format also in case like this one as in https://github.com/tensorflow/addons/issues/1825?

WindQAQ:
> @WindQAQ are we deprecating `data_format` also in case like this one as in #1825?

No, that issue is specific to `tfa.image`.

> I believe it makes the logic sligtly less obvious to readers. Also, pooling in higher dimensions can't be simplified as neatly as in case of 1D pooling, so this would break the trend of code applied in all layers.

I agree with @Susmit-A. IMO, splitting cases *clearly* is much more important than duplicated codes elimination.

bhack:
More in general, what I think is that when we have some component with a "family" already in Tensorflow we could check TF style before submit small style refactoring PR. Also if currently it is not a mandatory check in our contributing guidelines or PR template /cc @seanpmorgan. 
E.g. in this case https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/layers/pooling.py#L83

marload:
I will close this PR!

PR #2022, opened 2020-07-20
Bump to tf2.3rc2

seanpmorgan:

PR #2027, opened 2020-07-21
Update custom-ops image to 2.2.0

bhack:


bhack:
@seanpmorgan What we need to do when we have PR like https://github.com/tensorflow/addons/pull/2022? 
Just as reminder in https://github.com/tensorflow/build/issues/7 `reminder.sh` currently conditional skip `rc`.

seanpmorgan:
This docker image is 9GB vs. our image which has been cleaned up and sits at 3GB. There important contents of the image are the exact same (devtoolset-7 etc.) and we install the version of tensorflow ourselves. 

The correct path forward is to get the custom-op image to fork from the devel image, include python3.8 and remove a lot of the bloat. Until that happens there is no need to bump the docker image IMO

bhack:
Yes I know but this was just a build test to see if things still build with an updated upstream images as our custom image is 2.1.0.
If you are monitoring the custom-ops repository activity and you know that new upstream Dockerimage releases are currently useless by a compatibility point of view with new Tensorflow releases there is no interest to test them.

PR #2079, opened 2020-08-06
Allow Optimizers to accept a loss tensor

omalleyt12:
This brings the TensorFlow Addons Optimizers up to date with the built-in Optimizers


WindQAQ:
@omalleyt12 Thanks for the fix! Can you run `bash tools/pre-commit.sh` to format codes? Thanks again!

omalleyt12:
@WindQAQ Thanks for the review! Done

PR #2099, opened 2020-08-18
Added support for noisy dense layers.

LeonShams:
# Description

This PR adds support for noisy dense layers. Noisy dense layers are like dense layers but random noise is injected to help agents explore in reinforcement learning environments. This is very commonly used in Deep Q Learning and was first introduced by DeepMind in their [Noisy Networks for Exploration](https://arxiv.org/pdf/1706.10295.pdf) paper in 2017.
Fixes: [#2127](https://github.com/tensorflow/addons/issues/2127)

## Type of change

- [ ] Bug fix
- [ ] New Tutorial
- [ ] Additional Testing
- [x] New Layer and the changes conform to the [layer contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/layers/README.md#contribution-guidelines)

# Checklist:

# How Has This Been Tested?

I used PyTest to test the code and trained it on some reinforcement learning environments from the [gym](https://github.com/openai/gym) toolkit and compared my results to the results on the [Noisy Networks for Exploration](https://arxiv.org/pdf/1706.10295.pdf) paper.

### DIFF
--- a/tensorflow_addons/layers/noisy_dense.py
+++ b/tensorflow_addons/layers/noisy_dense.py
@@ -32,10 +32,9 @@ def _scale_noise(x):
 
 
 @tf.keras.utils.register_keras_serializable(package="Addons")
-class NoisyDense(tf.keras.layers.Layer):
-    r"""Like normal dense layer but random noise is added to the weights
-    matrix. As the network improves the random noise is decayed until
-    it is insignificant.
+class NoisyDense(tf.keras.layers.Dense):
+    r"""Noisy dense layer that inject random noise to
+    the weights of normal dense layer.
 
     A `NoisyDense` layer implements the operation:
     $$
@@ -60,31 +59,31 @@ class NoisyDense(tf.keras.layers.Layer):
     (None, 32)
 
     Arguments:
-    units: Positive integer, dimensionality of the output space.
-    sigma: A float between 0-1 used as a standard deviation figure and is
-      applied to the gaussian noise layer (`sigma_kernel` and `sigma_bias`).
-    activation: Activation function to use.
-      If you don't specify anything, no activation is applied
-      (ie. "linear" activation: `a(x) = x`).
-    use_bias: Boolean, whether the layer uses a bias vector.
-    kernel_regularizer: Regularizer function applied to
-      the `kernel` weights matrix.
-    bias_regularizer: Regularizer function applied to the bias vector.
-    activity_regularizer: Regularizer function applied to
-      the output of the layer (its "activation").
-    kernel_constraint: Constraint function applied to
-      the `kernel` weights matrix.
-    bias_constraint: Constraint function applied to the bias vector.
+      units: Positive integer, dimensionality of the output space.
+      sigma: A float between 0-1 used as a standard deviation figure and is
+        applied to the gaussian noise layer (`sigma_kernel` and `sigma_bias`).
+      activation: Activation function to use.
+        If you don't specify anything, no activation is applied
+        (ie. "linear" activation: `a(x) = x`).
+      use_bias: Boolean, whether the layer uses a bias vector.
+      kernel_regularizer: Regularizer function applied to
+        the `kernel` weights matrix.
+      bias_regularizer: Regularizer function applied to the bias vector.
+      activity_regularizer: Regularizer function applied to
+        the output of the layer (its "activation").
+      kernel_constraint: Constraint function applied to
+        the `kernel` weights matrix.
+      bias_constraint: Constraint function applied to the bias vector.
 
     Input shape:
-    N-D tensor with shape: `(batch_size, ..., input_dim)`.
-    The most common situation would be
-    a 2D input with shape `(batch_size, input_dim)`.
+      N-D tensor with shape: `(batch_size, ..., input_dim)`.
+      The most common situation would be
+      a 2D input with shape `(batch_size, input_dim)`.
 
     Output shape:
-    N-D tensor with shape: `(batch_size, ..., units)`.
-    For instance, for a 2D input with shape `(batch_size, input_dim)`,
-    the output would have shape `(batch_size, units)`.
+      N-D tensor with shape: `(batch_size, ..., units)`.
+      For instance, for a 2D input with shape `(batch_size, input_dim)`,
+      the output would have shape `(batch_size, units)`.
     """
 
     @typechecked
@@ -101,21 +100,20 @@ class NoisyDense(tf.keras.layers.Layer):
         bias_constraint: types.Constraint = None,
         **kwargs
     ):
-        super(NoisyDense, self).__init__(
-            activity_regularizer=activity_regularizer, **kwargs
+        super().__init__(
+            units=units,
+            activation=activation,
+            use_bias=use_bias,
+            kernel_regularizer=kernel_regularizer,
+            bias_regularizer=bias_regularizer,
+            activity_regularizer=activity_regularizer,
+            kernel_constraint=kernel_constraint,
+            bias_constraint=bias_constraint,
+            **kwargs,
         )
-
-        self.units = units
+        delattr(self, "kernel_initializer")
+        delattr(self, "bias_initializer")
         self.sigma = sigma
-        self.activation = activations.get(activation)
-        self.use_bias = use_bias
-        self.kernel_regularizer = regularizers.get(kernel_regularizer)
-        self.bias_regularizer = regularizers.get(bias_regularizer)
-        self.kernel_constraint = constraints.get(kernel_constraint)
-        self.bias_constraint = constraints.get(bias_constraint)
-
-        self.input_spec = InputSpec(min_ndim=2)
-        self.supports_masking = True
 
     def build(self, input_shape):
         # Make sure dtype is correct
@@ -189,10 +187,20 @@ class NoisyDense(tf.keras.layers.Layer):
         else:
             self.sigma_bias = None
             self.mu_bias = None
+        self._reset_noise()
         self.built = True
 
+    @property
+    def kernel(self):
+        return self.mu_kernel + (self.sigma_kernel * self.eps_kernel)
+
+    @property
+    def bias(self):
+        if self.use_bias:
+            return self.mu_bias + (self.sigma_bias * self.eps_bias)
+
     # Create the factorised Gaussian noise
-    def reset_noise(self):
+    def _reset_noise(self):
         dtype = self._compute_dtype_object
 
         # Generate random noise
@@ -208,74 +216,27 @@ class NoisyDense(tf.keras.layers.Layer):
         self.eps_kernel = _scale_noise(eps_i) * _scale_noise(eps_j)
         self.eps_bias = _scale_noise(eps_j)
 
-    def remove_noise(self):
+    def _remove_noise(self):
         dtype = self._compute_dtype_object
         self.eps_kernel = tf.zeros([self.last_dim, self.units], dtype=dtype)
-        self.eps_bias = tf.zeros([self.last_dim, self.units], dtype=dtype)
+        self.eps_bias = tf.zeros([self.units], dtype=dtype)
 
     def call(self, inputs, reset_noise=True, remove_noise=False):
-        dtype = self._compute_dtype_object
-        if inputs.dtype.base_dtype != dtype.base_dtype:
-            inputs = tf.cast(inputs, dtype=dtype)
-
         # Generate fixed parameters added as the noise
         if remove_noise:
-            self.remove_noise()
+            self._remove_noise()
         elif reset_noise:
-            self.reset_noise()
+            self._reset_noise()
 
-        r"""
-        Perform:
-        $$
-        y \stackrel{\text{def}}{=}
-        (\mu^w + \sigma^w \odot \eps^w)x + \mu^b + \sigma^b \odot \eps^b
-        $$
-        to calculate the output
-        """
-        kernel = self.mu_kernel + (self.sigma_kernel * self.eps_kernel)
-
-        if inputs.dtype.base_dtype != dtype.base_dtype:
-            inputs = tf.cast(inputs, dtype=dtype)
-
-        rank = inputs.shape.rank
-        if rank == 2 or rank is None:
-            if isinstance(inputs, tf.sparse.SparseTensor):
-                outputs = tf.sparse.sparse_dense_matmul(inputs, kernel)
-            else:
-                outputs = tf.linalg.matmul(inputs, kernel)
-        # Broadcast kernel to inputs.
-        else:
-            outputs = tf.tensordot(inputs, kernel, [[rank - 1], [0]])
-            # Reshape the output back to the original ndim of the input.
-            if not tf.executing_eagerly():
-                shape = inputs.shape.as_list()
-                output_shape = shape[:-1] + [kernel.shape[-1]]
-                outputs.set_shape(output_shape)
-
-        if self.use_bias:
-            noisy_bias = self.mu_bias + (self.sigma_bias * self.eps_bias)
-            outputs = tf.nn.bias_add(outputs, noisy_bias)
-
-        if self.activation is not None:
-            outputs = self.activation(outputs)
-
-        return outputs
-
-    def compute_output_shape(self, input_shape):
-        input_shape = tf.TensorShape(input_shape)
-        input_shape = input_shape.with_rank_at_least(2)
-        if tf.compat.dimension_value(input_shape[-1]) is None:
-            raise ValueError(
-                "The innermost dimension of input_shape must be defined"
-                ", but saw: %s" % input_shape
-            )
-        return input_shape[:-1].concatenate(self.units)
+        # TODO(WindQAQ): Replace this with `dense()` once public.
+        return super().call(inputs)
 
     def get_config(self):
-        config = super(NoisyDense, self).get_config()
+        config = super(tf.keras.layers.Dense, self).get_config()
         config.update(
             {
                 "units": self.units,
+                "sigma": self.sigma,
                 "activation": activations.serialize(self.activation),
                 "use_bias": self.use_bias,
                 "kernel_regularizer": regularizers.serialize(self.kernel_regularizer),
--- a/tensorflow_addons/layers/noisy_dense.py
+++ b/tensorflow_addons/layers/noisy_dense.py
@@ -1,4 +1,4 @@
-# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+# Copyright 2020 The TensorFlow Authors. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -11,9 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-# Orginal implementation from keras_contrib/layer/normalization
-# =============================================================================
+# ==============================================================================
 
 import tensorflow as tf
 from tensorflow.keras import (
@@ -28,8 +26,8 @@ from tensorflow.keras.layers import InputSpec
 
 @tf.keras.utils.register_keras_serializable(package="Addons")
 class NoisyDense(tf.keras.layers.Layer):
-    """Like normal dense layer but random noisy is added to the weights matrix. But
-  as the network improves the random noise is decayed until it is insignificant.
+    """Like normal dense layer (https://github.com/tensorflow/tensorflow/blob/v2.3.0/tensorflow/python/keras/layers/core.py#L1067-L1233)
+  but random noisy is added to the weights matrix. But as the network improves the random noise is decayed until it is insignificant.
 
   A `NoisyDense` layer implements the operation:
   `output = activation(dot(input, µ_kernel + (σ_kernel * ε_kernel)) + bias)`
--- a/tensorflow_addons/layers/noisy_dense.py
+++ b/tensorflow_addons/layers/noisy_dense.py
@@ -21,7 +21,7 @@ from tensorflow.python.ops import math_ops, nn_ops, sparse_ops, gen_math_ops, st
 from tensorflow.python.keras.engine.input_spec import InputSpec
 from tensorflow.python.framework import dtypes, tensor_shape, sparse_tensor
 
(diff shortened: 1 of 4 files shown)


tensorflow_addons/layers/tests/noisy_dense_test.py
```diff
@@ -0,0 +1,86 @@
...
+    inputs = tf.convert_to_tensor(
+        np.random.randint(low=0, high=7, size=(2, 2)), dtype=dtype
+    )
+    layer = NoisyDense(5, dtype=dtype)
+    outputs = layer(inputs)
+    np.testing.assert_array_equal(outputs.dtype, dtype)
+
+
+@pytest.mark.usefixtures("maybe_run_functions_eagerly")
+def test_noisy_dense_with_policy():
```
bhack, line 50: If I remember correctly we had a fixture like `@pytest.mark.usefixtures("run_with_mixed_precision_policy")` somewhere in the codebase

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,249 @@
...
+        if tf.compat.dimension_value(input_shape[-1]) is None:
+            raise ValueError(
+                "The innermost dimension of input_shape must be defined, but saw: %s"
+                % input_shape
+            )
+        return input_shape[:-1].concatenate(self.units)
+
+    def get_config(self):
+        config = super(NoisyDense, self).get_config()
+        config.update(
```
bhack, line 236: Are you missing something in the config dictionary e.g. like `use_bias` or initializes?

LeonShams: >Are you missing something in the config dictionary e.g. like `use_bias`

I added `use_bias` but for some reason, it is now giving me a formatting error even though I have formatted my code multiple times.

>or initializes?

I implemented the initializers used in the paper. If you would like I can make it so that the initializer can be changed to something other than the default, but this may hurt the performance.

tensorflow_addons/layers/tests/noisy_dense_test.py
LeonShams: Using the fixture would force the TensorSpec in the function to use a mixed policy when all I need is for the layer to do so.

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,249 @@
...
+    @staticmethod
+    def _scale_noise(x):
+        return tf.sign(x) * tf.sqrt(tf.abs(x))
+
+    def call(self, inputs):
+        dtype = self._compute_dtype_object
+        if inputs.dtype.base_dtype != dtype.base_dtype:
+            inputs = tf.cast(inputs, dtype=dtype)
+
+        # Fixed parameters added as the noise
```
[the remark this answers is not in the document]
LeonShams: In short no, it is not necessary to generate the factorized Gaussian noise every time. Instead, there could be a separate function that does this and the user would have to call that whenever they would like to recreate the noise. I decided to just put it here so that the user wouldn't have to worry about recreating the noise and have a seamless transition between Dense and Noisy dense layers for their agent. I chose simplicity over functionality and a little bit of speed, but if you think it would be better to have the functionality and speed then I will go ahead and separate it.

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,249 @@
...
+        dtype = tf.dtypes.as_dtype(self.dtype or K.floatx())
+        if not (dtype.is_floating or dtype.is_complex):
+            raise TypeError(
+                "Unable to build `Dense` layer with non-floating point "
+                "dtype %s" % (dtype,)
+            )
+
+        input_shape = tf.TensorShape(input_shape)
+        self.last_dim = tf.compat.dimension_value(input_shape[-1])
+        sqrt_dim = self.last_dim ** (1 / 2)
```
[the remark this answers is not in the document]
LeonShams: It is all the same, the only difference is the `sqrt_dim = ...` which is needed to create the initializers.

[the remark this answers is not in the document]
bhack: Ok it is only for storage after retrieving.

bhack (approved):
For me it is ok.. Waiting just for a CODEOWNER pass

tensorflow_addons/layers/noisy_dense.py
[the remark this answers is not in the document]
LeonShams: I modified it so that it will generate the noise every call by default, but can be disabled and generated separately with a function call for better performance.

bhack (dismissed):

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,274 @@
...
+)
+from tensorflow.keras import backend as K
+from tensorflow.keras.layers import InputSpec
+
+
+@tf.keras.utils.register_keras_serializable(package="Addons")
+class NoisyDense(tf.keras.layers.Layer):
+    """
+    Like normal dense layer (https://github.com/tensorflow/tensorflow/blob/v2.3.0/tensorflow/python/keras/layers/core.py#L1067-L1233)
+    but random noisy is added to the weights matrix. But as the network improves the random noise is decayed until it is insignificant.
```
WindQAQ, line 31: This does not work when rendering. The first line should be
```
"""Your Title here, do not break line.

foo bar
zzz xxx
```

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,274 @@
...
+
+    A `NoisyDense` layer implements the operation:
+    `output = activation(dot(input, µ_kernel + (σ_kernel * ε_kernel)) + bias)`
```
WindQAQ, line 34: Please get rid of unicode symbol, you can use mathjax expr for rendering math expression.
```
$$
\mathrm{NoisyDense}(x) = \mathrm{activation}(\mathrm{dot}(x, \mu + (\sigma \cdot \epsilon)) + \mathrm{bias})
$$
```

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,274 @@
...
+    where `activation` is the element-wise activation function
+    passed as the `activation` argument, `µ_kernel` is your average weights matrix
```
WindQAQ, line 36: ditto

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,274 @@
...
+    created by the layer, σ_kernel is a weights matrix that controls the importance of
```
WindQAQ, line 37: ditto

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,274 @@
...
+    the ε_kernel which is just random noise, and `bias` is a bias vector created by the layer
```
WindQAQ, line 38: ditto

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,274 @@
...
+    def __init__(
+        self,
+        units,
+        activation=None,
+        use_bias=True,
+        kernel_regularizer=None,
+        bias_regularizer=None,
+        activity_regularizer=None,
+        kernel_constraint=None,
+        bias_constraint=None,
```
WindQAQ, line 89: Please use typecheck for `__init__`. Similar usage can be found in other layers.

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,274 @@
...
+                trainable=True,
+            )
+        else:
+            self.σ_bias = None
+            self.µ_bias = None
+        self.built = True
+
+    @staticmethod
+    def _scale_noise(x):
+        return tf.sign(x) * tf.sqrt(tf.abs(x))
```
WindQAQ, line 184: I don't get the reason that it should be staticmethod. Maybe move it to the upper of class

```python
def _scale_noise(x):
    return tf.sign(x) * tf.sqrt(tf.abs(x))


class NoisyDense(xxx):
    xxx
```

WindQAQ (changes requested):
Some changes needed:

1. use typecheck
2. get rid of all unicode symbols

tensorflow_addons/layers/noisy_dense.py
bhack: Yes I think we need to add this somwhere contribution guidelines.

WindQAQ:
I am also interested in the design. When should users call `remove_noise`? Is it in testing phase? If it is, I propose we should pass `training` argument to `call`.

LeonShams:
> When should users call `remove_noise`

Removing the noise allows you to gauge an agent's performance without the element of chance, so it is often used when testing an agent.

> Is it in the testing phase? If it is, I propose we should pass `training` argument to call

Remove noise is not meant to be used during training only for testing. The idea is that you remove the noise before you start your testing games, then call `reset_noise` once they have finished and you are ready to train again.

WindQAQ:
> Removing the noise allows you to gauge an agent's performance without the element of chance, so it is often used when testing an agent.

Then maybe a better way to handle this is
```python
def call(inputs, training=None):
    # manipulate noise depend on training phase or not
    noise = tf.keras.backend.in_train_phase(_noise_train, _noise_test, training=training)
```
instead of using the kwarg `reset_noise`. Using `training` as name can benefit `model.evaluate` for setting the flag correctly.

bhack:
I suppose that `model(x, training=False)` will not isolate only `reset_noise` control right?

LeonShams:
What I can do is add training as well as remove noise and automatically reset the noise before training. This way if the agent is in testing the noise can be removed manually but reset noise will never need to be called and instead done automatically when training.

bhack:
> What I can do is add training as well as remove_noise and automatically reset the noise if training is true. This way if the agent is in testing the noise can be removed manually but reset_noise will never need to be called and instead done automatically when training.

If this a quite regular pipeline for this domain I think it is ok.

tensorflow_addons/layers/noisy_dense.py
LeonShams: How do you want me to add mathjax expressions to the comments? Or should I just use a raw string and leave it unrendered?

bhack: Generally it was suggested for Markdown files:
https://github.com/tensorflow/docs/blob/master/site/en/community/contribute/docs_style.md#math-in-markdown
But:
> MathJax does not render properly on GitHub.
This notation can be off-putting to unfamiliar developers.

@WindQAQ It would be nice to have something like http://asciimath.org/ to have both source code and rendered MathJax readability

WindQAQ: Either way in docstring is fine to me. For real python code, please use `mu` instead of `µ` for example.

bhack: Yes `mu` is also the asciimath style the then couldbe rendered as a MathJax /cc @lamberta just to know if he have already explored this option.

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,288 @@
...
+
+    Input shape:
+    N-D tensor with shape: `(batch_size, ..., input_dim)`.
+    The most common situation would be
+    a 2D input with shape `(batch_size, input_dim)`.
+
+    Output shape:
+    N-D tensor with shape: `(batch_size, ..., units)`.
+    For instance, for a 2D input with shape `(batch_size, input_dim)`,
+    the output would have shape `(batch_size, units)`.
```
WindQAQ, line 87:
```
    Output shape:
      N-D tensor xxx
```

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,288 @@
...
+      the `kernel` weights matrix.
+    bias_regularizer: Regularizer function applied to the bias vector.
+    activity_regularizer: Regularizer function applied to
+      the output of the layer (its "activation").
+    kernel_constraint: Constraint function applied to
+      the `kernel` weights matrix.
+    bias_constraint: Constraint function applied to the bias vector.
+
+    Input shape:
+    N-D tensor with shape: `(batch_size, ..., input_dim)`.
```
WindQAQ, line 80:
```
    Input shape:
      N-D tensor xxx
```

WindQAQ (changes requested):
I just figure out that lots of codes are copied from `tf.keras.layers.Dense`. Therefore, a better design pattern is that

```python
class NoisyDense(tf.keras.layers.Dense):
    @typechecked
    def __init__(
        self,
        units: int,
        sigma: float = 0.5,
        activation: types.Activation = None,
        use_bias: bool = True,
        kernel_regularizer: types.Regularizer = None,
        bias_regularizer: types.Regularizer = None,
        activity_regularizer: types.Regularizer = None,
        kernel_constraint: types.Constraint = None,
        bias_constraint: types.Constraint = None,
        **kwargs
    ):
        super().__init__(xxxx)
        # DO OTHER THINGS

    def build(self, input_shape):
        # OVERRIDE

    def reset_noise(self):
        # DO SOMETHING

    def remove_noise(self):
        # DO SOMETHING

    def call(self, inputs, reset_noise=True):
        if reset_noise:
            self.reset_noise()
        # TODO(WindQAQ): Replace this with `dense()` once public.
        # prepare your self.kernel, self.bias here.
        return super().call(inputs)

    def get_config(self):
        # OVERRIDE
```

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,290 @@
...
+        # Scale the random noise
+        self.eps_kernel = _scale_noise(eps_i) * _scale_noise(eps_j)
+        self.eps_bias = _scale_noise(eps_j)
+
+    def remove_noise(self):
+        dtype = self._compute_dtype_object
+        self.eps_kernel = tf.zeros([self.last_dim, self.units], dtype=dtype)
+        self.eps_bias = tf.zeros([self.last_dim, self.units], dtype=dtype)
+
+    def call(self, inputs, reset_noise=True, remove_noise=False):
```
WindQAQ, line 216: This is good to me 😃

bhack: Private python :wink:

WindQAQ: Not really private :-( Just hide them from doc LOL

LeonShams:
Regarding the `super().__init__(xxxx)`, what would you like me to do about `self.kernel_initializer` and `self.bias_initializer` since they are not used in `NoisyDense`?

WindQAQ:
> Regarding the `super().__init__(xxxx)`, what would you like me to do about `self.kernel_initializer` and `self.bias_initializer` since they are not used in `NoisyDense`?

Just pass them as `None`. and remove them with `delattr` after it.

LeonShams:
Also, what would you like me to do about `super(NoisyDense, self).get_config()` since it would now config the Dense layer?

WindQAQ:
> Also, what would you like me to do about `super(NoisyDense, self).get_config()` since it will now get the config of Dense?

```python
config = super().get_config()
config.pop("foo") # pop things you do not have
config["bar"] = bar # update things you have additionally
return config
```

LeonShams:
That will raise an error since `kernel_initializer` and `bias_initializer` are removed.

WindQAQ:
> That will raise an error since `kernel_initializer` and `bias_initializer` were deleted.
> 
> Dense `get_config()`:
> 
> ```
> config.update({
>     ...
>     'kernel_initializer':
>         initializers.serialize(self.kernel_initializer),
>     'bias_initializer':
>         initializers.serialize(self.bias_initializer),
>     ...
> }
> ```

Try this
```python
config = super(tf.keras.layers.Dense, self).get_config()
config["foo"] = foo # update your stuff
```

https://colab.research.google.com/drive/14VzUnIZIDadej8s_fHlSKx-ieTs9rttV?usp=sharing

LeonShams:
The first method works well!

>```
>config = super(tf.keras.layers.Dense, self).get_config()
>config["foo"] = foo # update your stuff
>```

But is there a reason why you used `config["foo"]` instead of `config.update({})`?

WindQAQ:
> The first method works well!
> 
> > ```
> > config = super(tf.keras.layers.Dense, self).get_config()
> > config["foo"] = foo # update your stuff
> > ```
> 
> But is there a reason why you used `config["foo"]` instead of `config.update({})`?

Basically no. Just type in rush. `config.update(xxx)` LGTM.

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+def _scale_noise(x):
+    return tf.sign(x) * tf.sqrt(tf.abs(x))
+
+
+@tf.keras.utils.register_keras_serializable(package="Addons")
+class NoisyDense(tf.keras.layers.Dense):
+    r"""Noisy dense layer that inject random noise to
+    the weights of normal dense layer.
```
WindQAQ, line 37: Is this formatted by our formatter? It should be one line so the doc rendering works correctly. Maybe we can use less words like
```
r"""Noisy dense layer.

Describe some details of the this layers in plain text. Maybe use the section 3 in the paper.
See [Noisy Networks for Explanation](https://arxiv.org/pdf/1706.10295.pdf) for more details.
```
And put the whole descriptions below it.

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,252 @@
...
+    def kernel(self):
+        return self.mu_kernel + (self.sigma_kernel * self.eps_kernel)
+
+    @property
+    def bias(self):
+        if self.use_bias:
+            return self.mu_bias + (self.sigma_bias * self.eps_bias)
+
+    # Create the factorised Gaussian noise
+    def _reset_noise(self):
```
WindQAQ, line 203:
```python
def _reset_noise(self):
    """Create the factorised Gaussian noise."""
```

tensorflow_addons/layers/tests/noisy_dense_test.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+import pytest
+import numpy as np
+
+import tensorflow as tf
+from tensorflow import keras
+from tensorflow_addons.utils import test_utils
+from tensorflow_addons.layers.noisy_dense import NoisyDense
+from tensorflow.keras.mixed_precision.experimental import Policy
```
WindQAQ, line 25:
```python
import pytest
import numpy as np
import tensorflow as tf
from tensorflow import keras
from tensorflow.keras.mixed_precision.experimental import Policy

from tensorflow_addons.layers.noisy_dense import NoisyDense
```

tensorflow_addons/layers/tests/noisy_dense_test.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def test_noisy_dense():
```
WindQAQ, line 28: Make this parameterized?
```python
@pytest.mark.parametrize(input_shape, [(3, 2), (3, 4, 2), (None, None, 2), (3, 4, 5, 2)])
```

WindQAQ (commented):
Some documentation improvement needed. Thanks.

tensorflow_addons/layers/noisy_dense.py
LeonShams: Flake8 was complaining that the line was too long so I split it into 2. 

I could make it:
```
Noisy dense layer that injects random noise to the weights of dense layer.
```

instead of:
```
Noisy dense layer that injects random noise to the weights of normal dense layer.
```

WindQAQ: Sure. LGTM :-) Don't forget to update the detailed descriptions and paper link down below.

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -33,16 +33,22 @@ def _scale_noise(x):
...
 
     A `NoisyDense` layer implements the operation:
     $$
     \mathrm{NoisyDense}(x) =
-    \mathrm{activation}(\mathrm{dot}(x, \mu + (\sigma \cdot \eps))
+    \mathrm{activation}(\mathrm{dot}(x, \mu + (\sigma \cdot \epsilon))
     + \mathrm{bias})
     $$
-    with bias only being added if `use_bias` is `True`.
+    Where $\mu$ is the standard weights layer, $\epsilon$ is the factorised
```
WindQAQ, line 48: NIT: lower case `w` -> `where`

LeonShams: Oops, I forgot to change this in the committed version. All fixed!

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,264 @@
...
+        sqrt_dim = self.last_dim ** (1 / 2)
+        if self.last_dim is None:
+            raise ValueError(
+                "The last dimension of the inputs to `Dense` "
+                "should be defined. Found `None`."
+            )
+        self.input_spec = InputSpec(min_ndim=2, axes={-1: self.last_dim})
+
+        self.sigma_init = initializers.Constant(value=self.sigma / sqrt_dim)
+        self.mu_init = initializers.RandomUniform(
```
WindQAQ, line 147: Maybe `sigma_init` and `mu_init` should be only alive in `build`, so just get rid of `self.` from them.

tensorflow_addons/layers/noisy_dense.py
```diff
@@ -0,0 +1,264 @@
...
+    \mathrm{NoisyDense}(x) =
+    \mathrm{activation}(\mathrm{dot}(x, \mu + (\sigma \cdot \epsilon))
+    + \mathrm{bias})
+    $$
+    where $\mu$ is the standard weights layer, $\epsilon$ is the factorised
+    Gaussian noise, and $\sigma$ is a second weights layer which controls
+    $\epsilon$.
+    Note: bias only added if `use_bias` is `True`.
+
+    Example:
```
WindQAQ, line 53: I think there should be an empty line between `Example:` and `>>>` so that rendering can work correctly. (Not 100% sure)

WindQAQ (commented):
@LeonShams Sorry that this should be the last comment. Others LGTM! Thank you.

LeonShams:
No worries, I made the changes, let me know if there is anything else.

bhack (approved):

WindQAQ (approved):

PR #2114, opened 2020-08-24
Added extra requirements for tensorflow, tensorflow-gpu, and tensorflow-cpu

### LINKED ISSUE #2111: List TensorFlow as an "extra" requirement to install the correct TFA for a given TF version
When installing `tensorflow_addons` without specifying an exact version, users may install a version that is incompatible with their existing TensorFlow version.  

As such, users are advised to check the [Python Op Compatibility Matrix](https://github.com/tensorflow/addons#python-op-compatibility-matrix) and manually install the correct version of TFA to be safe.

For most Python projects, this can be taken care of automatically by listing a specific `tensorflow` version(s) as a requirement in the `setup.py` file:

```
install_requires=['tensorflow>=2.2,<2.4']
```

I suspect that this project does not do this because of the various different pip package names TensorFlow uses (e.g., `tf-nightly`, `tensorflow-gpu`, `tensorflow-cpu`).  By having such a requirement, user could accidentally install two different versions of TensorFlow at the same time.

To get around this, we could introduce an extra requirement:

```
extras_require = {'tensorflow': ['tensorflow>=2.2,<2.4']}
```

This way, users will only get automatic dependency resolution when running `pip install tensorflow_addons[tensorflow]`.

The benefit of doing this is that if a new version of TFA is released that no longer supports TensorFlow 2.2 (for example), pip will fall back to installing the previous version of TFA if the user has TF 2.2 installed.

tgaddair:
# Description

Brief Description of the PR:

This restores the functionality removed by #974 but makes it optional, so users can get the behavior when installing:

```
pip install tensorflow_addons[tensorflow]
```

Fixes #2111

## Type of change

Install script enhancement.

# Checklist:

- [X] I've properly [formatted my code according to the guidelines](https://github.com/tensorflow/addons/blob/master/CONTRIBUTING.md#coding-style)
    - [X] By running Black + Flake8
    - [X] By running pre-commit hooks
- [X] This PR addresses an already submitted issue for TensorFlow Addons

# How Has This Been Tested?

```
pip install '.[tensorflow]'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,15 @@ we can only guarantee compatibility with the TensorFlow versions which it was te
 Warnings will be emitted when importing `tensorflow_addons` if your TensorFlow version does not match 
 what it was tested against.
 
+To ensure you install a compatible version of `tensorflow_addons` for your existing version of
+TensorFlow, you can install with the `tensorflow` extra requirement:
+
+```
+pip install tensorflow-addons[tensorflow]
+```
+
+Similar extras exist for `tensorflow-gpu` and `tensorflow-cpu`.
+
 #### Python Op Compatibility Matrix
 | TensorFlow Addons | TensorFlow | Python  |
 |:----------------------- |:---|:---------- |
--- a/README.md
+++ b/README.md
@@ -52,6 +52,15 @@ run the following:
 ```
 pip install tensorflow-addons
 ```
+
+To ensure you have a version of TensorFlow that is compatible with TensorFlow Addons, 
+you can specify the `tensorflow` extra requirement during install:
+
+```
+pip install tensorflow-addons[tensorflow]
+```
+
+Similar extras exist for the `tensorflow-gpu` and `tensorflow-cpu` packages.
  
 
 To use TensorFlow Addons:
@@ -67,15 +76,6 @@ However, there are still a few private API uses within the repository so at the 
 we can only guarantee compatibility with the TensorFlow versions which it was tested against. 
 Warnings will be emitted when importing `tensorflow_addons` if your TensorFlow version does not match 
 what it was tested against.
-
-To ensure you install a compatible version of `tensorflow_addons` for your existing version of
-TensorFlow, you can install with the `tensorflow` extra requirement:
-
-```
-pip install tensorflow-addons[tensorflow]
-```
-
-Similar extras exist for `tensorflow-gpu` and `tensorflow-cpu`.
 
 #### Python Op Compatibility Matrix
 | TensorFlow Addons | TensorFlow | Python  |
--- a/setup.py
+++ b/setup.py
@@ -76,6 +76,8 @@ class BinaryDistribution(Distribution):
 
 
 project_name, version = get_project_name_version()
+min_tf_version = version["MIN_TF_VERSION"]
+max_tf_version = version["MAX_TF_VERSION"]
 setup(
     name=project_name,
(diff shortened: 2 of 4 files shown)


bhack:
Do you need to add something in the README?

tgaddair:
Thanks for taking a look @bhack, I've added some additional docs to the README explaining the usage.

README.md
```diff
@@ -68,6 +68,15 @@ we can only guarantee compatibility with the TensorFlow versions which it was te
 Warnings will be emitted when importing `tensorflow_addons` if your TensorFlow version does not match 
 what it was tested against.
 
+To ensure you install a compatible version of `tensorflow_addons` for your existing version of
+TensorFlow, you can install with the `tensorflow` extra requirement:
+
+```
+pip install tensorflow-addons[tensorflow]
+```
```
seanpmorgan, line 76: Pending the answer to https://github.com/tensorflow/addons/issues/2111#issuecomment-679510888 I think we could even move this up to the primary "Install TFA" block

README.md
```diff
@@ -68,6 +68,15 @@ we can only guarantee compatibility with the TensorFlow versions which it was te
...
+
+Similar extras exist for `tensorflow-gpu` and `tensorflow-cpu`.
```
seanpmorgan, line 78: Would be happier if `tensorflow-gpu` was done away with but I think while it exists this is good to use for dependencies

tensorflow_addons/version.py
```diff
@@ -14,6 +14,10 @@
 # ============================================================================
 """Define TensorFlow Addons version information."""
 
+# Required TensorFlow version [min, max)
+MIN_TF_VERSION = "2.2.0"
+MAX_TF_VERSION = "2.4.0"
+
```
seanpmorgan, line 20: Very happy with the single location

README.md
tgaddair: As discussed in that issue, reworded this section to clarify that the effect will be to ensure a compatible version of TensorFlow is installed, not the reverse (a compatible TFA for your TF install).  Also moved to the `Installation` section as suggested.  Let me know what you think!

zh794390558:
When to merge this?

zh794390558:
https://github.com/tensorflow/addons/issues/864

tgaddair:
@seanpmorgan ready for review whenever you have some time.

setup.py
```diff
@@ -86,6 +88,11 @@ def has_ext_modules(self):
     packages=find_packages(),
     ext_modules=get_ext_modules(),
     install_requires=Path("requirements.txt").read_text().splitlines(),
+    extras_require={
+        "tensorflow": [f"tensorflow>={min_tf_version},<{max_tf_version}"],
+        "tensorflow-gpu": [f"tensorflow-gpu>={min_tf_version},<{max_tf_version}"],
+        "tensorflow-cpu": [f"tensorflow-cpu>={min_tf_version},<{max_tf_version}"],
```
seanpmorgan, line 94: This is failing CI and flake. Since we support py35+ could you use `.format` instead of f-strings

seanpmorgan: Also please merge with the latest master branch when you get a chance and it'll fix the other fails

tgaddair: Updated.  Thanks!

tgaddair:
Thanks for the review, @seanpmorgan!  Looks like all tests are passing now.

PR #2102, opened 2020-08-19
add test for reloading lookahead optimizer

BinyanHu:
# Description

Brief Description of the PR: add the test for saving and reloading a Lookahead optimizer. The added test function reveals the bug that the states of a lookahead's inner optimizer are not properly saved and reloaded (#2094). So the test can not pass currently.

## Type of change

- [ ] Bug fix
- [ ] New Tutorial
- [ ] Updated or additional documentation
- [x] Additional Testing
- [ ] New Activation and the changes conform to the [activation contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/activations/README.md#contribution-guidelines)
- [ ] New Callback and the changes conform to the [callback contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/callbacks/README.md#contribution-guidelines)
- [ ] New Image addition and the changes conform to the [image op contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/README.md#contribution-guidelines)
- [ ] New Layer and the changes conform to the [layer contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/layers/README.md#contribution-guidelines)
- [ ] New Loss and the changes conform to the [loss contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/losses/README.md#contribution-guidelines)
- [ ] New Metric and the changes conform to the [metric contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/metrics/README.md#contribution-guidelines)
- [ ] New Optimizer and the changes conform to the [optimizer contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/optimizers/README.md#contribution-guidelines)
- [ ] New RNN Cell and the changes conform to the [rnn contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/rnn/README.md#contribution-guidelines)
- [ ] New Seq2seq addition and the changes conform to the [seq2seq contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/seq2seq/README.md#contribution-guidelines)
- [ ] New Text addition and the changes conform to the [text op contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/text/README.md#contribution-guidelines)

# Checklist:

[...]

Issue #2094, opened 2020-08-15
Weights of Inner Optimizers Not Saved


BinyanHu:
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Linux Ubuntu 16.04 & Windows 10
- TensorFlow version and how it was installed (source or binary): 2.3.0 from source
- TensorFlow-Addons version and how it was installed (source or binary): 0.11.1 from source
- Python version: 3.7
- Is GPU used? (yes/no): yes

**Describe the bug**

Resume a training process needs the restoration of the optimizer states to continue training RIGHT from the previous state without any loss of accuracy. Currently, the keras interface of saving model `keras.Model.save_weights`  checkpoints both the network parameters and the optimizer weights. However, when an optimizer is wrapped inside another, its weights can not be saved by this mean.

For example, when I was trying to use the Ranger optimizer, which is constructed by wrapping RAdam with Lookahead:
```
optimizer = tfa.optimizers.Lookahead(
    tfa.optimizers.RectifiedAdam()
)
```
I noticed a performance drop on resuming training. I found that the weights of the inner RAdam were not saved into the checkpoint. (I checked the `.index` file in the checkpoint folder and there are no variable names like "m" and "v", only "slow", which is the weights of Lookahead). Therefore, after loading the weights from file and restart fitting, the weights of RAdam are randomly reinitialized. This could because the weights of the inner optimizer are not automatically tracked.

**Experiments**

I trained the two LeNets on the FashionMNIST dataset. All the configurations are the same except for the optimizers. Both training are interrupted in the middle and then resumed.

![image](https://user-images.githubusercontent.com/26186061/90313678-f46e6d00-df40-11ea-9000-3fa8e429818d.png)
*Fig. TensorBoard. Blue: Ranger (Lookahead+RAdam), orange: RAdam.*

Note the "bump" of the Ranger curve caused by the reinitialization of RAdam weights. Apparently, the weights of the inner optimizer are not correctly saved.

bhack:
Can you prepare a minimal PR with a new test to cover your case?

bhack:
So that we could check if It is similar to https://github.com/tensorflow/addons/issues/1911

BinyanHu:
> So that we could check if It is similar to #1911

Our issues are similar. But the real problem is on the missing weights of the inner RAdam optimizer. 

First, I reran my program with `status.assert_consumed()`, and the errors are as follows:
```
AssertionError:
[...]

PR #2126, opened 2020-08-27
Base optimizer tracking

bhack:
# Description

Brief Description of the PR:
Inner optimizer is not tracked for checkpoint

Fixes # (issue)
https://github.com/tensorflow/addons/issues/2094

## Type of change

- [x ] Bug fix
- [ ] New Tutorial
- [ ] Additional Testing

# Checklist:

- [X ] I've properly [formatted my code according to the guidelines](https://github.com/tensorflow/addons/blob/master/CONTRIBUTING.md#coding-style)
    - [X ] By running Black + Flake8
    - [ X] By running pre-commit hooks
- [X] This PR addresses an already submitted issue for TensorFlow Addons
- [X] I have added tests that prove my fix is effective or that my feature works

# How Has This Been Tested?
Tested in https://github.com/tensorflow/addons/pull/2102

It still need to pass `save` and `load_model` case.

### DIFF
--- a/tensorflow_addons/optimizers/average_wrapper.py
+++ b/tensorflow_addons/optimizers/average_wrapper.py
@@ -46,6 +46,7 @@ class AveragedOptimizerWrapper(tf.keras.optimizers.Optimizer, metaclass=abc.ABCM
             raise TypeError("sequential_update must be of bool type")
 
         self._optimizer = optimizer
+        self._track_trackable(self._optimizer, "awg_optimizer")
 
         if sequential_update is not None:
             warnings.warn(
--- a/tensorflow_addons/optimizers/lookahead.py
+++ b/tensorflow_addons/optimizers/lookahead.py
@@ -80,6 +80,7 @@ class Lookahead(tf.keras.optimizers.Optimizer):
         self._set_hyper("sync_period", sync_period)
         self._set_hyper("slow_step_size", slow_step_size)
         self._initialized = False
+        self._track_trackable(self._optimizer, 'base_optimizer')
 
     def _create_slots(self, var_list):
         self._optimizer._create_slots(
--- a/tensorflow_addons/optimizers/lookahead.py
+++ b/tensorflow_addons/optimizers/lookahead.py
@@ -80,7 +80,7 @@ class Lookahead(tf.keras.optimizers.Optimizer):
         self._set_hyper("sync_period", sync_period)
         self._set_hyper("slow_step_size", slow_step_size)
         self._initialized = False
-        self._track_trackable(self._optimizer, "base_optimizer")
+        self._track_trackable(self._optimizer, "lh_base_optimizer")
 
     def _create_slots(self, var_list):
         self._optimizer._create_slots(
--- a/tensorflow_addons/optimizers/lookahead.py
+++ b/tensorflow_addons/optimizers/lookahead.py
@@ -80,7 +80,7 @@ class Lookahead(tf.keras.optimizers.Optimizer):
         self._set_hyper("sync_period", sync_period)
         self._set_hyper("slow_step_size", slow_step_size)
         self._initialized = False
-        self._track_trackable(self._optimizer, 'base_optimizer')
+        self._track_trackable(self._optimizer, "base_optimizer")
 
     def _create_slots(self, var_list):
         self._optimizer._create_slots(


bhack:
/cc @reedwm Can you give us a feedback here. What we need to correctly support Keras `save` and `load_model`?
Need we to introduce something like your `_DelegatingTrackableMixin` in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/mixed_precision/experimental/loss_scale_optimizer.py#L58?

I see that in that case in Tensorflow your are just [testing checkpoint](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/mixed_precision/experimental/loss_scale_optimizer_test.py#L439) but not `save` and `load_model`.

See user proposed tests in https://github.com/tensorflow/addons/pull/2102

reedwm:
The purpose of `_DelegatingTrackableMixin` is so the checkpoint format of the `LossScaleOptimizer` is the same as the checkpoint format of the inner optimizer. This allows you to, e.g., save a checkpoint of float32 model with an Adam optimizer, then load the checkpoint into the equivalent mixed precision model with a LossScaleOptimizer wrapping an Adam optimizer.

I'm not very familiar with the `Lookahead` optimizer, but it seems this use case doesn't apply: I don't think people would want to train partially with `Lookahead` and partially without, and this would be hard anyway since `Lookahead` has an extra set of "slow weights". If I'm correct, the current approach in this PR is fine.

Note we test a SavedModel with LossScaleOptimizer [here](https://github.com/tensorflow/tensorflow/blob/b73c9f1236f405268a7406416ed1672db68c3d2e/tensorflow/python/keras/mixed_precision/experimental/keras_test.py#L1064).

bhack:
@reedwm Thank you for the feedback. I am going to check your `SaveModel` test that I missed.

bhack:
@reedwm I've tried to expand your test more in the style of the user proposed one and is passing on Tensorflow: https://github.com/tensorflow/tensorflow/pull/42749

In our Lookahead case at https://github.com/tensorflow/addons/pull/2102 instead is passing only with `Checkpoint`, with `save_weight/load_weight` cases or `model.save` and `model.load_model` but in h5. 
So I think that we are still missing something for the `model.save` in `tf` format.

Any hint?

reedwm:
Hmm I'm not sure. What is the error?

Just a guess, but maybe the issue is because you return additional weights from the `weights` property:

https://github.com/tensorflow/addons/blob/ef8b0c9944b363fd2a78fc4d973f5a69486b8098/tensorflow_addons/optimizers/lookahead.py#L135-L137

AFAIK, no other optimizer does this. Maybe it's causing SavedModel to be confused.

Also last time I checked, SavedModel does not save slot variables, but checkpoints do. Make sure your SavedModel test is not relying on slot variables being restored.

Also note LossScaleOptimizer used to use an approach similar to this PR. https://github.com/tensorflow/tensorflow/commit/f9e99f4dca729f6c21155641103165b6dfac4730 changed it, but before that commit SavedModel still worked, so this approach should be possible.

/CC @k-w-w any ideas what the issue could be?

bhack:
As you can see we have slot in https://github.com/tensorflow/addons/blob/ef8b0c9944b363fd2a78fc4d973f5a69486b8098/tensorflow_addons/optimizers/lookahead.py#L84

bhack:
I still see this @allenlavoie commit in the current master: 
https://github.com/tensorflow/tensorflow/commit/147e6777068ea569cbecbd270b8b07c648bae09b

Is it the motivation of this claim?
https://github.com/tensorflow/tensorflow/blame/master/tensorflow/python/keras/engine/training.py#L1986-L1987

allenlavoie:
I don't think that's related. IIRC it was because `Checkpoint(model=tf.keras.Model()).save` is not compatible with `Model().save_weights` (since `Model` is the root of the save in the second case).

What's the error/issue if you try to save with SavedModel?

bhack:
The optimizer or the base/inner optimizer (or both) don't seems to be in the correct state after `load_model` in `tf` format.
Instead It seems ok in h5, checkpoint or  `save_weights` and `load_weights`.

allenlavoie:
Hrm. Does tf.saved_model.save / tf.saved_model.load have the same issue? There's a bit of extra logic in Model.save/load_model that could be causing issues here.

bhack:
With `tf.saved_model.save` is the same.

bhack:
@allenlavoie What is exactly the `RestoredOptimizer` object?

bhack:
Cause I don't understand this case https://github.com/tensorflow/tensorflow/blame/master/tensorflow/python/keras/saving/saving_utils.py#L179

allenlavoie:
RestoredOptimizer is just a container for holding the restored slot variables. If nothing owned the tf.Variable objects they would just go out of scope immediately and be deleted. It doesn't implement any of the other Optimizer methods.

Maybe this line is the issue: https://github.com/tensorflow/tensorflow/blob/be2870d0db035d12a207039922d76ae908b1101c/tensorflow/python/keras/optimizer_v2/optimizer_v2.py#L1391

When we tell SavedModel to load something back as a Python object, it takes a Python function to set attributes with. It's calling _set_hyper and passing the inner optimizer, and maybe that's not doing anything reasonable. You could try using a more complicated setter that e.g. typechecks for OptimizerV2s and uses setattr+_track_trackable or something.

bhack:
I don't find so much entries of `ResotredOptimizer` object in the code. Where is it created?

What is the logic of:
```
      if isinstance(model.optimizer, optimizer_v2.RestoredOptimizer):
        raise NotImplementedError(
            'As of now, Optimizers loaded from SavedModel cannot be saved. '
            'If you\'re calling `model.save` or `tf.keras.models.save_model`,'
            ' please set the `include_optimizer` option to `False`. For '
            '`tf.saved_model.save`, delete the optimizer from the model.')
```

bhack:
Ok nevermind I've seen the logic of RestoredOptimizer.

bhack:
I've just extracted a very small code snippet removing `Lookahead` wrapping. 
If you see it is working with the plain `optimizer=SGD` but not with `optimizer=adam` so I suppose it is something related to slots:

```
import os
import numpy as np
import tensorflow as tf
import tempfile


from tensorflow_addons.optimizers import Lookahead

def _init_model(optimizer, init_w):
    dense = tf.keras.layers.Dense(input_shape=(3,), units=1)
    model.add(dense)
    dense.set_weights([init_w, np.zeros(1,)])
   print(dense.get_weights)
    model.compile(optimizer, loss="mse")
    return model

optimizer="adam"
np.random.seed(0)
x = np.random.standard_normal((10000, 3))
w = np.random.standard_normal((3, 1))
y = np.dot(x, w) + np.random.standard_normal((10000, 1)) * 1e-4

init_w = np.random.standard_normal((3, 1))

model = _init_model(optimizer, init_w)
model.fit(x, y, epochs=2, shuffle=False)
with tempfile.TemporaryDirectory() as ckpt_dir:
    new_model = _init_model(optimizer, init_w)
    new_model.fit(x, y, epochs=1, shuffle=False)
    ckpt_path = os.path.join(ckpt_dir)
    new_model.save(ckpt_path)
    new_model = _init_model(optimizer, init_w)
    new_model = tf.keras.models.load_model(ckpt_path)
    new_model.fit(x, y, epochs=1, shuffle=False)
    assert np.allclose(model.predict(x), new_model.predict(x))
```

bhack:
@reedwm With Adam Is also failing the tensorflow test on tour test if the change is correct: https://github.com/tensorflow/tensorflow/pull/42749.

reedwm:
I commented in https://github.com/tensorflow/tensorflow/pull/42749#issuecomment-683964931. Since slot variables are not restored and Adam has slot variables, the sample code you gave is expected to fail.

bhack:
Ok I commented also there.

bhack:
I think that this could be merged as is. 
Instead we are drafting a general warning on Tensorflow at https://github.com/tensorflow/tensorflow/pull/42846 for the `SavedModel` case.

PR #2137, opened 2020-08-30
Add experimental_aggregate_gradients support

fsx950223:
Similar #1924


tensorflow_addons/optimizers/tests/moving_average_test.py
```diff
@@ -259,3 +259,15 @@ def apply_gradients():
 
     np.testing.assert_allclose(var.read_value(), [0.8, 1.8])
     np.testing.assert_allclose(ema_var.read_value(), [0.9, 1.9])
+
+
+@pytest.mark.usefixtures("maybe_run_functions_eagerly")
+def test_model_mixed_precision():
+    tf.keras.mixed_precision.experimental.set_policy("mixed_float16")
```
WindQAQ, line 266: We should use pytest fixture here to reset policy after the test. See https://github.com/tensorflow/addons/blob/master/tensorflow_addons/seq2seq/tests/decoder_test.py#L177-L179

Issue #2004, opened 2020-07-15
Focal loss migration


AakashKumarNain:
@tanzhenyu moving the discussion regarding focal loss from #914  to here. Here is a working colab for the cases you asked.

https://colab.research.google.com/drive/1JIaUVOHC0rTsBWKK-oTBv5qZJ5zLJzTQ?usp=sharing

bhack:
Focal loss now is at https://github.com/tensorflow/models/blob/master/official/vision/keras_cv/losses/focal_loss.py

tanzhenyu:
Thanks @AakashKumarNain . I don't have access to the colab. We do provided a keras_cv focal loss. Let us know if the implementation serves your need.

AakashKumarNain:
Thanks @tanzhenyu . The implementation is good but I have one concern. This loss should be included in core keras because:
1. It is not limited to object detection or computer vision only. Anything that involves cross-entropy loss can leverage focal loss as well. Including it in a `cv` specific package sends a message that it works only for CV which isn't true.
2. Focal loss was introduced in 2017. It has enough citations to be included in the core.

Let me know what you think

seanpmorgan:
TensorFlow Addons is transitioning to a minimal maintenance and release mode. New features will not be added to this repository. For more information, please see our public messaging on this decision:
[TensorFlow Addons Wind Down](https://github.com/tensorflow/addons/issues/2807)

Please consider sending feature requests / contributions to other repositories in the TF community with a similar charters to TFA:
[Keras](https://github.com/keras-team/keras)
[Keras-CV](https://github.com/keras-team/keras-cv)
[Keras-NLP](https://github.com/keras-team/keras-nlp)

PR #2156, opened 2020-09-10
Update migration strategy

seanpmorgan:
# Description
Updates the migration strategy.

cc @dynamicwebpaige @karmel @tomerk @tanzhenyu

### DIFF
--- a/MIGRATION_TO_CORE.md
+++ b/MIGRATION_TO_CORE.md
@@ -1,28 +1,11 @@
-# Migration From TF-Addons To TensorFlow Core
+# Migration From TF-Addons To TensorFlow Core / Keras
 
-### In-Progress & Previous Migrations:
-https://github.com/tensorflow/addons/projects/2/
-
-### Process 
-1. Create an issue in TensorFlow Addons for a candidate that you think should be 
-migrated. 
-2. The SIG will evaluate the request and add it to the `Potential Candidates` section 
-of our GitHub project.
-3. If it's agreed that a migration makes sense, an RFC needs to be written to discuss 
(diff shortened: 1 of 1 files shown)


MIGRATION_TO_CORE.md
```diff
@@ -1,28 +1,11 @@
...
-4. If approved, a pull request must move the addition along with proper tests.
-5. After merging, the addition will be replaced with an alias to the core function 
-if possible.
-6. If an alias is not possible (e.g. large parameter changes), then a deprecation 
-warning will be added and eventual removed after 2 releases.
+# Migration From TF-Addons To TensorFlow Core / Keras
 
+**Given the challenges of external SIG coordinating with internal roadmaps, a new 
+process has been put in place for the core TF and Keras teams to handle migration 
+and deprecation of Addons components. If you believe there is a strong candidate for
```
bhack, line 5: I think the real problem here is about expecting a deprecation PR. 
I think it is plausible to expect this if we have a "real" migration approach but instead it will be very hard if we have a duplicative process in the "ecosystem". 
When you start to duplicate something it is hard to expect that you will care about a deprecation cause probably in the duplication act implicitly you didn't care about checking in the ecosystem. 
It is not totally correlated cause there could be real cases when you are constrained to reinvent the wheel but I suppose there is a very strong relationship between the duplication behaviour and the lack of deprecation activities.

bhack: Just to make a pratical use case:
https://github.com/tensorflow/addons/issues/2004#issuecomment-695345816

PR #2163, opened 2020-09-14
Fixing minor issues from #969

hyang0129:
Fixing minor issues from #969. 

- Update code owners 
- Fix typo in class name


hyang0129:
let's try this again. I think I did the pr wrong

PR #2169, opened 2020-09-17
Kendall's Tau metric, based loosely on scipy.

sorensenjs:
# Description

Adds Kendall's Tau, a measure of ordinal concordance as a Keras compatible metric,

## Type of change

- [ ] Bug fix
- [ ] New Tutorial
- [ ] Additional Testing
- [ x ] New Metric and the changes conform to the [metric contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/metrics/README.md#contribution-guidelines)

# Checklist:

- [ ] I've properly [formatted my code according to the guidelines](https://github.com/tensorflow/addons/blob/master/CONTRIBUTING.md#coding-style)

# How Has This Been Tested?

Multiple tests have been included, and because this metric exists in non-tf form in scipy it's easy
to verify correctness.

*

### DIFF
--- a/tensorflow_addons/metrics/kendalls_tau.py
+++ b/tensorflow_addons/metrics/kendalls_tau.py
@@ -14,10 +14,6 @@
 # ==============================================================================
 """Implements Kendall's Tau metric and loss."""
 
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
-
 import numpy as np
 import tensorflow.compat.v2 as tf
 from tensorflow_addons.metrics.utils import MeanMetricWrapper
--- a/tensorflow_addons/metrics/kendalls_tau.py
+++ b/tensorflow_addons/metrics/kendalls_tau.py
@@ -14,8 +14,10 @@
 # ==============================================================================
 """Implements Kendall's Tau metric and loss."""
 
+import warnings
+
 import numpy as np
-import tensorflow.compat.v2 as tf
+import tensorflow as tf
 from tensorflow_addons.metrics.utils import MeanMetricWrapper
 from tensorflow_addons.utils.types import TensorLike
 
@@ -84,10 +86,12 @@ def kendalls_tau(y_true: TensorLike, y_pred: TensorLike) -> tf.Tensor:
     y_pred = tf.reshape(y_pred, [-1])
     y_pred.shape.assert_is_compatible_with(y_true.shape)
     if tf.equal(tf.size(y_true), 0) or tf.equal(tf.size(y_pred), 0):
+        warnings.warn('y_true and y_pred tensors are not the same size.')
         return np.nan
     perm = tf.argsort(y_true)
     n = tf.shape(perm)[0]
     if tf.less(n, 2):
+        warnings.warn('Scalar tensors have no defined ordering.')
         return np.nan
 
     left = 0
--- a/tensorflow_addons/metrics/kendalls_tau.py
+++ b/tensorflow_addons/metrics/kendalls_tau.py
@@ -16,7 +16,6 @@
 
 from __future__ import absolute_import
 from __future__ import division
-from __future__ import google_type_annotations
 from __future__ import print_function
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


tensorflow_addons/metrics/kendalls_tau.py
```diff
@@ -0,0 +1,191 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Implements Kendall's Tau metric and loss."""
+
+import numpy as np
+import tensorflow.compat.v2 as tf
```
AakashKumarNain, line 18: Avoid using `compat` module

tensorflow_addons/metrics/kendalls_tau.py
```diff
@@ -0,0 +1,191 @@
...
+
+    Returns:
+        Kendell's Tau, the 1945 tau-b formulation that ignores ordering of
+        ties, as a scalar Tensor.
+    """
+    y_true = tf.reshape(y_true, [-1])
+    y_pred = tf.reshape(y_pred, [-1])
+    y_pred.shape.assert_is_compatible_with(y_true.shape)
+    if tf.equal(tf.size(y_true), 0) or tf.equal(tf.size(y_pred), 0):
+        return np.nan
```
AakashKumarNain, line 87: Raise a `warning` first

tensorflow_addons/metrics/kendalls_tau.py
```diff
@@ -0,0 +1,191 @@
...
+    perm = tf.argsort(y_true)
+    n = tf.shape(perm)[0]
+    if tf.less(n, 2):
+        return np.nan
```
AakashKumarNain, line 91: Same. Raise a warning first

AakashKumarNain:
Thanks for the contribution. I am not very familiar with `scipy` implementation. @WindQAQ @bhack are you familiar with this one?

tensorflow_addons/metrics/kendalls_tau.py
sorensenjs: Done, removed keras.compat and future imports as well.

tensorflow_addons/metrics/kendalls_tau.py
sorensenjs: Done

tensorflow_addons/metrics/kendalls_tau.py
sorensenjs: Done.

AakashKumarNain:
LGTM! Thanks again. Some test are failing but they are on our end and not yours. Once we fix them, I think we can merge this.

jvdillon:
Hello! I'd love to see this in TFP!

Id also love to see the top-k loss, as described here: https://arxiv.org/pdf/1206.5280.pdf

tensorflow_addons/metrics/kendalls_tau.py
```diff
@@ -0,0 +1,195 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
jvdillon, line 1: 2020

tensorflow_addons/metrics/kendalls_tau.py
```diff
@@ -0,0 +1,195 @@
...
+
+    Returns:
+        A tuple consisting of a int32 scalar that counts the number of
+        exchanges required to produce a sorted permutation, and a tf.int32
+        Tensor that contains the ordering of y values that are sorted.
+    """
+    exchanges = 0
+    num = tf.size(y)
+    k = tf.constant(1, tf.int32)
+    while tf.less(k, num):
```
jvdillon, line 40: Since we are a general purpose library, we need to write everything so that it does not presume eager execution.  Minimally this means no python controlflow (eg, `for`, `while`, and `if` are not allowed).

More generally, this has a code nature that would be quite inefficient in a SIMD regime.  Perhaps we could examine the full quadratic comparison? Eg,

`0.5 * tf.math.reduce_mean(score[..., tf.newaxis] < score[..., tf.newaxis, :], axis=-1)`

In general this would be quite large. I therefore recommend extending this idea to "chunks" using a `tf.while_loop` to reduce them. In so doing we can handle many cases in memory yet also bound the memory by way of chunks.

Thoughts?

jvdillon (commented):
Ive left some initial comments, should you be interested in moving it to TFP.

bhack:
@jvdillon Thanks, I have question It seems that the TFP repository currently doesn't hosts metrics right? We could have case like this or our [Matthews correlation](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/metrics/matthews_correlation_coefficient.py) impl where the user could need to consume it as inherited from `tf.keras.metrics.Metric` object.
So if the ops is in TFP but the metric is here we need to depend on TFP. Do you plan to support metrics directly?

jvdillon:
We dont have keras metrics but we do have standalone functions which might
be regarded as metrics. We strive to be as library agnostic as possible. If
possible we prefer simple functions and have a strict policy that all
functions must support batch input.  Take a look at other tfp.stats for
details.

bhack:
@jvdillon Ok but I see that you have Keras layers and optimizers in TFP. 
I think that it is ok to have an op like this in TFP but then why not to have also metric that wraps the operator (when it is useful to be consumed as metrics)?
I think that It will be hard for TF Addons to depend on TFP currently just to implement metrics.

jvdillon:
On Tue, Sep 22, 2020 at 10:53 AM bhack <notifications@github.com> wrote:

> @jvdillon <https://github.com/jvdillon> Ok but I see that you have Keras
> layers and optimizers in TFP.
> I think that it is ok to have an op like this in TFP but then why not to
> have also metric that wraps the operator (when it is useful to be consumed
> as metrics)?
> I think that It will be hard for TF Addons to depend on TFP currently just
> to implement metrics.
>

I will bring this question to the team. In the meantime, maybe proceed with
adding the core function?

bhack:
@jvdillon It is ok for me. 
@AakashKumarNain @sorensenjs what do you think?

P.s. Are you interested also in https://github.com/tensorflow/addons/blob/master/tensorflow_addons/metrics/matthews_correlation_coefficient.py?

brianwa84:
I guess keras metrics have some streaming functionality? At least the old tf.metrics did.
Note that we have a similar code for streaming reduction baking under tfp.experimental.mcmc (covariance, moments, etc). This uses tensors instead of Variables, though.

tensorflow_addons/metrics/kendalls_tau.py
sorensenjs: Only after I finished writing this did I fully examine the implementation of the AUC metrics in Keras, which use a bucketized approximation, not that dissimilar to the two-dimensional bucket approach presented in https://arxiv.org/abs/1712.01521 which would be much more amenable to streaming. (OTOH, I'm unclear why _Metrics_ need to be streaming, usually the classifier labels is a space much smaller than the input features.)

I also note that I should have looked at the completely rewritten scipy implementation which works differently, but still requires sorting by both of the full inputs which seems largely incompatible with tf Metrics design.
https://github.com/scipy/scipy/blob/01d8bfb6f239df4ce70c799b9b485b53733c9911/scipy/stats/stats.py#L4452

I suspect a reasonable thing to do here is to probably fork this into two separate efforts, one for tfp which would focus on the explicit exact solution with potential backoff to approximate, and a tensorflow addons submission that focuses on implementing a streaming approximation. I think I can manage that, but it's going to take some time - and at present I don't have a good sense of which of these would be easier to do first.

AakashKumarNain:
> I guess keras metrics have some streaming functionality? At least the old tf.metrics did.
> Note that we have a similar code for streaming reduction baking under tfp.experimental.mcmc (covariance, moments, etc). This uses tensors instead of Variables, though.

@brianwa84 Yes, the `Metric` api has streaming functionality. 
PS: I am okay with adding the core functionality for the time being. 

cc: @seanpmorgan

bhack:
@seanpmorgan Before we lost this PR I am still oriented to include this TFP.

tensorflow_addons/metrics/kendalls_tau.py
sorensenjs: Hi - sorry I'm back to this finally - I think from the discussion this PR should be moved to tfp,
most likely in the https://github.com/tensorflow/probability/tree/master/tensorflow_probability/python/stats directory.

However, I think I need to make a significant redesign to comply with the suggestion of removing the control flow.

I'm not quite able to distill what the consensus is here for next steps, maybe I should do all of these:

- Make a tfp PR?
- Reimplement this as a streaming algorithm per the previously mentioned arxiv paper.
- Change this current PR to use tf.while(), possibly with chunks.

bhack:
@jvdillon @brianwa84  Do you have permission to migrate this in TFP?

brianwa84:
I don't know what you mean by "migrate". I think you can just open a new PR against TFP.

For now, we could perhaps put something like this under `tfp.experimental.stats`, where we have other streaming metrics like covariance etc. Or you could modify to drop the streaming component and simply compute on a single tensor of observations, which would be more consistent w/ e.g. `tfp.stats.covariance`. We would likely want to eliminate the for loops for TFP. If you really need a loop, we usually use `tf.while_loop` or `tf.scan` to avoid bloating the `tf.function` graph.

bhack:
> I don't know what you mean by "migrate". 

Yes I supposed that we had a related issue about this features as it was mandatory in our contribution policy. But in this case we have a PR directly.

> For now, we could perhaps put something like this under tfp.experimental.stats, where we have other streaming metrics like covariance etc. Or you could modify to drop the streaming component and simply compute on a single tensor of observations, which would be more consistent w/ e.g. tfp.stats.covariance. We would likely want to eliminate the for loops for TFP. If you really need a loop, we usually use tf.while_loop or tf.scan to avoid bloating the tf.function graph.

The plan is ok for me 

@sorensenjs 
Can you open the PR at https://github.com/tensorflow/probability/ ?

sorensenjs:
Created https://github.com/tensorflow/probability/pull/1147, following up there with requested changes from this thread.

sorensenjs:
This was added to tensorflow probability
https://github.com/tensorflow/probability/commit/9999dd41df50157e09656d07b2d4dd49136a9288?branch=9999dd41df50157e09656d07b2d4dd49136a9288&diff=unified

See https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/stats/kendalls_tau.py for more details.

bhack:
It would be nice if it could be exposed as metrics.

sorensenjs:
Kendall's tau doesn't fit the metrics streaming style well. There are approximate versions (see https://arxiv.org/pdf/1712.01521.pdf) that are more interesting and well suited to the metrics. I might try doing that.

bhack:
> Kendall's tau doesn't fit the metrics streaming style well. There are approximate versions (see https://arxiv.org/pdf/1712.01521.pdf) that are more interesting and well suited to the metrics. I might try doing that.

Interesting. Thanks for the resoruce

sorensenjs:
I've created a completely new streaming based algorithm for this abandoned PR
https://github.com/tensorflow/addons/pull/2423

PR #2180, opened 2020-09-25
Fix MultiOptimizer list of layers

### LINKED ISSUE #2178: Error in MultiOptimizer when layers list are used
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Ubuntu 18.04
- TensorFlow version and how it was installed (source or binary): docker tensorflow 
- TensorFlow-Addons version and how it was installed (source or binary): binary
- Is GPU used? (yes/no):

**Describe the bug**

Documentation of MultiOptimizer explains that list of layers can be used but it results in an error.

https://github.com/tensorflow/addons/blob/c7b867ae6c01e9ac3dbb8c3408f1308b41acfc9b/tensorflow_addons/optimizers/discriminative_layer_training.py#L38

https://github.com/tensorflow/addons/blob/c7b867ae6c01e9ac3dbb8c3408f1308b41acfc9b/tensorflow_addons/optimizers/discriminative_layer_training.py#L52

```python

import tensorflow as tf
import numpy as np


model = tf.keras.Sequential(
        [tf.keras.Input(shape=[1]), tf.keras.layers.Dense(1), tf.keras.layers.Dense(1), tf.keras.layers.Dense(1)]
    )

x = np.array(np.ones([100]))
y = np.array(np.ones([100]))

weights_before_train = (
    model.layers[0].weights[0].numpy(),
    model.layers[1].weights[0].numpy(),
)

opt1 = tf.keras.optimizers.Adam(learning_rate=1e-3)
opt2 = tf.keras.optimizers.SGD(learning_rate=0)

opt_layer_pairs = [(opt1, model.layers[0]), (opt2, model.layers[1:])]

loss = tf.keras.losses.MSE
optimizer = tfa.optimizers.MultiOptimizer(opt_layer_pairs)

model.compile(optimizer=optimizer, loss=loss)
```
Error produced

```
  File "test_multi_optimizer.py", line 24, in <module>
    optimizer = tfa.optimizers.MultiOptimizer(opt_layer_pairs)
  File "/usr/local/lib/python3.6/dist-packages/typeguard/__init__.py", line 840, in wrapper
    retval = func(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/tensorflow_addons/optimizers/discriminative_layer_training.py", line 92, in __init__
    for opt, layer in optimizers_and_layers
  File "/usr/local/lib/python3.6/dist-packages/tensorflow_addons/optimizers/discriminative_layer_training.py", line 92, in <listcomp>
    for opt, layer in optimizers_and_layers
  File "/usr/local/lib/python3.6/dist-packages/tensorflow_addons/optimizers/discriminative_layer_training.py", line 142, in create_optimizer_spec
    ), "Object passed is not an instance of tf.keras.layers.Layer nor tf.keras.Model"
AssertionError: Object passed is not an instance of tf.keras.layers.Layer nor tf.keras.Model
```

WindQAQ:
# Description

Fixes #2178 

## Type of change

- [x] Bug fix
- [ ] New Tutorial
- [ ] Additional Testing

# Checklist:

# How Has This Been Tested?

Additional test.

### DIFF
--- a/tensorflow_addons/optimizers/discriminative_layer_training.py
+++ b/tensorflow_addons/optimizers/discriminative_layer_training.py
@@ -14,7 +14,7 @@
 # ==============================================================================
 """Discriminative Layer Training Optimizer for TensorFlow."""
 
-from typing import Union
+from typing import List, Union
 
 import tensorflow as tf
 from typeguard import typechecked
@@ -29,50 +29,46 @@ class MultiOptimizer(tf.keras.optimizers.Optimizer):
 
     Each optimizer will optimize only the weights associated with its paired layer. This can be used
     to implement discriminative layer training by assigning different learning rates to each optimizer
-    layer pair. (Optimizer, list(Layers)) pairs are also supported. Please note that the layers must be
-    instantiated before instantiating the optimizer.
+    layer pair. `(tf.keras.optimizers.Optimizer, List[tf.keras.layers.Layer])` pairs are also supported.
+    Please note that the layers must be instantiated before instantiating the optimizer.
 
     Args:
         optimizers_and_layers: a list of tuples of an optimizer and a layer or model. Each tuple should contain
-            exactly 1 instantiated optimizer and 1 object that subclasses tf.keras.Model or tf.keras.Layer. Nested
+            exactly 1 instantiated optimizer and 1 object that subclasses `tf.keras.Model` or `tf.keras.layers.Layer`. Nested
             layers and models will be automatically discovered. Alternatively, in place of a single layer, you can pass
             a list of layers.
         optimizer_specs: specialized list for serialization. Should be left as None for almost all cases. If you are
-            loading a serialized version of this optimizer, please use tf.keras.models.load_model after saving a
+            loading a serialized version of this optimizer, please use `tf.keras.models.load_model` after saving a
             model compiled with this optimizer.
 
     Usage:
 
-    ```python
-    model = get_model()
-
-    opt1 = tf.keras.optimizers.Adam(learning_rate=1e-4)
-    opt2 = tf.keras.optimizers.Adam(learning_rate=1e-2)
-
-    opt_layer_pairs = [(opt1, model.layers[0]), (opt2, model.layers[1:])]
-
-    loss = tf.keras.losses.MSE
-    optimizer = tfa.optimizers.MultiOpt(opt_layer_pairs)
-
-    model.compile(optimizer=optimizer, loss = loss)
-
-    model.fit(x,y)
-    '''
+    >>> model = tf.keras.Sequential([
+    ...     tf.keras.Input(shape=(4,)),
+    ...     tf.keras.layers.Dense(8),
+    ...     tf.keras.layers.Dense(16),
+    ...     tf.keras.layers.Dense(32),
+    ... ])
+    >>> optimizer1 = tf.keras.optimizers.Adam(learning_rate=1e-4)
+    >>> optimizer2 = tf.keras.optimizers.Adam(learning_rate=1e-2)
+    >>> optimizers_and_layers = [(optimizer1, model.layers[0]), (optimizer2, model.layers[1:])]
+    >>> optimizer = tfa.optimizers.MultiOptimizer(optimizers_and_layers)
+    >>> model.compile(optimizer=optimizer, loss="mse")
 
     Reference:
 
-    [Universal Language Model Fine-tuning for Text Classification](https://arxiv.org/abs/1801.06146)
-    [Collaborative Layer-wise Discriminative Learning in Deep Neural Networks](https://arxiv.org/abs/1607.05440)
+    - [Universal Language Model Fine-tuning for Text Classification](https://arxiv.org/abs/1801.06146)
+    - [Collaborative Layer-wise Discriminative Learning in Deep Neural Networks](https://arxiv.org/abs/1607.05440)
 
-    Notes:
+    Note:
 
-    Currently, MultiOpt does not support callbacks that modify optimizers. However, you can instantiate
-    optimizer layer pairs with tf.keras.optimizers.schedules.LearningRateSchedule instead of a static learning
-    rate.
+    Currently, MultiOpt does not support callbacks that modify optimizers.
+    However, you can instantiate optimizer layer pairs with
+    `tf.keras.optimizers.schedules.LearningRateSchedule`
+    instead of a static learning rate.
 
-    This code should function on CPU, GPU, and TPU. Apply the with strategy.scope() context as you
+    This code should function on CPU, GPU, and TPU. Apply the with `tf.distribute.Strategy().scope()` context as you
     would with any other optimizer.
-
     """
 
     @typechecked
@@ -131,29 +127,29 @@ class MultiOptimizer(tf.keras.optimizers.Optimizer):
         return config
 
     @classmethod
-    def create_optimizer_spec(cls, optimizer_instance, layer):
-
-        assert isinstance(
-            optimizer_instance, tf.keras.optimizers.Optimizer
-        ), "Object passed is not an instance of tf.keras.optimizers.Optimizer"
-
-        assert isinstance(layer, tf.keras.layers.Layer) or isinstance(
-            layer, tf.keras.Model
-        ), "Object passed is not an instance of tf.keras.layers.Layer nor tf.keras.Model"
-
-        if type(layer) == list:
+    def create_optimizer_spec(
+        cls,
+        optimizer: tf.keras.optimizers.Optimizer,
+        layer: Union[
+            tf.keras.Model,
+            tf.keras.Sequential,
+            tf.keras.layers.Layer,
+            List[tf.keras.layers.Layer],
+        ],
+    ):
+        if isinstance(layer, list):
             weights = [var.name for sublayer in layer for var in sublayer.weights]
         else:
             weights = [var.name for var in layer.weights]
 
         return {
-            "optimizer": optimizer_instance,
+            "optimizer": optimizer,
             "weights": weights,
         }
 
     @classmethod
     def maybe_initialize_optimizer_spec(cls, optimizer_spec):
-        if type(optimizer_spec["optimizer"]) == dict:
+        if isinstance(optimizer_spec["optimizer"], dict):
             optimizer_spec["optimizer"] = tf.keras.optimizers.deserialize(
                 optimizer_spec["optimizer"]
             )
--- a/tensorflow_addons/optimizers/discriminative_layer_training.py
+++ b/tensorflow_addons/optimizers/discriminative_layer_training.py
@@ -72,7 +72,7 @@ class MultiOptimizer(tf.keras.optimizers.Optimizer):
     `tf.keras.optimizers.schedules.LearningRateSchedule`
     instead of a static learning rate.
 
-    This code should function on CPU, GPU, and TPU. Apply the with `tf.distribute.Strategy().scope()` context as you
+    This code should function on CPU, GPU, and TPU. Apply with `tf.distribute.Strategy().scope()` context as you
     would with any other optimizer.
     """
 
--- a/tensorflow_addons/optimizers/discriminative_layer_training.py
+++ b/tensorflow_addons/optimizers/discriminative_layer_training.py
@@ -14,7 +14,7 @@
 # ==============================================================================
 """Discriminative Layer Training Optimizer for TensorFlow."""
 
-from typing import List, Union
+from typing import Iterable, List, Union
 
 import tensorflow as tf
 from typeguard import typechecked
@@ -74,8 +74,8 @@ class MultiOptimizer(tf.keras.optimizers.Optimizer):
     @typechecked
     def __init__(
         self,
-        optimizers_and_layers: Union[list, None] = None,
-        optimizer_specs: Union[list, None] = None,
+        optimizers_and_layers: Union[Iterable, None] = None,
+        optimizer_specs: Union[Iterable, None] = None,
         name: str = "MultiOptimzer",
         **kwargs
     ):
--- a/tensorflow_addons/optimizers/discriminative_layer_training.py
+++ b/tensorflow_addons/optimizers/discriminative_layer_training.py
@@ -106,8 +106,7 @@ class MultiOptimizer(tf.keras.optimizers.Optimizer):
     def apply_gradients(self, grads_and_vars, **kwargs):
         """Wrapped apply_gradient method.
 
-        Returns a list of tf ops to be executed.
-        Name of variable is used rather than var.ref() to enable serialization and deserialization.
+        Returns an operation to be executed.
         """
 
         for spec in self.optimizer_specs:
@@ -142,6 +141,10 @@ class MultiOptimizer(tf.keras.optimizers.Optimizer):
             List[tf.keras.layers.Layer],
         ],
     ):
+        """Creates a serializable optimizer spec.
+
+        The name of each variable is used rather than `var.ref()` to enable serialization and deserialization.
+        """
         if isinstance(layer, list):
             weights = [var.name for sublayer in layer for var in sublayer.weights]
         else:
--- a/tensorflow_addons/optimizers/discriminative_layer_training.py
+++ b/tensorflow_addons/optimizers/discriminative_layer_training.py
@@ -95,10 +95,10 @@ class MultiOptimizer(tf.keras.optimizers.Optimizer):
 
         else:
             raise RuntimeError(
-                "You must specify either an list of optimizers and layers or a list of optimizer_specs"
+                "Must specify either an list of optimizers and layers or a list of optimizer_specs"
             )
 
-    def apply_gradients(self, grads_and_vars, name=None, **kwargs):
+    def apply_gradients(self, grads_and_vars, **kwargs):
         """Wrapped apply_gradient method.
 
         Returns a list of tf ops to be executed.
--- a/tensorflow_addons/optimizers/discriminative_layer_training.py
+++ b/tensorflow_addons/optimizers/discriminative_layer_training.py
@@ -14,7 +14,7 @@
 # ==============================================================================
 """Discriminative Layer Training Optimizer for TensorFlow."""
 
-from typing import Iterable, List, Union
+from typing import List, Union
 
 import tensorflow as tf
 from typeguard import typechecked
@@ -79,8 +79,8 @@ class MultiOptimizer(tf.keras.optimizers.Optimizer):
     @typechecked
     def __init__(
         self,
-        optimizers_and_layers: Union[Iterable, None] = None,
-        optimizer_specs: Union[Iterable, None] = None,
+        optimizers_and_layers: Union[list, None] = None,
+        optimizer_specs: Union[list, None] = None,
         name: str = "MultiOptimizer",
         **kwargs
     ):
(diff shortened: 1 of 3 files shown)


tensorflow_addons/optimizers/tests/discriminative_layer_training_test.py
```diff
@@ -22,32 +22,9 @@
...
-    # TODO(WindQAQ): Clean up this in TF2.4
-
-    if use_gpu:
-        return [tf.float32, tf.float64]
-    else:
-        return [tf.half, tf.float32, tf.float64]
-
-
 @pytest.mark.with_device(["cpu", "gpu"])
-@pytest.mark.parametrize("dtype", [tf.float16, tf.float32, tf.float64])
```
WindQAQ, line 43: This is not being used in the original tests, so I remove it

tensorflow_addons/optimizers/discriminative_layer_training.py
```diff
@@ -131,29 +130,33 @@ def get_config(self):
...
+            List[tf.keras.layers.Layer],
+        ],
+    ):
+        """Creates a serializable optimizer spec.
 
-        if type(layer) == list:
+        The name of each variable is used rather than `var.ref()` to enable serialization and deserialization.
+        """
+        if isinstance(layer, list):
             weights = [var.name for sublayer in layer for var in sublayer.weights]
```
bhack, line 148: Is here trainable_weights?

tensorflow_addons/optimizers/discriminative_layer_training.py
```diff
@@ -131,29 +130,33 @@ def get_config(self):
...
         else:
             weights = [var.name for var in layer.weights]
```
bhack, line 150: Ditto?

bhack (changes requested):
From a quick overview It seems that we are not covering the `optimizer, model case`?

tensorflow_addons/optimizers/discriminative_layer_training.py
WindQAQ: This is discussed in https://github.com/tensorflow/addons/pull/969#discussion_r487591038.

WindQAQ:
> From a quick overview It seems that we are not covering the `optimizer, model case`?

Not sure what this case is. Can you elaborate it?

tensorflow_addons/optimizers/discriminative_layer_training.py
WindQAQ: I would respect the decision of code owner. Change in design is not part of this PR. /cc @hyang0129

bhack:
I meant it seems that `tf.Model` could be and input right? Do we have a test for this input case?

tensorflow_addons/optimizers/discriminative_layer_training.py
bhack: If it was interpreted as we have `weights` and we set the same `weights` at semantic level is ok

WindQAQ:
> I meant it seems that `tf.Model` could be and input right? Do we have a test for this input case?

Thanks for the info. Updated the related tests :-)

tensorflow_addons/optimizers/discriminative_layer_training.py
```diff
@@ -131,29 +130,33 @@ def get_config(self):
...
-            optimizer_instance, tf.keras.optimizers.Optimizer
-        ), "Object passed is not an instance of tf.keras.optimizers.Optimizer"
-
-        assert isinstance(layer, tf.keras.layers.Layer) or isinstance(
-            layer, tf.keras.Model
-        ), "Object passed is not an instance of tf.keras.layers.Layer nor tf.keras.Model"
+    def create_optimizer_spec(
+        cls,
+        optimizer: tf.keras.optimizers.Optimizer,
+        layer: Union[
```
bhack, line 136: Also the naming here It Is a little bit confusing also in the internal code cause if `layers` with `s` It could cover the `list` case but what about `tf.keras.Model`?

tensorflow_addons/optimizers/discriminative_layer_training.py
WindQAQ: Yep. We can wait the response from the code owner. I see the underlying `fit` only takes trainable variables into account. As you recommend, we can change it to `trainable_weights` after dicussion.

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/engine/training.py#L737

tensorflow_addons/optimizers/discriminative_layer_training.py
WindQAQ: Sure, the naming is not so good to me either, but I cannot come up with one... do you have any suggestion?

bhack: Ugly as ugly we already have `optimizers_and_layers`. 
With another boolean we could have `layers_or_model` :smile:

hyang0129:
Sorry was on vacation. I'll take a look tomorrow.

tensorflow_addons/optimizers/discriminative_layer_training.py
hyang0129: I would prefer to carry over any features/bugs inherent in the original tensorflow. I believe that other optimizers do not make the distinction between weights and trainable weights. Rather, the weights themselves contain a trainable attribute that the graph seems to understand.

WindQAQ: Correct. a common use case in custom training loop is

```python
multi_optimizer.apply_gradient(grads_and_vars, model.trainable_variables)
```

in this case, even though we record the name of non-trainable variables, they will not be used. I would also vote for `weights` instead of `trainable_weights`

tensorflow_addons/optimizers/discriminative_layer_training.py
WindQAQ: Can you share the input signature you propose?

hyang0129: that's a good idea. please go ahead with the renaming.

tensorflow_addons/optimizers/discriminative_layer_training.py
bhack: As we have `apply_gradients` here at the end I don't know If for readibility we are more in a case like  https://github.com/tensorflow/tensorflow/blob/ba6d71b874c954bdd4219da118318fe2710515cf/tensorflow/python/keras/integration_test/function_test.py#L61

bhack: Eg in Adam what is the receiving `var_list` https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/optimizer.py#L400 ?

tensorflow_addons/optimizers/discriminative_layer_training.py
bhack: `layer` -> `layers_or_model`

tensorflow_addons/optimizers/discriminative_layer_training.py
bhack: My opinion is that as we are only overriding/wrapping `apply_gradients`  our `grads_and_vars` arg is expected to be passed in the base class as a returned by `compute_gradients()`:
>   grads_and_vars: List of (gradient, variable) pairs as returned by
        `compute_gradients()`.
 
But base class `compute_gradients` has a default behavior when `var_list`  is  Null as you can see at:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/optimizer.py#L432-L434

So it is not mandatory but we expect that by default `var_list` is variables.trainable_variables() as you can see from https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/optimizer.py#L495

But as we are using directly our weights in apply_gradient it is like we are not using the default behavior or at least like as we have called `computer_gradients()` with all the weights in `var_list` .

tensorflow_addons/optimizers/tests/discriminative_layer_training_test.py
```diff
@@ -95,8 +72,78 @@ def test_fit_layer_optimizer(dtype, device, serialize):
...
+                test_utils.assert_allclose_according_to_type(w_before, w_after)
+
+
+def test_model():
+    input = tf.keras.Input(shape=(4,))
+    output = tf.keras.layers.Dense(16)(input)
+    output = tf.keras.layers.Dense(16)(output)
+    output = tf.keras.layers.Dense(32)(output)
+    output = tf.keras.layers.Dense(32)(output)
+    model = tf.keras.Model(input, output)
```
bhack, line 127: Is the case to test a nested model here or do you think is It useless?

tensorflow_addons/optimizers/tests/discriminative_layer_training_test.py
AakashKumarNain: But we should add this test, right?

tensorflow_addons/optimizers/tests/discriminative_layer_training_test.py
hyang0129: I recommend having two models to match the common pretraining structure. 

pretrained = get_pretrained_model()

model = tf.keras.Sequential([pretrained, tf.keras.layers.Dense(num_classes)])

hyang0129: and just confirm that the pretrained model is not training

bhack: I meant more something like submodels:
https://github.com/keras-team/keras-io/blob/fc24a909dacc9036d374be5adb168a2daafc8b2c/guides/functional_api.py#L277:L292

hyang0129: I believe that is unnecessary based on my understanding on how models become graphs. Both the functional and sequential models are subclasses of the Model class and they should convert to a graph in the same way. I haven't confirmed this, but I'm guessing that sequential is just an wrapper of the functional model. I also haven't ever experienced any differences between functional and sequential models except during model construction. 

However, it wouldn't hurt to put it in.

bhack: So this loop is ok also in that case?

https://github.com/WindQAQ/addons/blob/d334fabbbee0ae6ef8c35bb124dcec5fee09dd81/tensorflow_addons/optimizers/discriminative_layer_training.py#L150

tensorflow_addons/optimizers/discriminative_layer_training.py
WindQAQ: Hi, seems that `tf.keras.Model` is a subclass of `tf.keras.layers.Layer`. Do we still need to do this?

tensorflow_addons/optimizers/tests/discriminative_layer_training_test.py
bhack: So with current loop we cover also that case?
```weights = [var.name for var in layer.weights]```

Comment (WindQAQ) [reply tensorflow_addons/optimizers/tests/discriminative_layer_training_test.py]: > So with current loop we cover also that case?
> `weights = [var.name for var in layer.weights]`

Correct. Updated tests.

File: tensorflow_addons/optimizers/discriminative_layer_training.py (hunk 2 above)
Comment (WindQAQ) [reply tensorflow_addons/optimizers/discriminative_layer_training.py:148]: > My opinion is that as we are only overriding/wrapping `apply_gradients` our `grads_and_vars` arg is expected to be passed in the base class as returned by `compute_gradients()`:
> 
> > grads_and_vars: List of (gradient, variable) pairs as returned by
> > `compute_gradients()`.
> 
> But base class `compute_gradients` has a default behavior when `var_list` is Null as you can see at:
> https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/optimizer.py#L432-L434
> 
> So it is not mandatory but we expect that by default `var_list` is variables.trainable_variables() as you can see from https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/optimizer.py#L495
> 
> But as we are using directly our weights in apply_gradient it is like we are not using the default behavior or at least like as we have called `computer_gradients()` with all the weights in `var_list` .

This is V1 optimizer?

File: tensorflow_addons/optimizers/tests/discriminative_layer_training_test.py (hunk 5 above)
Comment (bhack) [reply tensorflow_addons/optimizers/tests/discriminative_layer_training_test.py]: Ops sorry they was the same It was just 1 single case It Is a duplicated comment.

File: tensorflow_addons/optimizers/discriminative_layer_training.py (hunk 4 above)
Comment (bhack) [reply tensorflow_addons/optimizers/discriminative_layer_training.py:136]: It is a borderline case cause is multi inheritance:
```class Model(base_layer.Layer, version_utils.ModelVersionSelector):```
We are only using `layers` so a single base classe features but I don't know about readibility.

File: tensorflow_addons/optimizers/discriminative_layer_training.py (hunk 2 above)
Comment (bhack) [reply tensorflow_addons/optimizers/discriminative_layer_training.py:148]: Yes this is V1. 
In V2 you need to pass a variable list or a callable as it has no default but they suppose as default example:
https://github.com/tensorflow/tensorflow/blob/v2.3.1/tensorflow/python/keras/optimizer_v2/optimizer_v2.py#L118-L120

Instead in our `apply_gradients` we are controlling internally the variable list to pass to the real `apply_gradiennts` of optimizers. It seems to me that we are performing a little bit like `fit->minimize` method or not?

File: tensorflow_addons/optimizers/tests/discriminative_layer_training_test.py [hunk 6]
```diff
@@ -95,8 +72,167 @@ def test_fit_layer_optimizer(dtype, device, serialize):
...
+    y1 = model1(inputs)
+    y2 = model2(inputs)
+    y3 = model3(inputs)
+    outputs = tf.keras.layers.Average()([y1, y2, y3])
+    model = tf.keras.Model(inputs, outputs)
+
+    optimizers_and_layers = [
+        (tf.keras.optimizers.SGD(), model1),
+        (tf.keras.optimizers.SGD(learning_rate=0.0), model2),
+        (tf.keras.optimizers.SGD(), model3),
```
bhack, line 198: Does it will work here using the final `model`?

tensorflow_addons/optimizers/discriminative_layer_training.py
WindQAQ: We are not “controlling the var list” internally but fetching only vars that users pass in.

bhack: If we look at the argument ok we receive an user  argument but internally we access to our prepared `spec["weights"]`:
` for name in spec["weights"]:`

tensorflow_addons/optimizers/tests/discriminative_layer_training_test.py
WindQAQ: yes

bhack: I meant something like `(tf.keras.optimizers.SGD(), model)` is supported right?

WindQAQ: yes

tensorflow_addons/optimizers/discriminative_layer_training.py
bhack: So as you want cause we cannot upcast to `Layer` in python or It Is generally a [borderline pratice](https://docs.python.org/3.8/reference/datamodel.html#id1)

WindQAQ: Updated the naming. See if this is better now :-)

tensorflow_addons/optimizers/discriminative_layer_training.py
hyang0129: the multi optimizer stores a reference of weight names. 

when optimizing, it uses this reference to allocate grad var pairs to the correct optimizer. 

it does not determine what weights are passed to the multi optimizer. 

So you may have a situation where 5 vars are passed but spec['weights'] contains 10 vars and only 3 of them match. Thus, only 3 are passed on. 

The original design is intended to be used in the model.fit function, not a custom optimization loop. Even in a custom optimization loop, the user is expected to specify the var list to include only trainable weights. I am fairly certain that when an optimizer is called in a fit loop, the var list passed includes only trainable variables. Thus, the var list passed to the multi optimizer includes only trainable variables. Finally, the multi optimizer only passes on to its sub optimizers the variables in var list it received that match a variable in spec['weights'] for that particular optimizer. If this is true, then a non trainable variable will never make it to the multi optimizer or sub optimizers. 

This behavior has been tested in a [colab notebook](https://colab.research.google.com/drive/1NKAsrjj1qcls6p8S5dvJT1PkdC2Ieazc?usp=sharing).

tensorflow_addons/optimizers/tests/discriminative_layer_training_test.py
```diff
@@ -95,8 +72,167 @@ def test_fit_layer_optimizer(dtype, device, serialize):
...
+    ]
+
+    weights_before_train = [
+        [weight.numpy() for weight in layer.weights] for layer in model.layers
+    ]
+
+    multi_optimizer = MultiOptimizer(optimizers_and_layers)
+    model.compile(multi_optimizer, loss="mse")
+
+    x = np.random.rand(128, 4)
```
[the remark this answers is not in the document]
WindQAQ: Nice suggestion :-) Updated

tensorflow_addons/utils/test_utils.py
```diff
@@ -217,6 +217,25 @@ def pytest_collection_modifyitems(items):
                 item.add_marker(pytest.mark.skip("The gpu is not available."))
 
 
+def assert_not_allclose(a, b, **kwargs):
```
hyang0129, line 220: nice

hyang0129 (dismissed):

tensorflow_addons/optimizers/discriminative_layer_training.py
```diff
@@ -24,72 +24,70 @@
...
+    Generally useful for transfer learning of deep networks.
 
-    Each optimizer will optimize only the weights associated with its paired layer. This can be used
-    to implement discriminative layer training by assigning different learning rates to each optimizer
-    layer pair. (Optimizer, list(Layers)) pairs are also supported. Please note that the layers must be
-    instantiated before instantiating the optimizer.
+    Each optimizer will optimize only the weights associated with its paired layer.
+    This can be used to implement discriminative layer training by assigning
+    different learning rates to each optimizer layer pair.
+    `(tf.keras.optimizers.Optimizer, List[tf.keras.layers.Layer])` pairs are also supported.
```
hyang0129, line 33: List[tf.keras.layers.Layer]) -> List([tf.keras.layers.Layer]). Was missing a (

hyang0129:
@bhack I think this is good to go.

tensorflow_addons/optimizers/discriminative_layer_training.py
WindQAQ: It's ( optimizer, List[layer] ), where () stands for Tuple.

tensorflow_addons/optimizers/discriminative_layer_training.py
bhack: If it Is a case where weight is a superset to check the name probably ok.

But I still don't catch  when `if var.name == name` needs to handle not trainable weights.

bhack (approved):

PR #2191, opened 2020-10-06
Bump TF 2.3.1

WindQAQ:
Bump TF 2.3.1

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,7 +40,7 @@ jobs:
       matrix:
         os: ['macos-latest', 'windows-latest', 'ubuntu-18.04']
         py-version: ['3.5', '3.6', '3.7', '3.8']
-        tf-version: ['2.3.0']
+        tf-version: ['2.3.0', '2.3.1']
       fail-fast: false
     steps:
       - uses: actions/github-script@0.3.0
@@ -80,7 +80,7 @@ jobs:
       matrix:
         os: ['macOS', 'Windows', 'Linux']
         py-version: ['3.5', '3.6', '3.7', '3.8']
-        tf-version: ['2.3.0']
+        tf-version: ['2.3.0', '2.3.1']
       fail-fast: false
     if: (github.event_name == 'push' && github.ref == 'refs/heads/master') || github.event_name == 'release'
     steps:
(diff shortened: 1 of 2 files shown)


bhack (commented):
Is this going to break https://github.com/tensorflow/addons/blob/master/tools/update_release_version.sh?

WindQAQ:
> Is this going to break https://github.com/tensorflow/addons/blob/master/tools/update_release_version.sh?

Oops, I don't even notice that we have this amazing tool. And yes, it will break it.

bhack:
> > Is this going to break https://github.com/tensorflow/addons/blob/master/tools/update_release_version.sh?
> 
> Oops, I don't even notice that we have this amazing tool. And yes, it will break it.

Do you think It could be extended to cover this?

WindQAQ:
It's not so easy...
But seems that we have never published the wheel again patch versioned tensorflow so we should discuss if we should let this PR go forward.

bhack:
> It's not so easy...
> But seems that we have never published the wheel against patch versioned tensorflow so we should discuss if we should let this PR go forward.

This was cause generally wed didn't have Tensorflow patch releases but this case was about security fixes.

WindQAQ:
So, should we modify that script to make it compatible with multiple versions?

bhack:
Yes

WindQAQ:
Let me close it first as I don't have good implementation of that script in mind.

bhack:
What strings we will have on 2.4.0 release?

WindQAQ:
We usually pin to rc version first.

bhack:
Yes I meant to expand regular expression I think we need to figure out what strings we need to handle. After this change what strings we will need to maintain? Need we need to handle arbitrary list of releases e.g. 2.3.0,2.3.1,2.4.0rc etc..

WindQAQ:
> Need we need to handle arbitrary list of releases e.g. 2.3.0,2.3.1,2.4.0rc etc..

Yes

bhack:
Only in tf-version right? Are the single entries  just the first element of this list?

WindQAQ:
> Only in tf-version right? Are the single entries just the first element of this list?

Yes, I think so.

PR #2192, opened 2020-10-06
Update ABI compatibility version

WindQAQ:
Fix https://colab.research.google.com/drive/19AFVhLDcFa1OHMkaZm5K3alytnxODaGh?usp=sharing. This is not the desired warning. We are built custom ops against TF2.3.*.

### DIFF
--- a/tensorflow_addons/utils/resource_loader.py
+++ b/tensorflow_addons/utils/resource_loader.py
@@ -20,8 +20,8 @@ import warnings
 
 import tensorflow as tf
 
-MIN_TF_VERSION_FOR_ABI_COMPATIBILITY = "2.2.0"
-MAX_TF_VERSION_FOR_ABI_COMPATIBILITY = "2.3.0"
+MIN_TF_VERSION_FOR_ABI_COMPATIBILITY = "2.3.0"
+MAX_TF_VERSION_FOR_ABI_COMPATIBILITY = "2.4.0"
 abi_warning_already_raised = False
 SKIP_CUSTOM_OPS = False
 


tensorflow_addons/utils/resource_loader.py
```diff
@@ -20,8 +20,8 @@
 
 import tensorflow as tf
 
-MIN_TF_VERSION_FOR_ABI_COMPATIBILITY = "2.2.0"
-MAX_TF_VERSION_FOR_ABI_COMPATIBILITY = "2.3.0"
+MIN_TF_VERSION_FOR_ABI_COMPATIBILITY = "2.3.0"
+MAX_TF_VERSION_FOR_ABI_COMPATIBILITY = "2.4.0"
```
bhack, line 24: Is 2.4.0 ok?

WindQAQ: Not okay, so we check the version via `min_version <= version < max_version`.

https://github.com/tensorflow/addons/blob/master/tensorflow_addons/utils/resource_loader.py#L115

bhack: Do you remember why we had that? It Is a little bit less intuitive that the MAX compatibility It is excluded

WindQAQ: Actually, I don't know about that too much. I guess it is because

```python
>>> LooseVersion("2.3.1") <= LooseVersion("2.3")
False
```

seanpmorgan: Yeah strange variable name that should be updated. Also we should update the release_version.sh to update resource_loader

bhack: Ok but we I see we use extended version with the final `.0`

seanpmorgan (approved):
Going to merge this PR as this is a current bug. Happy to see a PR with different variable namings though.

I've updated the release process to make sure this gets checked during release:
https://github.com/tensorflow/community/pull/276

Will make sure that gets merged this week.

PR #2197, opened 2020-10-09
Improve the AttentionMechanism base class

guillaumekln:
# Description

The `AttentionWrapper` takes `AttentionMechanism` instances as arguments. However, it used methods that were not defined in the `AttentionMechanism` base class.

This PR improves the base class definition with the actual interface expected by `AttentionWrapper`.

## Type of change

- [ ] Bug fix
- [ ] New Tutorial
- [ ] Additional Testing
- [x] New Seq2seq addition and the changes conform to the [seq2seq contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/seq2seq/README.md#contribution-guidelines)

# Checklist:

# How Has This Been Tested?

Ran the existing tests.

### DIFF
--- a/tensorflow_addons/seq2seq/attention_wrapper.py
+++ b/tensorflow_addons/seq2seq/attention_wrapper.py
@@ -38,17 +38,93 @@ from typing import Optional, Callable, Union, List
 from tensorflow.python.keras.engine import base_layer_utils
 
 
-class AttentionMechanism:
+class AttentionMechanism(tf.keras.layers.Layer):
+    """Base class for attention mechanisms.
+
+    Subclasses should implement the logic of registering the attention `keys`
+    and `values` and computing the attention scores in `call`.
+    """
+
+    def __init__(self, **kwargs):
+        """Initializes the attention mechanism base class.
+
+        Args:
+          kwargs: Other layer arguments.
+        """
+        super().__init__(**kwargs)
+        self.keys = None
+        self.values = None
+
     @property
     def alignments_size(self):
-        raise NotImplementedError
+        """Size of the attention keys and values time dimension."""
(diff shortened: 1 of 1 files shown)


seanpmorgan (approved):
Refactor LGTM and I don't see any other need for `self._memory_initialized`. @qlzh727 any issues with merging?

qlzh727:
Thanks for the update and sorry for the late reply.

The concern I have for this PR is that the implementation of the attention is now split into two classes, especially self.key and self.value are created by used in two different class. When only looking at the AttentionMechanism, it is hard to reason about what is the self.key/value, unless we look at how they are initialized in _BaseAttentionMechanism.setup_memory().

If you feel AttentionMechanism should declare more methods for AttentionWrapper to use, maybe we should just create them but still keep the implementation in _BaseAttentionMechanism?

guillaumekln:
Thanks for the comment.

The initial observation was that `AttentionWrapper` expected a `_BaseAttentionMechanism` and not a `AttentionMechanism`. So my idea was to move some methods in the base class to fix this inconsistency and at the same time make it easier for users to write custom attention mechanisms. By following this route, would it be enough to define a base implementation of `setup_memory` in `AttentionMechanism` to clarify how the attributes are initialized?

Or do you think it's better for `AttentionMechanism` to just be the interface?

qlzh727:
I guess the AttentionWrapper won't work if user just implement the method in AttentionMechanism, so I would expect _BaseAttentionMechanism to just graduate to be a public class to override. Maybe we should just rename the _BaseAttentionMechanism and remove the AttentionMechanism?

guillaumekln:
Ok, let's rename `_BaseAttentionMechanism` to `AttentionMechanism`. I will make a new PR. Thanks!

PR #2200, opened 2020-10-11
Add support to releases list

bhack:
Add support to update releases list
*


bhack:
Please check if you want to enforce single entries with the specific pattern for inputs validation and for a more specific matching of the version strings in our files.
See https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

WindQAQ (commented):
Thanks!

WindQAQ (approved):

PR #2205, opened 2020-10-18
Add TF2.2 testing

seanpmorgan:
# Description

Per our compatibility matrix we want to test both TF2.2 and TF2.3 (TF2.4rc soon as well). This change will make it so wheels are built and tested against 2.2 and 2.3, but we will only publish the 2.3 built wheels as per the compatibility matrix. 

The testing is kind of buried in the build so I updated the Action name to be more clear about what it's doing.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,7 +40,7 @@ jobs:
       matrix:
         os: ['macos-latest', 'windows-latest', 'ubuntu-18.04']
         py-version: ['3.5', '3.6', '3.7', '3.8']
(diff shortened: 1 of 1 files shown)


seanpmorgan:
Looks like there is some functionality that is not available in TF2.2 like `tf.keras.random_bernoulli`. @tensorflow/sig-addons-maintainers any thoughts on if we can/should refactor or just release compatible with TF > 2.3 going forward

seanpmorgan:
Corrective changes done in https://github.com/tensorflow/addons/pull/2224

PR #2208, opened 2020-10-18
Standardized Testing for All Optimizers

hyang0129:
References #992. 

Aim is to create a single test file that tests all relevant classes within the optimizer module of tfa. The test will discover classes that inherit from tf.keras.optimizers.Optimizer then execute a set of standard tests on each of those classes. 

This will test for common behaviors that optimizers should have (eg. serialization, minimizes, etc.).

### DIFF
--- a/tensorflow_addons/optimizers/tests/standard_test.py
+++ b/tensorflow_addons/optimizers/tests/standard_test.py
@@ -6,21 +6,26 @@ from tensorflow_addons import optimizers
 import inspect
 import sys
 
-classes_to_test = ['RectifiedAdam', 'LazyAdam']
+classes_to_test = ["RectifiedAdam", "LazyAdam"]
+
 
 def discover_classes(module, parent):
 
-    classes = [m[1] for m in inspect.getmembers(module, inspect.isclass)
-               if issubclass(m[1], parent) and m[0] in classes_to_test ]
+    classes = [
+        m[1]
+        for m in inspect.getmembers(module, inspect.isclass)
+        if issubclass(m[1], parent) and m[0] in classes_to_test
+    ]
 
     return classes
 
-@pytest.mark.parametrize("optimizer", discover_classes(optimizers, tf.keras.optimizers.Optimizer))
+
+@pytest.mark.parametrize(
+    "optimizer", discover_classes(optimizers, tf.keras.optimizers.Optimizer)
+)
 def test_optimizer_minimize(optimizer):
 
-    model = tf.keras.Sequential(
-        [tf.keras.Input(shape=[1]), tf.keras.layers.Dense(1)]
-    )
+    model = tf.keras.Sequential([tf.keras.Input(shape=[1]), tf.keras.layers.Dense(1)])
 
     x = np.array(np.ones([1]))
     y = np.array(np.zeros([1]))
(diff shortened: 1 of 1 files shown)


hyang0129:
@seanpmorgan the optimizer module contains many classes that do not behave in the same way as an optimizer object. For example, multi_optimizer is a wrapper instead of an optimizer. 

Perhaps we could have a class signature that exempts certain classes from the standard test? 

Anyways, I've written some code to discover classes and it should work for any module. I think this can extend to layers and other modules that all contain classes that inherit from a common parent.

hyang0129:
Maybe we should have a list of behaviors that each class in a module should have. So for optimizers, I would imagine that list includes minimizes and serializes. Ironically, I feel like the process of defining tests for optimizers is defining what a tfa optimizer should be.

hyang0129:
on failure we get something like this. 

standard_test.py FF                                                      [100%]

================================== FAILURES ===================================
____________________ test_optimizer_minimize[cpu-LazyAdam] ____________________

tensorflow_addons/optimizers/tests/standard_test.py
```diff
@@ -0,0 +1,49 @@
+import pytest
```
seanpmorgan, line 1: Need to add license

tensorflow_addons/optimizers/tests/standard_test.py
```diff
@@ -0,0 +1,49 @@
...
+    loss_values = history.history["loss"]
+
+    np.testing.assert_array_less(loss_values[-1], loss_values[0])
+
+
+if __name__ == "__main__":
+    sys.exit(pytest.main(["standard_test.py"]))
+
+    # python -m flake8 tensorflow_addons/optimizers/tests/standard_test.py
+    # python -m black tensorflow_addons/optimizers/tests/standard_test.py
```
seanpmorgan, line 49: Should be able to remove all of this once we've finalized since it'll be picked up from `optimizers/test/run_all_tests.py`

tensorflow_addons/optimizers/tests/standard_test.py
```diff
@@ -0,0 +1,49 @@
...
+
+def discover_classes(module, parent):
+
+    classes = [
+        m[1]
+        for m in inspect.getmembers(module, inspect.isclass)
+        if issubclass(m[1], parent) and m[0] in classes_to_test
+    ]
+
+    return classes
```
seanpmorgan, line 20: I like this, but can we make it so it discovers all optimizers within the folder and then add a list of exemptions instead of a list of things to test. IMO it is likely that newly added optimizers will not be added to the "to_test" list

seanpmorgan (changes requested):
Really like this and they way we can extend it to other subpackages as you mentioned. Couple of thoughts below.

tensorflow_addons/optimizers/tests/standard_test.py
hyang0129: yeah that's just for me to test locally. will definitely remove before merge

tensorflow_addons/optimizers/tests/standard_test.py
hyang0129: that's true, an exception list will allow us to automatically catch new opts. I'll change that.

hyang0129:
@seanpmorgan I've updated the code based on your feedback. 

I've defined two standard behaviors for opts which is minimizes and serializes. I feel like there should be more standard behaviors but idk. 

Also, half the classes appear to be wrappers. Once wrappers are standardized, we may need a way to select only optimizers and exclude wrappers because the testing criteria for wrappers could be different (or not because aside from wrapping, the wrapper should have the minimize and serialize behaviors. maybe just change how the test instances the wrappers?).

hyang0129:
Also, if we want to apply this to other modules, I think we should move the discover classes function to something more common like test_utils

hyang0129:
@seanpmorgan 

please review

seanpmorgan (approved):
LGTM thanks @hyang0129! I agree with you that this could apply to more subpackages and so the discovery util could be moved into a more generic place.

Each subpackage will have their own exception list though so it could get a bit messy. For now I think this is good, but happy to consider a proposed refactor

Issue #2222, opened 2020-11-01
Nightly windows build is failing due to missing commit datetime


seanpmorgan:
Our builds are failing due to the windows wheels not getting their commit datetime appended to the whl names:

```
Uploading tensorflow_addons-0.12.0.dev0-cp36-cp36m-win_amd64.whl

  0%|          | 0.00/910k [00:00<?, ?B/s]
 62%|██████▏   | 568k/910k [00:00<00:00, 5.42MB/s]
100%|██████████| 910k/910k [00:06<00:00, 150kB/s] 
NOTE: Try --verbose to see response content.
HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/
File already exists. See https://pypi.org/help/#file-name-reuse for more information.
```
https://github.com/tensorflow/addons/runs/1317714967?check_suite_focus=true

seanpmorgan:
EDIT --- Actually the full `NIGHTLY_FLAG` is missing for the windows builds. Not sure why this isn't working anymore:
https://github.com/tensorflow/addons/blob/master/.github/workflows/release.yml#L54

Thx @bhack

PR #2223, opened 2020-11-01
PyPi upload skip if exists

seanpmorgan:
# Description

This partially addresses #2222 but does not actually fix windows nightlies. It will come in handy though we we restart builds due to some other flaky failure.

https://github.com/pypa/gh-action-pypi-publish/blob/master/README.md#tolerating-release-package-file-duplicates


seanpmorgan:
Will close this for now since it may cause blindspots in our CI

PR #2226, opened 2020-11-02
Update testing dependency

seanpmorgan:
# Description

MacOS py3.8 is failing to pip install scikit-image for testing. Should be no issue in upgrading this testing dependency.

## Type of change

- [x] Bug fix


AakashKumarNain:
LGTM! Thanks for the fix

AakashKumarNain (approved):

PR #2233, opened 2020-11-08
Standardized Testing Module

hyang0129:
See #992 

This refactors a previous commit to make the discover classes function accessible to other modules for the creation of future standardized testing of classes within a module


hyang0129:
@seanpmorgan please review the refactoring

tensorflow_addons/optimizers/tests/standard_tests.py
```diff
@@ -18,7 +18,7 @@
 import tensorflow as tf
```
WindQAQ, line 18: Why the file name is changed?

WindQAQ (commented):
Sorry for the late review. It's a good idea to me. One minor question that I want to clarify. Thank you!

tensorflow_addons/optimizers/tests/standard_tests.py
hyang0129: I think I changed it because there would be multiple tests? We can change it back?

WindQAQ: Let's change it back. I think it's better to have a standard naming convention for file!

WindQAQ (approved):
Thank you!

PR #2241, opened 2020-11-16
Change GroupNormalization default groups to 32.

### LINKED ISSUE #2240: Set GroupNormalization default `groups` parameter to 32.
Currently, the default `groups` parameter for [`tfa.layers.GroupNormalization`](https://www.tensorflow.org/addons/api_docs/python/tfa/layers/GroupNormalization) is set to 2.

According to the original [Group Normalization paper](https://arxiv.org/abs/1803.08494) it would make more sense to set that default to 32. For reference, here is `Table 3 (top)` from the paper:

| `groups` | 64 | 32 | 16 | 8 | 4 | 2 |1 (=LN) |
| --- | --- | --- | --- | --- | --- | --- | --- |
| ResNet-50 ImageNet validation error `%` | 24.6 | ***24.1*** | 24.6 | 24.4 | 24.6 | 24.7 | 25.3 |
| Difference to `groups=32` | 0.5|  - | 0.5 | 0.3 | 0.5 | 0.6 | 1.2 |

Note that most experiments in the paper were performed using `groups=32` as well.

aaronmondal:
This changes the default `groups` parameter to `32`, as recommended by the authors of [Group Normalization](https://arxiv.org/abs/1803.08494).

Fixes #2240

### DIFF
--- a/tensorflow_addons/layers/normalizations.py
+++ b/tensorflow_addons/layers/normalizations.py
@@ -74,7 +74,7 @@ class GroupNormalization(tf.keras.layers.Layer):
     @typechecked
     def __init__(
         self,
-        groups: int = 2,
+        groups: int = 32,
         axis: int = -1,
         epsilon: float = 1e-3,
         center: bool = True,
--- a/tensorflow_addons/layers/normalizations.py
+++ b/tensorflow_addons/layers/normalizations.py
@@ -25,6 +25,9 @@ from tensorflow_addons.utils import types
 @tf.keras.utils.register_keras_serializable(package="Addons")
 class GroupNormalization(tf.keras.layers.Layer):
     """Group normalization layer.
+    
+    Source: "Group Normalization" (Yuxin Wu & Kaiming He, 2018)
+    https://arxiv.org/abs/1803.08494
 
     Group Normalization divides the channels into groups and computes
     within each group the mean and variance for normalization.
@@ -46,6 +49,7 @@ class GroupNormalization(tf.keras.layers.Layer):
         groups: Integer, the number of groups for Group Normalization.
             Can be in the range [1, N] where N is the input dimension.
             The input dimension must be divisible by the number of groups.
+            Defaults to 32.
         axis: Integer, the axis that should be normalized.
         epsilon: Small float added to variance to avoid dividing by zero.
         center: If True, add offset of `beta` to normalized tensor.
--- a/tensorflow_addons/layers/normalizations.py
+++ b/tensorflow_addons/layers/normalizations.py
@@ -25,7 +25,7 @@ from tensorflow_addons.utils import types
 @tf.keras.utils.register_keras_serializable(package="Addons")
 class GroupNormalization(tf.keras.layers.Layer):
     """Group normalization layer.
-    
+
     Source: "Group Normalization" (Yuxin Wu & Kaiming He, 2018)
     https://arxiv.org/abs/1803.08494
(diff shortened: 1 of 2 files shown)


aaronmondal:
To clarify: `test_regularizations` was modified to accommodate the default of 32. A group size of 32 was already tested in [`test_initializer`](https://github.com/tensorflow/addons/blob/38cdaa71ef44b078608527bc067f23a510311e80/tensorflow_addons/layers/tests/normalizations_test.py#L205)

Harsh188 (commented):
Hi! Thanks for contributing. Could you add to the [documentation](https://github.com/tensorflow/addons/blob/38cdaa71ef44b078608527bc067f23a510311e80/tensorflow_addons/layers/normalizations.py#L27) by specifying the default number of groups used?

aaronmondal:
Edited the `Arguments` section and added the source paper as well in the same way it was done for [`tf.nn.swish`](https://github.com/tensorflow/tensorflow/blob/fcc4b966f1265f466e82617020af93670141b009/tensorflow/python/ops/nn_impl.py#L512-L546). Linking to papers was done differently for  e.g. [`tf.nn.silu`](https://github.com/tensorflow/tensorflow/blob/6468a1d17ae43c161e836bd3d1af081e8e44ef0d/tensorflow/python/ops/nn_impl.py#L512), but I think it is better to display that link explicitly so users immediately see the link target.

tensorflow_addons/layers/normalizations.py
```diff
@@ -25,6 +25,9 @@
 @tf.keras.utils.register_keras_serializable(package="Addons")
 class GroupNormalization(tf.keras.layers.Layer):
     """Group normalization layer.
+    
+    Source: "Group Normalization" (Yuxin Wu & Kaiming He, 2018)
```
Harsh188, line 29: Reference is [mentioned](https://github.com/tensorflow/addons/blob/174fc18ed03d9a30ddbbe2a9d6c3087a4766973d/tensorflow_addons/layers/normalizations.py#L74) at the end.

Harsh188 (changes requested):

tensorflow_addons/layers/normalizations.py
Harsh188: Could you modify it such that only one exists. It might be redundant otherwise.

aaronmondal: My bad. Didn't notice it since it was out of my viewport :D

aaronmondal: Ill change it again and put it at the top. Leaving it at the bottom seems to be inconsistent with the way this was done in other papers.

tensorflow_addons/layers/normalizations.py
```diff
@@ -66,15 +70,12 @@ class GroupNormalization(tf.keras.layers.Layer):
 
     Output shape:
         Same shape as input.
-
```
aaronmondal, line 69: I removed this as this way of referencing did not seem to be consistent with citations in other functions. It also makes it difficult to find the reference.

Harsh188 (approved):

AakashKumarNain:
LGTM! Thanks @aaronmondal and @Harsh188

AakashKumarNain (approved):

aaronmondal:
Thank you @Harsh188 and @AakashKumarNain for your quick review and approval :)

PR #2264, opened 2020-12-03
Remove tf.function decorator in tfa.image.equalize

### LINKED ISSUE #2263: Remove @tf.function in tfa.image.equalize
`tfa.image.equalize` uses an internal `scale_channel` function which triggers excessive retracing when using it with datasets and probably in other cases as well. To illustrate this point the code snippet below benchmarks a naive `tf.data` loop once with the current implementation and once with a copy of the code in `tfa.image.equalize` with `@tf.function` removed.

On my machine I get:
```
[Insert a gazillion retracing warnings here]
With @tf.function: 0.6684637546539307
Without @tf.function: 0.08414626121520996
```

```python
import time
import statistics
from functools import partial

import tensorflow as tf
import tensorflow_datasets as tfds

from tensorflow_addons.image.utils import to_4D_image, from_4D_image


def equalize_image(image, data_format: str = "channels_last") -> tf.Tensor:
    """Implements Equalize function from PIL using TF ops."""

    # @tf.function  <-- This triggers retracing.
    def scale_channel(image, channel):
        """Scale the data in the channel to implement equalize."""
        image_dtype = image.dtype

        if data_format == "channels_last":
            image = tf.cast(image[:, :, channel], tf.int32)
        elif data_format == "channels_first":
            image = tf.cast(image[channel], tf.int32)
        else:
            raise ValueError(
                "data_format can either be channels_last or channels_first"
            )
        # Compute the histogram of the image channel.
        histo = tf.histogram_fixed_width(image, [0, 255], nbins=256)

        # For the purposes of computing the step, filter out the nonzeros.
        nonzero = tf.where(tf.not_equal(histo, 0))
        nonzero_histo = tf.reshape(tf.gather(histo, nonzero), [-1])
        step = (tf.reduce_sum(nonzero_histo) - nonzero_histo[-1]) // 255

        def build_lut(histo, step):
            # Compute the cumulative sum, shifting by step // 2
            # and then normalization by step.
            lut = (tf.cumsum(histo) + (step // 2)) // step
            # Shift lut, prepending with 0.
            lut = tf.concat([[0], lut[:-1]], 0)
            # Clip the counts to be in range.  This is done
            # in the C code for image.point.
            return tf.clip_by_value(lut, 0, 255)

        # If step is zero, return the original image.  Otherwise, build
        # lut from the full histogram and step and then index from it.

        if step == 0:
            result = image
        else:
            result = tf.gather(build_lut(histo, step), image)

        return tf.cast(result, image_dtype)

    idx = 2 if data_format == "channels_last" else 0
    image = tf.stack([scale_channel(image, c) for c in range(image.shape[idx])], idx)

    return image


def equalize(
    image, data_format: str = "channels_last", name=None
) -> tf.Tensor:
    with tf.name_scope(name or "equalize"):
        image_dims = tf.rank(image)
        image = to_4D_image(image)
        fn = partial(equalize_image, data_format=data_format)
        image = tf.map_fn(fn, image)
        return from_4D_image(image, image_dims)


dset = tfds.load(name='mnist', data_dir='tensorflow_datasets', split='train')
dset = dset.map(lambda data: data['image']).cache()

# Caching for a fair comparison.
for image in dset:
    img = image
print('Cached dataset')

function_times = list()
for _ in range(10):
    start = time.time()
    for image in dset.take(10):
        augmented_image = tfa.image.equalize(tf.constant(image))
    function_times.append(time.time() - start)
print(f'With @tf.function: {statistics.mean(function_times)}')

non_function_times = list()
for _ in range(10):
    start = time.time()
    for image in dset.take(10):
        augmented_image = equalize(tf.constant(image))
    non_function_times.append(time.time() - start)
print(f'Without @tf.function: {statistics.mean(non_function_times)}')    
```

aaronmondal:
Fixes #2263

### DIFF
--- a/tensorflow_addons/image/color_ops.py
+++ b/tensorflow_addons/image/color_ops.py
@@ -27,7 +27,7 @@ from typing import Optional
 from functools import partial
 
 
-def scale_channel(image: TensorLike, channel: int) -> tf.Tensor:
+def _scale_channel(image: TensorLike, channel: int) -> tf.Tensor:
     """Scale the data in the channel to implement equalize."""
     image_dtype = image.dtype
     image = tf.cast(image[:, :, channel], tf.int32)
@@ -46,18 +46,14 @@ def scale_channel(image: TensorLike, channel: int) -> tf.Tensor:
     else:
         lut_values = (tf.cumsum(histo, exclusive=True) + (step // 2)) // step
         lut_values = tf.clip_by_value(lut_values, 0, 255)
-        lookup_initializer = tf.lookup.KeyValueTensorInitializer(
-            keys=tf.range(256), values=lut_values
-        )
-        lookup_table = tf.lookup.StaticHashTable(lookup_initializer, 255)
-        result = lookup_table.lookup(image)
+        result = tf.gather(lut_values, image)
 
     return tf.cast(result, image_dtype)
 
 
-def equalize_image(image: TensorLike) -> tf.Tensor:
+def _equalize_image(image: TensorLike) -> tf.Tensor:
     """Implements Equalize function from PIL using TF ops."""
-    image = tf.stack([scale_channel(image, c) for c in range(image.shape[-1])], -1)
+    image = tf.stack([_scale_channel(image, c) for c in range(image.shape[-1])], -1)
     return image
 
 
@@ -68,9 +64,7 @@ def equalize(image: TensorLike, name: Optional[str] = None) -> tf.Tensor:
     Args:
       images: A tensor of shape
           `(num_images, num_rows, num_columns, num_channels)` (NHWC), or
-          `(num_images, num_channels, num_rows, num_columns)` (NCHW), or
           `(num_rows, num_columns, num_channels)` (HWC), or
-          `(num_channels, num_rows, num_columns)` (CHW), or
           `(num_rows, num_columns)` (HW). The rank must be statically known (the
           shape is not `TensorShape(None)`).
       name: The name of the op.
@@ -80,7 +74,7 @@ def equalize(image: TensorLike, name: Optional[str] = None) -> tf.Tensor:
     with tf.name_scope(name or "equalize"):
         image_dims = tf.rank(image)
         image = to_4D_image(image)
-        fn = partial(equalize_image)
+        fn = partial(_equalize_image)
         image = tf.map_fn(fn, image)
         return from_4D_image(image, image_dims)
 
--- a/tensorflow_addons/image/color_ops.py
+++ b/tensorflow_addons/image/color_ops.py
@@ -27,48 +27,41 @@ from typing import Optional
 from functools import partial
 
 
+def scale_channel(image: TensorLike, channel: int) -> tf.Tensor:
+    """Scale the data in the channel to implement equalize."""
+    image_dtype = image.dtype
+    image = tf.cast(image[:, :, channel], tf.int32)
+
+    # Compute the histogram of the image channel.
+    histo = tf.histogram_fixed_width(image, [0, 255], nbins=256)
+
+    # For the purposes of computing the step, filter out the nonzeros.
+    nonzero_histo = tf.boolean_mask(histo, histo != 0)
+    step = (tf.reduce_sum(nonzero_histo) - nonzero_histo[-1]) // 255
+
+    # If step is zero, return the original image.  Otherwise, build
+    # lut from the full histogram and step and then index from it.
+    if step == 0:
+        result = image
+    else:
+        lut_values = (tf.cumsum(histo, exclusive=True) + (step // 2)) // step
+        lut_values = tf.clip_by_value(lut_values, 0, 255)
+        lookup_initializer = tf.lookup.KeyValueTensorInitializer(
+            keys=tf.range(256), values=lut_values
+        )
+        lookup_table = tf.lookup.StaticHashTable(lookup_initializer, 255)
+        result = lookup_table.lookup(image)
+
+    return tf.cast(result, image_dtype)
+
+
 def equalize_image(image: TensorLike) -> tf.Tensor:
     """Implements Equalize function from PIL using TF ops."""
-
-    def scale_channel(image, channel):
-        """Scale the data in the channel to implement equalize."""
-        image_dtype = image.dtype
-        image = tf.cast(image[:, :, channel], tf.int32)
-
-        # Compute the histogram of the image channel.
-        histo = tf.histogram_fixed_width(image, [0, 255], nbins=256)
-
-        # For the purposes of computing the step, filter out the nonzeros.
-        nonzero = tf.where(tf.not_equal(histo, 0))
-        nonzero_histo = tf.reshape(tf.gather(histo, nonzero), [-1])
-        step = (tf.reduce_sum(nonzero_histo) - nonzero_histo[-1]) // 255
-
-        def build_lut(histo, step):
-            # Compute the cumulative sum, shifting by step // 2
-            # and then normalization by step.
-            lut = (tf.cumsum(histo) + (step // 2)) // step
-            # Shift lut, prepending with 0.
-            lut = tf.concat([[0], lut[:-1]], 0)
-            # Clip the counts to be in range.  This is done
-            # in the C code for image.point.
-            return tf.clip_by_value(lut, 0, 255)
-
-        # If step is zero, return the original image.  Otherwise, build
-        # lut from the full histogram and step and then index from it.
-
-        if step == 0:
-            result = image
-        else:
-            result = tf.gather(build_lut(histo, step), image)
-
-        return tf.cast(result, image_dtype)
-
-    idx = 2  # Channels last format
-    image = tf.stack([scale_channel(image, c) for c in range(image.shape[idx])], idx)
-
+    image = tf.stack([scale_channel(image, c) for c in range(image.shape[-1])], -1)
     return image
 
 
+@tf.function
 def equalize(image: TensorLike, name: Optional[str] = None) -> tf.Tensor:
     """Equalize image(s)
 
--- a/tensorflow_addons/image/color_ops.py
+++ b/tensorflow_addons/image/color_ops.py
@@ -30,7 +30,6 @@ from functools import partial
 def equalize_image(image: TensorLike) -> tf.Tensor:
     """Implements Equalize function from PIL using TF ops."""
 
-    @tf.function
     def scale_channel(image, channel):
         """Scale the data in the channel to implement equalize."""
         image_dtype = image.dtype


aaronmondal:
There are other functions such as `tf.image.translate_xy` which have the same problem. I think these can be "unfunctionized" as well. Users can always create their own wrappers like 
```python
@tf.function
def functionized_equalize(*args):
    return tfa.image.equalize(*args)
```
Please let me know wether I should change the other cases as well, or wether I should drop this PR because the current behaviour was intended.

WindQAQ:
> There are other functions such as `tf.image.translate_xy` which have the same problem. I think these can be "unfunctionized" as well. Users can always create their own wrappers like
> 
> ```python
> @tf.function
> def functionized_equalize(*args):
>     return tfa.image.equalize(*args)
> ```
> 
> Please let me know wether I should change the other cases as well, or wether I should drop this PR because the current behaviour was intended.

Thanks for the PR. Agree +1. We should let users decorate it instead of aggressively decorating in our libraries. Pull @seanpmorgan and @bhack into discussion.

bhack:
I think that we need to make a refresh pass on our old thread at https://github.com/tensorflow/addons/issues/13

My main concern Is that we are going to require to our users to check the internal implementation (source code) of our functions to evaluate if there could be a performance improvment for the specific function with wrapping.

E.g. An impl with many lines of python code Is different from an impl of just few native TF ops.

WindQAQ:
>  If one uses some features that are needed to be inside `@tf.function`, then at least put all functions used in that scope to public space, do not make it a local or anonymous function (like in this case). So that it will not trigger excessive graph re-tracing.

The colab link to support my argument. Avoid closure/lambda/local function in function when inside the scope of `tf.function`.

https://colab.research.google.com/drive/1WfehLDRpujQqvVimlXbqSKB7iu2e_i9d?usp=sharing

bhack:
> > If one uses some features that are needed to be inside `@tf.function`, then at least put all functions used in that scope to public space, do not make it a local or anonymous function (like in this case). So that it will not trigger excessive graph re-tracing.
> 
> The colab link to support my argument. Avoid closure/lambda/local function in function when inside the scope of `tf.function`.
> 
> https://colab.research.google.com/drive/1WfehLDRpujQqvVimlXbqSKB7iu2e_i9d?usp=sharing

If you `grep` TF source code it seems that as a policy it doesn't use `@tf.function` internally.

WindQAQ:
> If you `grep` TF source code it seems that as a policy it doesn't use `@tf.function` internally.

Nope.

```
tensorflow/python/compiler/xla/xla.py:    @def_function.function
tensorflow/python/tpu/tpu_embedding_v2.py:        @def_function.function
tensorflow/python/tpu/tpu_embedding_v2.py:  @def_function.function
tensorflow/python/tpu/tpu_embedding_v2.py:  @def_function.function
tensorflow/python/ops/image_ops_impl.py:      # def_function.function seems to lose shape information, so set it here.
tensorflow/python/ops/image_ops_impl.py:# def_function.function.
tensorflow/python/ops/image_ops_impl.py:@def_function.function(
tensorflow/python/ops/parallel_for/pfor.py:    concrete_function = def_function.function(fn).get_concrete_function(
tensorflow/python/ops/parallel_for/pfor.py:  @def_function.function
tensorflow/python/ops/parallel_for/pfor.py:    @def_function.function
tensorflow/python/ops/parallel_for/control_flow_ops.py:    functions_run_eagerly = def_function.functions_run_eagerly()
tensorflow/python/ops/parallel_for/control_flow_ops.py:    f = def_function.function(f)
tensorflow/python/ops/nccl_ops.py:    return def_function.function(_all_reduce)()
tensorflow/python/data/benchmarks/from_tensor_slices_benchmark.py:        @def_function.function
tensorflow/python/data/ops/dataset_ops.py:      if def_function.functions_run_eagerly():
tensorflow/python/distribute/distribute_lib.py:          self._reduce_sum_fns[axis] = def_function.function(reduce_sum_fn)
tensorflow/python/distribute/distribute_lib.py:        self._mean_reduce_helper_fns[axis] = def_function.function(
tensorflow/python/distribute/tpu_strategy.py:      tpu_function = def_function.function(tpu_function)
tensorflow/python/training/tracking/tracking.py:    @def_function.function(input_signature=[], autograph=False)
tensorflow/python/training/tracking/tracking.py:    @def_function.function(input_signature=[], autograph=False)
tensorflow/python/training/tracking/tracking.py:    @def_function.function(input_signature=[], autograph=False)
tensorflow/python/training/saving/functional_saver.py:  @def_function.function(
tensorflow/python/training/saving/functional_saver.py:  @def_function.function(
tensorflow/python/training/saving/functional_saver.py:      @def_function.function(experimental_compile=False)
tensorflow/python/training/saving/functional_saver.py:      @def_function.function(experimental_compile=False)
tensorflow/python/training/saving/saveable_object_util.py:  @def_function.function(
tensorflow/python/training/saving/saveable_object_util.py:  @def_function.function(input_signature=restored_type_specs)
tensorflow/python/eager/forwardprop.py:# Using function.defun rather than def_function.function avoids
tensorflow/python/eager/context.py:  @def_function.function
tensorflow/python/keras/engine/training.py:      train_function = def_function.function(
tensorflow/python/keras/engine/training.py:      predict_function = def_function.function(
tensorflow/python/keras/optimizer_v2/adam.py:  @def_function.function(experimental_compile=True)
tensorflow/python/keras/optimizer_v2/adam.py:  @def_function.function(experimental_compile=True)
tensorflow/python/keras/saving/saving_utils.py:  @def_function.function(input_signature=input_signature, autograph=False)
tensorflow/python/keras/metrics.py:        update_state_fn = def_function.function(obj.update_state)
tensorflow/python/keras/mixed_precision/experimental/loss_scale_benchmark.py:        run_fn = def_function.function(run_fn)
tensorflow/python/keras/distribute/distributed_training_utils.py:      distributed_function = def_function.function(distributed_function)
tensorflow/python/keras/distribute/simple_models.py:  @def_function.function
tensorflow/python/saved_model/signature_serialization.py:    @def_function.function
tensorflow/python/saved_model/save.py:          def_function.function(
```

bhack:
Yes if you see e.g.:
> tensorflow/python/ops/image_ops_impl.py:@def_function.function(

Is `non_max_suppression_padded_v2` that also needs a specific compilation hack:

```python
# TODO(b/158709815): Improve performance regression due to
# def_function.function.
@def_function.function(
    experimental_implements='non_max_suppression_padded_v2') 
```

I suppose that for cases like `non_max_suppression_padded_v2` the user not need to be aware if it is a long compositional op or it is just a wrapper of a native TF op e.g. this normal NMS has not a function wrapper:

https://github.com/tensorflow/tensorflow/blob/v2.3.1/tensorflow/python/ops/image_ops_impl.py#L2997-L3047

WindQAQ:
> Yes if you see e.g.:
> 
> > tensorflow/python/ops/image_ops_impl.py:@def_function.function(
> 
> Is `non_max_suppression_padded_v2` that also needs a specific compilation hack:
> 
> ```python
> # TODO(b/158709815): Improve performance regression due to
> # def_function.function.
> @def_function.function(
>     experimental_implements='non_max_suppression_padded_v2') 
> ```
> 
> I suppose that for cases like `non_max_suppression_padded_v2` the user not need to be aware if it is a long compositional op or it is just a wrapper of a native TF op e.g. this normal NMS has not a function wrapper:
> 
> https://github.com/tensorflow/tensorflow/blob/v2.3.1/tensorflow/python/ops/image_ops_impl.py#L2997-L3047

Yep, they are used internally but not being used in public APIs.

bhack:
is `def scale_channel` public?

WindQAQ:
> is `def scale_channel` public?

Ohoh, I see. Seems that we are misunderstanding each other :-(
It is a local function inside `equalize_image`. I mean using nested functions (defining functions inside function) will retrace the graph multiple times because the scope of them is not visible public, and the graphs cannot be cached.

Discourage:

```python
def fun()
    @tf.function
    def fun_inside_fun():
        pass
    local_fn()
```

Encourage:

```python
@tf.function
def fun_inside_fun():
    pass

def fun():
    fun_inside_fun()
```

bhack:
Yes

bhack:
And also  if `fun` is not public:

```
@tf.function
def fun():
    def fun_inside_fun():
        pass
    fun_inside_fun()
```

aaronmondal:
Another thing i just noticed  is that that in the above `grep` output the usages of `@tf.function`/`def_function` are usually used **contditionally**, i.e. after checking weather running in eager context. e.g. from `keras/engine/training.py`:
```python
    if not self.run_eagerly:
      train_function = def_function.function(
          train_function, experimental_relax_shapes=True)
```
It may make sense to do this in `tfa` as well.

WindQAQ:
> Another thing i just noticed is that that in the above `grep` output the usages of `@tf.function`/`def_function` are usually used **contditionally**, i.e. after checking weather running in eager context. e.g. from `keras/engine/training.py`:
> 
> ```python
>     if not self.run_eagerly:
>       train_function = def_function.function(
>           train_function, experimental_relax_shapes=True)
> ```
> 
> It may make sense to do this in `tfa` as well.

That's a good point! Make sense to me too. It will take some time to finalize the conclusion. If it's ok, can you open an issue for it? Thank you!

As for this PR, I suggest we can move `scale_channel` and `build_lut` outside any function. Do not define functions in functions. Given `tf.function` indeed grants some performance when decorating on `scale_channel`, maybe just leave `tf.function` on `scale_channel`. We do not want to see performance degradation after any PRs!

aaronmondal:
If it's ok with you @WindQAQ I'd change this PR to the way you proposed in your colab. For the other thing ill create a new issue.

Btw If i understand correctly the behaviour of `tf.function` can be controlled via [`tf.config.run_functions_eagerly`](https://www.tensorflow.org/api_docs/python/tf/config/run_functions_eagerly), and the current context can be checked via [`tf.executing_eagerly`](https://www.tensorflow.org/api_docs/python/tf/executing_eagerly), but i gotta check on how this actually works :D

AakashKumarNain:
> > is `def scale_channel` public?
> 
> Ohoh, I see. Seems that we are misunderstanding each other :-(
> I'm not saying public and private API. I mean using nested functions (defining functions inside function) will retrace the graph multiple times because the scope of them is not visible public, and the graphs cannot be cached.
> 
> Discourage:
> 
> ```python
> def fun():
>     @tf.function
>     def fun_inside_fun():
>         pass
>     fun_inside_fun()
> ```
> 
> Encourage:
> 
> ```python
> @tf.function
> def fun_inside_fun():
>     pass
> 
> def fun():
>     fun_inside_fun()
> ```

+1. Any function that is being decorated by `tf.function` should be separately defined.

fsx950223:
> > is `def scale_channel` public?
> 
> Ohoh, I see. Seems that we are misunderstanding each other :-(
> I'm not saying public and private API. I mean using nested functions (defining functions inside function) will retrace the graph multiple times because the scope of them is not visible public, and the graphs cannot be cached.
> 
> Discourage:
> 
> ```python
> def fun():
>     @tf.function
>     def fun_inside_fun():
>         pass
>     fun_inside_fun()
> ```
> 
> Encourage:
> 
> ```python
> @tf.function
> def fun_inside_fun():
>     pass
> 
> def fun():
>     fun_inside_fun()
> ```

Interesting! Tensorflow also decorates some inner functions with Function. Should they be changed too?
E.g:
https://github.com/fsx950223/tensorflow/blob/9aeb8f695c9b6658d710240989a55515d2a6ab1f/tensorflow/python/eager/def_function.py#L698

WindQAQ:
> Interesting! Tensorflow also decorates some inner functions with Function. Should they be changed too?
> E.g:
> https://github.com/fsx950223/tensorflow/blob/9aeb8f695c9b6658d710240989a55515d2a6ab1f/tensorflow/python/eager/def_function.py#L698

@fsx950223 It's ok if users do not define it in the loop like the example you post (at least there is no retracing warning). But in `equalize_image`, `scale_channel` seems to be defined repeatedly due to `tf.map_fn`.

In general, it's only my personal thought, I don't like define such stuff due to unclear python function in function/lambda/variable scope 😆

aaronmondal:
@fsx950223 I found loosening the shape constraints quite useful to get rid of inner `tf.function`s in some cases, e.g.
```python
@tf.function(
    input_signature=(
        tf.TensorSpec(shape=[None, None, 3], dtype=tf.float32),
        tf.TensorSpec(shape=[None, 4], dtype=tf.float32),
        tf.TensorSpec(shape=None, dtype=tf.int32),
        tf.TensorSpec(shape=None, dtype=tf.float32)
    )
)
def rand_augment(img, boxes, n_layers, magnitude):
    ...
```
lets you use `tf.function` at the top-level of augmentation pipelines where images  have variable shapes, instead of wrapping every augmentation in its own `tf.function`.

fsx950223:
> ```python
> @tf.function(
>     input_signature=(
>         tf.TensorSpec(shape=[None, None, 3], dtype=tf.float32),
>         tf.TensorSpec(shape=[None, 4], dtype=tf.float32),
>         tf.TensorSpec(shape=None, dtype=tf.int32),
>         tf.TensorSpec(shape=None, dtype=tf.float32)
>     )
> )
> def rand_augment(img, boxes, n_layers, magnitude):
> ```

``` python
@tf.function(experimental_relax_shapes=True)
def rand_augment(img, boxes, n_layers, magnitude):
```

aaronmondal:
@fsx950223 Thanks for the info! I did not realize that.

@WindQAQ Ok I refactored the code, changed some indexing to `tf.boolean_mask` and used a `tf.lookup.StaticHashTable` to map the image to the `lut`. In my tests this was faster and I found it more readable. In the current version I am using `@tf.function` at the top-level of `equalize`, which made it run ~40% faster in my tests. See this [colab](https://colab.research.google.com/gist/aaronmondal/bf7abd4f37520d401109d10448b404ad/untitled2.ipynb) for comparisons.

aaronmondal:
Just tested it with internal code. TFA tests pass and the the image calculations seem to be correct, but for some reason I am now getting different retracing errors. I gotta check whether this is a problem with the internal  code or the PR.

bhack:
@tomerk As we request to the contributor to spent time on profiling/refactor image equalize do you think that we could find a way to organize this in the ecosystem? 

We already have this one and 2 Equalize implementations:
https://github.com/tensorflow/models/blob/master/official/vision/beta/ops/augment.py#L475-L513
https://github.com/tensorflow/models/blob/master/official/vision/image_classification/augment.py#L480-L518

+ the equalize in the TF GSOC: 
https://github.com/tanzhenyu/image_augmentation/blob/master/image_augmentation/image/image_ops.py#L170-L215

WindQAQ:
> @WindQAQ Ok I refactored the code, changed some indexing to `tf.boolean_mask` and used a `tf.lookup.StaticHashTable` to map the image to the `lut`. In my tests this was faster and I found it more readable. In the current version I am using `@tf.function` at the top-level of `equalize`, which made it run ~25%-40% faster in my tests. See this [colab](https://colab.research.google.com/gist/aaronmondal/bf7abd4f37520d401109d10448b404ad/untitled2.ipynb) for comparisons.

That's awesome! I haven't thought that we can implement like that! 

> I also tried using `@tf.function` at the `scale_channel` and at the `equalize_image` level, but the further I moved it 'inside', the slower `equalize` became. I think since `equalize_image` and `scale_channel` are only used by `equalize` and not by any other functions, it is safe to simply decorate `equalize` at the top level.

I'm fine to this if there is no performance regression and graph retracing.

> Just tested it with internal code. TFA tests pass and the the image calculations seem to be correct, but for some reason I am now getting different retracing errors. I gotta check whether this is a problem with the internal code or the PR.

Feel free to ping me if all if done. Note that some retracing might occur between different unittests.

tensorflow_addons/image/color_ops.py
```diff
@@ -27,49 +27,41 @@
...
-            return tf.clip_by_value(lut, 0, 255)
-
-        # If step is zero, return the original image.  Otherwise, build
-        # lut from the full histogram and step and then index from it.
-
-        if step == 0:
-            result = image
-        else:
-            result = tf.gather(build_lut(histo, step), image)
+def scale_channel(image: TensorLike, channel: int) -> tf.Tensor:
```
WindQAQ, line 30: Let's prefix the internal function with `_` so change `scale_channel` to `_scale_channel`.

tensorflow_addons/image/color_ops.py
```diff
@@ -27,49 +27,41 @@
...
+        lookup_table = tf.lookup.StaticHashTable(lookup_initializer, 255)
+        result = lookup_table.lookup(image)
 
-        return tf.cast(result, image_dtype)
+    return tf.cast(result, image_dtype)
 
-    idx = 2  # Channels last format
-    image = tf.stack([scale_channel(image, c) for c in range(image.shape[idx])], idx)
 
+def equalize_image(image: TensorLike) -> tf.Tensor:
```
WindQAQ, line 58: ditto

aaronmondal:
The retracing error was a bug in the main code.

I tested various different versions to increase performance and realized the `tf.lookup` thing was a bad idea. It is slightly faster on smaller images significantly slower on larger images,
so I changed it back to `tf.gather` which is more readable anyways (contrary to what I stated earlier oO).

Shifting the `dtype` conversion around did not help improve performance. Interestingly, performance **decreased** when converting the whole image instead of the individual channels.

Finally, the one thing that **did** improve performance was using `tf.vectorized_map` instead of `tf.function`, however this currently only works on my locally compiled TF 2.5 and not in TF 2.3 since that version is missing a kernel for `tf.histogram_fixed_width`, so I wont change it for now. It should be changed in the future since TF 2.5 apparently has that kernel.

aaronmondal:
Doc build is failing It seems to be an issue with a missing `dataclasses` package in the testcontainer.

WindQAQ (approved):
It should be upstream issue. I'll take a look at it latter. Thank you for providing various benchmarks!

PR #2256, opened 2020-11-27
Remove GELU

aaronmondal:
Fixes #2252

As the above issue explains, GELU will be part of [`tf.keras.activations.gelu`](https://www.tensorflow.org/api_docs/python/tf/keras/activations/gelu).

PR #2250, opened 2020-11-23
Build against TF2.4

seanpmorgan:
# Description

Update TFA to build against TF2.4 

Fixes #2110

PR #2265, opened 2020-12-03
Add tf.nn.gelu alias for TF >= 2.4

### LINKED ISSUE #2252: Drop GELU for 0.13 release
https://www.tensorflow.org/api_docs/python/tf/keras/activations/gelu will be available in TF2.4. Deprecation warning is already set for our upcming 0.12 release

aaronmondal:
Fixes #2252

See discussion in the above mentioned issue and the related, dropped PR #2256


aaronmondal:
Mentioning @seanpmorgan .

seanpmorgan (changes requested):
Looks very good thanks! Mind just modifying the version check to be consistent with elsewhere in the repo:
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/losses/tests/quantiles_test.py#L88

tensorflow_addons/activations/gelu.py
```diff
@@ -71,7 +72,9 @@ def gelu(x: types.TensorLike, approximate: bool = True) -> tf.Tensor:
 
     x = tf.convert_to_tensor(x)
 
-    return _gelu_py(x, approximate)
+    gelu_op = tf.nn.gelu if LooseVersion(tf.__version__) >= "2.4" else _gelu_py
```
WindQAQ, line 75: We should also warn users that the default value of `approximate` is changed from `True` to `False` if `tf.nn.gelu` is being used. Like

```python
warnings.warn("Default value of `approximate` is changed from `True` to `False`")
```

seanpmorgan:
Thanks! Marking blocked since this should be tested with https://github.com/tensorflow/addons/pull/2250 . It shouldn't be long before the release now.

aaronmondal:
Added a warning in the docstring as well.

seanpmorgan:
@googlebot I consent.

seanpmorgan:
Hmm very slight difference w/ the keras implementation:
```
#23 357.3 =================================== FAILURES ===================================
#23 357.3 ______________ [doctest] tensorflow_addons.activations.gelu.gelu _______________
#23 357.3 [gw0] linux -- Python 3.6.12 /usr/local/bin/python
#23 357.3 051     Note that `approximate` will default to `False` from TensorFlow version 2.4 onwards.
#23 357.3 052     Consider using `tf.nn.gelu` instead.
#23 357.3 053 
#23 357.3 054     Usage:
#23 357.3 055 
#23 357.3 056     >>> tfa.options.TF_ADDONS_PY_OPS = True
#23 357.3 057     >>> x = tf.constant([-1.0, 0.0, 1.0])
#23 357.3 058     >>> tfa.activations.gelu(x, approximate=False)
#23 357.3 059     <tf.Tensor: shape=(3,), dtype=float32, numpy=array([-0.15865529,  0.        ,  0.8413447 ], dtype=float32)>
#23 357.3 060     >>> tfa.activations.gelu(x, approximate=True)
#23 357.3 Expected:
#23 357.3     <tf.Tensor: shape=(3,), dtype=float32, numpy=array([-0.158808,  0.      ,  0.841192], dtype=float32)>
#23 357.3 Got:
#23 357.3     <tf.Tensor: shape=(3,), dtype=float32, numpy=array([-0.15880796,  0.        ,  0.841192  ], dtype=float32)>
```

Is there anyway to increase doc test tolerance? Perhaps we just get rid of this doctest since gelu is on its way out? @WindQAQ

WindQAQ:
I think we can just change the value of doctest. Actually, old c++ impl also gets different values of pure python ops. There is slight difference in casting and constant precision between c++ impl/python in addons/python in tf.nn.gelu.

seanpmorgan (approved):
LGTM thanks again @aaronmondal !

PR #2267, opened 2020-12-05
[*.py] Rename "Arguments:" to "Args:"

SamuelMarks:
I've written custom parsers and emitters for everything from docstrings to classes and functions. However, I recently came across an issue with the TensorFlow codebase: inconsistent use of `Args:` and `Arguments:` in its docstrings. It is easy enough to extend my parsers to support both variants, howevever it looks like `Arguments:` is wrong anyway, as per:

  - https://google.github.io/styleguide/pyguide.html#doc-function-args @ [`ddccc0f`](https://github.com/google/styleguide/blob/ddccc0f/pyguide.md)

  - https://chromium.googlesource.com/chromiumos/docs/+/master/styleguide/python.md#describing-arguments-in-docstrings @ [`9fc0fc0`](https://chromium.googlesource.com/chromiumos/docs/+/9fc0fc0/styleguide/python.md)

  - https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html @ [`c0ae8e3`](https://github.com/sphinx-contrib/napoleon/blob/c0ae8e3/docs/source/example_google.rst)

 

Therefore, only `Args:` is valid. This PR replaces them throughout the TensorFlow Addons codebase.

PS: For related PRs, see tensorflow/tensorflow/pull/45420

### DIFF
--- a/tensorflow_addons/layers/crf.py
+++ b/tensorflow_addons/layers/crf.py
@@ -46,13 +46,13 @@ class CRF(tf.keras.layers.Layer):
     >>> chain_kernel.shape
     TensorShape([4, 4])
 
-    Arguments:
+    Args:
         units: Positive integer, dimensionality of the reservoir.
         chain_initializer: Orthogonal matrix. Default to `orthogonal`.
         use_boundary: `Boolean`, whether the layer uses a boundary vector. Default to `True`.
         boundary_initializer: Tensors initialized to 0. Default to `zeros`.
         use_kernel: `Boolean`, whether the layer uses a kernel weights. Default to `True`.
-    Call Arguments:
+    Call Args:
         inputs: Positive integer, dimensionality of the output space.
         mask: A boolean `Tensor` of shape `[batch_size, sequence_length]`
             or `None`. Default to `None`.
--- a/tensorflow_addons/callbacks/tqdm_progress_bar.py
+++ b/tensorflow_addons/callbacks/tqdm_progress_bar.py
@@ -223,7 +223,7 @@ class TQDMProgressBar(Callback):
     def format_metrics(self, logs={}, factor=1):
         """Format metrics in logs into a string.
 
-        Arguments:
+        Args:
             logs: dictionary of metrics and their values. Defaults to
                 empty dictionary.
             factor (int): The factor we want to divide the metrics in logs
--- a/tensorflow_addons/layers/adaptive_pooling.py
+++ b/tensorflow_addons/layers/adaptive_pooling.py
@@ -26,7 +26,7 @@ class AdaptivePooling1D(tf.keras.layers.Layer):
 
     This class only exists for code reuse. It will never be an exposed API.
 
-    Arguments:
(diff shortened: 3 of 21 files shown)


tensorflow_addons/layers/crf.py
```diff
@@ -46,13 +46,13 @@ class CRF(tf.keras.layers.Layer):
...
 
-    Arguments:
+    Args:
         units: Positive integer, dimensionality of the reservoir.
         chain_initializer: Orthogonal matrix. Default to `orthogonal`.
         use_boundary: `Boolean`, whether the layer uses a boundary vector. Default to `True`.
         boundary_initializer: Tensors initialized to 0. Default to `zeros`.
         use_kernel: `Boolean`, whether the layer uses a kernel weights. Default to `True`.
-    Call Arguments:
+    Call Args:
```
SamuelMarks, line 55: :thinking: `Call Arguments:` also seems like a custom thing

WindQAQ:
Thanks for the PR! Although Tensorflow doc generator can parse each listing like `Foo:`, `Bar zoo:`, it's a great idea to clean them up at least in terms of consistency in our repo.

cc @seanpmorgan and @bhack to see if we should follow what core TF does.

bhack:
LGTM

WindQAQ:
BTW, we use `Arguments` because it's keras.io docs semantics.

SamuelMarks:
@WindQAQ Thanks for the pointer, just sent similar PRs to the [keras-team](https://github.com/keras-team) organisation.

AakashKumarNain:
LGTM

bhack:
We need an update of the linter like https://github.com/tensorflow/tensorflow/pull/45427 when It works

AakashKumarNain:
@bhack I didn't get it. Are you suggesting to refactor the docstrings and align them with TF core?

bhack:
No to add configuration in pylint also here so that It could be automatically detected also by our CI

bhack:
Here probably we need something like https://github.com/terrencepreilly/darglint#flake8 cause we are not using pylint in this repo.

AakashKumarNain:
> Here probably we need something like https://github.com/terrencepreilly/darglint#flake8 cause we are not using pylint in this repo.

Yes we can do that

qlzh727 (approved):
LGTM for RNN related code.

Google's internal python style guide also use "Args".

seanpmorgan (commented):
LGTM thanks @SamuelMarks !

bhack:
I think we need to have linting support if we want to enforce this as requested at https://github.com/tensorflow/addons/pull/2267#issuecomment-739557920 and in TF.

bhack:
I would have preferred to give priority to a working linting PR then merging this directly. @SamuelMarks Can you take in charge also a linting PR?

SamuelMarks:
@bhack Sure thing. I've got this tab open, just at the end of a deadline then will get right on it

Issue #2279, opened 2020-12-09
Reduce nested function definitions via @tf.function


aaronmondal:
As already discussed, in #2264 there are some inconsistencies with the usage of `@tf.function` and undesirable nested functions which may cleaned up by using `@tf.function`.

After running `grep -r -n "tf.function" .`, removing `test` files and `README` files I found the following:

```bash
./text/crf.py
```
This file was already addressed in #598 by @seanpmorgan, but the PR was closed because it was unclear how to handle variable shapes (which seem kind of necessary for text data :D). Looser shape constraints using explicit `input_signature`s in `@tf.function` may be a simple way to improve performance while staying flexible. **Note** that using looser shape constraints may result in slower execution than when using static shapes as pointed out by @fsx950223 [here](https://github.com/tensorflow/addons/pull/2264#issuecomment-740428815). I expect this would still be faster than no decoration at all.

```bash
./metrics/cohens_kappa.py
```
This file is rather inconsistent. Some functions are decorated, others are not. I assume a refactorization will lead to increased readability/maintainability and `@tf.function` decorations at the top level should give a nice performance gain.

```bash
./image/connected_components.py
```
This uses `tf.cond` internally. As far as I understand, `if` statements are translated to `tf.cond` anyways, provided that the condition is a `tf.Tensor` and the branches are compatible (see [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/autograph/g3doc/reference/control_flow.md#if-statements)). Changing this to a regular `if` statement may therefore improve readability. It may also make it possible to get rid of the nested function definitions in this file.
```bash
./image/color_ops.py
```
While `equalize` was refactored in #2264, `sharpness` is currently undecorated. Decorating may lead to a nice performance improvement.

```bash
./image/translate_ops.py
```
While `translate` is decorated, `translate_xy` is not. Since `translate_xy` makes use of `translate` internally, my assumption is that `translate` will excessively retrace in some cases, while `translate_xy` on its own is rather slow since it is untraced when used via the public API.

### `tf.cond` and `tf.while`
After checking the outputs of `grep -r -n "tf.cond" .` I noticed the following:

```bash
./layers/wrappers.py
./seq2seq/sampler.py
./seq2seq/beam_search_decoder.py
./optimizers/novograd.py
```
These use `tf.cond`. With `tf.function` we may be able to get rid of the nested function definitions and improve readability.

[...]

PR #2287, opened 2020-12-12
Refactor sharpness

aaronmondal:
Part of #2279.

Changes:
1. Made `sharpness_image` private (`_sharpness_image` now), as `sharpness` already handles single images and `_equalize_image` is private as well.
2. Made `sharpness` compatible with grayscale images (previously only worked with RGB).
3. Added the `name` argument and applied a `tf.name_scope`.
4. Decorated with `@tf.function`. **Now runs ~60% faster.** ([colab](https://colab.research.google.com/drive/1uZzz4G-Bfme1-9jaBsMPYTzykA25GsTu#scrollTo=mFA_HXxMmvhd))
5. Removed the `clip_by_value` invocation since the sharpening kernel is already normalized.
6. Refactored the tests. `sharpness` and `equalize` now both handle all input shapes.

Also removed a faulty test for `equalize`. That test was for the now unsupported `channels_first` and was passing because it was just renaming the `name_scope` to `channels_first` :D.

### DIFF
--- a/tensorflow_addons/image/color_ops.py
+++ b/tensorflow_addons/image/color_ops.py
@@ -79,39 +79,44 @@ def equalize(image: TensorLike, name: Optional[str] = None) -> tf.Tensor:
         return from_4D_image(image, image_dims)
 
 
-def sharpness_image(image: TensorLike, factor: Number) -> tf.Tensor:
+def _sharpness_image(image: TensorLike, factor: Number) -> tf.Tensor:
     """Implements Sharpness function from PIL using TF ops."""
     orig_image = image
     image_dtype = image.dtype
-    # SMOOTH PIL Kernel.
+    image_channels = image.shape[-1]
     image = tf.cast(image, tf.float32)
+
+    # SMOOTH PIL Kernel.
     kernel = (
         tf.constant(
             [[1, 1, 1], [1, 5, 1], [1, 1, 1]], dtype=tf.float32, shape=[3, 3, 1, 1]
         )
         / 13.0
     )
-    # Tile across channel dimension.
-    kernel = tf.tile(kernel, [1, 1, 3, 1])
-    strides = [1, 1, 1, 1]
+    kernel = tf.tile(kernel, [1, 1, image_channels, 1])
+
+    # Apply kernel channel-wise.
+    # TODO: Use tf.nn.conv2d as soon as grouped convolutions have CPU support.
     degenerate = tf.nn.depthwise_conv2d(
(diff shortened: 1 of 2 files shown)


WindQAQ (approved):

WindQAQ:
@aaronmondal can you make the colab link public to "anyone who knows the link can access it"? Other than this, the PR LGTM.

aaronmondal:
Link: https://colab.research.google.com/drive/1uZzz4G-Bfme1-9jaBsMPYTzykA25GsTu?usp=sharing
Sorry I always forget that the default link is not public :O

tensorflow_addons/image/color_ops.py
```diff
@@ -79,51 +79,56 @@ def equalize(image: TensorLike, name: Optional[str] = None) -> tf.Tensor:
...
-    kernel = tf.tile(kernel, [1, 1, 3, 1])
-    strides = [1, 1, 1, 1]
+    kernel = tf.tile(kernel, [1, 1, image_channels, 1])
+
+    # Apply kernel channel-wise.
     degenerate = tf.nn.depthwise_conv2d(
-        image, kernel, strides, padding="VALID", dilations=[1, 1]
+        image, kernel, strides=[1, 1, 1, 1], padding="VALID", dilations=[1, 1]
     )
-    degenerate = tf.clip_by_value(degenerate, 0.0, 255.0)
```
WindQAQ, line 100: I'd say we have to keep the clip here because we cann't know if users pass a valid image into the function.

WindQAQ: NVM. I think you're right.

aaronmondal: @WindQAQ Ah I see this is inconsistent with `blend`. My thought was if someone were to use an image scaled to `(-1, 1)` or `[-127.5, 127.5]`, this would cut off the bottom half and removing it would enable these scaled images as well. It should work anyways right now, because `blend` clips to `[0, 255]`, but it may make sense to remove the clip in blend to enable images in more scaling ranges. Gotta see whether this works with other ops that make use of `blend` though.

aaronmondal: Edit: Things still works with the "supported" image ranges, but zero-centered ranges will be cut off by `blend`.

WindQAQ: I think the clip should be removed because people might want to apply these ops during post-processing. In this case, those images are not supposed to be VALID IMO. Got to check if CV2/scipy have similar functionality.

PR #2293, opened 2020-12-15
Use ImageProjectiveTransformV3 for TF >= 2.4.0

WindQAQ:
# Description

Support fill_value and nearest fill_mode with ImageProjectiveTransformV3.

## Type of change

- [ ] Bug fix
- [ ] New Tutorial
- [ ] Additional Testing

# Checklist:

# How Has This Been Tested?

Test new features.

### DIFF
--- a/tensorflow_addons/image/transform_ops.py
+++ b/tensorflow_addons/image/transform_ops.py
@@ -144,7 +144,7 @@ def transform(
                 )
             if fill_mode == "NEAREST":
                 raise ValueError(
-                    "NEAREST fill mode is not supported for TensorFlow < 2.4.0."
+                    "NEAREST fill_mode is not supported for TensorFlow < 2.4.0."
                 )
             output = tf.raw_ops.ImageProjectiveTransformV2(
                 images=images,
--- a/tensorflow_addons/image/transform_ops.py
+++ b/tensorflow_addons/image/transform_ops.py
@@ -37,6 +37,7 @@ def transform(
     transforms: TensorLike,
     interpolation: str = "NEAREST",
     fill_mode: str = "CONSTANT",
+    fill_value: TensorLike = 0.0,
     output_shape: Optional[list] = None,
     name: Optional[str] = None,
 ) -> tf.Tensor:
@@ -57,7 +58,7 @@ def transform(
       interpolation: Interpolation mode.
         Supported values: "NEAREST", "BILINEAR".
       fill_mode: Points outside the boundaries of the input are filled according
-        to the given mode (one of `{'constant', 'reflect', 'wrap'}`).
+        to the given mode (one of `{'constant', 'reflect', 'wrap', 'nearest'}`).
         - *reflect*: `(d c b a | a b c d | d c b a)`
           The input is extended by reflecting about the edge of the last pixel.
         - *constant*: `(k k k k | a b c d | k k k k)`
@@ -65,6 +66,10 @@ def transform(
           same constant value k = 0.
         - *wrap*: `(a b c d | a b c d | a b c d)`
           The input is extended by wrapping around to the opposite edge.
+        - *nearest*: `(a a a a | a b c d | d d d d)`
+          The input is extended by the nearest pixel.
+      fill_value: a float represents the value to be filled outside the
+        boundaries when `fill_mode` is "constant".
       output_shape: Output dimesion after the transform, [height, width].
         If None, output is the same size as input image.
 
@@ -115,13 +120,16 @@ def transform(
                 % len(transforms.get_shape())
             )
 
-        # TODO(WindQAQ): Support "nearest" `fill_mode` and `fill_value` in TF2.4.
-        output = tf.raw_ops.ImageProjectiveTransformV2(
+        fill_value = tf.convert_to_tensor(
+            fill_value, dtype=tf.float32, name="fill_value"
+        )
+        output = tf.raw_ops.ImageProjectiveTransformV3(
             images=images,
             transforms=transforms,
             output_shape=output_shape,
             interpolation=interpolation.upper(),
             fill_mode=fill_mode.upper(),
+            fill_value=fill_value,
         )
         return img_utils.from_4D_image(output, original_ndims)
 
@@ -281,6 +289,7 @@ def rotate(
     angles: TensorLike,
     interpolation: str = "NEAREST",
     fill_mode: str = "CONSTANT",
+    fill_value: TensorLike = 0.0,
     name: Optional[str] = None,
 ) -> tf.Tensor:
     """Rotate image(s) counterclockwise by the passed angle(s) in radians.
@@ -296,7 +305,7 @@ def rotate(
       interpolation: Interpolation mode. Supported values: "NEAREST",
         "BILINEAR".
       fill_mode: Points outside the boundaries of the input are filled according
-        to the given mode (one of `{'constant', 'reflect', 'wrap'}`).
+        to the given mode (one of `{'constant', 'reflect', 'wrap', 'nearest'}`).
         - *reflect*: `(d c b a | a b c d | d c b a)`
           The input is extended by reflecting about the edge of the last pixel.
         - *constant*: `(k k k k | a b c d | k k k k)`
@@ -304,6 +313,10 @@ def rotate(
           same constant value k = 0.
         - *wrap*: `(a b c d | a b c d | a b c d)`
           The input is extended by wrapping around to the opposite edge.
+        - *nearest*: `(a a a a | a b c d | d d d d)`
+          The input is extended by the nearest pixel.
+      fill_value: a float represents the value to be filled outside the
+        boundaries when `fill_mode` is "constant".
       name: The name of the op.
 
     Returns:
@@ -327,6 +340,7 @@ def rotate(
             angles_to_projective_transforms(angles, image_height, image_width),
             interpolation=interpolation,
             fill_mode=fill_mode,
+            fill_value=fill_value,
         )
         return img_utils.from_4D_image(output, original_ndims)
 
--- a/tensorflow_addons/image/transform_ops.py
+++ b/tensorflow_addons/image/transform_ops.py
@@ -14,6 +14,9 @@
 # ==============================================================================
 """Image transform ops."""
 
+import warnings
+from distutils.version import LooseVersion
+
 import tensorflow as tf
 from tensorflow_addons.image import utils as img_utils
 from tensorflow_addons.utils.types import TensorLike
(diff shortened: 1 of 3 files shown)


tensorflow_addons/image/transform_ops.py
```diff
@@ -37,6 +37,7 @@ def transform(
     transforms: TensorLike,
     interpolation: str = "NEAREST",
     fill_mode: str = "CONSTANT",
+    fill_value: TensorLike = 0.0,
```
WindQAQ, line 40: I have no idea of the where should we put the new argument. Should we maintain backward compatibility so we have to put it in the end of args? @seanpmorgan WDYT?

WindQAQ:
@seanpmorgan Hi Sean, this is only for TF2.4. Our next release will pin to both TF2.3 and TF2.4 or 2.4 only?

seanpmorgan:
> @seanpmorgan Hi Sean, this is only for TF2.4. Our nightly version will pin to both TF2.3 and TF2.4, right?

Trying to expand our python compatibility to be 2.3+. The wheel will be built against TF2.4 so custom-ops will need TF2.4, but everything else would be great to keep compatibility. 

Raising a NotImplementedError for <=2.4 is an option

bhack:
Tracking many TF core issues and PRs my opinion Is that currently TF has a quite strict policy to not backport not only features but also bug fixes from nightly/master. Patch releases are sparse and for security fixes only.

So I think that we need to start to support and promote only the last official TF release.

WindQAQ:
> Trying to expand our python compatibility to be 2.3+. The wheel will be built against TF2.4 so custom-ops will need TF2.4, but everything else would be great to keep compatibility.
> 
> Raising a NotImplementedError for <=2.4 is an option

I can do a warning for TF<2.4 that `fill_value` is not supported and is always 0.

> Tracking many TF core issues and PRs my opinion Is that currently TF has a quite strict policy to not backport not only features but also bug fixes from nightly/master. Patch releases are sparse and for security fixes only.
> 
> So I think that we need to start to support and promote only the last official TF release.

I am more with this proposal though, there is always a loop from "I want the new feature" -> "Please use latest TF" -> "codes does not work in latest TF" -> "Fallback to older TF" -> "I want the new feature" XD

bhack:
> I am more with this proposal though, there is always a loop from "I want the new feature" -> "Please use latest TF" -> "codes does not work in latest TF" -> "Fallback to older TF" -> "I want the new feature" XD

Often it Is more like:
There Is a bug -> It Is fixed in the last version, or in nightly/master -> please use the last official or nightly wheels.

tensorflow_addons/image/transform_ops.py
```diff
@@ -120,17 +123,36 @@ def transform(
...
+                images=images,
+                transforms=transforms,
+                output_shape=output_shape,
+                interpolation=interpolation.upper(),
+                fill_mode=fill_mode.upper(),
+                fill_value=fill_value,
+            )
+        else:
+            fill_mode = fill_mode.upper()
+            # TODO(WindQAQ): Get rid of the check once we drop TensorFlow < 2.4 support.
```
bhack, line 140: Can we automatically/programmatically fail with https://github.com/tensorflow/addons/blob/master/setup.py#L79?
What do you think? I suppose that this kind of things will stay forever in the code :wink:

tensorflow_addons/image/transform_ops.py
```diff
@@ -37,6 +40,7 @@ def transform(
     transforms: TensorLike,
     interpolation: str = "NEAREST",
     fill_mode: str = "CONSTANT",
```
aaronmondal, line 42: Could we set these set these strings to small caps `"nearest"` and `"constant"` (in the `Args` below as well)? This would be more consistent with the docstrings in `tf.keras.layers` and we have an `.upper()` in the code anyways.

tensorflow_addons/image/transform_ops.py
```diff
@@ -57,14 +61,18 @@ def transform(
...
-        to the given mode (one of `{'constant', 'reflect', 'wrap'}`).
+        to the given mode (one of `{'constant', 'reflect', 'wrap', 'nearest'}`).
         - *reflect*: `(d c b a | a b c d | d c b a)`
           The input is extended by reflecting about the edge of the last pixel.
         - *constant*: `(k k k k | a b c d | k k k k)`
           The input is extended by filling all values beyond the edge with the
           same constant value k = 0.
         - *wrap*: `(a b c d | a b c d | a b c d)`
           The input is extended by wrapping around to the opposite edge.
+        - *nearest*: `(a a a a | a b c d | d d d d)`
```
aaronmondal, line 72: Does this actually work? According to the docs for `ProjectiveTransformV3` the `'nearest'` option is for the `interpolation` arg, not the `fill_value` arg. There is no test for `nearest` in `transform_ops_test.py` either.

tensorflow_addons/image/translate_ops.py
```diff
@@ -76,6 +76,7 @@ def translate(
     translations: TensorLike,
     interpolation: str = "NEAREST",
     fill_mode: str = "CONSTANT",
```
aaronmondal, line 78: Same here.

tensorflow_addons/image/translate_ops.py
```diff
@@ -92,14 +93,18 @@ def translate(
...
-        to the given mode (one of `{'constant', 'reflect', 'wrap'}`).
+        to the given mode (one of `{'constant', 'reflect', 'wrap', 'nearest'}`).
         - *reflect*: `(d c b a | a b c d | d c b a)`
           The input is extended by reflecting about the edge of the last pixel.
         - *constant*: `(k k k k | a b c d | k k k k)`
           The input is extended by filling all values beyond the edge with the
           same constant value k = 0.
         - *wrap*: `(a b c d | a b c d | a b c d)`
           The input is extended by wrapping around to the opposite edge.
+        - *nearest*: `(a a a a | a b c d | d d d d)`
```
aaronmondal, line 104: And here.

aaronmondal (changes requested):
Thanks I need this so much! Was going crazy over the inconsistency between the different fill options for geometric transformations :D.

The only thing I am not so sure about is whether the `nearest` fill option actually does work.

There should probably be tests for the `"nearest"` and `"bilinear"` **interpolation** modes of `transform` as well. Currently, the `interpolation` argument is untested.

The `test_bilinear` and `test_uint8` tests in `transform_ops_test.py` apply to `rotate` only, which does make use of `transform` internally, but I don't think this relation should be assumed by the test case.

tensorflow_addons/image/transform_ops.py
WindQAQ: Yes it works, I forgot to add cc documentation in old PR :-(

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/image/image_ops.h#L113-L119

fsx950223:
The function should be transfered to official repo? There is a similar function https://github.com/tensorflow/tensorflow/blob/2a7d45c6800a99e4158526e25ec128ddd90e37f7/tensorflow/python/keras/layers/preprocessing/image_preprocessing.py#L625.

WindQAQ:
> The function should be transfered to official repo? There is a similar function https://github.com/tensorflow/tensorflow/blob/2a7d45c6800a99e4158526e25ec128ddd90e37f7/tensorflow/python/keras/layers/preprocessing/image_preprocessing.py#L625.

They do not make it public. In fact, they migrate it to KPL without informing us early this year 😭

bhack:
> 
> They do not make it public. In fact, they migrate it to KPL without informing us early this year sob So we are just doing a fork of private API now.

Lets see :wink: ... https://github.com/tensorflow/tensorflow/pull/45742

tensorflow_addons/image/transform_ops.py
seanpmorgan: Yes this looks like the best position

tensorflow_addons/image/transform_ops.py
seanpmorgan: Eh, we can clean up the codebase as our compatibility matrix moves. See https://github.com/tensorflow/addons/pull/2294

seanpmorgan:
> > They do not make it public. In fact, they migrate it to KPL without informing us early this year sob So we are just doing a fork of private API now.
> 
> Lets see 😉 ... [tensorflow/tensorflow#45742](https://github.com/tensorflow/tensorflow/pull/45742)

Agree with the PR, though I don't want to have to wait for a response prior to 0.12 release.

bhack:
> Agree with the PR, though I don't want to have to wait for a response prior to 0.12 release.

It is impossible to achieve this for 0.12 cause this will got in TF 2.5 but you can bypass the policy and depend on a private API for this case, as you want.

seanpmorgan:
> > Agree with the PR, though I don't want to have to wait for a response prior to 0.12 release.
> 
> It is impossible to achieve this for 0.12 cause this will got in TF 2.5 but you can bypass the policy and depend on a private API for this case, as you want.

Understand the TF core PR wouldn't be until TF2.5, but what do you mean depend on a private API and bypass the policy? There is no private API in this PR right?

bhack:
> > > Agree with the PR, though I don't want to have to wait for a response prior to 0.12 release.
> > 
> > 
> > It is impossible to achieve this for 0.12 cause this will got in TF 2.5 but you can bypass the policy and depend on a private API for this case, as you want.
> 
> Understand the TF core PR wouldn't be until TF2.5, but what do you mean depend on a private API and bypass the policy? There is no private API in this PR right?

I meant as an alternative to maintain a (function) fork.

tensorflow_addons/image/transform_ops.py
WindQAQ: Updated.

tensorflow_addons/image/transform_ops.py
WindQAQ: Updated.

tensorflow_addons/image/transform_ops.py
```diff
@@ -35,10 +38,11 @@
 def transform(
     images: TensorLike,
     transforms: TensorLike,
-    interpolation: str = "NEAREST",
-    fill_mode: str = "CONSTANT",
+    interpolation: str = "nearest",
+    fill_mode: str = "constant",
```
WindQAQ, line 42: This will not break the compatibility as we always uppercase it when passing into raw ops.

aaronmondal:
@WindQAQ Thanks for changing to lowercase. So much prettier :). Would you like me to file a PR for the missing testcases in `interpolation`?

WindQAQ:
> @WindQAQ Thanks for changing to lowercase. So much prettier :). Would you like me to file a PR for the missing testcases in `interpolation`?

It's more than welcome to do that!

aaronmondal (approved):

seanpmorgan (approved):
LGTM Thanks! Thanks @aaronmondal for the review!

tensorflow_addons/image/transform_ops.py
bhack: Yes but It could be easier if we are failing in different points in the sourfe when we move the minimal version or the single version we support. Do you understand what I meant?

bhack:
> > > > Agree with the PR, though I don't want to have to wait for a response prior to 0.12 release.
> > > 
> > > 
> > > It is impossible to achieve this for 0.12 cause this will got in TF 2.5 but you can bypass the policy and depend on a private API for this case, as you want.
> > 
> > 
> > Understand the TF core PR wouldn't be until TF2.5, but what do you mean depend on a private API and bypass the policy? There is no private API in this PR right?
> 
> I meant as an alternative to maintain a (function) fork.

It seems that we need to maintain the fork but if the new policy Is to maintain the public API surface minimal isn't better to handle private API occasional breakage instead of maintain fork in multiple points in the source code?

PR #2299, opened 2020-12-17
Add AVX2 support

### LINKED ISSUE #1148: Enable SSE4.2 and AVX support during build
So the pip installed TF does not support these instruction sets by default, but modern-ish CPUs do. (Roughly CPUs after 2012).

We could try this and see if there are any improvements in test times and weight the benefits.  If nothing else we can add it as a flag for building from source. Currently TF-IO does this by default:
https://github.com/tensorflow/io/blob/master/.github/workflows/build.yml#L13

@perfinion do we know if this is on the roadmap for default TF installations?

seanpmorgan:
# Description

Support AVX2 compilation by default

Fixes #1148

Don't think we need to support anything older than AVX2. Also did some cleanup on our builds

### DIFF
--- a/.github/workflows/make_wheel_Windows.sh
+++ b/.github/workflows/make_wheel_Windows.sh
@@ -6,8 +6,9 @@ export BAZEL_VC="C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/
 python -m pip install --default-timeout=1000 wheel setuptools tensorflow==$TF_VERSION
 bash ./tools/testing/build_and_run_tests.sh
 
+python configure.py
+
 bazel.exe build \
-  -c opt \
   --enable_runfiles \
   --noshow_progress \
   --noshow_loading_progress \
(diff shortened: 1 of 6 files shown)


.github/workflows/make_wheel_Windows.sh
```diff
@@ -6,8 +6,9 @@ export BAZEL_VC="C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/
 python -m pip install --default-timeout=1000 wheel setuptools tensorflow==$TF_VERSION
 bash ./tools/testing/build_and_run_tests.sh
 
+python configure.py
+
 bazel.exe build \
-  -c opt \
```
WindQAQ, line 10: Why `-c opt` is removed?

WindQAQ:
Can we also enable FMA with `-mfma`? It's separated from AVX2.

.github/workflows/make_wheel_Windows.sh
seanpmorgan: It's written by default in configure:
https://github.com/tensorflow/addons/blob/master/configure.py#L108

Just wanted to set that as the central source of build args

WindQAQ (approved):
LGTM. Thank you!

svobora:
My i7-3930k and i7-4930k send their regards.

seanpmorgan:
> My i7-3930k and i7-4930k send their regards.

We'll release a 0.12.1 path version in the near future:
https://github.com/tensorflow/addons/pull/2355

PR #2301, opened 2020-12-17
Move enable_runfiles flag to configure.py

seanpmorgan:
# Description

Cleanup where `--enable_runfiles` flag is set. Previously we just put this in the documentation for all OS since it didn't hurt but now configure will handle it and users don't need to worry about it.


WindQAQ (approved):
LGTM. Thanks!

PR #2329, opened 2020-12-31
Improve version check variable names

seanpmorgan:
# Description

Small improvement on the variable names. Previously it was difficult to know that MAX_VERISON was not included.

# Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,8 +76,8 @@ class BinaryDistribution(Distribution):
 
 
 project_name, version = get_project_name_version()
-min_tf_version = version["MIN_TF_VERSION"]
-max_tf_version = version["MAX_TF_VERSION"]
+inclusive_min_tf_version = version["INCLUSIVE_MIN_TF_VERSION"]
+exclusive_max_tf_version = version["EXCLUSIVE_MAX_TF_VERSION"]
 setup(
     name=project_name,
     version=version["__version__"],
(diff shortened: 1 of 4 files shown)


seanpmorgan:
Do we think we need to store the old variable name?
`MAX_TF_VERSION = EXCLUSIVE_MAX_TF_VERSION = "2.5.0"`

I haven't seen anyone utilize these variables outside of Addons, and we are on 0.x release but it is possible others are using these.

WindQAQ:
> Do we think we need to store the old variable name?
> e.g. `MAX_TF_VERSION = EXCLUSIVE_MAX_TF_VERSION = "2.5.0"`
> 
> I haven't seen anyone utilize these variables outside of Addons, and we are on 0.x release but it is possible others are using these.

I don't think we should. Users should afford the consequence of using non-public API. I am thinking to prefix those variables with `_`. Though there is nothing really private in Python, it should be enough to tell users that one shouldn't use those stuff.

seanpmorgan:
> > Do we think we need to store the old variable name?
> > e.g. `MAX_TF_VERSION = EXCLUSIVE_MAX_TF_VERSION = "2.5.0"`
> > I haven't seen anyone utilize these variables outside of Addons, and we are on 0.x release but it is possible others are using these.
> 
> I don't think we should. Users should afford the consequence of using non-public API. I am thinking to prefix those variables with `_`. Though there is nothing really private in Python, it should be enough to tell users that one shouldn't use those stuff.

Generally agree with all above, but since we import these into a different modules (e.g. `ensure_tf_install`) maybe we should just keep them as global syntax as done in this pr?

WindQAQ:
> Generally agree with all above, but since we import these into a different modules (e.g. `ensure_tf_install`) maybe we should just keep them as global syntax as done in this pr?

Looks good to me :-)

WindQAQ (approved):

PR #2344, opened 2021-01-13
Add experimental_implements attribute to dense_image_warp

thaink:
This attribute will help the TFLite conversion convert the whole block to a single TFLite custom op, significantly reduce the runtime.

# Description

Add experimental_implements attribute to dense_image_warp

## Type of change

- [ ] Bug fix
- [ ] New Tutorial
- [ ] Additional Testing
- [x] New Image addition and the changes conform to the [image op contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/README.md#contribution-guidelines)

# Checklist:

# How Has This Been Tested?

### DIFF
--- a/tensorflow_addons/image/dense_image_warp.py
+++ b/tensorflow_addons/image/dense_image_warp.py
@@ -158,7 +158,7 @@ def _get_dim(x, idx):
     return x.shape[idx] or tf.shape(x)[idx]
 
 
-@tf.function(experimental_implements="addons:DenseImageWarp")
+@tf.function
 def dense_image_warp(
     image: types.TensorLike, flow: types.TensorLike, name: Optional[str] = None
 ) -> tf.Tensor:
@@ -222,3 +222,18 @@ def dense_image_warp(
         interpolated = interpolate_bilinear(image, query_points_flattened)
         interpolated = tf.reshape(interpolated, [batch_size, height, width, channels])
         return interpolated
+
+
+@tf.function(experimental_implements="addons:DenseImageWarp")
+def dense_image_warp_annotated(
+    image: types.TensorLike, flow: types.TensorLike, name: Optional[str] = None
+) -> tf.Tensor:
+    """Similar to dense_image_warp but annotated with experimental_implements.
+
+    This annotation make the serialized function detectable by the TFLite MLIR
+    converter and allow the converter to convert it to corresponding TFLite op.
+
+    However, with the annotation, this function cannot be used with backprop
+    under `tf.GradientTape` objects.
+    """
+    return dense_image_warp(image, flow, name)
--- a/tensorflow_addons/image/dense_image_warp.py
+++ b/tensorflow_addons/image/dense_image_warp.py
@@ -158,7 +158,7 @@ def _get_dim(x, idx):
     return x.shape[idx] or tf.shape(x)[idx]
 
 
-@tf.function
+@tf.function(experimental_implements="addons:DenseImageWarp")
 def dense_image_warp(
     image: types.TensorLike, flow: types.TensorLike, name: Optional[str] = None
 ) -> tf.Tensor:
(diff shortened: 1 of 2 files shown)


WindQAQ:
Hi @thaink, thanks for bringing this up. Seems that it breaks the backprop and unranked tensor inference, can you figure out why?

Also, it would be great to add yourself to CODEOWNER to get the notification when the implementation changes (In case you need to rewrite mlir patterns).

Thanks!

thaink:
It is weird to break those as this PR only add a name to the tf.function.
Let me investigate why.

thaink:
@WindQAQ I fixed the error by adding a new function. WDYT?

I am not really understand the code so well. So might not fit to put myself to CODEOWERS for this file.

WindQAQ (approved):
LGTM if we don't export it.

Is there any bug or somthing else in backprop? Will it hurt for any function annotated with `experimental_implements`, such as  `max_unpool2d`?

thaink:
According to https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/eager/function.py#L1544, Function has attr cannot depend on any tensors outside of its signature or modify variables. But the backprop implicitly add some captured input to it.

@marksandler2 is my understand correct? Although the dense_image_warp function here does not use any outside variables but look like some are added under tf.GradientTape?

marksandler2:
I think this is intended though unfortunate behavior. When a function is executed under gradient tape it now produces additional outputs that no longer have stable signature (e.g. they depend on the internal implementation of tf.function).
Thus it can't have the experimental_implement annotation. One option is to drop the annotation in those situations, which would make total sense in situations like you've encountered, however i am afraid that this would lead to silent dropping of annotations when the change is accidental. 

@thaink can you open a separate issue on Tensorflow side and cc me?

marksandler2:
FYI: the issue has been fixed at tensorflow head

thaink:
Hi @WindQAQ,
The since the fix is in tf-nightly, can I make a PR to merge dense_image_warp and dense_image_warp_annotated?

thaink:
@WindQAQ In my test the test_unknown_shapes in tensorflow_addons/image/tests/dense_image_warp_test.py failed after the fix. Is it OK drop the unknown shape support?

@marksandler2 The error I mentioned above is at here:
fn = dense_image_warp.get_concrete_function(
        tf.TensorSpec(shape=None, dtype=image_type),
        tf.TensorSpec(shape=None, dtype=flow_type),
    )
Does adding experimental_implements change the behavior of get_concrete_function as well?

marksandler2:
Thanks @Thai Nguyen <thaink@google.com>  - this looks unrelated to the
original issue. The problem seems to be in calling get_concrete_function
with TensorSpec rather than Tensor - i don't any documentation reflecting
this functionality.  It should however be easy to fix though - can you file
another bug?

WindQAQ:
> Hi @WindQAQ,
> The since the fix is in tf-nightly, can I make a PR to merge dense_image_warp and dense_image_warp_annotated?

I'm afraid that our nightly version only pin to stable version of TensorFlow, say 2.3 and 2.4 now, so it's unlikely to merge it. I suggest that we can do that after TensorFlow 2.5 releases.

> @WindQAQ In my test the test_unknown_shapes in tensorflow_addons/image/tests/dense_image_warp_test.py failed after the fix. Is it OK drop the unknown shape support?

I think that test failed even before the fix? There should be another bug in TensorFlow. Unfortunately, we cannot drop it as some `tf.data.Dataset` pipeline will generate unranked tensor.

> The problem seems to be in calling get_concrete_function with TensorSpec rather than Tensor - i don't any documentation reflecting this functionality.

The functionality is documented here.

https://www.tensorflow.org/guide/function#obtaining_concrete_functions

marksandler2:
Yeah - that's an unrelated bug! There were actually several other
alternative codepaths broken with
TensorSpec provided instead of a tensor. Will follow up with tensorflow.

thaink:
Thanks Mark for the quick investigation.
@WindQAQ I'll make a PR saying dense_image_warp_annotated will be removed.

Issue #2201, opened 2020-10-13
EmbeddingBag and Product-Key Memory Layers


Rocketknight1:
**Describe the feature and the current behavior/state.**
FAIR have a cool paper where they introduce [Product-Key Memory Layers](https://arxiv.org/abs/1907.05242) - these are layers that can add a huge number of parameters (100M-1B) to a network with a very minimal compute overhead.

Unfortunately, implementing them efficiently depends on the [EmbeddingBag layer](https://pytorch.org/docs/stable/generated/torch.nn.EmbeddingBag.html) from Pytorch. This layer basically does a gather op followed by a weighted sum across the final dimension of the gather indices.

It is trivial to implement this op as a composition of two or three ops in Tensorflow, but doing so requires you to materialize the output of the gather, which in the case of Product-Key Memory layers is enormous, and usually blows out my GPU RAM. By combining these ops into a single efficient call, EmbeddingBag avoids ever materializing the extremely large pre-sum gather output. There's no efficient way to do the same in Tensorflow without a custom op.

I've already gotten a CUDA and (single-threaded) CPU implementation of EmbeddingBag working locally using the custom-op repo and associated docker image. I've verified correctness by comparing outputs and gradients to those from the manual composition of ops, and speed and memory usage are vastly improved. I could also contribute a TF implementation of the Product-Key Memory layer itself if desired.

**Relevant information**
- Are you willing to contribute it (yes/no): yes
- Are you willing to maintain it going forward? (yes/no): yes
- Is there a relevant academic paper? (if so, where):  https://arxiv.org/abs/1907.05242
- Is there already an implementation in another framework? (if so, where): Yes, EmbeddingBag is already a PyTorch layer 
- Was it part of tf.contrib? (if so, where):

**Which API type would this fall under (layer, metric, optimizer, etc.)**
Layer

**Who will benefit with this feature?**
People who want to squeeze loads of parameters into their model while maintaining fast throughput and aren't worried about overfitting. The paper used it for big autoregressive NLP Transformers, but I suspect you could deploy it in a lot of other places too.

**Any other info.**
I have only implemented the portions of EmbeddingBag necessary for Product-Key Memory layers.

bhack:
Have you tried to see if the subgraoh could be already fused https://www.tensorflow.org/lite/convert/operation_fusion?hl=en#wrap_the_composite_operation_in_a_tffunction?

It could be nice to have fusion in TF and the layer here.

bhack:
[...]

PR #2352, opened 2021-01-18
EmbeddingBag op and layer

Rocketknight1:
# Description

Brief Description of the PR:
This is a PR for the EmbeddingBag op. Please don't merge it yet! Although it works, testing is incomplete and the file structure needs to be cleaned up. I'm opening it now just to get some initial feedback. I'll keep working on several of these issues (particularly 1, 3, 4 and 6 see below), but I'll need some feedback on 2) and 5), plus any other feedback you have for the rest of it!

Fixes # (issue)
#2201 

## Type of change

New layer and associated C++/CUDA op

# Comments

There are a few issues that need to be resolved before I'd feel comfortable with this being merged. In no particular order, they are:

1) The CUDA/C++ code is split with the forward and backward passes in separate files, which is not how other Tensorflow or Addons ops do it. This is just a style thing - I'll merge them soon. 

2) There are really two different entrypoints for users here, the function/op (analogous to tf.gather) and the layer (analogous to tf.keras.layers.Embedding). Like Embedding, the layer instantiates its own embeddings tensor and expects to be passed only indices and weights, whereas the function needs to be passed embeddings as well. Following PyTorch's naming conventions, I called the op embeddingbag and the layer EmbeddingBag, but this is almost certainly not what you want. What is the right way to name these two? Should I make the function/op a stateless Layer rather than just a function?

3) No support for float16/bfloat16 yet.

4) Because context->AllocateTemp continuously segfaulted for me when I was compiling in the custom-op repo, I used AllocateOutput to make some dummy outputs and then just used them as temp arrays. Compiling in tensorflow_addons itself seems much more stable, but I still need to go back and set that properly to AllocateTemp.

5) The CUDA/C++ ops expect a weight tensor. When no weights are passed, the Python wrapper instantiates dummy weights with `tf.ones_like()`. Is this acceptable?

6) More tests! I don't have any gradient tests at all yet, and I should probably add additional tests with weird shapes.


Rocketknight1:
@googlebot I signed it!

bhack:
Can you add the python compositional impl as https://github.com/tensorflow/addons/blob/master/tensorflow_addons/seq2seq/beam_search_decoder.py#L237 ?

bhack:
Here some other examples about custom + python ops alternative impl (now python only on master) https://github.com/tensorflow/addons/issues/1114#issuecomment-596211326

Rocketknight1:
Sure, I'll add that soon while I'm fixing everything else up.

tensorflow_addons/layers/embeddingbag.py
```diff
@@ -0,0 +1,130 @@
...
+    See [PyTorch op](https://pytorch.org/docs/stable/generated/torch.nn.EmbeddingBag.html).
+
+    Equivalent to tf.gather() followed by tf.reduce_sum() across the last dimension, with optional
+    weights. Fusing these into a single op has massive benefits for execution speed and particularly
+    memory usage, as the intermediate output of the gather never needs to be materialized.
+
+    Input Shapes:
+      indices: An int32 or int64 `Tensor` of the indices to gather from
+          `values`. Must be at least 2-dimensional, as the last dimension
+          will be summed out. Maximum value must be less than values.shape[0].
```
WindQAQ, line 74: If I understand correctly, there are only two cases in PyTorch:

1.  `(batch_size, sequence_length)`
2. `(batch_size,)`:  in this case, it has same effects as `(batch_size, 1)`

Any reason or use case that you want to  support higher dimension indices?

tensorflow_addons/layers/embeddingbag.py
```diff
@@ -0,0 +1,130 @@
...
+      values: A float32 `Tensor` from which to gather values. Must be rank 2.
+      weights: A float32 `Tensor` of weights which will be applied to each of
+          the gathered embedding vectors before the sum step.
+
+    Output shape:
+        indices.shape[:-1], values.shape[-1]
+    """
+
+    def __init__(self, num_embeddings, embedding_dim, embeddings_initializer=None,
```
WindQAQ, line 83: in keras semantics:

- num_embeddings -> input_dim
- embedding_dim -> output_dim

https://www.tensorflow.org/api_docs/python/tf/keras/layers/Embedding

tensorflow_addons/layers/embeddingbag.py
```diff
@@ -0,0 +1,130 @@
...
+                 **kwargs):
+        super(EmbeddingBag, self).__init__(**kwargs)
+        if num_embeddings <= 0 or embedding_dim <= 0:
+            raise ValueError('Both `num_embeddings` and `embedding_dim` should be positive, '
+                             'found num_embeddings {} and embedding_dim {}'.format(
+                num_embeddings, embedding_dim))
+        self.num_embeddings = num_embeddings
+        self.embedding_dim = embedding_dim
+        if embeddings_initializer is None:
+            self.embeddings_initializer = initializers.normal(mean=0, stddev=embedding_dim ** (-0.5))
```
WindQAQ, line 94: Better to follow how keras initializes embedding. In input argument, `embeddings_initializer="uniform"`.

https://www.tensorflow.org/api_docs/python/tf/keras/layers/Embedding

tensorflow_addons/layers/embeddingbag.py
```diff
@@ -0,0 +1,130 @@
...
+        self.embeddings = self.add_weight(
+          shape=(self.num_embeddings, self.embedding_dim),
+          initializer=self.embeddings_initializer,
+          name='embeddings',
+          regularizer=self.embeddings_regularizer,
+          constraint=self.embeddings_constraint,
+          experimental_autocast=False)
+        self.built = True
+
+    def call(self, indices, weights):
```
WindQAQ, line 111: in keras semantics:

```python
def call(self, indices, weights):
```

->

```python
def call(self, inputs, sample_weights=None):
```

tensorflow_addons/layers/embeddingbag.py
```diff
@@ -0,0 +1,130 @@
...
+    def get_config(self):
+        config = {
+            'num_embeddings': self.num_embeddings,
+            'embeddings_dim': self.embeddings_dim,
+            'embeddings_initializer':
+                initializers.serialize(self.embeddings_initializer),
+            'embeddings_regularizer':
+                regularizers.serialize(self.embeddings_regularizer),
+            'activity_regularizer':
+                regularizers.serialize(self.activity_regularizer),
```
WindQAQ, line 123: This should be a typo? There is no `activity_regularizer`.

WindQAQ (commented):
Hi @Rocketknight1, thanks for the contribution. This is a huge PR. I have a suggestion: I can help you do C++ CPU impl (with threading and vectorization) and python interface. This can reduce back and forth communication during review. If you can accept it, I'll work on your branch and comment out GPU build (I'll modify style and input/output), or I can file another PR for CPU ops. Thank you!

tanguycdls:
Hi @Rocketknight1 thanks for your PR. We recently switched from Torch to TF and Embedding Bag was missing !

In our use case we often work with list of non constant len called Ragged tensors in TF which uses a similar data format as Sparse CSR matrix: 

https://www.tensorflow.org/guide/ragged_tensor

We have ragged tensors such as : 
````
offsets = [0, 0, 0, 1, 5, 7]
indices = [12, 13, 14, 15, 16, 78, 16]
tf.RaggedTensor.from_row_splits(indices, offsets)
````
> <tf.RaggedTensor [[], [], [12], [13, 14, 15, 16], [78, 16]]>


For now we currently replace embedding bag by converting our ragged to a Sparse Tensor (indices being the rows and y the nbr of item in each row) and values being the indices we want to gather. We also use a second sparse Tensor which will have the weights instead of the indices.
We can then use tf.nn.safe_embedding_lookup_sparse and get better result than a simple gather then reduce. I'm not very clear on  the ram usage of that one but it does the embedding lookup on unique indices and then apply a gather on it. ([see](https://github.com/tensorflow/tensorflow/blob/582c8d236cb079023657287c318ff26adb239002/tensorflow/python/ops/embedding_ops.py#L511))

Another workaround we found is to create a sparse tensor Indicator: the x coordinate will be the rows of your batch, the y the indices and values being the weight: you can then consider your embedding + sum as a sparse_dense_matmul between the embeddings matrix and your sparse indicator. The sparse_dense_matmul itself is very fast the issue is more on creating the sparse indicator. I'm not sure how that option behaves on memory since the internals are handled by TF.

I did a few tests here:

https://colab.research.google.com/gist/tanguycdls/9c696097642844fc1e548c0cade48e11/sparseembeddings.ipynb 

the performance depends a lot on the sparsity of indices and nbr of items to compute in the ragged case.

I'll try to compile your branch to compare the performance between sparse matmul and your EmbeddingBag ! We did a few months ago a benchmark Pytorch vs TF and embeddingbag was slightly better than matmul in some cases. 

Would be happy to help on benchmarks if you need some in that PR !

**Comment by bhack** [comment]
@tanguycdls Thanks you this seems interesting to explore.

**Comment by Rocketknight1** [comment]
Hi @tanguycdls, your workaround with sparse multiplications is really interesting! I'm also curious to know how it compares in terms of memory usage to the CUDA EmbeddingBag. Please note that my CPU implementations are not very optimized right now (as @WindQAQ pointed out), but the CUDA should be at least reasonably performant.

File: tensorflow_addons/layers/embeddingbag.py (hunk 1 above)
Comment (Rocketknight1) [reply tensorflow_addons/layers/embeddingbag.py:74]: You're correct that I'm not precisely following the PyTorch semantics. The reason was to more closely follow the semantics of tf.gather, where the indices tensor can be N-dimensional. My op treats the final dimension as the 'bag' dimension, which is analogous to PyTorch when N = 2, and any other dimensions are all treated as 'batch' dimensions. The C++/CUDA code is completely blind to these details - it only cares about the total size of the indices array and the length of the final dimension, and treats all other dimensions as one single 'batch' dimension.

If people prefer, I can try to mimic the PyTorch semantics more closely, but it would probably take some hacking to support a separate offset tensor.

File: tensorflow_addons/layers/embeddingbag.py (hunk 4 above)
Comment (Rocketknight1) [reply tensorflow_addons/layers/embeddingbag.py:111]: Please note the 'weights' here are not sample weights, but the bag weights that the op will use when computing its output. My function syntax is possibly wrong here - is it acceptable for a Keras layer to take two arguments like that, or should that be a single tuple argument?

File: tensorflow_addons/layers/embeddingbag.py (hunk 5 above)
Comment (Rocketknight1) [reply tensorflow_addons/layers/embeddingbag.py:123]: This is a mistake, caused by me copying this from another op in the repo. I'll remove it!

File: tensorflow_addons/layers/embeddingbag.py (hunk 2 above)
Comment (Rocketknight1) [reply tensorflow_addons/layers/embeddingbag.py:83]: Will fix!

File: tensorflow_addons/layers/embeddingbag.py (hunk 3 above)
Comment (Rocketknight1) [reply tensorflow_addons/layers/embeddingbag.py:94]: Will fix!

**Comment by Rocketknight1** [comment]
Also yes @WindQAQ, if you want to improve the CPU implementations feel free to make those changes. I'm aware that they are currently 'reference' implementations that are not high-performance. Thank you!

File: tensorflow_addons/layers/embeddingbag.py (hunk 3 above)
Comment (Rocketknight1) [reply tensorflow_addons/layers/embeddingbag.py]: This is now fixed.

File: tensorflow_addons/layers/embeddingbag.py (hunk 2 above)
Comment (Rocketknight1) [reply tensorflow_addons/layers/embeddingbag.py]: Fixed.

File: tensorflow_addons/layers/embeddingbag.py (hunk 5 above)
Comment (Rocketknight1) [reply tensorflow_addons/layers/embeddingbag.py]: Fixed.

**Comment by bhack** [comment]
@tanguycdls It could be interesting to benchmark also against TF-nightly as the Compiler/Stack some optimizations on sparse are really on the development edge. See https://llvm.discourse.group/t/mlir-support-for-sparse-tensors/2020/12

File: tensorflow_addons/layers/embeddingbag.py (hunk 1 above)
Comment (WindQAQ) [reply tensorflow_addons/layers/embeddingbag.py:74]: I know that your impl is independent to the first (N-1)-dim, but note that if you want to support higher dimension, there will be less compiler optimization kickin for Eigen compilation (If you don't specialize each rank's functor).

For 1-D indices, my thought is that we can implement later :-)

Comment (Rocketknight1) [reply tensorflow_addons/layers/embeddingbag.py:74]: I'm happy to constrain the implementation to 2D if you think there's a significant benefit - users can just flatten the first N-1 dimensions and unflatten them afterwards if they have multidimensional input.

**Comment by WindQAQ** [comment]
@googlebot I consent.

**Comment by WindQAQ** [comment]
Done

- Forward CPU part with threading and vectorization
- Pass combiner attr (only support sum now)
- Runtime shape check
- values dtype specialization
- python interface

Note that I comment out GPU build so far. Please follow template specialization and naming convention (snake case for variable, camel case for template).

/cc @Rocketknight1

File: tensorflow_addons/layers/embeddingbag.py (hunk 1 above)
Comment (WindQAQ) [reply tensorflow_addons/layers/embeddingbag.py]: Sure, I will modify interface later.

**Comment by tanguycdls** [comment]
Hello, Thanks for the links @bhack  I'll check the performance against nightlies for sparse !

About the cuda implem I dont have any gpu to work on I will try to dot it on collab. 

I managed to compile your work on my computer to run some tests on CPU on a dense indices matrix to compare the perf:

```
def test_sparse_dense_matmul(indices, values_params, weights, safe=True):
    from scipy.sparse import coo_matrix
    # Creates indicator matrix:
    x = np.tile(np.arange(indices.shape[0]), indices.shape[1]).astype(int)
    y = indices.ravel().astype(int)
    values = weights.ravel()
    sparse = tf.SparseTensor(np.array([x, y]).T, values , (indices.shape[0], input_dim))
    if safe:
        sparse_coo = coo_matrix((values, (x, y)), (indices.shape[0], input_dim))
        sparse_coo.sum_duplicates() # very slow
        x = sparse_coo.row
        y, values = sparse_coo.col, sparse_coo.data
        sparse =  tf.sparse.reorder(tf.SparseTensor(np.array([x, y]).T, values , (indices.shape[0], input_dim))) # reorder is slow as well not sure if it's needed for dense matmul ?
    return tf.sparse.sparse_dense_matmul(sparse, values_params) 
`````
on a large (100_000, 100) indices and 10_000 input dim: the sparse dense matmul without any safety check (meaning with possible column duplicates that I believe wont be considered by Tensorflow when doing the matmul) your cpu implem is almost 3.5 times faster than sparse matmul (0.48 seconds (sparse dense) vs 0.14 seconds (custom embedding bag)).

If I check the duplicates it gets worse because of scipy and the reordering needed (6 seconds ). The manual bag is only 1.18 seconds. 

I'm not very familiar with the c++ code base but i'll try if I can modify at least your cpu implem to support flattened indices and an offset array instead of the 2D indices. That would make that implem compatible with sparse application (non constant sequence) which would improve memory efficiency against at least safe_embedding_lookup_sparse. While keeping dense array as fast I believe.

**Comment by bhack** [comment]
@tanguycdls Thanks, can you share a minimal colab to reproduce this numbers on your sparse impl?

**Comment by Rocketknight1** [comment]
Hi @WindQAQ, thanks for all the work you've done on this! I probably won't be able to work on it much for a week or so, so I'm going to back off and let you finish with the CPU implementation before I push any more changes.

Also, does anyone know what the correct way is to assert that the input indices and weights have the same shape? Should I do that in Python, or is it best to do it in the C++ InferenceContext?

**Comment by WindQAQ** [comment]
> Hi @WindQAQ, thanks for all the work you've done on this! I probably won't be able to work on it much for a week or so, so I'm going to back off and let > Also, does anyone know what the correct way is to assert that the input indices and weights have the same shape? Should I do that in Python, or is it best to do it in the C++ InferenceContext?

I'll do that.

**Comment by WindQAQ** [comment]
> I'm not very familiar with the c++ code base but i'll try if I can modify at least your cpu implem to support flattened indices and an offset array instead of the 2D indices. That would make that implem compatible with sparse application (non constant sequence) which would improve memory efficiency against at least safe_embedding_lookup_sparse. While keeping dense array as fast I believe.

After scrutinizing ragged tensor kernels, I think it would be better to get another op, say `RaggedEmbeddingBag`, which has another input `splits`. In this way, we can detect the ragged input and call

```python
_embedding_bag_so.ops.addons_ragged_embedding_bag(ragged_tensor.values, ragged_tensor.row_splits, ...)
```

The underlying implementation can be very similar. If we can support ragged tensor, the gradient of `EmbeddingBag` can be composed of `RaggedEmbeddingBag` and other python ops I guess.

bhack:
@WindQAQ as we have already accumulated more then 1.000 lines of code in this PR do we have any example to benchmark the memory overhead between a compositional Embeddingbag and this Custom implementation?

@tanguycdls shared a gist in a comment but I don't think It is enought to verify the memory overhead.

Rocketknight1:
Hey all, just an update - this weekend I'm planning to continue making patches to the CUDA code, and I'll also try to provide benchmark data against a composition of ops (i.e. gather + einsum or gather + multiply + reduce_sum), and also against a dense_sparse_matmul implementation.

bhack:
@Rocketknight1 There is an internal Google ticket for this missing `gather + reduce` fusion https://github.com/google/jax/issues/3206#issuecomment-766408613

Rocketknight1:
I found some time this evening for the comparison and made a gist here: https://gist.github.com/Rocketknight1/8b3ec932995be18220e869c47a8d40e7

The conclusion is that sparse_dense_matmul works **much** better than a gather -> reduce, but performance still seems to be worse than the custom op, although not orders of magnitude worse like it was before. I tried compiling with XLA, but sparse-dense matmul is not supported by XLA (and neither is embeddingbag, lol).

Script output below (AMD Threadripper 1950X, Geforce 3090, Ubuntu 18.04 LTS):
```
Running with composite op...
50 runs took 1.9536158019909635 seconds
Running with sparse-dense matmul...
50 runs took 2.729604701977223 seconds
```

Also, the sparse-dense matmul approach errored out at much smaller sizes than the custom-op approach, so it seems like the custom op is more memory-efficient too.

bhack:
@Rocketknight1Is your gist running whit the last commit in this PR?

Rocketknight1:
This is compiled with the version before @WindQAQ's edits - I haven't tried compiling since then because I think he was going to comment out some of the GPU code while working on the CPU ops.

Rocketknight1:
Hi @WindQAQ, thanks for the incredibly in-depth work on the CPU ops! I just looked over it there - it's vastly better than my very manual loop stuff. Are you done - am I safe to start making changes to the GPU code again?

WindQAQ:
> Hi @WindQAQ, thanks for the incredibly in-depth work on the CPU ops! I just looked over it there - it's vastly better than my very manual loop stuff. Are you done - am I safe to start making changes to the GPU code again?

I'm done. But I'm thinking to support RaggedEmbeddingBag and it's grad only. Probably like this, so that we can minimize codes. I test on my PC and there is almost 0 overhead to deal with dense tensor on CPU.

```python
def _embedding_bag(indices, params, weights):
    if isinstance(indices, tf.RaggedTensor):
        return _ragged_embedding_bag(indices.values, indices.row_splits, params, weights)

    indices_shape = tf.shape(indices)
    bags = indices_shape[0]
    sequence_length = indices_shape[1]
    indices = tf.reshape(indices, (-1,))
    weights = tf.reshape(weights, (-1,))
    output_dim = params.shape[1]
    splits = tf.range(0, bag * sequence_length + 1, sequence_length)
    output = _ragged_embedding_bag(indices, splits, params, weights)
    return output
```

For backward part, I am thinking to return sparse gradients, that is, we only deal with non-zero gradients to minimize memory usage (and can be faster for optimizer when it comes to larger vocab size).

```python
def _ragged_embedding_bag_grad(indices, params, weights):
    # params_grads_indices is non-zero indices
    params_grads_indices, params_grads_values, weights_grads = _ragged_embedding_bag_grad(indices, splits, params, weights, grads)
    return [None, None, tf.IndexedSlices(values=params_grads_values, indices=params_grads_indices, dense_shape=tf.shape(params)), weights_grads]
```

Any though on this?

Rocketknight1:
That looks interesting, but the ops as written assume that the number of indices per row (i.e. bag_dim) is constant, right? My CUDA op certainly does - if we add support for RaggedTensor, that assumption doesn't hold anymore, so I'll have to rewrite the CUDA, or add a separate op for the RaggedTensor case, right? Just making sure I understand.

Rocketknight1:
We could, however, use the `sparse_dense_matmul` fallback for that - so then for systems where the op can be compiled, and when the indices are dense, they could use the custom op, but when they need a fallback, or when they have a RaggedTensor of indices we use `sparse_dense_matmul`. That might work better?

WindQAQ:
Yes, both you and I should rewrite in order to support ragged tensor. That is the only way that I can come up with to use minimal codebase to support full spectrum. Does sparse_dense_matmul has lower memory usage than naive method?

WindQAQ:
BTW, I think your CUDA code can go further. Each thread is better to handle multiple elelements to improve instruction-level parallelism. I haven't benchmarked it, but usually it can be faster with a tradeoff of occupancy and ILP.

Rocketknight1:
@WindQAQ that's really interesting! I was mostly imitating the Pytorch code, where each thread handled one (bag x feature), which meant there was little instruction-level parallelism that I could see. Would you suggest each thread loading a block of 2-4 values and operating on them as vectors? Would we need to ensure all dimensions were multiples of 2 or 4 to avoid awkward performance-killing conditional execution?

WindQAQ:
Thanks for investigation. You can always do it latter in another PR. Let's finalize backward part first.

I have already made your cuda forward code work and test it. For backward, note that you have to decorate it with `with_device`, and run pytest. See our test file for example. Also, we do not trigger GPU testing on CI by default.

Rocketknight1:
@WindQAQ Understood! I'll try to finish tests for the backward code. I think this is a good plan:

1) Finish tests
2) Replace allocate_output with allocate_temp
3) Write pure TF sparse_dense_matmul fallback
4) Add RaggedTensor support with sparse_dense_matmul

That will finish this PR without needing more CUDA/C++ code. Then I'll test a CUDA implementation for RaggedTensor and see if the performance is better than sparse_dense_matmul. If it is, I'll submit an extra PR for that implementation.

Rocketknight1:
If you want to contribute any of that, let me know - I can add you as a contributor to my fork after this PR. But I'm happy to try it myself if you're busy!

WindQAQ:
> 2. Replace allocate_output with allocate_temp

If you want to allocate temp tensor, feel free to modify functor's input signature and pass ctx into it.

> 3. Write pure TF sparse_dense_matmul fallback

I think just fallback to gather reduce? sparse_dense_matmul does not have GPU kernels, so it might be slow if you test on GPU I guess.

bhack:
> I think just fallback to gather reduce? sparse_dense_matmul does not have GPU kernels, so it might be slow if you test on GPU I guess.

On a Colab run of the shared Gist with `tf.debugging.set_log_device_placement(True)` I see `Executing op __inference_sparse_dense_matmul_run_74 in device /job:localhost/replica:0/task:0/device:GPU:0`

WindQAQ:
Thanks for testing. I overlooked one file 😭

bhack:
What is the current CUDA benchmark cause I see on the sparse on Colab GPU 50 runs took 8.199049400999996 seconds

Rocketknight1:
Hi @WindQAQ or anyone else, do you know the right way to handle the case when the CPU functor doesn't need temp allocations but the GPU functor does? Is there a way to only do ctx->allocate_temp() for GPU ops?

WindQAQ:
You can pass ctx into functor as well.

Rocketknight1:
Ah, sorry, I misread what you wrote earlier. Thanks!

Rocketknight1:
@WindQAQ Quickly checking with you before I make a change - I appreciate you adding the 'combiner' option to match the 'mode' option in PyTorch's embeddingbag, but I want to change the implementation. The reason is that the current implementation allows some illegal combinations, such as providing a weights tensor when the combiner mode is 'mean', which results in incorrect outputs.

A simpler version is to remove the combiner from the C++/CUDA code entirely, which makes them smaller and more maintainable. Then, in the Python wrapper, we can check that no weights are supplied if the mode is 'mean', and then make a full weight tensor with value 1/bag_dim and pass it to the C++/CUDA op, or do a fallback to dense_sparse_matmul or nn.embedding_lookup_sparse.

Alternatively, I can write a second kernel that handles the case when there are no weights, which should support both the 'sum' (without weights) or 'mean' modes. What do you think?

WindQAQ:
I think we should fix it rather than remove it. Btw, Im not matching pytorch but tf.nn.embedding_lookup_sparse.

Rocketknight1:
We should raise an error if the user passes weights in mode 'mean', because this doesn't make sense. So that means any weights we pass to the op in that mode will be ones we generate in the Python wrapper - it should be easy enough to generate weights equal to 1/bag_dim, right?

WindQAQ:
Why it does not make sense?

WindQAQ:
I mean something like this

```python
def _embedding_bag():
    # Call C++ kernel, which only implements sum combiner now.
    output = embedding_bag(combiner="SUM")
    if has_weights:
        output /= tf.reduce_sum(weights, axis=1)
```

Rocketknight1:
I think that would work, but I think we should raise an error if the user passes weights and the combiner mode is set to 'mean', because otherwise we just silently ignore a parameter that they pass, right?

WindQAQ:
> I think that would work, but I think we should raise an error if the user passes weights and the combiner mode is set to 'mean', because otherwise we just silently ignore a parameter that they pass, right?

I don't get it. Can you give me an example? Thanks.

Rocketknight1:
I was thinking something like this, to match the PyTorch behaviour:

```
def _embedding_bag():
    has_weights = weights is not None
    if has_weights and combiner.upper() == "MEAN":
        raise ValueError("Weights are only supported for mode 'SUM'.")
    elif combiner.upper() == "MEAN"
        # Set weight tensor appropriately
    # Call C++ kernel, which only implements and accepts sum combiner now.
    output = embedding_bag(combiner="SUM")
```

But I understand wanting to match `nn.embedding_lookup_sparse`, especially if we ever want to get this into Tensorflow Core. I think your approach is probably better - but I think if we do that, we should default to 'SUM', because if we don't then a user who expects the PyTorch behaviour will supply weights and then get a really strange output (because combiner == 'MEAN'), with no message or warning to explain what went wrong. Do you think that's okay?

WindQAQ:
I have no idea why weights can not be used with mean combiner... Shouldn't weighted mean be common enough?

Rocketknight1:
I guess in Pytorch, they just assume that you can normalize the weights yourself if you want to do that. But you're right that your approach matches `nn.embedding_lookup_sparse`, so I guess we should do that. Is it okay if we make 'SUM' the default combiner mode, though, so as not to surprise people who expect the PyTorch behaviour?

WindQAQ:
That's ok if we document it.

Rocketknight1:
Sure! Also, are you sure we need to keep the combiner argument in the C++/CUDA code? It feels easier to just fix things in the Python wrapper e.g. by normalizing weights.

WindQAQ:
It's fine to remove it if there won't be any extension in the future. Adding an input or attribute to an existing kernel is not an easy work when it comes to backward compatibility.

Rocketknight1:
I think we're safe - modes like 'mean' and 'sqrt' can always be handled in the wrapper. The PyTorch EmbeddingBag also has a 'max' mode, but if we ever implement it, it will need a completely different calculation and gradient computation, so that should definitely be a different kernel.

WindQAQ:
The just go ahead

Rocketknight1:
Cool! And sorry for all the questions - you just did a lot of work on the op that I really appreciate, so I don't want to start breaking things without making sure it's okay with you :p

WindQAQ:
Btw, backward compatibility means that we have to maintain v2 Kernel in order not to break old graphdef. We definitely need different implementation for different combiner. What you want to say might be having a brand new op kernel for new max combiner. That is not encouraged to do so in terms of compiled binary size and different c++ kernel codepath. But anyway, I cannot find any usage of max combiner in the community. Let's come up with other ideas if anyone requests.

Rocketknight1:
@WindQAQ that all sounds good!

Also, do you (or anyone) know how to correctly use AllocateTemp in GPU ops? The code I'm using right now is below, but it segfaults. I've tried with and without the gpu_allocator, and I still get segfaults. I can't see any documentation for how to do this properly, and copying code from existing ops in Tensorflow doesn't seem to be working.
```
    tensorflow::AllocatorAttributes gpu_allocator;
    gpu_allocator.set_on_host(false);
    gpu_allocator.set_gpu_compatible(true);

    Tensor sortedIndicesTensor;
    Tensor sortedIndicesCounterTensor;
    OP_REQUIRES_OK(context, context->allocate_temp(
                          DataTypeToEnum<Tindices>::value, TensorShape({indices.size()}),
                          &sortedIndicesTensor, gpu_allocator));
    OP_REQUIRES_OK(context, context->allocate_temp(
                          DataTypeToEnum<Tindices>::value, TensorShape({indices.size()}),
                          &sortedIndicesCounterTensor, gpu_allocator));
```

WindQAQ:
It should be pretty fine. Can you push the change so that I can see the whole program? Thanks.

Rocketknight1:
Pushed!

WindQAQ:
Unfortunately, I cannot get it worked either.  There seems to be some problem when using thrust and I don't know why. What you want to do is probably like

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/dynamic_partition_op_gpu.cu.cc#L334-L366

Can you try to use cub instead?

Rocketknight1:
Is the problem with thrust? I think the error occurs as soon as `allocate_temp` is called, even if I remove the later calls to thrust, right? Or did you get different results?

WindQAQ:
Sorry that I overlook the log. Let me dig into it again

WindQAQ:
I play around with CI. The segfault seems to occur before allocate_temp.

without calling functor: no segfault ecccf4b
Call functor with nothing inside: segfault 6f03884

Have you used gdb to local segfault? Sorry that I do not setup gpu on the my machine.

Rocketknight1:
Hey! I'm sorry for the delay. The company I was working for got acquired and I accepted a new job, so everything was a bit chaotic for a week or two.

I checked and you're right, though - the segfault is occurring at the point when the GPU functor is called, not when the temp allocation is performed. I'm trying to break it down to figure out what exactly is causing the issue, but C++ debugging isn't my strongest point so it's a little slow going.

Rocketknight1:
Is there a recommended way to get something like a stack trace in tf-addons so I can see exactly where the error is coming from? All I can work out is that the call to the functor is the issue, but why exactly I have no idea.

WindQAQ:
I think cuda-gdb YOUR_PYTHON_PROGRAM.py can do the job.

Rocketknight1:
It works! Sorry again for the delay - with the job move and everything, I didn't really have time to familiarize myself with C++ debugging. The error in the end turned out to be very simple once I was able to get the code into gdb - I was registering a functor templated with CPUDevice as a GPU kernel, which caused the segfault the moment the functor was called. When I templated it with GPUDevice instead, everything worked, including allocate_temp.

Rocketknight1:
At this point the code looks like it's functional and reasonably performant - should I focus on just adding some tests and getting the PR approved, or is it worth adding in float16/bfloat16 support at this point?

bhack:
c++ linting is failing. Can you check?

bhack:
Can you check also the naming conventions at https://www.tensorflow.org/guide/create_op?

Rocketknight1:
It looks like some tests are still failing. What's the right way to try the tests locally so I can check I pass them here before I push another commit?

bhack:
> It looks like some tests are still failing. What's the right way to try the tests locally so I can check I pass them here before I push another commit?

https://github.com/tensorflow/addons/blob/master/CONTRIBUTING.md#code-testing

Rocketknight1:
It's all working now! The build for 18.04 + Py3.6 failed, but in the logs this seems to be a dependency issue unrelated to my commits.

The final-final question is naming and entry points for users. The way it's set up right now is that there's a Keras layer that's analogous to `tf.keras.layers.Embedding`, and there's also the raw op `_embedding_bag`. I wasn't sure how to name both of these, since users will probably want to use both of them. I couldn't see anything in the naming conventions to answer the question, but I should probably remove the leading underscore for the raw op at least. Can you suggest what the names should be?

Also, thanks for all your help so far, we're almost done!

fsx950223 (dismissed):
LGTM, but I can't find ubuntu-gpu in ci/cd

fsx950223:
Could you add a ```tf_function``` test case?

aartbik:
> /cc @aartbik if he is interested on the sparse code-path.

can you please provide some pointers on where to start looking? (this is a huge PR)

bhack:
> > /cc @aartbik if he is interested on the sparse code-path.
> 
> can you please provide some pointers on where to start looking? (this is a huge PR)

I meant @tanguycdls exploring sparse impl workaround at https://github.com/tensorflow/addons/pull/2352#issuecomment-763042721

Rocketknight1:
> 
> 
> Could you add a `tf_function` test case?

Sure! I'll try to get to that this weekend.

Rocketknight1:
Hey! Little late but I the tests are in - can you give me workflow approval to make sure it's okay? I'll try testing here too.

fsx950223 (dismissed):

Rocketknight1:
Hi @bhack can you approve running workflows? It's awkward for me to run all the tests locally and it's really helpful if I can quickly check via CI!

bhack (dismissed):

Rocketknight1:
I have run into a problem - the code added by @WindQAQ to convert the parameter gradients to an IndexedSlices object does not seem to work in graph mode. I don't think this is a problem with the code - it looks good to me! However, the IndexedSlices tensor is returned as a dense tensor when I wrap _embedding_bag with tf.function().

I googled around and couldn't find a cause for this, but there's a suggestive issue here - it's possible this is because of a different TF bug that got monkey-patched: https://github.com/tensorflow/tensorflow/issues/36236

Either way, I've edited the tests and hopefully it should all work now.

tensorflow_addons/layers/embedding_bag.py
```diff
@@ -0,0 +1,165 @@
...
+    combiner = op.get_attr("combiner")
+    value_grads, weight_grads = _embedding_bag_so.ops.addons_embedding_bag_grad(
+        indices, params, weights, grads, combiner=combiner
+    )
+    # Because value grads are sparse, returning IndexedSlices can be faster for optimizer.
+    unique_indices = tf.unique(tf.reshape(indices, (-1,)))[0]
+    sorted_unique_indices = tf.sort(unique_indices)
+    return [
+        None,
+        tf.IndexedSlices(
```
fsx950223, line 76: Maybe just return value_grads is a better choice, according to https://github.com/foxik/tensorflow/blob/b0e08f68d84476b941e8a821d653ab08129c92ce/tensorflow/python/eager/function.py#L1241.

Rocketknight1: Good idea, I made the change!

Rocketknight1:
@bhack I believe this PR should have fixed it! Can you approve running the tests?

bhack:
You have a lint error.

Rocketknight1:
Fixed! I think.

Rocketknight1:
We did it!

fsx950223 (approved):

bhack (changes requested):
Can you add yourself as Codeowner in https://github.com/tensorflow/addons/blob/master/.github/CODEOWNERS ?

Rocketknight1:
@bhack Ah, sorry, just saw that. I've added myself!

Rocketknight1:
@bhack Can we run tests? I think this is ready to merge now!

Rocketknight1:
@bhack I made the requested change (adding myself to CODEOWNERS) so I think this is ready to merge!

bhack:
@fsx950223 Anything else?

fsx950223 (approved):

bhack (approved):

PR #2355, opened 2021-01-20
Default compile with AVX support

### LINKED ISSUE #2336: Compile with AVX only
Seems that TF2.4.0 is accidentally compiled with AVX2 (or more, not sure what's the CPU spec on TF release CI), and we follow it in https://github.com/tensorflow/addons/pull/2299. We should fallback to subset of ISAs, probably AVX, once there is a new release.

https://github.com/tensorflow/tensorflow/pull/46229

/cc @seanpmorgan

seanpmorgan:
# Description
Default compile with only AVX (not AVX2) support. We can release a 0.12.1 patch in the coming weeks

Fixes #2336


WindQAQ (approved):
Thanks!

Issue #2350, opened 2021-01-18
[Metric] Adding AdjustedRsquared Metric


jonpsy:
**Describe the feature and the current behavior/state.**
   The issue #265 dealt with adding innovative metric functions to the repository. Some of which were not completed, one of that, is AdjustedRsquared metric. The author has already finished Rsquared metric, so I decided to add the adjusted R squared metric as an optional feature of that class. Much like how NMS function of tensorflow supports soft NMS as an additional feature https://www.tensorflow.org/api_docs/python/tf/image/non_max_suppression_with_scores.

Here's a slight demonstration of the PR I want to submit.
https://colab.research.google.com/drive/1Kn1TGFEKrGry8uTI5OLhHvbKRx6fXwum?usp=sharing

**Relevant information**
- Are you willing to contribute it (yes/no): Yes
- Are you willing to maintain it going forward? (yes/no): yes
- Is there a relevant academic paper? (if so, where): https://onlinelibrary.wiley.com/doi/abs/10.1002/9781118445112.stat06627
- Is there already an implementation in another framework? (if so, where): None that I know of
- Was it part of tf.contrib? (if so, where):

**Which API type would this fall under (layer, metric, optimizer, etc.)** Metric

**Who will benefit with this feature?**

Expert statisticians who utilize these features who are dependent on tensorflow for their research. Beginners who are exploring data science could also benefit from this.

**Any other info.**

Harsh188:
Hey, @jonpsy unfortunately it looks like the academic paper you have referenced does not meet our minimum requirements of 50 citations.  We are still trying to improve our feature/contribution guidelines but for the time being this is the metric we are using to filter out feature requests.

jonpsy:
I see, so shall we  put this PR on hold till it passes 50 citations +?  I can work on other metrics meanwhile.

bhack:
That one is only an article and not the original paper.
You need to find some Deep learning papers that use this metric.

jonpsy:
@bhack  I haven't been able to find a research paper which explicitly introduces this idea, though I have had some luck finding papers which use them. For ex: Harel, Ofer. "The estimation of R 2 and adjusted R 2 in incomplete data sets using multiple imputation." Journal of Applied Statistics 36.10 (2009): 1109-1118. 

Citation count: 214.

Would this suffice?

bhack:
@jonpsy Do you have something applied in the Deep learning context?

jonpsy:
[...]

PR #2364, opened 2021-01-24
- Add adjr2

jonpsy:
# Description

Brief Description of the PR:

Fixes # (issue)
#2350 

## Type of change

- [ ] Bug fix
- [ ] New Tutorial
- [ ] Additional Testing
- [x] New Metric and the changes conform to the [metric contribution guidelines](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/metrics/README.md#contribution-guidelines)

# Checklist:

- [ ] I've properly [formatted my code according to the guidelines](https://github.com/tensorflow/addons/blob/master/CONTRIBUTING.md#coding-style)
- [x] I have made corresponding changes to the documentation
- [ ] 
@bhack


jonpsy:
All the tests seem to be passing, would appreciate a review :)

tensorflow_addons/metrics/r_square.py
```diff
@@ -125,18 +132,25 @@ def update_state(self, y_true, y_pred, sample_weight=None) -> None:
             tf.reduce_sum((y_true - y_pred) ** 2 * sample_weight, axis=0)
         )
         self.count.assign_add(tf.reduce_sum(sample_weight, axis=0))
+        self.num_examples.assign_add(tf.size(y_true))
 
     def result(self) -> tf.Tensor:
         mean = self.sum / self.count
         total = self.squared_sum - self.sum * mean
         raw_scores = 1 - (self.res / total)
+        n = tf.cast(self.num_examples, dtype=tf.float32)
```
bhack, line 141: It is better that you move the code in an internal function that converts R2 in an adjusted R2 so that it is computed only if it is required the adjusted version.

bhack (changes requested):

bhack (dismissed):

bhack:
Can you resolve conflicts?

bhack:
Please fix tests

jonpsy:
Done :)

bhack (approved):

AakashKumarNain:
Overall LGTM but I have one doubt which I have pointed out in the code itself. Can you please clarify that?

tensorflow_addons/metrics/r_square.py
```diff
@@ -83,6 +96,8 @@ def __init__(
         dtype: AcceptableDTypes = None,
         y_shape: Tuple[int, ...] = (),
         multioutput: str = "uniform_average",
+        penalize: bool = False,
+        num_preds: tf.int32 = 0,
```
AakashKumarNain, line 100: Why is it set to `0`? By definition, it should be equal to number of independent variables

bhack: Does num preds arg need to be `tf.int32`?

bhack: Does num preds arg need to be `tf.int32`?

AakashKumarNain: By definition, yes but it would get casted anyways. My point is that this argument represents the number of independent variables used for the model. I don't know why it is set to zero. Check the `p` argument [here](https://en.wikipedia.org/wiki/Coefficient_of_determination#Adjusted_R2)

bhack: Can you take a look at pytorch lighting https://pytorch-lightning.readthedocs.io/en/latest/_modules/pytorch_lightning/metrics/functional/r2score.html ?

bhack: I also think that the extra checks on "num_preds" as in pytorch lighting are useful also here.

bhack (changes requested):
I also think that the extra checks on "num_preds" as in pytorch lighting are useful also here.

tensorflow_addons/metrics/r_square.py
jonpsy: The num_preds should be provided by the user himself,  there's no way of knowing otherwise. I've set it to zero as default because it R2 doesn't require it, only AdjR2. The pytorch_lightning also keeps it to 0 as default.

jonpsy:
Thanks, I forgot to address some corner cases the PyTorch codes were of great help!

jonpsy:
@bhack, How about I remove the boolean ```penalize```, and keep only ```num_preds``` defaulting to zero, but if it has some value our result will calculate adjr2 in the result function itself?

bhack:
If you want to do this refactor is ok

jonpsy:
Great, I've also added checks against these corner cases.

tensorflow_addons/metrics/r_square.py
AakashKumarNain: I agree that it should be provided by the user, that's why I am a bit hesitant for a default value for this parameter. Though that is more of a API choice.

bhack: Now 0 has a specific semantic

bhack:
Tests are failing

tensorflow_addons/metrics/r_square.py
AakashKumarNain: Worst case scenario: If the user isn't aware of the default argument, then the adjusted r2 score would be much lower, which the end user wouldn't be expecting

jonpsy: Setting ```num_preds```  to a value > 0  is the only way to use AdjR2. So, it acts as both a flag and a parameter.

AakashKumarNain: I agree. I propose this: If the user didn't provide a value for `num_preds`, then we should raise a warning that the default value of `0` is used. This way the end user would know why the score is much lower than expected. 

This is just my suggestion. Please let me know if you have something better in mind

jonpsy: I see your point. The API mainly calculates R2 but allows AdjR2 as an additional(optional) feature. I'm using num_preds as a flag to calculate AdjR2. Incase its default, it falls back to standard R2 :). 

Maybe my naming of ```num_preds``` is a bit off, shall I rename it to ```adjusted``` so that it's more clear that its both a flag and a parameter. Pytorch lightning does the same

AakashKumarNain: Another suggestion: Include two parameters, a boolean `adjusted` to indicate whether to calculate adjusted r2 or not. And another `num_features` to indicate the number of independent variables with a default value of `0`. 

```
a warning shouldn't be needed because that's the expected behaviour.
```

IMO, it is needed. Until unless we tell the user that they didn't set the correct value of `num_features`, the calculation will fall back to standard r2 score (the expected behavior)

bhack: Do we have other use case in TF Addons API where optional parameters are activated only with a dependecy on a another boolean parameter flag?

AakashKumarNain: IIRC, `cohen's kappa` has certain bool parameters for specific use cases

bhack: It is a different from Kappa cause It Is not only a [flag argument case] (https://ardalis.com/are-boolean-flags-on-methods-a-code-smell/) but It will has another argument that depend on the flag one.

AakashKumarNain: Okay but is there any problem with the dependency, if at all we include it? I am looking for the `cons` basically

bhack: It is about a potential dead param that you can change but It is inactivated when the flag is false.

But we could search if this is an anti-pattern or not.

bhack: E.g. often just the flag parameter is evaluated as code smell

jonpsy: I've added warnings for potential incorrect values of num_features, such that it falls back to original R2 score (and there are test to check it). 

Making num_preds both as a flag and param hits two birds with one stone doesn't it? 

Instead of setting "penalize = true" And then check if the user put "num_preds" correctly( num_preds > 0),  we can just have "adjusted" which accounts for both the validity & calculation. As bhack said, a potential dead parameter.

AakashKumarNain: Hmm got your point. Thanks for the clarification @bhack and @jonpsy I am okay with it. Just one last thing. Can we rename `num_preds` to `num_features`?

bhack: Yes it is `number of independent regressors` so probably we could use something better then "num_preds".

jonpsy: Sure, I wrote ```num_preds``` because it means "number of predictors". I'm open to a better name tho, anything specific in mind?

bhack: Sometimes is the number of indipendet regressors/predictor sometimes is "number of explanatory/indipendent variables".
So for me as you like

AakashKumarNain: @jonpsy let us know if you have something better in mind. It's the only thing left to do and then we can merge this

jonpsy: ```num_regressors``` has a nice ring to it :)

AakashKumarNain: Let's go with it then

AakashKumarNain (approved):

AakashKumarNain (approved):

AakashKumarNain:
Thanks for the changes. LGTM!

bhack (approved):

PR #2368, opened 2021-01-25
Lookahead optimizer: Initialize slow vars when adding slot

### LINKED ISSUE #2351: Lookahead might divide weights with mixed_float16
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Ubuntu 18.04
- TensorFlow version and how it was installed (source or binary): 2.3.0-rc2 (source)
- TensorFlow-Addons version and how it was installed (source or binary): Just lookahead extracted from the source (master)
- Python version: 3.6.9
- Is GPU used? (yes/no): yes

When tf.keras.mixed_precision.experimental.set_policy('mixed_float16') is set, using LookAhead divides the layers weights after a few gradient updates (for example 7 in my case).

After investigating, I think the reason is that when mixed_float16 is used with automatic loss scaling, the loss scale is initialized very high. Thus the first few gradient updates do nothing (see https://github.com/tensorflow/tensorflow/blob/582c8d236cb079023657287c318ff26adb239002/tensorflow/python/keras/mixed_precision/loss_scale_optimizer.py#L714 for how updates are skipped).

As a result, LookAhead ends up with its 'slow vars' initialized to 0 instead of the original layer weight. Indeed the vars are initialized to the correct value only on the first iteration during the gradient update (https://github.com/tensorflow/addons/blob/master/tensorflow_addons/optimizers/lookahead.py#L106).

Thus when the gradient update finally does something, the first variable update does average them with the slow variables (initialized to zero). Which would explain the weight division I'm seeing.

I don't have a short code to replicate the bug, but I think the above explanation and links should highlight well enough the issue with the current way LookAhead's slow variables are initialized.

olesalscheider:
# Description

Initialize the "slow" vars of the lookahead optimizer when adding the slot instead of during the first iteration.
Sine variables might not be present during the first initialization, so that they are never be initialized to the correct value with the old behavior.

Fixes #2351.

## Type of change

- [x] Bug fix
- [ ] New Tutorial
- [ ] Additional Testing

# Checklist:

    - x] By running Black + Flake8

# How Has This Been Tested?

@axeldavy tested the change in #2351. I ran a short training with the modified optimizer and confirmed that it still works.

### DIFF
--- a/tensorflow_addons/optimizers/lookahead.py
+++ b/tensorflow_addons/optimizers/lookahead.py
@@ -87,7 +87,7 @@ class Lookahead(tf.keras.optimizers.Optimizer):
             var_list=var_list
         )  # pylint: disable=protected-access
         for var in var_list:
-            self.add_slot(var, "slow")
+            self.add_slot(var, "slow", initializer=var)
 
     def _create_hypers(self):
         self._optimizer._create_hypers()  # pylint: disable=protected-access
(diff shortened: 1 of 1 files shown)


WindQAQ (approved):
Thanks.

PR #2374, opened 2021-01-30
Update 2.4.1

seanpmorgan:
# Description

Typically we don't need to build against minor versions, but since 2.4.1 changes compilation settings we should align. Planning on a 0.12.1 release for TFA after these go through


bhack (dismissed):

bhack:
Just a note without using the automatic release update script we are going to have 2.4.0 in the devel container.

seanpmorgan:
> Just a note without using the automatic release update script we are going to have 2.4.0 in the devel container.

That's a good point lets open a new PR to keep the master branch correctly updated. Shouldn't be an issue for the 0.12.1 release though on the other branch

PR #2381, opened 2021-02-07
variable histogram bins added

### LINKED ISSUE #2379: Ideally the bins in histogram equalization is variable rather than limited to 256
https://github.com/tensorflow/addons/blob/d26e2ed5f68092aed57016a7005ce534b1be3dce/tensorflow_addons/image/color_ops.py#L36

whatwilliam:
# Description

Changes the equalization method in color_ops.py to take in a variable number of histogram bins.

Fixes #2379

## Type of change

- [x] Bug fix

# Checklist:

- [x] I have made corresponding changes to the documentation

# How Has This Been Tested?

Function has been sanity tested to be the same as the old equalize_image and past tests should still pass. This PR adds `bins` as an optional `int` parameter which defaults to `256` to ensure backwards compatibility. The expected use case of this parameter is for imaging data which has greater dynamic range than 8-bit.

### DIFF
--- a/tensorflow_addons/image/color_ops.py
+++ b/tensorflow_addons/image/color_ops.py
@@ -31,7 +31,6 @@ def _scale_channel(
     image: TensorLike, channel: int, bins: Optional[int] = 256
 ) -> tf.Tensor:
     """Scale the data in the channel to implement equalize."""
-    assert bins >= 256, "Number of histogram bins should be at least 256"
     image_dtype = image.dtype
     image = tf.cast(image[:, :, channel], tf.int32)
 
--- a/tensorflow_addons/image/color_ops.py
+++ b/tensorflow_addons/image/color_ops.py
@@ -28,7 +28,7 @@ from functools import partial
 
 
 def _scale_channel(
-    image: TensorLike, channel: int, bins: Optional[int] = 256
+    image: TensorLike, channel: int, bins: int = 256
 ) -> tf.Tensor:
     """Scale the data in the channel to implement equalize."""
     image_dtype = image.dtype
@@ -53,7 +53,7 @@ def _scale_channel(
     return tf.cast(result, image_dtype)
 
 
-def _equalize_image(image: TensorLike, bins: Optional[int] = 256) -> tf.Tensor:
+def _equalize_image(image: TensorLike, bins: int = 256) -> tf.Tensor:
     """Implements Equalize function from PIL using TF ops."""
     image = tf.stack(
         [_scale_channel(image, c, bins) for c in range(image.shape[-1])], -1
@@ -63,7 +63,7 @@ def _equalize_image(image: TensorLike, bins: Optional[int] = 256) -> tf.Tensor:
 
 @tf.function
 def equalize(
-    image: TensorLike, bins: Optional[int] = 256, name: Optional[str] = None
+    image: TensorLike, bins: int = 256, name: Optional[str] = None
 ) -> tf.Tensor:
     """Equalize image(s)
 
--- a/tensorflow_addons/image/color_ops.py
+++ b/tensorflow_addons/image/color_ops.py
@@ -39,7 +39,10 @@ def _scale_channel(
 
     # For the purposes of computing the step, filter out the nonzeros.
     nonzero_histo = tf.boolean_mask(histo, histo != 0)
-    step = (tf.reduce_sum(nonzero_histo) - nonzero_histo[-1]) // bins - 1
+    if (bins - 1) == 0:
+        step = tf.reduce_sum(nonzero_histo) - nonzero_histo[-1]
+    else:
+        step = (tf.reduce_sum(nonzero_histo) - nonzero_histo[-1]) // bins - 1
 
     # If step is zero, return the original image.  Otherwise, build
     # lut from the full histogram and step and then index from it.
--- a/tensorflow_addons/image/color_ops.py
+++ b/tensorflow_addons/image/color_ops.py
@@ -31,6 +31,7 @@ def _scale_channel(
     image: TensorLike, channel: int, bins: Optional[int] = 256
 ) -> tf.Tensor:
     """Scale the data in the channel to implement equalize."""
+    assert bins >= 256, "Number of histogram bins should be at least 256"
     image_dtype = image.dtype
     image = tf.cast(image[:, :, channel], tf.int32)
 
@@ -39,10 +40,7 @@ def _scale_channel(
 
     # For the purposes of computing the step, filter out the nonzeros.
     nonzero_histo = tf.boolean_mask(histo, histo != 0)
-    if (bins - 1) == 0:
-        step = tf.reduce_sum(nonzero_histo) - nonzero_histo[-1]
-    else:
-        step = (tf.reduce_sum(nonzero_histo) - nonzero_histo[-1]) // (bins - 1)
+    step = (tf.reduce_sum(nonzero_histo) - nonzero_histo[-1]) // (bins - 1)
 
     # If step is zero, return the original image.  Otherwise, build
     # lut from the full histogram and step and then index from it.
--- a/tensorflow_addons/image/color_ops.py
+++ b/tensorflow_addons/image/color_ops.py
@@ -27,17 +27,17 @@ from typing import Optional
 from functools import partial
 
 
-def _scale_channel(image: TensorLike, channel: int) -> tf.Tensor:
+def _scale_channel(image: TensorLike, channel: int, bins: Optional[int] = 256) -> tf.Tensor:
     """Scale the data in the channel to implement equalize."""
     image_dtype = image.dtype
     image = tf.cast(image[:, :, channel], tf.int32)
 
     # Compute the histogram of the image channel.
-    histo = tf.histogram_fixed_width(image, [0, 255], nbins=256)
+    histo = tf.histogram_fixed_width(image, [0, bins-1], nbins=bins)
 
     # For the purposes of computing the step, filter out the nonzeros.
     nonzero_histo = tf.boolean_mask(histo, histo != 0)
-    step = (tf.reduce_sum(nonzero_histo) - nonzero_histo[-1]) // 255
+    step = (tf.reduce_sum(nonzero_histo) - nonzero_histo[-1]) // bins-1
 
     # If step is zero, return the original image.  Otherwise, build
     # lut from the full histogram and step and then index from it.
@@ -45,20 +45,20 @@ def _scale_channel(image: TensorLike, channel: int) -> tf.Tensor:
         result = image
     else:
         lut_values = (tf.cumsum(histo, exclusive=True) + (step // 2)) // step
-        lut_values = tf.clip_by_value(lut_values, 0, 255)
+        lut_values = tf.clip_by_value(lut_values, 0, bins-1)
         result = tf.gather(lut_values, image)
 
     return tf.cast(result, image_dtype)
 
 
-def _equalize_image(image: TensorLike) -> tf.Tensor:
+def _equalize_image(image: TensorLike, bins: Optional[int] = 256) -> tf.Tensor:
     """Implements Equalize function from PIL using TF ops."""
-    image = tf.stack([_scale_channel(image, c) for c in range(image.shape[-1])], -1)
+    image = tf.stack([_scale_channel(image, c, bins) for c in range(image.shape[-1])], -1)
     return image
 
 
 @tf.function
-def equalize(image: TensorLike, name: Optional[str] = None) -> tf.Tensor:
+def equalize(image: TensorLike, bins: Optional[int] = 256, name: Optional[str] = None) -> tf.Tensor:
     """Equalize image(s)
 
     Args:
@@ -67,6 +67,7 @@ def equalize(image: TensorLike, name: Optional[str] = None) -> tf.Tensor:
           `(num_rows, num_columns, num_channels)` (HWC), or
           `(num_rows, num_columns)` (HW). The rank must be statically known (the
           shape is not `TensorShape(None)`).
+      bins: The number of bins in the histogram.
       name: The name of the op.
     Returns:
       Image(s) with the same type and shape as `images`, equalized.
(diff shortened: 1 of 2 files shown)


Ouwen:
@whatwilliam tests fail for equalize likely some small addition error 

```
=================================== FAILURES ===================================
____________________ test_equalize_with_PIL[eager_mode-cpu] ____________________

    @pytest.mark.usefixtures("maybe_run_functions_eagerly")
    def test_equalize_with_PIL():
        np.random.seed(0)
        image = np.random.randint(low=0, high=255, size=(4, 3, 3, 3), dtype=np.uint8)
        equalized = np.stack([ImageOps.equalize(Image.fromarray(i)) for i in image])
>       np.testing.assert_equal(color_ops.equalize(tf.constant(image)).numpy(), equalized)
E       AssertionError: 
E       Arrays are not equal
E       
E       Mismatched elements: 108 / 108 (100%)
E       Max absolute difference: 252
E       Max relative difference: 50.4
E        x: array([[[[0, 1, 0],
E                [0, 0, 0],
E                [0, 0, 0]],...
E        y: array([[[[172,  10, 127],
E                [140,  47, 170],
E                [196, 151, 117]],...

tensorflow_addons/image/tests/color_ops_test.py:50: AssertionError
___________________ test_equalize_with_PIL[tf_function-cpu] ____________________

    @pytest.mark.usefixtures("maybe_run_functions_eagerly")
    def test_equalize_with_PIL():
        np.random.seed(0)
        image = np.random.randint(low=0, high=255, size=(4, 3, 3, 3), dtype=np.uint8)
        equalized = np.stack([ImageOps.equalize(Image.fromarray(i)) for i in image])
>       np.testing.assert_equal(color_ops.equalize(tf.constant(image)).numpy(), equalized)
E       AssertionError: 
E       Arrays are not equal
E       
E       Mismatched elements: 108 / 108 (100%)
E       Max absolute difference: 252
E       Max relative difference: 50.4
E        x: array([[[[0, 1, 0],
E                [0, 0, 0],
E                [0, 0, 0]],...
E        y: array([[[[172,  10, 127],
E                [140,  47, 170],
E                [196, 151, 117]],...
```

tensorflow_addons/image/color_ops.py
```diff
@@ -27,38 +27,44 @@
...
     image = tf.cast(image[:, :, channel], tf.int32)
 
     # Compute the histogram of the image channel.
-    histo = tf.histogram_fixed_width(image, [0, 255], nbins=256)
+    histo = tf.histogram_fixed_width(image, [0, bins - 1], nbins=bins)
 
     # For the purposes of computing the step, filter out the nonzeros.
     nonzero_histo = tf.boolean_mask(histo, histo != 0)
-    step = (tf.reduce_sum(nonzero_histo) - nonzero_histo[-1]) // 255
+    step = (tf.reduce_sum(nonzero_histo) - nonzero_histo[-1]) // bins - 1
```
Ouwen, line 42: likely

tensorflow_addons/image/color_ops.py
```diff
@@ -27,38 +27,47 @@
...
+        step = (tf.reduce_sum(nonzero_histo) - nonzero_histo[-1]) // (bins - 1)
 
     # If step is zero, return the original image.  Otherwise, build
     # lut from the full histogram and step and then index from it.
     if step == 0:
         result = image
     else:
         lut_values = (tf.cumsum(histo, exclusive=True) + (step // 2)) // step
-        lut_values = tf.clip_by_value(lut_values, 0, 255)
+        lut_values = tf.clip_by_value(lut_values, 0, bins - 1)
```
bhack, line 53: For line 43 could it clip with 0 ?

tensorflow_addons/image/color_ops.py
whatwilliam: new commit has been made to catch it, all checks now pass.

tensorflow_addons/image/color_ops.py
whatwilliam: If I understand your question correctly, I suppose so, then all you would be left with is a Tensor of zeros. But that would also mean that the equalization in question would be in one histogram bin, which would not be the intended use case.

bhack: Yes, probably you could test also 0, 256, and not 256 bins with https://scikit-image.org/docs/dev/api/skimage.exposure.html#skimage.exposure.equalize_hist

tensorflow_addons/image/color_ops.py
```diff
@@ -31,6 +31,7 @@ def _scale_channel(
     image: TensorLike, channel: int, bins: Optional[int] = 256
 ) -> tf.Tensor:
     """Scale the data in the channel to implement equalize."""
+    assert bins >= 256, "Number of histogram bins should be at least 256"
```
bhack, line 34: I suppose this is not the case for one of the API that expose bins params like scikit-image that we could already use in tests here https://scikit-image.org/docs/dev/api/skimage.exposure.html#skimage.exposure.equalize_hist

whatwilliam: assertion has been removed

bhack (changes requested):
I think that you can switch the test from `ImageOps.equalize` to skimage.exposure.equalize_hist to test with different bins values

whatwilliam:
> I think that you can switch the test from `ImageOps.equalize` to `skimage.exposure.equalize_hist` to test with different bins values

The purpose of this PR is to add variable histogram bins to Tensorflow Addon's Image Ops.

bhack:
> The purpose of this PR is to add variable histogram bins to Tensorflow Addon's Image Ops.

Yes but as its related test uses PIL `ImageOps.equalize` it doesn't support bins testing:

https://github.com/tensorflow/addons/blob/704c247a6d322e19cb07640aa1861c0366b43ce3/tensorflow_addons/image/tests/color_ops_test.py#L49

whatwilliam:
> > The purpose of this PR is to add variable histogram bins to Tensorflow Addon's Image Ops.
> 
> Yes but as its related test uses PIL `ImageOps.equalize` it doesn't support bins testing:
> 
> https://github.com/tensorflow/addons/blob/704c247a6d322e19cb07640aa1861c0366b43ce3/tensorflow_addons/image/tests/color_ops_test.py#L49

That will not work because the `equalize` function on the original `tensorflow:master` doesn't return the same array as using the `skimage.exposure.equalize_hist` given the same image at 256 bins.

whatwilliam:
> > > The purpose of this PR is to add variable histogram bins to Tensorflow Addon's Image Ops.
> > 
> > 
> > Yes but as its related test uses PIL `ImageOps.equalize` it doesn't support bins testing:
> > https://github.com/tensorflow/addons/blob/704c247a6d322e19cb07640aa1861c0366b43ce3/tensorflow_addons/image/tests/color_ops_test.py#L49
> 
> That will not work because the `equalize` function on the original `tensorflow:master` doesn't return the same array as using the `skimage.exposure.equalize_hist` given the same image at 256 bins. It also seems that skimage normalizes their result from [0,1] but even when the original equalize function from tensorflow's result is normalized, the two arrays differ by about 0.018

Since both implementations don't give the same answer, would it be alright to create a test that tests for a acceptable range of difference using skimage?

bhack:
> Since both implementations don't give the same answer, would it be alright to create a test that tests for a acceptable range of difference using skimage?

Just if the numerical tolerance is minimal and related to differences in precision. 
Or you need to find another way to refactor the test to verify bins.

/cc @tomerk We have a duplicated impl at https://github.com/tensorflow/models/blob/master/official/vision/beta/ops/augment.py#L476-L506

tensorflow_addons/image/color_ops.py
```diff
@@ -27,38 +27,44 @@
 from functools import partial
 
 
-def _scale_channel(image: TensorLike, channel: int) -> tf.Tensor:
+def _scale_channel(
+    image: TensorLike, channel: int, bins: Optional[int] = 256
```
WindQAQ, line 31: `bins: int = 256`

tensorflow_addons/image/color_ops.py
```diff
@@ -27,38 +27,44 @@
...
         lut_values = (tf.cumsum(histo, exclusive=True) + (step // 2)) // step
-        lut_values = tf.clip_by_value(lut_values, 0, 255)
+        lut_values = tf.clip_by_value(lut_values, 0, bins - 1)
         result = tf.gather(lut_values, image)
 
     return tf.cast(result, image_dtype)
 
 
-def _equalize_image(image: TensorLike) -> tf.Tensor:
+def _equalize_image(image: TensorLike, bins: Optional[int] = 256) -> tf.Tensor:
```
WindQAQ, line 56: `bins: int = 256`

tensorflow_addons/image/color_ops.py
```diff
@@ -27,38 +27,44 @@
...
+    image = tf.stack(
+        [_scale_channel(image, c, bins) for c in range(image.shape[-1])], -1
+    )
     return image
 
 
 @tf.function
-def equalize(image: TensorLike, name: Optional[str] = None) -> tf.Tensor:
+def equalize(
+    image: TensorLike, bins: Optional[int] = 256, name: Optional[str] = None
```
WindQAQ, line 66: `bins: int = 256`

`Optional[int]` means `Union[None, int]`. I believe it does not work when bins is passed as `None`.

WindQAQ (changes requested):

tensorflow_addons/image/color_ops.py
whatwilliam: `Optional` has been removed. It is now `bins: int = 256`

tensorflow_addons/image/tests/color_ops_test.py
```diff
@@ -50,6 +51,22 @@ def test_equalize_with_PIL():
...
+    bins = np.random.randint(low=256, dtype=np.int32)
+    equalized = tf.cast(color_ops.equalize(tf.constant(image), bins), np.float16)
+    equalized = tf.math.divide(equalized, 255)
+    skiequalized = equalize_hist(image, bins)
+    assert (
+        tf.reduce_mean(
+            tf.math.abs(tf.reshape(tf.math.subtract(equalized, skiequalized), [-1]))
+        )
+        <= 0.015
+    ), "Greater than 1.5% difference in arrays"
```
bhack, line 67: I didn't go into details but I'm still a little bit concerned about these numerical differences on a such small input matrix. @WindQAQ What do you think?

tensorflow_addons/image/tests/color_ops_test.py
```diff
@@ -50,6 +51,22 @@ def test_equalize_with_PIL():
     np.testing.assert_equal(color_ops.equalize(tf.constant(image)).numpy(), equalized)
 
 
+@pytest.mark.usefixtures("maybe_run_functions_eagerly")
```
bhack, line 54: Probably we could add `@pytest.mark.parametrize` to test different bins

tensorflow_addons/image/tests/color_ops_test.py
whatwilliam: The 1.5% metric comes from the difference between the implementations of the proven `PIL.ImageOps.equalize` and the test I'm using with `skimage.exposure.equalize_hist`.

Here's a quick showcase I made
```
#compares PIL and skimage
import tensorflow as tf
import numpy as np
import PIL
from skimage.exposure import equalize_hist
from PIL.ImageOps import *

np.random.seed(0)

low_image = np.random.randint(low=0, high=256, size=(4, 3, 3, 3), dtype=np.uint8)
PIL_equalized = np.stack([PIL.ImageOps.equalize(PIL.Image.fromarray(i)) for i in low_image])
PIL_equalized = tf.math.divide(tf.cast(PIL_equalized, tf.float32), 255) #normalized
skiequalized = equalize_hist(low_image, 256)
low_avg = tf.reduce_mean(
            tf.math.abs(tf.reshape(tf.math.subtract(PIL_equalized, skiequalized), [-1]))
)

high_image = np.random.randint(low=0, high=256, size=(28, 28, 3, 3), dtype=np.uint8)
PIL_equalized = np.stack([PIL.ImageOps.equalize(PIL.Image.fromarray(i)) for i in high_image])
PIL_equalized = tf.math.divide(tf.cast(PIL_equalized, tf.float32), 255) #normalized
skiequalized = equalize_hist(high_image, 256)
high_avg = tf.reduce_mean(
            tf.math.abs(tf.reshape(tf.math.subtract(PIL_equalized, skiequalized), [-1]))
)
print(low_avg, high_avg)
```
```
OUTPUT:
tf.Tensor(0.036022358, shape=(), dtype=float32) tf.Tensor(0.0061728833, shape=(), dtype=float32)
```
Here the differences between the two ranges from 3.6-0.6% depending on the amount of variance allowed by size of the image/number of pixels compared. So 1.5% is a reasonable metric that stands between this range that accounts for the different implementations

tensorflow_addons/image/tests/color_ops_test.py
Ouwen: @bhack the two most common would probably be 256 and 2^16. I'm not sure if it is a good idea to include more as it would slow the tests down. Additionally, since the test matrices are randomly generated there is a chance a low error tolerance would cause intermittent failures.

I think having 1 test to match the current PIL function and another test to match skimage within some tolerance that doesn't cause transient errors would be ideal. Otherwise we may need to adopt skimage equalize implementation which may be a much larger change than the current PR.

Another way to solve the transient error issue is to load a static test image.

bhack: Having the two most common is ok. How many bins are testing skimage?

whatwilliam: @bhack Originally only one random bin size was tested. The most recent commit adds `@pytest.mark.parametrize` to test for bin sizes 256 and 2^16.

tensorflow_addons/image/tests/color_ops_test.py
bhack: Please add a comment in the test why we need this tolerance

whatwilliam: comment has been added

bhack (approved):
I approve but I will not merge this until I have a second approval from another reviewer for the current  tolerance approach over an alternative implementation in tests

WindQAQ (approved):

Issue #2403, opened 2021-02-24
Lookahead fails with mixed precision and `ReduceLROnPlateau`


rroseandrieux:
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): 20.04
- TensorFlow version and how it was installed (source or binary): pip install tensorflow==2.4.1
- TensorFlow-Addons version and how it was installed (source or binary): pip install tensorflow-addons==0.12.1 
- Python version: 3.8.5
- Is GPU used? (yes/no): no

**Describe the bug**

Lookahead is not compatible with mixed precision and lr callbacks. The attribute 'lr' or 'learning_rate' is missing.
This may be due to the learning_rate being an attribute in Lookahead while being search with get_hyper by the LossScaleOptimizer when using mixed precision. 
It will fail with any callbacks that try to access the learning rate. 

**Code to reproduce the issue**
```
import tensorflow as tf
from tensorflow.keras import layers
from tensorflow.keras import mixed_precision
from tensorflow_addons.optimizers import Lookahead, RectifiedAdam
from tensorflow.keras.callbacks import ReduceLROnPlateau

policy = mixed_precision.Policy('mixed_float16')
mixed_precision.set_global_policy(policy)

model = tf.keras.Sequential()
model.add(layers.Dense(10))

test_ones = tf.ones([1, 20])

optimizer = Lookahead(RectifiedAdam(0.003))
model.compile(
    optimizer=optimizer,
    loss='binary_crossentropy',
    metrics=['accuracy'],
    run_eagerly=False,
)
model.fit(test_ones, tf.constant([0]), 8, epochs=100, callbacks=[ReduceLROnPlateau(monitor="accuracy")])
```

**Other info / logs**
```
Traceback (most recent call last):
  File "/home/raphael/Workspace/classification-api/test_r.py", line 23, in <module>
    model.fit(test_ones, tf.constant([0]), 8, epochs=100, callbacks=[ReduceLROnPlateau(monitor="accuracy")])
  File "/home/raphael/Workspace/envs/test/lib/python3.8/site-packages/tensorflow/python/keras/engine/training.py", line 1145, in fit
    callbacks.on_epoch_end(epoch, epoch_logs)
  File "/home/raphael/Workspace/envs/test/lib/python3.8/site-packages/tensorflow/python/keras/callbacks.py", line 432, in on_epoch_end
    callback.on_epoch_end(epoch, numpy_logs)
[...]

Issue #2332, opened 2021-01-04
Stocastic Weight Averaging doesn't work with Learning Rate scheduling and mixed precision


grofte:
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Google Colab / Linux Ubuntu 18.04
- TensorFlow version and how it was installed (source or binary): TF 2.4, binaries from pip
- TensorFlow-Addons version and how it was installed (source or binary): TFA 0.12, binaries from pip
- Python version: 3.6.9 / 3.8
- Is GPU used? (yes/no): yes

**Describe the bug**

When I combine _learning rate reduction on plateau_ with _stocastic weight averaging_ and _mixed precision_ the training fails after the first epoch. 

**Code to reproduce the issue**

I have commented out the two lines related to SWA to make it easier to find them. Toggle them back on and run and the error will occur. Comment them out and the training fails. 

https://gist.github.com/grofte/8142164f6b8be3d912cad9f2f1660f05 

**Other info / logs**

```
AttributeError                            Traceback (most recent call last)
<ipython-input-229-7c4428eca112> in <module>()
----> 1 model.fit(x_train, y_train, epochs=6, callbacks=[lr_reduce])

4 frames
/usr/local/lib/python3.6/dist-packages/tensorflow/python/keras/mixed_precision/loss_scale_optimizer.py in __getattribute__(self, name)
    882   def __getattribute__(self, name):
    883     try:
--> 884       return object.__getattribute__(self, name)
    885     except AttributeError as e:
    886       if name == '_optimizer' or name == '_hyper':

AttributeError: 'LossScaleOptimizer' object has no attribute 'lr'
```

grofte:
Oops, sorry. Google Colab hides some of the error message

```
AttributeError                            Traceback (most recent call last)
<ipython-input-229-7c4428eca112> in <module>()
----> 1 model.fit(x_train, y_train, epochs=6, callbacks=[lr_reduce])

4 frames
/usr/local/lib/python3.6/dist-packages/tensorflow/python/keras/engine/training.py in fit(self, x, y, batch_size, epochs, verbose, callbacks, validation_split, validation_data, shuffle, class_weight, sample_weight, initial_epoch, steps_per_epoch, validation_steps, validation_batch_size, validation_freq, max_queue_size, workers, use_multiprocessing)
   1143           epoch_logs.update(val_logs)
   1144 
-> 1145         callbacks.on_epoch_end(epoch, epoch_logs)
[...]

PR #2404, opened 2021-02-26
[WIP]  LookAhead fix for mixed

jonpsy:
# Description

Brief Description of the PR:
A naive solution.

Fixes # (issue)
#2403

I still think delegating the ```_hyper``` dictionary of the inner optimizer(in our case Rectified Adam) would be a better solution, but I'm getting an infinite loop in that case. That'd be the best solution imo, I'd like to discuss how to fix that.

A few test fails as of now

### DIFF
--- a/tensorflow_addons/optimizers/lookahead.py
+++ b/tensorflow_addons/optimizers/lookahead.py
@@ -186,7 +186,7 @@ class Lookahead(tf.keras.optimizers.Optimizer):
         # Backwards compatibility with Keras optimizers.
         if name == "lr":
             name = "learning_rate"
-            self._optimizer._set_hyper(name, value)
+            self._optimizer._set_hyper(name, value)  # Delegate to internal optimizer
         if hasattr(self, "_hyper") and name in self._hyper:
             self._set_hyper(name, value)
         else:
--- a/tensorflow_addons/optimizers/lookahead.py
+++ b/tensorflow_addons/optimizers/lookahead.py
@@ -15,6 +15,7 @@
 
 import tensorflow as tf
 from tensorflow_addons.utils import types
+from tensorflow_addons.utils.types import FloatTensorLike
 
 from typeguard import typechecked
 
@@ -45,6 +46,7 @@ class Lookahead(tf.keras.optimizers.Optimizer):
         self,
         optimizer: types.Optimizer,
         sync_period: int = 6,
+        learning_rate: FloatTensorLike = 0.001,
         slow_step_size: types.FloatTensorLike = 0.5,
         name: str = "Lookahead",
         **kwargs,
@@ -56,6 +58,8 @@ class Lookahead(tf.keras.optimizers.Optimizer):
                 and apply the gradients.
             sync_period: An integer. The synchronization period of lookahead.
                 Enable lookahead mechanism by setting it with a positive value.
+            learning_rate: A `Tensor` or a floating point value.
+                The learning rate.
             slow_step_size: A floating point value.
                 The ratio for updating the slow weights.
             name: Optional name for the operations created when applying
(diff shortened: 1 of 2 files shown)


jonpsy:
Hi @reedwm I really like your [approach](https://github.com/tensorflow/addons/issues/2332#issuecomment-754999736)  on this. However, I face two problems

1. If we relay our ```_hyper``` dict to the internal_optimizer (say RectifiedAdam), then we will lose our original ```_hyper``` dict. For ex: In LookAhead the original ```_hyper``` dict has these two params ```sync_period``` and ```slow_step_size```.
2. I went ahead and did it anyway, and I got maximum recursion depth reached error.

```
  File "/home/psykik/PycharmProjects/addons/tensorflow_addons/optimizers/lookahead.py", line 175, in _hyper
    return self._optimizer.hyper
  File "/home/psykik/PycharmProjects/addons/venv/lib/python3.8/site-packages/tensorflow/python/keras/optimizer_v2/optimizer_v2.py", line 788, in __getattribute__
    return super(OptimizerV2, self).__getattribute__(name)
RecursionError: maximum recursion depth exceeded while calling a Python object
```

reedwm:
Hmmm, I suspected delegating `_hyper` might be iffy. I will take a look next week and try to come up with a solution.

A public API to expose hyperaparameters would fix this, as then LossScaleOptimizer could use that API instead of using `_hyper`, and Lookahead could override the API. But it's much better to avoid adding a public API if possible.

jonpsy:
Acked, but the LossScaleOptimizer is inside the official tensorflow repo. Imo, there's no cleaner way of doing it without meddling with the official repo. 

I have a stand-in solution though, how about we add another hyper parameter named ```learning_rate``` to the ```LookAhead```. Basically, it's a copy of the ```learning_rate``` parameter of the internal_optimizer, and we will update both of them parallely.  This way ```learning_rate``` hyper param will be found inside the ```_hyper``` dict of LookAhead.

bhack:
I've just triggered the CI to investigate side effects on tests with https://github.com/tensorflow/tensorflow/pull/47449

jonpsy:
All the tests should pass.

My approach:

1) Create a stand-in hyperparameter inside ```LookAhead``` class, with the same name ```learning_rate```. This way ```learning_rate``` will be found in ```_hyper``` dictionary.

2) Override the __set_attr__ of the OptimizerV2.py inside LookAhead so that it sets the ```learning_rate``` of the internal optimizer.

reedwm:
The issue with creating a `learning_rate` hyperparameter in Lookahead is that setting the Lookahead learning rate does not change the original optimizer's learning rate. For example, try running the following with this PR:

```python
import tensorflow as tf

orig_optimizer = tf.keras.optimizers.Adam(learning_rate=0.002)
optimizer = tfa.optimizers.Lookahead(orig_optimizer)
optimizer = tf.keras.mixed_precision.LossScaleOptimizer(optimizer)
print(optimizer.lr)  # 0.001
print(orig_optimizer.lr)  # 0.002
optimizer.learning_rate = 0.003
print(optimizer.lr)  # 0.003
print(orig_optimizer.lr)  # 0.002
```

In this case, `orig_optimizer.lr` is the actual learning rate that will be used during training, but setting `optimizer.lr` does not affect `orig_optimizer.lr`. This means the LearningRateSchecduler will have no impact.

I'll try to come up with a workaround, but I dont have any yet. /CC @omalleyt12 let me know if you have any ideas

jonpsy:
This is why I've overriden the set attribute, so that it changes both of them. Very well, shall I add this test case?

reedwm:
Unfortunately, overriding `__setattr__` doesn't work, since LossScaleOptimizer delegates \_\_setattr\_\_ to call `self._optimizer._set_hyper_` instead of `self._optimizer.__setattr__`. See [here](https://github.com/tensorflow/tensorflow/blob/19c6728118c25f02f4a6fa754ca53ca6aa1044e4/tensorflow/python/keras/mixed_precision/loss_scale_optimizer.py#L919-L921).

Perhaps overriding `_set_hyper` in Lookahead, similar to what you do in `__setattr__`, would work.

jonpsy:
Interesting..... I'll try that . Parallelly, bhack is working on fixing it in the official repo itself. So all our hardworking maybe for naught, haha.

jonpsy:
Closing this PR in favor of  tensorflow/tensorflow#47449

PR #2486, opened 2021-05-18
Fix using NoisyNet with .fit() or .train_on_batch()

markub3327:
The standard Keras layer `call()` method contains only **inputs, training and mask arguments**. Using reset_noise() and remove_noise() outside the call() is an effective way when we decide to save the model in h5 format (We don't just have to use a custom model).
 
For loading model:             
`model = tf.keras.models.load_model(fileName, compile=False, custom_objects={'NoisyDense': NoisyDense})`

When eps_bias and eps_kernel weren't defined by `add_weight()` method the learning process doesn't work for me and agent not converge (the agent not achieve a game goal).

Thanks.

### DIFF
--- a/tensorflow_addons/layers/noisy_dense.py
+++ b/tensorflow_addons/layers/noisy_dense.py
@@ -237,7 +237,7 @@ class NoisyDense(tf.keras.layers.Dense):
             self.sigma_bias = None
             self.mu_bias = None
             self.eps_bias = None
-        self._reset_noise()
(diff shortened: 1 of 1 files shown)


markub3327:
@bhack Yes, I can maintain this layer. Thanks.

bhack:
Sorry, please commit here

markub3327:
@bhack  Sorry, now it's done ...

PR #2406, opened 2021-03-01
Matthew Fix

jonpsy:
# Description

Brief Description of the PR:

Fixes # (issue)
#2339 

## Type of change

- [x] Bug fix
- [ ] New Tutorial
- [x] Updated or additional documentation
- [x] Additional Testing

# Checklist:

- [ x] I've properly [formatted my code according to the guidelines](https://github.com/tensorflow/addons/blob/master/CONTRIBUTING.md#coding-style)
    - [x] By running Black + Flake8
    - [ ] By running pre-commit hooks
- [x] This PR addresses an already submitted issue for TensorFlow Addons
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] This PR contains modifications to C++ custom-ops

# How Has This Been Tested?

The issue created by the user is perfectly solved by this

If you're adding a bugfix or new feature please describe the tests that you ran to verify your changes:

PR #2495, opened 2021-06-02
Fix support for sparse labels in MatthewsCorrelationCoefficient

lgeiger:
# Description

This PR adds support for sparse labels and predictions in `MatthewsCorrelationCoefficient`. This re-enables the use of `num_classes=1` which was broken in #2406.

Fixes #2494

## Type of change

- [x] Bug fix
- [ ] New Tutorial
- [ ] Additional Testing

# Checklist:

### DIFF
--- a/tensorflow_addons/metrics/matthews_correlation_coefficient.py
+++ b/tensorflow_addons/metrics/matthews_correlation_coefficient.py
@@ -69,7 +69,7 @@ class MatthewsCorrelationCoefficient(tf.keras.metrics.Metric):
     ):
         """Creates a Matthews Correlation Coefficient instance."""
         super().__init__(name=name, dtype=dtype)
-        self.num_classes = num_classes
+        self.num_classes = max(2, num_classes)
         self.conf_mtx = self.add_weight(
             "conf_mtx",
             shape=(self.num_classes, self.num_classes),
@@ -82,9 +82,19 @@ class MatthewsCorrelationCoefficient(tf.keras.metrics.Metric):
         y_true = tf.cast(y_true, dtype=self.dtype)
         y_pred = tf.cast(y_pred, dtype=self.dtype)
 
+        if y_true.shape[-1] == 1:
+            labels = tf.squeeze(tf.round(y_true), axis=-1)
+        else:
+            labels = tf.argmax(y_true, 1)
+
+        if y_pred.shape[-1] == 1:
+            predictions = tf.squeeze(tf.round(y_pred), axis=-1)
+        else:
+            predictions = tf.argmax(y_pred, 1)
+
         new_conf_mtx = tf.math.confusion_matrix(
-            labels=tf.argmax(y_true, 1),
-            predictions=tf.argmax(y_pred, 1),
+            labels=labels,
+            predictions=predictions,
             num_classes=self.num_classes,
             weights=sample_weight,
             dtype=self.dtype,
@@ -126,7 +136,4 @@ class MatthewsCorrelationCoefficient(tf.keras.metrics.Metric):
         """Resets all of the metric state variables."""
 
         for v in self.variables:
-            K.set_value(
-                v,
-                np.zeros((self.num_classes, self.num_classes), v.dtype.as_numpy_dtype),
-            )
+            K.set_value(v, np.zeros(v.shape, v.dtype.as_numpy_dtype))
--- a/tensorflow_addons/metrics/tests/matthews_correlation_coefficient_test.py
+++ b/tensorflow_addons/metrics/tests/matthews_correlation_coefficient_test.py
@@ -23,17 +23,28 @@ from sklearn.metrics import matthews_corrcoef as sklearn_matthew
 
 def test_config():
     # mcc object
-    mcc1 = MatthewsCorrelationCoefficient(num_classes=1)
-    assert mcc1.num_classes == 1
+    mcc1 = MatthewsCorrelationCoefficient(num_classes=2)
+    assert mcc1.num_classes == 2
     assert mcc1.dtype == tf.float32
     # check configure
     mcc2 = MatthewsCorrelationCoefficient.from_config(mcc1.get_config())
-    assert mcc2.num_classes == 1
+    assert mcc2.num_classes == 2
     assert mcc2.dtype == tf.float32
 
 
 def check_results(obj, value):
     np.testing.assert_allclose(value, obj.result().numpy(), atol=1e-6)
+
+
+def test_binary_classes_sparse():
+    gt_label = tf.constant([[1.0], [1.0], [1.0], [0.0]], dtype=tf.float32)
+    preds = tf.constant([[1.0], [0.0], [1.0], [1.0]], dtype=tf.float32)
+    # Initialize
+    mcc = MatthewsCorrelationCoefficient(1)
+    # Update
+    mcc.update_state(gt_label, preds)
+    # Check results
+    check_results(mcc, [-0.33333334])
 
 
 def test_binary_classes():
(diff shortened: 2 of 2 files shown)


tensorflow_addons/metrics/tests/matthews_correlation_coefficient_test.py
```diff
@@ -23,19 +23,30 @@
...
 
+def test_binary_classes_sparse():
+    gt_label = tf.constant([[1.0], [1.0], [1.0], [0.0]], dtype=tf.float32)
+    preds = tf.constant([[1.0], [0.0], [1.0], [1.0]], dtype=tf.float32)
+    # Initialize
+    mcc = MatthewsCorrelationCoefficient(1)
+    # Update
+    mcc.update_state(gt_label, preds)
+    # Check results
+    check_results(mcc, [-0.33333334])
```
lgeiger, line 47: This restores the behaviour of v0.12

jonpsy: The mentioned example isn't sparse, can you report the result of the following input?

```python 
 gt_label = tf.constant([1, 1, 1, 0], dtype=tf.float32)
 preds = tf.constant([1, 0, 1, 1], dtype=tf.float32)
```

tensorflow_addons/metrics/matthews_correlation_coefficient.py
```diff
@@ -82,9 +82,19 @@ def update_state(self, y_true, y_pred, sample_weight=None):
         y_true = tf.cast(y_true, dtype=self.dtype)
         y_pred = tf.cast(y_pred, dtype=self.dtype)
```
jonpsy, line 83: Instead of rounding in the future, I suggest cast it to tf.int64

jonpsy:
The test doesn't "break" per se, it wasn't intended to be used for sparse-labels in the first place. See this [comment](https://github.com/tensorflow/addons/pull/2406#issuecomment-788145938). I gather that you're trying to add sparse support with multi and binary labels? This [code](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/metrics/cohens_kappa.py)  will serve as a good reference for library convention.

Most importantly, it doesn't make sense to have ```numclasses=1```, even if you have one class, it still represented with 2; one indicating its presence and other its lack thereof ( think of it like Yes/No).  So the current code **rightly** "breaks" your edge case.

tensorflow_addons/metrics/matthews_correlation_coefficient.py
```diff
@@ -69,7 +69,7 @@ def __init__(
     ):
         """Creates a Matthews Correlation Coefficient instance."""
         super().__init__(name=name, dtype=dtype)
-        self.num_classes = num_classes
+        self.num_classes = max(2, num_classes)
         self.conf_mtx = self.add_weight(
             "conf_mtx",
             shape=(self.num_classes, self.num_classes),
```
jonpsy, line 75: Instead of ```max(2, num_classes)```, add a check to detect if num_classes is less than 2 and appropriately throw error. See cohen_kappa.py for this.

jneeven:
> Most importantly, it doesn't make sense to have numclasses=1, even if you have one class, it still represented with 2; one indicating its presence and other its lack thereof ( think of it like Yes/No). So the current code rightly "breaks" your edge case.

@jonpsy I strongly disagree. It is very common to, if there is just a single class, have only a single (sigmoid) output representing the confidence that the input is a positive. It's the reason the [BinaryCrossentropy loss](https://www.tensorflow.org/api_docs/python/tf/keras/losses/BinaryCrossentropy) exists and it's a totally valid usecase. In fact, I'd argue that it doesn't make sense to have *two* outputs if you only have a single class, because you can represent the same prediction with just a single number without losing any information. Additionally, `num_classes=1` was actually the "default" used in [the original docstring](https://github.com/tensorflow/addons/pull/2406/files#diff-6fd6d41de218f9cda2bdde7ec26183ecf47670697b10fafee35396193987771bL53-L55), so saying that this update has rightly broken that usecase seems strange to me. Changing the concept of binary predictions from `num_classes=1` to `num_classes=2` unnecessarily breaks backward compatibility, as both concepts were previously supported (ignoring the fact that the multi-class case was apparently bugged).

tensorflow_addons/metrics/tests/matthews_correlation_coefficient_test.py
lgeiger: Yeah I don't think this would work correctly. But the motivation of this PR was to restore the behaviour of v0.12 and earlier. I'd be happy to change it to support the above example. Is there a place where the conventions of the metrics are documented /cc @seanpmorgan

lgeiger:
> The test doesn't "break" per se, it wasn't intended to be used for sparse-labels in the first place. See this [comment](https://github.com/tensorflow/addons/pull/2406#issuecomment-788145938). I gather that you're trying to add sparse support with multi and binary labels?

To be honest, I am not very familiar with this code and the conventions for TFA metrics. But I disagree with your assessment that this behaviour was never intended. It has been the behaviour in all releases between v0.7 and v0.12 including it being tested in the unittest. And the functionality was removed in #2406 while changing the unittest.
I think there is definitely a point to discuss in how to handle this in a consistent way across the other metrics. But as far as I can tell this functionality was removed in #2406 without additional discussion mentioning that this is a breaking change for users, or am I missing something here?

I don't want to advocate for not ever making breaking changes (in fact semver definitely allows for that when moving from 0.12 to 0.13), but then it needs to be documented in the release notes and ideally the code should throw an actionable error message. Which currently isn't the case since the release notes only mention that there was a "fix for MCC" and the metric will now return `0` for `num_classes=1` which can be tricky to debug in userspace.

I also agree with the point @jneeven raises about supporting `num_classes=1` being a sensible thing.

For me primarily there are two main questions about this PR.
- Do we want to restore the behaviour of v0.12 and earlier or do we want to support the behaviour mentioned in https://github.com/tensorflow/addons/pull/2495#discussion_r644428974
- Do we want to support `num_classes=1` or not. I do think it makes sense since it is an easy to support thing. But if we are breaking backwards compatibility anyway, we could also make the choice of only supporting `num_classes=2` with sparse labels which would be an easy to do code change in user space however would break with the conventions of e.g. `BinaryCrossentropy` loss which @jneeven detailed above.

jonpsy:
> @jonpsy I strongly disagree. It is very common to, if there is just a single class, have only a single (sigmoid) output representing the confidence that the input is a positive. It's the reason the [BinaryCrossentropy loss](https://www.tensorflow.org/api_docs/python/tf/keras/losses/BinaryCrossentropy) exists and it's a totally valid usecase. In fact, I'd argue that it doesn't make sense to have _two_ outputs if you only have a single class, because you can represent the same prediction with just a single number without losing any information.

It looks like you have confused the probability of class occurring with the number of distinguishable classes. In your example itself, please note the word "Binary" which itself suggests two-class problem. Label representation (via sparse or OHE) and probability of class occurring are two different things, the former is used in metric implementations. For instance, consider the following example,

```python
import tensorflow as tf
m = tf.keras.metrics.MeanIoU(num_classes=1)
m.update_state([1.0, 1.0, 1.0, 1.0], [0.3, 0.7, 0.8, 0.9])
m.result().numpy()
```
As per your line of reasoning, this is perfectly fine since we're using a "single number for representing class", As expected, this throws errors. I believe this should suffice unless of course, official TensorFlow implementation is wrong.

**But what about backward compatibility?**

> But I disagree with your assessment that this behaviour was never intended. It has been the behaviour in all releases between v0.7 and v0.12 including it being tested in the unittest. And the functionality was removed in #2406 while changing the unittest.

There's no "backward compatibility" issue here, the previous implementation was invaild in the doctest, simple as that. It's not an assessment, it's a fact. Bluntly put, you've restored an invalid case and changed the current code to "work" on that case. Although, I concede that we should've discussed it in our PR.

I also agree that we should've thrown an error when this invalid case is invoked, for example: cohen_kappa.py does this 

https://github.com/tensorflow/addons/blob/97eb293d8b085cec4ba1429e1d466a1a8e10c2f0/tensorflow_addons/metrics/cohens_kappa.py#L112-L120

Which I've already suggested in a previous comment on this PR.

I've done my best to explain this and if you're still having trouble understanding, I suppose @bhack  could help you out here or you can post it to StackOverflow.

If you want to add support for sparse labels, I would be happy to help you in that regard :+1:

tensorflow_addons/metrics/tests/matthews_correlation_coefficient_test.py
jonpsy: In tensorflow (and in most libraries), sparse labels are tensors of rank 1. That means it's dimensionality is ```(numclasses, )``` and not ```(numclasses, 1)```.  

As for conventions, you can check [README.md](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/metrics/README.md) in metrics to get a brief idea on how things work. You can also check [metrics.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/metrics.py) from tensorflow official for code conventions as we follow the same standard. There's also an exquisite set of metrics available in our ```metrics``` folders which will give you a good idea. 

Let me know if this helps, :+1:

seanpmorgan:

PR #2515, opened 2021-07-01
Utilize SIG Build Docker Images

seanpmorgan:
# Description

Move to managed SIG Build images.


bhack:
@angerson A TF 2.6 has still a python 2.6 wheel do you plan to push also a 2.6 image?

bhack:
OK.. green lights here..

bhack:
https://discuss.tensorflow.org/t/adopting-open-source-dockerfiles-for-official-tf-nightly-ci/6050

bhack:
We are having some issues with the `auditwheel` patch now. 
What was the origin of this patch? Is it still required?

```
#27 0.585 + sed -i 's/libresolv.so.2"/libresolv.so.2", "libtensorflow_framework.so.2"/g' /usr/local/lib/python3.8/dist-packages/auditwheel/policy/policy.json
#27 0.588 sed: can't read /usr/local/lib/python3.8/dist-packages/auditwheel/policy/policy.json: No such file or directory
#27 ERROR: executor failed running [/bin/sh -c bash tools/releases/tf_auditwheel_patch.sh]: exit code: 2
------
```

seanpmorgan:
> We are having some issues with the `auditwheel` patch now. What was the origin of this patch? Is it still required?
> 
> ```
> #27 0.585 + sed -i 's/libresolv.so.2"/libresolv.so.2", "libtensorflow_framework.so.2"/g' /usr/local/lib/python3.8/dist-packages/auditwheel/policy/policy.json
> #27 0.588 sed: can't read /usr/local/lib/python3.8/dist-packages/auditwheel/policy/policy.json: No such file or directory
> #27 ERROR: executor failed running [/bin/sh -c bash tools/releases/tf_auditwheel_patch.sh]: exit code: 2
> ------
> ```

Was solved in https://github.com/tensorflow/addons/pull/2623

seanpmorgan:
> We are having some issues with the `auditwheel` patch now. What was the origin of this patch? Is it still required?
> 
> ```
> #27 0.585 + sed -i 's/libresolv.so.2"/libresolv.so.2", "libtensorflow_framework.so.2"/g' /usr/local/lib/python3.8/dist-packages/auditwheel/policy/policy.json
> #27 0.588 sed: can't read /usr/local/lib/python3.8/dist-packages/auditwheel/policy/policy.json: No such file or directory
> #27 ERROR: executor failed running [/bin/sh -c bash tools/releases/tf_auditwheel_patch.sh]: exit code: 2
> ------
> ```

Just for visibility the reason for this [auditwheel patch](https://github.com/tensorflow/addons/blob/master/tools/releases/tf_auditwheel_patch.sh) is to specify that tensorflow shared library should be added to the approved manylinux2010 libraries. If we remove this it will fail to export a manylinux2010 package.

seanpmorgan:
Nice to be able to cut a lot of this tech debt from the build file

tools/docker/dev_container.Dockerfile
```diff
@@ -1,11 +1,9 @@
 #syntax=docker/dockerfile:1.1.5-experimental
-FROM gcr.io/tensorflow-testing/nosla-cuda11.2-cudnn8.1-ubuntu18.04-manylinux2010-multipython as dev_container_cpu
+ARG PY_VERSION
+FROM tensorflow/build:latest-python$PY_VERSION as dev_container_cpu
```
bhack, line 3: I think that we could need to add comment here that this is a GPU only image with a reference to https://github.com/tensorflow/build/pull/47 or any related ticket that you like (on TF or Addons).

bhack, line 3: If you can add here a fake but future-proof `IMAGE_TYPE` arg I think that we could remove the overhead of maintaining another separated Dockerfile (if you suppose that the our devel and our Vscode/Codespaces image could be almost the same one):
https://github.com/tensorflow/addons/blob/4617bc2e7e2a90b27ed9e070c36fc0724c99e149/.devcontainer/Dockerfile#L1-L2

The main thing we are doing on that additional Dockerfile, other then let the user controlling the `IMAGE_TYPE` when they need the GPU image (currently not available)  is that we are adding a `default` (guid/uid) non root user in the image build.

If we exclude that unification we need to evaluate if without a default non root user, we can still official suggest that we prefer/support only [Docker in rootless mode](https://docs.docker.com/engine/security/rootless/) this is currently the SIG Build policy that doesn't want to add a "default" user in the image.

It is hard to support both as rootless Docker vs root Docker are mutually exclusive for root and non root users permission on the host mounted volumes:

https://github.com/moby/moby/issues/41497
https://discuss.tensorflow.org/t/adopting-open-source-dockerfiles-for-official-tf-nightly-ci/6050/9

bhack (commented):
If https://github.com/tensorflow/addons/blob/master/.devcontainer/Dockerfile#L1-L2

tools/docker/dev_container.Dockerfile
seanpmorgan: Done.

seanpmorgan: Added the holder for `IMAGE_TYPE`. And updated to push it correctly as gpu dev container. I think it should be good to merge and we can revisit once we have a means to get a cpu image.

seanpmorgan:
Made a new PR for the release:
https://github.com/tensorflow/community/pull/406

bhack (approved):

PR #2598, opened 2021-11-09
Use multipython image for dev container

seanpmorgan:
# Description

Previous container only supports up to py3.6 which TF2.7 is not built for. TF team has stopped publishing these custom-op images since TF2.5.

Using the CUDA image makes the container larger, but it still uses the cpu only version of TF so it's accurate to call it dev_cpu. Building the gpu image would be pretty trivial now, but also this limits the maintenance burden on our team.

There is no multipython container without GPU installs:
https://console.cloud.google.com/gcr/images/tensorflow-testing/GLOBAL

### DIFF
--- a/tools/docker/dev_container.Dockerfile
+++ b/tools/docker/dev_container.Dockerfile
@@ -1,11 +1,11 @@
 #syntax=docker/dockerfile:1.1.5-experimental
-FROM tensorflow/tensorflow:2.1.0-custom-op-ubuntu16 as dev_container_cpu
+FROM gcr.io/tensorflow-testing/nosla-cuda11.2-cudnn8.1-ubuntu18.04-manylinux2010-multipython as dev_container_cpu
 ARG TF_PACKAGE
 ARG TF_VERSION
 
 # Temporary until custom-op container is updated
-RUN ln -sf /usr/bin/python3 /usr/bin/python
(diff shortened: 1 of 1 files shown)


bhack:
Is this image used only to produce wheels?

seanpmorgan:
> Is this image used only to produce wheels?

It is the image we use to create wheels. It has the proper gcc, bazel, and python installations. Currently our build it broken due to the dev container not being able to be built with TF2.5 + Python3.6:
https://github.com/tensorflow/addons/actions/runs/1437774966

This is gaiting a 0.15 release

bhack (approved):

bhack:
The image is very large now:

https://hub.docker.com/r/tfaddons/dev_container/tags

This is really going to slow down the bootstrap of the devcontainer to pull the image in Vsocde and also sto start a Github Codespaces

bhack:
As this is linked in https://github.com/tensorflow/addons/blob/master/.devcontainer/Dockerfile#L2

Do we have any alternative for the developer env?

seanpmorgan:
Not that I know of, the old container stopped being updated by TF dev infra. We can either request they resume that -- but alternatively they mentioned managed dev containers at the TF community day. We should get an ETA.

bhack:
That it is part of https://github.com/tensorflow/addons/pull/2515

PR #2599, opened 2021-11-10
Drop dev from 0.15

seanpmorgan:
# Description

Prepping for release


tensorflow_addons/version.py
```diff
@@ -28,7 +28,7 @@
 # stable release (indicated by `_VERSION_SUFFIX = ''`). Outside the context of a
 # release branch, the current version is by default assumed to be a
 # 'development' version, labeled 'dev'.
```
bhack, line 30: Do we need to remove this string comment?

seanpmorgan: Not sure why? Outside of a release branch (r0.15 being a release branch) it's assumed dev.

PR #2600, opened 2021-11-10
Drop macOS ARM wheels from 0.15 release

seanpmorgan:
Per discussion in SIG meeting we will not gait the Addons releases until tensorflow-macos provides an SLA. We will continue to publish tfa-nightly m1 wheels.


bhack:
Just to notify again @kulinseth  about this

bhack (approved):

PR #2623, opened 2021-12-14
Improve CI testing and release process

seanpmorgan:
# Description

* New auditwheel policy file: https://github.com/pypa/auditwheel/issues/305
* Use bazeliskrc instead of .bazelversion to avoid un-needed rigidness causing GPU test failure.

Going forward, need to align with the open source SIG Build containers, looking to return CI to production right now though.


.bazeliskrc
```diff
@@ -0,0 +1 @@
+USE_BAZEL_VERSION=3.7.2
```
fsx950223, line 1: Tensorflow uses 4.2.1 now.

seanpmorgan: Thanks, lets merge this as CI fix and then can make separate PR to increase bazel version

fsx950223 (approved):

Issue #2653, opened 2022-01-29
F1Score is not working with sparse labels


markub3327:
Hello,

   I tried to use F1Score metrics instead of accuracy during training. I get an error with incompatible shapes because I used sparse_categorical_crossentropy and my labels are sparsely defined.

Please can I or you implement a sparse version of F1Score?

Thanks.
Have a nice day.

PR #2654, opened 2022-01-29
f_scores with sparse labels

markub3327:
The solution of #2653.


bhack:
F-score is going to be implemented in https://github.com/keras-team/keras-cv/issues/393. 
Please add any additional feature there.

Issue #17, opened 2019-01-15
Build out examples with working colab notebooks


seanpmorgan:
Would be nice to mirror the TF2 doc style of colab notebooks being first-class tutorials.

I think it's also a good idea to require colab notebooks for new contributions in order to showcase their benefits and best practices. Certainly won't be possible/necessary for everything.

armando-fandango:
Let us refactor this issue with separate issue for each specific example that we propose to build. That way it would be easier to know and close out when the issue the done.

Smokrow:
Should we showcase the benefit of the implementations with some kind of before/after comparison and some kind of explanation? Or more like a basic usage?

seanpmorgan:
I think anytime we can compare it to some control that'd be great. For example optimizers could be compared to SGD on some demo mnist model.

This won't always be appropriate though, and basic usage is definitely useful to have examples for.

seanpmorgan:
Closing this as we'll create specific example requests in the coming days.

PR #2671, opened 2022-02-15
:arrow_up: `scikit-image`

### LINKED ISSUE #2666: Fix Sklearn and Skimage Testing Dependencies
The build for TF2.8 and specifically python 3.10 requires us to increment the version of sklearn and skimage.  Unfortunately this causes conflicts with the lowest version of NumPy we need to support for our TensorFlow compatibility matrix. 

As each TF release occurs, we should check to see if we can return the 5 tests that were skipped as part of:
https://github.com/tensorflow/addons/pull/2635

Alternatively it would be nice to remove these testing dependencies all together if possible.

lgeiger:
This PR updates `scikit-image` to fix #2666.


bhack:
Please check the CI logs:

> 2022-02-16T00:14:46.9665119Z #17 12.52 ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
2022-02-16T00:14:46.9665718Z #17 12.52 tensorflow 2.6.3 requires numpy~=1.19.2, but you have numpy 1.21.5 which is incompatible.

seanpmorgan:
> Please check the CI logs:
> 
> > 2022-02-16T00:14:46.9665119Z #17 12.52 ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
> > 2022-02-16T00:14:46.9665718Z #17 12.52 tensorflow 2.6.3 requires numpy~=1.19.2, but you have numpy 1.21.5 which is incompatible.

Not seeing why this is an issue? We install 1.21 and then are able to run through the tests even though TF 2.6 wants 1.19

bhack:
What kind of side effects we could have? At least we rely on TF.

bhack:
I remember that we had a lot of efforts and failing PRs for numpy 1.20 in TF:

https://github.com/tensorflow/tensorflow/issues/50204#issuecomment-1007075682

Issue #2631, opened 2021-12-22
MacOS Arm64 support for stable TF-Addons Releases


seanpmorgan:
Previously, we've added support for M1 ARM64 wheels and added them to our nightly publishing:
https://github.com/tensorflow/addons/pull/2559

We've excluded this from the stable releases since we're gaited on the release of tensorflow-macos, and we didn't want this to slow down our release process with an unknown SLA:
https://github.com/tensorflow/addons/pull/2600

We've received word of a 1 week SLA for the publishing of tensorflow-macos so going forward with TFA 0.16 we will be gaiting our release in order to include macos wheels as part of our stable release.

poinwater:
Hi, I want to ask a few questions about the mac arm support.
1. What is SLA and why you need to get it for supporting tensorflow-macos?
2. Is getting SLA a requirement for supporting m1 platform?
2. Will the TFA 0.16 release support tensorflow 2.5.0 - 2.7.0 on mac Arm64? It seems like the new release is against tensorflow 2.8.0. 

If you could answer my questions, I really appreciate it.

seanpmorgan:
https://pypi.org/project/tensorflow-addons/0.16.1/#files

seanpmorgan:
> Hi, I want to ask a few questions about the mac arm support.
> 
> 1. What is SLA and why you need to get it for supporting tensorflow-macos?
> 2. Is getting SLA a requirement for supporting m1 platform?
> 3. Will the TFA 0.16 release support tensorflow 2.5.0 - 2.7.0 on mac Arm64? It seems like the new release is against tensorflow 2.8.0.
> 
> If you could answer my questions, I really appreciate it.

1. Service level agreement, it means they will aim to release tensorflow-macos 1 week after tf releases.

PR #2696, opened 2022-04-18
Build against TF 2.9

seanpmorgan:
Still needed:

- [x] TF2.9 tensorflow release
- [x] TF2.9 tensorflow-macos release


.github/workflows/release.yml
```diff
@@ -42,19 +42,19 @@ jobs:
...
         py-version: ['3.7', '3.8', '3.9', '3.10']
-        tf-version: ['2.6.3', '2.8.0']
+        tf-version: ['2.7.1', '2.9.0rc0']
         cpu: ['x86']
         exclude:
           - py-version: '3.10'
-            tf-version: '2.6.3'
+            tf-version: '2.7.1'
         include:
           - os: 'macos-11'
```
bhack, line 51: These are not released yet for 2.9.x https://pypi.org/project/tensorflow-macos/#files

tools/install_so_files.sh
```diff
@@ -1,7 +1,7 @@
 set -e -x
 
 if [ "$TF_NEED_CUDA" == "1" ]; then
-  CUDA_FLAG="--crosstool_top=//build_deps/toolchains/gcc7_manylinux2010-nvcc-cuda11:toolchain"
+  CUDA_FLAG="--crosstool_top=@ubuntu20.04-gcc9_manylinux2014-cuda11.2-cudnn8.1-tensorrt7.2_config_cuda//crosstool:toolchain"
```
bhack, line 4: What we want to do with the new manylinux2014  toolchains? 
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/toolchains/remote_config/configs.bzl

bhack: https://github.com/tensorflow/tensorflow/blob/master/.bazelrc#L608-L619

bhack: @angerson @perfinion  What we need to switch from the custom_ops repo toolchain to `@ubuntu20.04-gcc9_manylinux2014-cuda11.2-cudnn8.1-tensorrt7.2_config_cuda//crosstool:toolchain` ?

bhack: As custom the original toolchain was just cloned from TF custom ops repo, now that it is unmaintained, I suppose we need to import all the TF workspaces [like in TF IO](https://github.com/tensorflow/io/blob/master/WORKSPACE#L115-L144) and make some other refactoring to remove our dependency from the custom ops repo infrastructure.

tensorflow_addons/utils/test_utils.py
```diff
@@ -26,7 +26,9 @@
 from tensorflow_addons.utils import resource_loader
 
 # TODO: copy the layer_test implementation in Addons.
-if tf.__version__[:3] > "2.5":
+if tf.__version__[:3] > "2.8":
+    from keras.testing_infra.test_utils import layer_test  # noqa: F401
+elif tf.__version__[:3] > "2.5":
```
fsx950223, line 31: We don't need this condition anymore.

fsx950223, line 31: We don't need this condition anymore.

bhack: Yes probably we have code from different epochs

https://github.com/tensorflow/addons/search?q=packaging.Version&type=

.github/workflows/release.yml
seanpmorgan: Yes, will leave this in the PR though because we don't release until tensorflow-macos has:
https://github.com/tensorflow/addons/issues/2631

tensorflow_addons/utils/test_utils.py
seanpmorgan: Removed unnecessary check

tools/install_so_files.sh
seanpmorgan: GPU tests passing using the remote toolchains

WORKSPACE
```diff
@@ -18,3 +18,28 @@ tf_configure(
 )
 
 cuda_configure(name = "local_config_cuda")
+
+http_archive(
+    name = "org_tensorflow",
+    sha256 = "7d736fa5ff3868516359d8370e7b57251b8080243fc38e8089ee0ceb8ee90264",
+    strip_prefix = "tensorflow-2.9.0-rc0",
```
seanpmorgan, line 25: Need to update once we have a stable release

tools/install_so_files.sh
bhack: So it Is what more or less whatintroduced here in the last days.

 But just for the record:
 now users are implcitly going to download the Tensorflow tarball, llvm tarball etc..

seanpmorgan: Just during build, not for the packaged whl. Totally fine with that given we're loosing ~1900 lines of tech debt

bhack: Yes on build and also to contribute a PR we download many things. This is a SIG-build techincal debt with SIGs that are custom ops "orphan".

configure.py
```diff
@@ -129,8 +129,14 @@ def create_build_configuration():
 
     write("build --spawn_strategy=standalone")
     write("build --strategy=Genrule=standalone")
+    write("build  --experimental_repo_remote_exec")
     write("build -c opt")
 
+    if Version(tf.__version__) >= Version("2.9.0"):
```
bhack, line 135: @fsx950223 @seanpmorgan Do we need to maintain this?

.github/workflows/release.yml
bhack: Still waiting for this

denadai2: @bhack released! :D

seanpmorgan:
@bhack @fsx950223 looks like changes to configure.py break custom op build compatibility. Maybe the last 3 commits or I saw another land in #2700 . I can take a look Sunday if it's not clear

bhack:
> @bhack @fsx950223 looks like changes to configure.py break custom op build compatibility. Maybe the last 3 commits or I saw another land in #2700 . I can take a look Sunday if it's not clear

Are the Tensorflow <2.9 wheels rebuilt upstream on manylinux_2014 and so the have `D_GLIBCXX_USE_CXX11_ABI=1`?

Case if they are still the old ones released they are not using manylinux_2014...

bhack:
@seanpmorgan Yes confirmed we need to conditional build the old container 2.7.2 and 2.8.1.

``` _ZN10tensorflow8internal21CheckOpMessageBuilder9NewStringB5cxx11Ev``` see the missing cxx11 symbol.

bhack:
Let see if ovverriding of the CX11 abi flags is enough for the old versions also if we build on manylinux2014.

bhack:
>  I can take a look Sunday

Green lights here :green_circle: :green_circle: :green_circle: 

It is better to take a sun-day :sun_with_face: :beach_umbrella:

bhack (approved):

seanpmorgan:
merging so our nightly is compatible with 2.9. Will have a full release by EOW

PR #2697, opened 2022-04-20
FC loss - clip focal_weight & set reduction to AUTO

johnypark:
# Description

Brief Description of the PR: 

The current focal loss has a bug that induces a complex number when gamma <1, which may be due to leakage in floating number calculation that results in negative numbers in probability. 

Also, with the default reduction =NONE,  exploding gradient case is observed with an adequate learning rate.

Made the simple following changes:
- Use the keras backend clip function to set minimum focal_weight values with epsilon, which solves the negative number problem.
-  Set reduction = AUTO, which mitigates the exploding gradient problem. 

## Type of change

- [c ] Bug fix

# How Has This Been Tested?

Tested it on a [Colab notebook](https://colab.research.google.com/drive/1tp4e-WtN39HjL9jS7aPulE0Bp_U1kFtX?usp=sharing) with TPU and tensorflow=2.8 and tensorflow_addons=0.16.1.

Please let me know if we need a more exhaustive demonstration, I can provide some more tests on Colab. Allowing edits by maintainers just in case.

### DIFF
--- a/tensorflow_addons/losses/focal_loss.py
+++ b/tensorflow_addons/losses/focal_loss.py
@@ -68,7 +68,7 @@ class SigmoidFocalCrossEntropy(LossFunctionWrapper):
         from_logits: bool = False,
         alpha: FloatTensorLike = 0.25,
         gamma: FloatTensorLike = 2.0,
-        reduction: str = tf.keras.losses.Reduction.NONE,
+        reduction: str = tf.keras.losses.Reduction.AUTO,
         name: str = "sigmoid_focal_crossentropy",
     ):
         super().__init__(
@@ -136,7 +136,8 @@ def sigmoid_focal_crossentropy(
 
     if gamma:
         gamma = tf.cast(gamma, dtype=y_true.dtype)
-        modulating_factor = tf.pow((1.0 - p_t), gamma)
-
+        focal_weight = K.clip((1.0 - p_t), K.epsilon(), 1.0) 
+        modulating_factor = tf.pow(focal_weight, gamma)
+        
     # compute the final loss and return
     return tf.reduce_sum(alpha_factor * modulating_factor * ce, axis=-1)


bhack:
Do you have a test to cover this case?

bhack:
Do you have a test to cover this case?

johnypark:
> Do you have a test to cover this case?

Hi,  I did the test here [Colab notebook](https://colab.research.google.com/drive/1tp4e-WtN39HjL9jS7aPulE0Bp_U1kFtX?usp=sharing) ... the class fixed_SigmoidFocalCrossEntropy includes the two simple changes made to the code and the notebook shows that it works. 

I am new to contributing to tensorflow addons, so not familiar of what is the standard way of doing the test. Please direct me what kinds of test needs to be done to cover this case.

AakashKumarNain:
@bhack what's the plan of incorporating Focal loss from Keras-cv?

bhack:
> @bhack what's the plan of incorporating Focal loss from Keras-cv?
> 

We have a contribuor interested https://github.com/keras-team/keras-cv/issues/319#issuecomment-1104592455

AakashKumarNain:
> > @bhack what's the plan of incorporating Focal loss from Keras-cv?
> 
> We have a contribuor interested [keras-team/keras-cv#319 (comment)](https://github.com/keras-team/keras-cv/issues/319#issuecomment-1104592455)

Let's try to get that done and move this one away slowly

johnypark:
> > > @bhack what's the plan of incorporating Focal loss from Keras-cv?
> > 
> > 
> > We have a contribuor interested [keras-team/keras-cv#319 (comment)](https://github.com/keras-team/keras-cv/issues/319#issuecomment-1104592455)
> 
> Let's try to get that done and move this one away slowly

Can you review and incorporate this changes to addons if you like them, in the meantime? They are just simple two changes in the code.

bhack:
Can you add a test for this uncovered case in https://github.com/tensorflow/addons/blob/master/tensorflow_addons/losses/tests/focal_loss_test.py

PR #2702, opened 2022-05-15
AUTOTUNE 'y_shape'

markub3327:
Hello,

   `y_shape` can be determine automatically, not manually typed by user.

Thanks.

### DIFF
--- a/tensorflow_addons/metrics/r_square.py
+++ b/tensorflow_addons/metrics/r_square.py
@@ -86,13 +86,11 @@ class RSquare(Metric):
         self,
         name: str = "r_square",
         dtype: AcceptableDTypes = None,
-        y_shape: Tuple[int, ...] = (),
         multioutput: str = "uniform_average",
         num_regressors: tf.int32 = 0,
         **kwargs,
     ):
         super().__init__(name=name, dtype=dtype, **kwargs)
-        self.y_shape = y_shape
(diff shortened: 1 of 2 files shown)


bhack:
Can you keep the arg and emit a deprecation if passed? So that we don't brake the API so fast.

We have deprecation examples in the source code

markub3327:
@bhack 

Ok. I added a warning about it.

tensorflow_addons/metrics/r_square.py
```diff
@@ -86,13 +86,17 @@ def __init__(
...
         multioutput: str = "uniform_average",
         num_regressors: tf.int32 = 0,
         **kwargs,
     ):
         super().__init__(name=name, dtype=dtype, **kwargs)
-        self.y_shape = y_shape
+
+        if "y_shape" in kwargs:
+            warnings.warn(
+                "y_shape has been removed, because it's automatically derived.",
```
bhack, line 97: Generally we indicate a release margin:

https://github.com/tensorflow/addons/blob/37a368adfea1d0ec4cb85998946f458e9d565818/tensorflow_addons/optimizers/lamb.py#L83

/cc @seanpmorgan Do you remember how much releases?

bhack (changes requested):

PR #2704, opened 2022-05-18
Fix documentation of adabelief

denadai2:
This PR fixes the documentation of `total_steps`, which enables the warmup only if the value is bigger than zero. However, the documentation says "positive value". Zero (the default value) is positive though

### DIFF
--- a/tensorflow_addons/optimizers/adabelief.py
+++ b/tensorflow_addons/optimizers/adabelief.py
@@ -113,7 +113,7 @@ class AdaBelief(tf.keras.optimizers.Optimizer):
             rectify: boolean. Whether to apply learning rate rectification as
                 from RAdam.
             total_steps: An integer. Total number of training steps. Enable
-                warmup by setting a positive value.
+                warmup by setting a value greater than zero.
             warmup_proportion: A floating point value. The proportion of
                 increasing steps.
             min_lr: A floating point value. Minimum learning rate after warmup.


tensorflow_addons/optimizers/adabelief.py
```diff
@@ -113,7 +113,7 @@ def __init__(
             rectify: boolean. Whether to apply learning rate rectification as
                 from RAdam.
             total_steps: An integer. Total number of training steps. Enable
-                warmup by setting a positive value.
+                warmup by setting a value greater than zero.
```
bhack, line 116: I think that it was correct:
http://oeis.org/A000027
http://oeis.org/A001477

denadai2: nice. However, it requires less knowledge to just say "greater than zero". I did not know that discussion

bhack: There could be some confusion with [signed zero](https://en.wikipedia.org/wiki/Signed_zero) but on integer I think it is hard to be confused. I will merge as it is just an alternative.

denadai2: thanks

juntang-zhuang:
Looks great to me. The same issue is with `rectified_adam` document https://github.com/tensorflow/addons/blob/953d84817a6f8a3973eed69a3505739c6fa5ff97/tensorflow_addons/optimizers/rectified_adam.py#L107 , you might consider modifying it as well.

PR #2706, opened 2022-05-19
Point optimizer to tf.keras.optimizer.legacy.Optimizer to be compatib…

chenmoneygithub:
# Description

Brief Description of the PR:
Keras has made a new version of optimizer, check the link [here](https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/experimental), and in a future TF/Keras release, `tf.keras.optimizers.XXX` will point to the new optimizer, and old optimizers will continue to be supported under `legacy` namespace until further notice. To avoid code breakage, this PR is replacing `tf.keras.optimizer.Optimizer` with `tf.keras.optimizer.legacy.Optimizer`. 

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New Tutorial
- [ ] Additional Testing

# Checklist:

# How Has This Been Tested?

No new test cases are required, just making sure the current test is not broken due to the change.

### DIFF
--- a/tensorflow_addons/optimizers/__init__.py
+++ b/tensorflow_addons/optimizers/__init__.py
@@ -14,7 +14,7 @@
 # ==============================================================================
 """Additional optimizers that conform to Keras API."""
 
-from tensorflow_addons.optimizers.constants import BaseOptimizerClass
+from tensorflow_addons.optimizers.constants import KerasLegacyOptimizer
 from tensorflow_addons.optimizers.average_wrapper import AveragedOptimizerWrapper
 from tensorflow_addons.optimizers.conditional_gradient import ConditionalGradient
 from tensorflow_addons.optimizers.cyclical_learning_rate import CyclicalLearningRate
--- a/tensorflow_addons/optimizers/__init__.py
+++ b/tensorflow_addons/optimizers/__init__.py
@@ -14,6 +14,7 @@
 # ==============================================================================
 """Additional optimizers that conform to Keras API."""
 
+from tensorflow_addons.optimizers.constants import BASE_OPTIMIZER_CLASS
 from tensorflow_addons.optimizers.average_wrapper import AveragedOptimizerWrapper
 from tensorflow_addons.optimizers.conditional_gradient import ConditionalGradient
 from tensorflow_addons.optimizers.cyclical_learning_rate import CyclicalLearningRate
--- a/tensorflow_addons/optimizers/__init__.py
+++ b/tensorflow_addons/optimizers/__init__.py
@@ -14,7 +14,7 @@
 # ==============================================================================
 """Additional optimizers that conform to Keras API."""
 
-from tensorflow_addons.optimizers.constants import BASE_OPTIMIZER_CLASS
+from tensorflow_addons.optimizers.constants import BaseOptimizerClass
 from tensorflow_addons.optimizers.average_wrapper import AveragedOptimizerWrapper
 from tensorflow_addons.optimizers.conditional_gradient import ConditionalGradient
 from tensorflow_addons.optimizers.cyclical_learning_rate import CyclicalLearningRate
--- a/tensorflow_addons/optimizers/adabelief.py
+++ b/tensorflow_addons/optimizers/adabelief.py
@@ -17,12 +17,12 @@
 import tensorflow as tf
 from tensorflow_addons.utils.types import FloatTensorLike
 
-from tensorflow_addons.optimizers import BaseOptimizerClass
+from tensorflow_addons.optimizers import KerasLegacyOptimizer
 from typing import Union, Callable, Dict
 
 
 @tf.keras.utils.register_keras_serializable(package="Addons")
-class AdaBelief(BaseOptimizerClass):
+class AdaBelief(KerasLegacyOptimizer):
     """Variant of the Adam optimizer.
 
     It achieves fast convergence as Adam and generalization comparable to SGD.
--- a/tensorflow_addons/optimizers/adabelief.py
+++ b/tensorflow_addons/optimizers/adabelief.py
@@ -17,12 +17,12 @@
 import tensorflow as tf
 from tensorflow_addons.utils.types import FloatTensorLike
 
-from tensorflow_addons.optimizers import BASE_OPTIMIZER_CLASS
+from tensorflow_addons.optimizers import BaseOptimizerClass
 from typing import Union, Callable, Dict
 
 
 @tf.keras.utils.register_keras_serializable(package="Addons")
-class AdaBelief(BASE_OPTIMIZER_CLASS):
+class AdaBelief(BaseOptimizerClass):
     """Variant of the Adam optimizer.
 
     It achieves fast convergence as Adam and generalization comparable to SGD.
--- a/tensorflow_addons/optimizers/adabelief.py
+++ b/tensorflow_addons/optimizers/adabelief.py
@@ -19,9 +19,14 @@ from tensorflow_addons.utils.types import FloatTensorLike
 
 from typing import Union, Callable, Dict
 
+if tf.__version__[:3] > "2.8":
+    optimizer_class = tf.keras.optimizers.legacy.Optimizer
+else:
+    optimizer_class = tf.keras.optimizers.Optimizer
+
 
 @tf.keras.utils.register_keras_serializable(package="Addons")
-class AdaBelief(tf.keras.optimizers.legacy.Optimizer):
+class AdaBelief(optimizer_class):
     """Variant of the Adam optimizer.
 
     It achieves fast convergence as Adam and generalization comparable to SGD.
--- a/tensorflow_addons/optimizers/adabelief.py
+++ b/tensorflow_addons/optimizers/adabelief.py
@@ -21,7 +21,7 @@ from typing import Union, Callable, Dict
 
 
 @tf.keras.utils.register_keras_serializable(package="Addons")
-class AdaBelief(tf.keras.optimizers.Optimizer):
+class AdaBelief(tf.keras.optimizers.legacy.Optimizer):
     """Variant of the Adam optimizer.
 
     It achieves fast convergence as Adam and generalization comparable to SGD.
--- a/tensorflow_addons/optimizers/adabelief.py
+++ b/tensorflow_addons/optimizers/adabelief.py
@@ -17,16 +17,12 @@
 import tensorflow as tf
 from tensorflow_addons.utils.types import FloatTensorLike
 
+from tensorflow_addons.optimizers import BASE_OPTIMIZER_CLASS
 from typing import Union, Callable, Dict
-
-if tf.__version__[:3] > "2.8":
-    optimizer_class = tf.keras.optimizers.legacy.Optimizer
-else:
-    optimizer_class = tf.keras.optimizers.Optimizer
 
 
 @tf.keras.utils.register_keras_serializable(package="Addons")
-class AdaBelief(optimizer_class):
+class AdaBelief(BASE_OPTIMIZER_CLASS):
     """Variant of the Adam optimizer.
 
     It achieves fast convergence as Adam and generalization comparable to SGD.
--- a/tensorflow_addons/optimizers/average_wrapper.py
+++ b/tensorflow_addons/optimizers/average_wrapper.py
@@ -32,7 +32,12 @@ class AveragedOptimizerWrapper(BASE_OPTIMIZER_CLASS, metaclass=abc.ABCMeta):
         if isinstance(optimizer, str):
             optimizer = tf.keras.optimizers.get(optimizer)
 
-        if tf.__version__[:3] > "2.8":
+        if tf.__version__[:3] <= "2.8":
+            if not isinstance(optimizer, tf.keras.optimizers.Optimizer):
+                raise TypeError(
+                    "optimizer is not an object of tf.keras.optimizers.Optimizer."
+                )
+        else:
             if not isinstance(
                 optimizer,
                 (tf.keras.optimizers.legacy.Optimizer, tf.keras.optimizers.Optimizer),
@@ -40,11 +45,6 @@ class AveragedOptimizerWrapper(BASE_OPTIMIZER_CLASS, metaclass=abc.ABCMeta):
                 raise TypeError(
                     "optimizer is not an object of tf.keras.optimizers.legacy.Optimizer "
                     "or tf.keras.optimizers.Optimizer."
-                )
-        else:
-            if not isinstance(optimizer, tf.keras.optimizers.Optimizer):
-                raise TypeError(
-                    "optimizer is not an object of tf.keras.optimizers.Optimizer."
                 )
 
         self._optimizer = optimizer
--- a/tensorflow_addons/optimizers/average_wrapper.py
+++ b/tensorflow_addons/optimizers/average_wrapper.py
@@ -22,9 +22,13 @@ from tensorflow_addons.utils import types
 from typeguard import typechecked
 
 
-class AveragedOptimizerWrapper(
-    tf.keras.optimizers.legacy.Optimizer, metaclass=abc.ABCMeta
-):
+if tf.__version__[:3] > "2.8":
+    optimizer_class = tf.keras.optimizers.legacy.Optimizer
+else:
+    optimizer_class = tf.keras.optimizers.Optimizer
+
+
+class AveragedOptimizerWrapper(optimizer_class, metaclass=abc.ABCMeta):
     @typechecked
     def __init__(
         self, optimizer: types.Optimizer, name: str = "AverageOptimizer", **kwargs
--- a/tensorflow_addons/optimizers/average_wrapper.py
+++ b/tensorflow_addons/optimizers/average_wrapper.py
@@ -17,18 +17,12 @@ import abc
 import warnings
 
 import tensorflow as tf
+from tensorflow_addons.optimizers import BASE_OPTIMIZER_CLASS
 from tensorflow_addons.utils import types
-
 from typeguard import typechecked
 
 
-if tf.__version__[:3] > "2.8":
-    optimizer_class = tf.keras.optimizers.legacy.Optimizer
-else:
-    optimizer_class = tf.keras.optimizers.Optimizer
-
-
-class AveragedOptimizerWrapper(optimizer_class, metaclass=abc.ABCMeta):
+class AveragedOptimizerWrapper(BASE_OPTIMIZER_CLASS, metaclass=abc.ABCMeta):
     @typechecked
     def __init__(
         self, optimizer: types.Optimizer, name: str = "AverageOptimizer", **kwargs
--- a/tensorflow_addons/optimizers/average_wrapper.py
+++ b/tensorflow_addons/optimizers/average_wrapper.py
@@ -17,12 +17,12 @@ import abc
 import warnings
 
 import tensorflow as tf
-from tensorflow_addons.optimizers import BaseOptimizerClass
+from tensorflow_addons.optimizers import KerasLegacyOptimizer
 from tensorflow_addons.utils import types
 from typeguard import typechecked
 
 
-class AveragedOptimizerWrapper(BaseOptimizerClass, metaclass=abc.ABCMeta):
+class AveragedOptimizerWrapper(KerasLegacyOptimizer, metaclass=abc.ABCMeta):
     @typechecked
     def __init__(
         self, optimizer: types.Optimizer, name: str = "AverageOptimizer", **kwargs
@@ -33,7 +33,7 @@ class AveragedOptimizerWrapper(BaseOptimizerClass, metaclass=abc.ABCMeta):
             optimizer = tf.keras.optimizers.get(optimizer)
 
         if not isinstance(
-            optimizer, (tf.keras.optimizers.Optimizer, BaseOptimizerClass)
+            optimizer, (tf.keras.optimizers.Optimizer, KerasLegacyOptimizer)
         ):
             raise TypeError(
                 "optimizer is not an object of tf.keras.optimizers.Optimizer "
--- a/tensorflow_addons/optimizers/average_wrapper.py
+++ b/tensorflow_addons/optimizers/average_wrapper.py
@@ -32,20 +32,13 @@ class AveragedOptimizerWrapper(BASE_OPTIMIZER_CLASS, metaclass=abc.ABCMeta):
         if isinstance(optimizer, str):
             optimizer = tf.keras.optimizers.get(optimizer)
 
-        if tf.__version__[:3] <= "2.8":
-            if not isinstance(optimizer, tf.keras.optimizers.Optimizer):
-                raise TypeError(
-                    "optimizer is not an object of tf.keras.optimizers.Optimizer."
-                )
-        else:
-            if not isinstance(
-                optimizer,
-                (tf.keras.optimizers.legacy.Optimizer, tf.keras.optimizers.Optimizer),
-            ):
-                raise TypeError(
-                    "optimizer is not an object of tf.keras.optimizers.legacy.Optimizer "
-                    "or tf.keras.optimizers.Optimizer."
-                )
+        if not isinstance(
+            optimizer, (tf.keras.optimizers.Optimizer, BASE_OPTIMIZER_CLASS)
+        ):
+            raise TypeError(
+                "optimizer is not an object of tf.keras.optimizers.Optimizer "
+                "or tf.keras.optimizers.legacy.Optimizer (if you have tf version >= 2.9.0)."
+            )
 
         self._optimizer = optimizer
         self._track_trackable(self._optimizer, "awg_optimizer")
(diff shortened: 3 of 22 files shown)


manzilz:
changes to yogi.py: LGTM

juntang-zhuang:
changes to adabelief.py LGTM!

rehanguha (commented):
tensorflow_addons/optimizers/cocob.py changes LGTM.

bhack:
@chenmoneygithub To understand this a little bit: 
- With this PR all the Addons optimizers are legacy only
- Some of our (now) legacy only optimizers are also reimplemented in Keras experimental (as non legacy?):
  - https://github.com/keras-team/keras/blob/master/keras/optimizers/__init__.py#L37-L45
- Keras have already some v2 ready optimizers:
  - https://github.com/keras-team/keras/blob/master/keras/optimizers/__init__.py#L26-L33 

/cc @seanpmorgan

chenmoneygithub:
@bhack Thanks for your questions! Re your questions:

- Yes, after this PR addon optimizers will be based on `tf.keras.optimizers.legacy.Optimizer`, which is an alias to current `tf.keras.optimizers.Optimizer`. We will later check how to make them compatible with the new optimizer, which would need some refactor work.
- Do you mean AdamW? We are supporting AdamW in Keras now because it seems to be very popular.
- V2 optimizer is what people use now. I apologize the the misleading name, but basically V2 optimizers were made at the same time as TF2, so V2 means it supports TF2. The experimental optimizer is made to replace all those V2 optimizers. 

Hope this would answer your question!

bhack:
Yes, thanks for the confirmation.

> We will later check how to make them compatible with the new optimizer, which would need some refactor work.

I have some doubt on this point as it really depends on the Keras, Keras-CV, Keras-NLP roadmap and how that will overlap existing "legacy" components here. 

If a refactor it is needed probably it is better to evaluate if the single optimizer will be absorbed, with the new inheritance, in any of these Keras-* repos.

chenmoneygithub:
@bhack Yes, I have the exactly the same thought - new optimizer-based implementation should be done in Keras repo rather than doing in-place code changes to addon.optimizers.

bhack:
Ok. 
For this PR you need to handle a bit all the conditionals as we are going to support and produce wheels for the last 3 TF releases.

chenmoneygithub:
Got it, will reflect it in the code.

seanpmorgan:
Hi @chenmoneygithub , thanks for the PR and keeping us in the loop. As you can see from our CI this appears to be a significant breaking change.  TF Addons supports 3 TensorFlow/Keras versions during each release, so implementing this change would require some work arounds on our end. I'm wondering if this is implemented as intended on the Keras side? 

Has legacy.optimizer been available / has the new optimizer been part of experimental for a few releases?  Typically Keras is quite good at backwards compatibility. 

cc @fchollet

seanpmorgan:
I didn't see an RFC in keras governance or anything on this, so I suspect more downstream consumers (other than TFA) will be surprised.

bhack:
Also isn't not explained what these legacy optimizers are in the official documentation:
https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/legacy

chenmoneygithub:
@seanpmorgan yes, your concern is very reasonable. 

For some background context: In this release (TF/Keras 2.9), our goal is to push out the experimental optimizer, mainly for internal adoptions. Making experimental optimizer the default one will happen in a later release (2.10 or 2.11), and old optimizer deprecation would happen in a much later release (we have not decided which one yet!), but we made this legacy optimizer, which is a mirror of `tf.keras.optimizers.XXX` to prepare for the future deprecation. 

For the question about breakage: the actions failed due to symbol missing, not functionality. I added the version control, which means only use `legacy` namespace for version after 2.8. We don't want to generate workloads for OSS contributors, so we  prefer to handle the implementation based on our side. 

@bhack Yea, this is something we will fix. The page should be more clear.

chenmoneygithub:
@bhack Thanks!

Yes, we will make public notice before taking migration/deprecation actions, now we are focusing on internal optimizer adoptions.

Yea at this moment it's hard to edit the page.

bhack:
> Yea at this moment it's hard to edit the page.

What do you mean? I was mentioning about our dependency on a stable TF release in master instead of TF nightly.

chenmoneygithub:
@bhack Sorry I misread your previous comment, I thought you mean the documentation of `legacy.optimizer`.

Re you previous comment - I don't think cherrypicking is required? I am saying there are two options here:
1. addons contributors change the code to implement optimizers based on `tf.keras.optimizers.experimental.XXX`.
2. We move addons optimizers into main Keras by reimplementing them based on `tf.keras.optimizers.experimental.XXX`. e.g., `tf.keras.optimizers.lamb`.

1. is actually the better option, but it puts on workloads on contributors.

bhack:
> I added the version control.

I was talking about this. Can you point me on this version control?

bhack:
> My previous push failed due to some merge conflict, I will ping you once I get the latest code uploaded.
Ok thanks.

More in general:
> We move addons optimizers into main Keras by reimplementing them based on tf.keras.optimizers.experimental.XXX. e.g., tf.keras.optimizers.lamb.

It could be ok for us but you need to consider:

1. A little bit of deprecation margin https://github.com/tensorflow/addons#periodic-evaluation-of-subpackages
2. I suppose you will eventually fix the Keras.io repository  yourself for TFA depending tutorial/notebooks in case of breaking optimizers API
3. We are a small project but we have accumulated over time a little bit of downstream dependencies:
immagine

bhack:
Ok, now I see that we are back [on the original idea](https://github.com/tensorflow/addons/pull/2706#issuecomment-1132077769) to conditional control the namespace downstream (in TFA directly).

seanpmorgan:
> Ok, now I see that we are back [on the original idea](https://github.com/tensorflow/addons/pull/2706#issuecomment-1132077769) to conditional control the namespace downstream (in TFA directly).

Yes, this is what I was afraid of. This PR is sub-optimal because it makes downstream consumers inherit the debt of a quickly implemented public api change to Keras.

@chenmoneygithub @fchollet is there anywhere publicly facing from the Keras perspective where we can discuss this? TFA is not the only one impacted by this.

cc @joanafilipa

chenmoneygithub:
@bhack I just updated the code with the version control part.

Yes, I am aware of the large user group of tf addons, and never want to break their workflows. On Keras it remains unclear to us that when we can completely remove the code of the old optimizer, personally I feel it is a hard task. If you are curious on why we are making the new optimizer even though we are not removing the old code in the short term, briefly we feel the old optimizer's logic and layout are bad, so we replace things like slot variables and fused ops with more understandable code, and split out the distributed training part to a separate class.

bhack:
> Yes, I am aware of the large user group of tf addons, and never want to break their workflows. On Keras it remains unclear to us that when we can completely remove the code of the old optimizer, personally I feel it is a hard task. If you are curious on why we are making the new optimizer even though we are not removing the old code in the short term, briefly we feel the old optimizer's logic and layout are bad, so we replace things like slot variables and fused ops with more understandable code, and split out the distributed training part to a separate class.

Personally I agree with Sean's vision. 
As these are public symbols  it seems to me a little bit too late to present an RFC in Keras/governance or in TF/community only when effectively we want to remove these symbols (than at the RFC time they will be already `legacy`). 

For a top namespace change like `tf.keras.*` I suppose that an early RFC is required to involve the community, SIGs, and more in general downstream projects.

This time it went like this, probably for a good reasons (I don't know) but I would not feel like promoting this process in general for our community / ecosystem.

chenmoneygithub:
We discussed about RFC when we kicked off the project, and did not go for it because main changes happen on internal logic, while the public API mostly remains identical. 

Our rollout journey would be hidden for main users, basically the `tf.keras.optimizers` namespace will point to `tf.keras.optimizers.experimental` in a future release (this info is available in the release note of TF 2.9), and users will switch to new optimizer silently. In the window of 2.9, we will mainly work with internal teams to fix unseen errors during testing, and later on we will publish a public notice for required actions. Addons optimizer is the most special one, it is written based on the Keras optimizer, and itself has a large user group, so I am prioritizing this action now.

bhack:
I could understand this only eventually for end-users but not for derived/ecosystem projects. 
Do you meant that using Keras public API for creating derived objects in a third party library and in this case a TF ecosystem library is not a supported use case anymore?

I don't think that TFA is a special case for derived projects.

chenmoneygithub:
@seanpmorgan Sorry for the inconvenience, but I don't think this is something that can be avoided? Migration/deprecation changes usually require actions from the downstream system. Specifically for this optimizer rework, the current action is to point the symbols to a safer route, i.e., `tf.keras.optimizers.legacy.xxx`.

@bhack That's not what I meant. My point is on our side, we want to do the right things, e.g., the old optimizer was a bit in a chaos and we have to clean it up. For downstream libraries, we make sure it is not broken in any release, and provide guidance on how actions should be made. I am saying TFA is special because each optimizer has its own owner right? In this case it is hard to let all of them take actions, so we are closely monitoring how we should handle TFA optimizers. IIUC, you are asking for an RFC on this optimizer rework. There were lots of experimental things we did within this work (mostly within internal logic), which iterate fast once the previous trial failed, so we felt that an RFC won't actually affect the work. 

But I agree with your concern! I think the right thing to do here is we should make our big moves visible to the public, either via an RFC for design concerns, or a general FYI for future required actions.

szutenberg:
I strongly support the views of @bhack and @seanpmorgan . We should avoid introducing breaking changes, it will also affect many other projects... 

@chenmoneygithub Did you think about introducing `keras.optimizers.OptimizerV3` for the new class? This PR wouldn't be required.

Small off-topic about changes in Keras (IIUC there's no other, public discussion about it):

I had a look at the [experimental Adam optimizer](https://github.com/keras-team/keras/blob/master/keras/optimizers/optimizer_experimental/adam.py) and I don't understand what's the benefit of not using fused kernel [ResourceApplyAdam](https://www.tensorflow.org/api_docs/python/tf/raw_ops/ResourceApplyAdam). Do you plan to introduce similar changes to layers.BatchNormalization as well?

From my perspective, it's related to https://github.com/keras-team/keras/pull/16537 . Here it was also argued that using `tf.reduce_sum(tf.square(x))` is more understandable than `2.0 * tf.nn.l2_loss(x)` but the second option is more end-user friendly thanks to better performance and lower memory peak. Was performance and device memory peak measured on different accelerators when proposing this change?

chenmoneygithub:
@szutenberg Thanks for the comment!

I don't think breaking changes should be completely avoided. Breaking changes without a time buffer for actions are bad and annoying. But this optimizer rework leaves a time period for taking actions, so nothing is broken right now. In this case, this is like a tradeoff between benefits and eng efforts.

I don't get why this PR is not required if we make OptimizerV3 (actually I believe what we are doing is exactly what you mean by OptimizerV3, just name diff). Eventually if we need to have one single source of truth for optimizer, we have to make this PR. 

Fused kernel is fast but opaque. For users customizing their optimizer, fused ops is not very much readable. We believe that the source code should be readable and as-it-is for our users. We use XLA + python code to achieve a similar performance as fused ops.

bhack:
> I am saying TFA is special because each optimizer has its own owner right? 

But there is a SIG lead, SIG maintainers, code owners etc.  Who do you think is going to fix a bug when a codeowner is no longer available? 

A SIG is guaranteed, best effort, by its lead and maintainers.

> Migration/deprecation changes usually require actions from the downstream system. Specifically for this optimizer rework, the current action is to point the symbols to a safer route, i.e., tf.keras.optimizers.legacy.xxx.

This is why generally we discuss with partners and derived projects with open RFCs also to eventually plan the downstream work that it is required.
 
If you are convinced that it is the right way to handle the process I am not here to insist on the opposite as I don't think that at the end the RFC process is mandatory for anyone (/cc @joanafilipa).

Anyway, thanks for the PR, it took away some work for us. PIng us when you see the green light in the CI jobs so we will merge it.

chenmoneygithub:
@bhack Sorry I am not quite familiar with how addons works, I assumed it is the code owner that handles all those issues. My bad!

For RFC, my opinion is it's case by case. For changes causing significant user journey changes we definitely do a RFC. This optimizer rework is sitting a bit in the grey zone - on one side it imports new symbols, on the other side it only changes user journey in a very limited manner. 

No problem, as said before we want to do the right things, and keeping TFA optimizers unbroken is definitely something we will make sure.

szutenberg:
@chenmoneygithub This PR wouldn't be required if API of keras.optimizers.Optimizer remains unchanged :wink: 

I believe that it's better to write a good tutorial for writing custom optimizers than to write a clean but a bit slower code for the most popular optimizers. Performance matters for users. If it's even 1% difference in training step duration then they may pay 1% more for renting GPU only thanks to the clean code which they won't ever look at. Is it worth?

> Fused kernel is fast but opaque.

Some time ago I rewrote the equation in [tf.raw_ops.ResourceApplyAdam docs](https://www.tensorflow.org/api_docs/python/tf/raw_ops/ResourceApplyAdam). Please let me know if something is still not clear for users.

chenmoneygithub:
@szutenberg Thanks!

We are not changing `tf.keras.optimizers.Optimizer` now, the `legacy` is a mirror we make for future migration. `tf.keras.optimizers.Optimizer` is a path not the actual class, which represents our default optimizer. It's not a good practice to keep two sources of truth, so `tf.keras.optimizers.Optimizer` should always point to the symbol we want users to use. 

Not really, the actual time elapse for a new optimizer step is less than the previous optimizer, or say the new optimizer has a better performance indeed. Another thing to notice is optimizer time only contributes to a small portion of the whole execution time, so 1% increase or drop in optimizer performance is negligible to end users. 

thanks for the documentation, it looks nice! our concern is not on the documentation though, but instead on the fact that users need to read c++ code to get the internal mechanism, while it can be written in python ops. Another thing is from the documentation or optimizer code, it's hard to find the c++ code, that is why I said it is opaque.

bhack:
@szutenberg It is a little bit off topic but many of the canned Optimizers are going to bit JIT compiled by default with XLA:

https://github.com/keras-team/keras/commit/d730bca0f772f2ec8d0940465b3f99ec8fc1e019

So performance/optimizations will rely by default on the compiler maturity/performance for a given device on a give sequence of TF ops used to compose the optimizer.

chenmoneygithub:
@bhack Hi, I am still hitting testing failures, but from the [report](https://github.com/tensorflow/addons/runs/6615218648?check_suite_focus=true) I cannot locate the error. I am suspecting it is the last line 
```
ERROR tensorflow_addons/optimizers/tests/standard_test.py - AttributeError: m...
```
But the error message is incomplete, so before trying to reproduce the error locally with the exactly same environment (which can take some time...), I want to check if you have any suggestion on this kind of errors? Thanks!

chenmoneygithub:
Thanks! All tests have passed, please take a review, thanks again!

tensorflow_addons/optimizers/adabelief.py
```diff
@@ -19,9 +19,14 @@
 
 from typing import Union, Callable, Dict
 
+if tf.__version__[:3] > "2.8":
```
bhack, line 22: I'have some doubt to repeat this block in every optimizer file 

@seanpmorgan What do you think? Can we define this as `optimizer_legacy` in the optimizers init.py?

szutenberg: I think we should avoid duplicating this code.

Please note that the condition is wrong because it fails for TF 2.10:
```
>>> tf.__version__
'2.10.0-dev20220531'
>>> tf.__version__[:3]
'2.1'
>>> tf.__version__[:3] > "2.8"
False
```

chenmoneygithub:
@bhack Sure thing, to clarify, you are suggesting having a global variable `base_optimizer_class` in `__init__.py`, which points to legacy.optimizer after 2.9 and default optimizer before 2.8? 

@szutenberg Thanks for catching it!

bhack:
Yes

tensorflow_addons/optimizers/__init__.py
```diff
@@ -14,6 +14,7 @@
 # ==============================================================================
 """Additional optimizers that conform to Keras API."""
 
+from tensorflow_addons.optimizers.constants import BASE_OPTIMIZER_CLASS
```
szutenberg, line 17: this breaks alphabetic order

tensorflow_addons/optimizers/average_wrapper.py
```diff
@@ -32,10 +32,20 @@ def __init__(
         if isinstance(optimizer, str):
             optimizer = tf.keras.optimizers.get(optimizer)
 
-        if not isinstance(optimizer, tf.keras.optimizers.Optimizer):
-            raise TypeError(
-                "optimizer is not an object of tf.keras.optimizers.Optimizer"
-            )
+        if tf.__version__[:3] <= "2.8":
```
szutenberg, line 35: why do we need this condition? What if we would write `(tf.keras.optimizers.Optimizer, BASE_OPTIMIZER_CLASS)`?

tensorflow_addons/optimizers/constants.py
```diff
@@ -0,0 +1,5 @@
+import tensorflow as tf
```
szutenberg, line 1: missing copyright header

tensorflow_addons/optimizers/constants.py
```diff
@@ -0,0 +1,5 @@
...
+
+BASE_OPTIMIZER_CLASS = tf.keras.optimizers.legacy.Optimizer
```
szutenberg, line 3: Why not if ... else? It will crash if tf.keras.optimizers.legacy doesn't exist (TF 2.8), right?

Are you sure it has to be CAPITAL_LETTER?

I'd rename it to KerasLegacyOptimizer (or KERAS_LEGACY_OPTIMIZER - I'm not an expert in conventions :smile: ) and rename the file to keras.py.

tensorflow_addons/optimizers/constants.py
```diff
@@ -0,0 +1,5 @@
...
+if tf.__version__[:3] <= "2.8":
```
szutenberg, line 4: As I wrote previously:
```
>>> tf.__version__
'2.10.0-dev20220531'
>>> tf.__version__[:3]
'2.1'
>>> tf.__version__[:3] > "2.8"
False
```

Please fix this condition (also in other files) or just check if `tf.keras.optimizers.legacy` exists, maybe code would be cleaner then...

tensorflow_addons/optimizers/lookahead.py
```diff
@@ -16,11 +16,12 @@
 import tensorflow as tf
 from tensorflow_addons.utils import types
```
szutenberg, line 17: isn't required to update types.Optimizer?

tensorflow_addons/optimizers/conditional_gradient.py
```diff
@@ -17,12 +17,13 @@
 import tensorflow as tf
 from tensorflow_addons.utils.types import FloatTensorLike
 
+from tensorflow_addons.optimizers import BASE_OPTIMIZER_CLASS
```
szutenberg, line 20: this line should be added after line 17

szutenberg (changes requested):

tensorflow_addons/optimizers/__init__.py
chenmoneygithub: This order matters, if this line does not go before other optimizers it creates a cyclic importing.

tensorflow_addons/optimizers/average_wrapper.py
chenmoneygithub: I did this for the error message to be accurate. but rethinking about it, I am doing (tf.keras.optimizers.Optimizer, BASE_OPTIMIZER_CLASS) check and imply that after 2.9 you should expect tf.keras.optimizers.legacy.Optimizer.

tensorflow_addons/optimizers/constants.py
chenmoneygithub: done

tensorflow_addons/optimizers/constants.py
chenmoneygithub: Thanks! I changed the check condition.

For the naming, yea, we probably want to do camel case since it is a class itself. But I don't want to imply Legacy in the name, which is not yet 100% correct - we still use `tf.keras.optimizers.Optimizer` in many places. I am renaming to `BaseOptimizerClass`, wdyt?

tensorflow_addons/optimizers/constants.py
chenmoneygithub: good catch! done

tensorflow_addons/optimizers/lookahead.py
chenmoneygithub: good catch! done

tensorflow_addons/optimizers/tests/weight_decay_optimizers_test.py
```diff
@@ -401,13 +401,17 @@ def test_var_list_with_exclude_list_sgdw(dtype):
     )
 
 
+if tf.__version__[:3] > "2.8":
```
fsx950223, line 404: Use ```from packaging.version import Version```

tensorflow_addons/optimizers/lazy_adam.py
```diff
@@ -27,8 +27,14 @@
 from typing import Union, Callable
 
 
+if tf.__version__[:3] > "2.8":
```
fsx950223, line 30: Same as above

tensorflow_addons/optimizers/weight_decay_optimizers.py
```diff
@@ -261,10 +261,17 @@ def _do_use_weight_decay(self, var):
         return var.ref() in self._decay_var_list
 
 
+optimizer_class = Union[
```
fsx950223, line 264: Why not BaseOptimizerClass?

tensorflow_addons/utils/types.py
```diff
@@ -46,7 +47,12 @@
 Regularizer = Union[None, dict, str, Callable, tf.keras.regularizers.Regularizer]
 Constraint = Union[None, dict, str, Callable, tf.keras.constraints.Constraint]
 Activation = Union[None, str, Callable]
-Optimizer = Union[tf.keras.optimizers.Optimizer, str]
+if importlib.util.find_spec("tensorflow.keras.optimizers.legacy") is not None:
+    Optimizer = Union[
+        tf.keras.optimizers.Optimizer, tf.keras.optimizers.legacy.Optimizer, str
```
fsx950223, line 52: Same as above

tensorflow_addons/optimizers/constants.py
szutenberg: @chenmoneygithub - I expect that after refactor in keras is done, we will be gradually switching back to `tf.keras.optimizers.Optimizer`. `BaseOptimizerClass` will be confusing then because it'll be name for the legacy class.

chenmoneygithub:
Hi there, sorry for not updating this for a while. I have been terribly sick for the last a few days, will update this PR when I can work. Thanks!

tensorflow_addons/optimizers/constants.py
chenmoneygithub: sg, renamed to KerasLegacyOptimizer.

tensorflow_addons/optimizers/tests/weight_decay_optimizers_test.py
chenmoneygithub: thanks! I am switching to check if "optimizers.legacy" exists to keep consistency.

tensorflow_addons/optimizers/weight_decay_optimizers.py
chenmoneygithub: I am renaming this to keras_legacy_optimizer to keep aligned with changes suggested by Michal.

tensorflow_addons/utils/types.py
chenmoneygithub: This I am using the name "Optimizer" to keep it unchanged.

fsx950223:
Please check ci/cd.

chenmoneygithub:
@fsx950223 I checked the failure log, and it seems to ask for type annotation from source Keras optimizer code not changed I made to addons. Can I ask what should we do here?

chenmoneygithub:
Hi maintainers, the current error message is:
```
'keras.optimizers.optimizer_v2.optimizer_v2.Optimizer.__init__' has not complete type annotations in its signature (it's missing the type hint for 'name'). We would like this functions to be typed.If you are not familiar with adding type hints in functions, you can look at functions already typed inthe codebase. 
```

Which does not look like a new issue imported by this PR. Could anyone help check and give some suggestions on how we should handle it? https://github.com/tensorflow/addons/runs/6836413149?check_suite_focus=true

I would like to have this PR submitted soon to avoid potential breakage of addons.optimizers in the next TF release.

bhack:
@chenmoneygithub https://github.com/tensorflow/addons/blob/master/CONTRIBUTING.md#about-type-hints

chenmoneygithub:
@bhack Thanks! But the error is pointing to main Keras repo rather than some classes I am changing in addons repo. Is there any way to exempt the check?

bhack:
If can add an exclusion at:

https://github.com/tensorflow/addons/blob/master/tools/testing/source_code_test.py#L41

fsx950223 (approved):

PR #2714, opened 2022-05-26
Increment MultiOptimizer iterations after each apply_gradients

lucasdavid:
MultiOptimizer does not call `super().apply_gradients`, and therefore does not update its `iterations: tf.Variable<type=int, shape=[]>` as the other optimizers in Tensorflow.
This leads to buggy behavior when training with a more complex `train_step` that relies on this variable for scheduling *learning rate* or other hyper-parameters such as *increasing/decreasing weights* for individual losses.

### DIFF
--- a/tensorflow_addons/optimizers/discriminative_layer_training.py
+++ b/tensorflow_addons/optimizers/discriminative_layer_training.py
@@ -19,6 +19,9 @@ from typing import List, Union
 import tensorflow as tf
 from typeguard import typechecked
 
+from keras import backend
+from keras.utils import tf_utils
+
 
 @tf.keras.utils.register_keras_serializable(package="Addons")
 class MultiOptimizer(tf.keras.optimizers.Optimizer):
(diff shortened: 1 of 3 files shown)


bhack:
Do you have a test to cover this use case?

lucasdavid:
I can write it. Do I post it in the comments or as a new unit test?

Unrelated to my problem: the testcase `tensorflow_addons/optimizers/tests/discriminative_layer_training_test:test_model` does not have any checkings, and is just a sanity test.

lucasdavid:
The failing check is telling me `using tf.control_dependencies() is not allowed in the codebase.`

I'm not sure if removing it in this scenario is a good idea though.

lucasdavid:
Another way to fix this (and maybe cleaner) would be to just defer to the inner optimizer's iterations:
```
class MultiOptimizer(...):
  @property
  def iterations(self):
    return self.specs[0]['optimizer'].iterations
```

AakashKumarNain:
@bhack I will look into this today or tomorrow if I get some time

fsx950223:
Add the file to the list https://github.com/tensorflow/addons/blob/ee5798c189f3903463b21f0d7189f480c47c365a/tools/testing/source_code_test.py#L177

lucasdavid:
@fsx950223 done. All checks pass now. Thank you. :-)

fsx950223 (approved):

PR #2744, opened 2022-08-07
Build Against TF 2.10

seanpmorgan:
# Description

- [ ] Need to update windows builds
- [ ] Waiting for arm macos build

Brief Description of the PR:


bhack:
@nicolaspi Can you check https://github.com/tensorflow/addons/runs/7999702173?check_suite_focus=true#step:6:5205

nicolaspi:
@bhack I am on it.

nicolaspi:
@bhack Please check https://github.com/tensorflow/addons/pull/2751. It should fix it.

bhack:
> @bhack Please check #2751. It should fix it.

Thanks, let's see the new CI run..

bhack:
@seanpmorgan An extra check was introduced 3 months ago in TF https://github.com/tensorflow/tensorflow/commit/6c6fe3d2f55d90ef59447c378ff7970559a5d3a6#diff-b0331d3a83e4697c835bec57bc002586de8a3f93b520cad43edf07218dc0ea0eR964-R973

So we need to adapt https://github.com/tensorflow/addons/blob/master/tensorflow_addons/utils/test_utils.py#L50-L65

configure.py
```diff
@@ -127,6 +136,9 @@ def create_build_configuration():
     write_action_env("TF_SHARED_LIBRARY_NAME", get_shared_lib_name())
     write_action_env("TF_CXX11_ABI_FLAG", tf.sysconfig.CXX11_ABI_FLAG)
 
+    # This should be replaced with a call to tf.sysconfig if it's added
```
bhack, line 139: https://github.com/tensorflow/tensorflow/pull/57468

bhack: Merged but I don't it will be available with the release

tensorflow_addons/utils/test_utils.py
```diff
@@ -48,35 +48,35 @@ def is_gpu_available():
...
+#         tf.config.set_logical_device_configuration(gpus[0], virtual_gpus)
+#         logical_gpus = tf.config.list_logical_devices("GPU")
+#         print(len(gpus), "Physical GPUs,", len(logical_gpus), "Logical GPUs")
+#     except RuntimeError as err:
+#         # Virtual devices must be set before GPUs have been initialized
+#         print(err)
+#
+#     configs = tf.config.get_logical_device_configuration(gpus[0])
+#     assert len(configs) == 2
+#     print(configs)
```
bhack, line 68: @seanpmorgan I think this is risky If we don't understand the impact on the runner... Probably it is better to check if we could only limit the visible device on the first GPU:

`tf.config.set_visible_devices`

.github/workflows/release.yml
```diff
@@ -136,7 +136,7 @@ jobs:
     needs: [release-wheel, test-with-bazel]
     runs-on: ubuntu-18.04
     env:
-      PY_VERSION: '3.9'
+      PY_VERSION: '3.10.rc3'
```
bhack, line 139: What python version we have here?

tensorflow_addons/utils/test_utils.py
bhack: I've tried to limit the visibility to the first device. Let's see the CI

bhack: Ok It worked.

bhack:
:green_circle:  Green lights here, waiting for the Macos delayed release:
https://pypi.org/project/tensorflow-macos/#history

lgeiger:
> 🟢 Green-light here, waiting for the MacOS delayed release:

Looks like 2.10.0 was just released yesterday: https://pypi.org/project/tensorflow-macos

bhack:
I don't the story about arm and non arm MacOS diversification we had.
Can we align all to MacOS 12? Al the new tensorflow-macos wheels are 12 only.

seanpmorgan:
> I don't the story about arm and non arm MacOS diversification we had. Can we align all to MacOS 12? Al the new tensorflow-macos wheels are 12 only.

The tensorflow wheels are still supporting 10_14:
https://pypi.org/project/tensorflow/2.10.0/#files

bhack:
This policy is confusing me a bit. 
What are the differences between TF MacOS  and Tensorflow-macos for X64? Are they having the same feature?

As tensorflow-macos is MacOS 12 only.

seanpmorgan:
> 



> This policy is confusing me a bit. What are the differences between TF MacOS and Tensorflow-macos for X64? Do they have the same features?
> 
> As tensorflow-macos is MacOS 12 only for TF 2.10

Same. tensorflow-macos just started adding x86 wheels on TF2.9. I'm not sure what optimization they provide for those chips or why you would install that package instead of TF. For now I'm just going to keep our macos_x86_64 built with macOS 10 since TF is still providing those wheels and I don't have justification for telling users to force an OS upgrade and switch to tensorflow-macos for non-ARM chipsets.

Hopefully they align on what's supported.

lgeiger:
> For now I'm just going to keep our macos_x86_64 built with macOS 10 since TF is still providing those wheels and I don't have justification for telling users to force an OS upgrade and switch to tensorflow-macos for non-ARM chipsets.

Sounds great.
Just for reference since I am not sure how long GitHub will still support 10.15 macOS images: it is possible to build backwards compatible wheels on the latest macOS by setting `--copt=-mmacosx-version-min=10.14 --linkopt=-mmacosx-version-min=10.14` as bazel flags or via the `MACOSX_DEPLOYMENT_TARGET` env variable.

bhack:
It would be nice if @jhavukainen or @kulinseth could clarify on this.

seanpmorgan:
> > For now I'm just going to keep our macos_x86_64 built with macOS 10 since TF is still providing those wheels and I don't have justification for telling users to force an OS upgrade and switch to tensorflow-macos for non-ARM chipsets.
> 
> Sounds great. Just for reference since I am not sure how long GitHub will still support 10.15 macOS images: it is possible to build backwards compatible wheels on the latest macOS by setting `--copt=-mmacosx-version-min=10.14 --linkopt=-mmacosx-version-min=10.14` as bazel flags or via the `MACOSX_DEPLOYMENT_TARGET` env variable.

Ack, it appears to be imminent:
https://github.com/actions/runner-images/issues/5583

Will upgrade the build OS, since the compilation settings are still here:
https://github.com/tensorflow/addons/blob/master/.github/workflows/make_wheel_macOS_x86.sh#L19-L20

.github/workflows/make_wheel_macOS_x86.sh
```diff
@@ -16,8 +16,8 @@ bazel clean
 python configure.py
 
 bazel build \
-  --copt -mmacosx-version-min=10.13 \
-  --linkopt -mmacosx-version-min=10.13 \
+  --copt -mmacosx-version-min=10.14 \
+  --linkopt -mmacosx-version-min=10.14 \
```
lgeiger, line 20: I think there might be a `=` missing here. Just wondering why the platform name [currently is `10.15`](https://github.com/tensorflow/addons/pull/2744/files#diff-c0354faba8e95ccf3419158e87d1d2f027b38a82d4bf88b14305e9f9f69d077dL83) which makes me think that this option might be silently ignored.
```suggestion
  --copt=-mmacosx-version-min=10.14 \
  --linkopt=-mmacosx-version-min=10.14 \
```

seanpmorgan: I believe we were building 10.13 and labeling 10.15 with delocate

bhack (approved):

.github/workflows/make_wheel_macOS_x86.sh
kulinseth: > It would be nice if @jhavukainen or @kulinseth could clarify on this.
@bhack , what's the question ?

bhack: https://github.com/tensorflow/addons/pull/2744#issuecomment-1252165641

bhack: @kulinseth see the link above

PR #2749, opened 2022-08-21
Fix wheel building for macos

nicolaspi:
The rpaths in the `custom_ops` libraries are invalidated once packaged into the wheel, this prevents delocate from retrieving the tensorflow dependency. This PR solves the issue by setting the env variable `DYLD_LIBRARY_PATH`.

`delocate==0.9.1` was failing silently with older version of `setuptools` (the packaged libraries were not processed). Upgrading to the lastest version solves the issue.

This fixes the current CI issue.


bhack (approved):

seanpmorgan:
Thanks @nicolaspi !

PR #2759, opened 2022-09-21
Update Compat Matrix

seanpmorgan:


fsx950223:
Please update bazel version too.

seanpmorgan:
We don't currently track bazel version so I would prefer we merge this so our compatibility info is up to date. If someone would like to go back through time and record the bazel versions I'm fine with that, but I'm not sure I see a great benefit to doing that.

PR #2782, opened 2022-11-22
Build with TF 2.11.0

bhack:
Build the release with TF 2.11.0


bhack:
@chenmoneygithub  Can you check the optimizer 2.11 errors?

chenmoneygithub:
@bhack I made #2783 to fix optimizer failures. The error log is too long to read, so I might miss some cases.

bhack:
> We need to remove `GroupNormalization` layer as it has been incorporated in the core now

Yes but I think that we need to give the same deprecation warming. Can you send a PR against `build_2.11.0` branch?

AakashKumarNain:
> > We need to remove `GroupNormalization` layer as it has been incorporated in the core now
> 
> Yes but I think that we need to give the same deprecation warning like other "duplicated" components. Can you send a PR against `build_2.11.0` branch?

I can do it over the weekend. Got a lot on my plate right now. Will that work?

bhack:
> I can do it over the weekend. Got a lot on my plate right now. Will that work?

Yes it is ok as we are waiting also for the Macos release https://pypi.org/project/tensorflow-macos/#history

.github/workflows/release.yml
```diff
@@ -101,20 +101,20 @@ jobs:
       matrix:
         os: ['macOS', 'Windows', 'Linux']
         py-version: ['3.7', '3.8', '3.9', '3.10']
-        tf-version: ['2.10.0']
+        tf-version: ['2.9.3', '2.10.1', '2.11.0']
```
seanpmorgan, line 104: We only publish wheels that were built with the most recent TF version.

bhack: Yes we had some drift with https://github.com/tensorflow/addons/blob/master/tools/update_release_version.sh.
I think that we could fix that utility that I've contributed as it is still useful to quickly prepare a release version PR for the "repetitive" tasks.

apivovarov:
TF 2.11 still not supported because `EXCLUSIVE_MAX_TF_VERSION = "2.11.0"` in https://github.com/tensorflow/addons/blob/master/tensorflow_addons/version.py

PR #2816, opened 2023-02-28
Fix Test Suite

seanpmorgan:
# Description

Test suite has several uses of np.bool and np.int which are no longer in the most recent version of numpy.

### DIFF
--- a/build_deps/toolchains/gpu/crosstool/BUILD.tpl
+++ b/build_deps/toolchains/gpu/crosstool/BUILD.tpl
@@ -8,12 +8,12 @@ load(":cc_toolchain_config.bzl", "cc_toolchain_config")
 toolchain(
     name = "toolchain-linux-x86_64",
     exec_compatible_with = [
-        "@bazel_tools//platforms:linux",
-        "@bazel_tools//platforms:x86_64",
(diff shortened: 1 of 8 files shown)


seanpmorgan:
@MarkDaoust when time allows can you take a look at the failure for doc build:
`ImportError: cannot import name 'builder' from 'google.protobuf.internal' (/usr/local/lib/python3.9/site-packages/google/protobuf/internal/__init__.py)`

Seems like protobuf changes recently.

bhack:
@seanpmorgan https://github.com/protocolbuffers/protobuf/issues/9778#issuecomment-1097524415

MarkDaoust:
Yeah, sorry about that. 

TF 2.12 is switching to the new version of protobuf. 

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/pip_package/setup.py#L100

I updated the docs tools a few weeks ago. 

You're going to need to follow for the 2.12 compatible release.

But if you do want this to just pass you could point the requirement to the commit just before the update: `git+https://github.com/tensorflow/docs@c1510dc5ad73aa829249822798e6263360b0f62a`

PR #2829, opened 2023-04-11
Fix Keras imports.

reedwm:
# Description

Brief Description of the PR:

Recently, Keras changed how you must import internal symbols. Now, you must import private symbols from `keras.src` instead of `keras`, e.g. you must use `from keras.src.utils import tf_utils` instead of `from keras.utils import tf_utils`. See slide 12 of the [April Keras community meeting slides](https://drive.google.com/file/d/1uY4aJtYYnHOgmvBQO5yz_o36dsah4_8u/view) for details Because old versions of Keras, including the current stable version, require the old import form, this PR changes every file importing internal Keras symbols to try both import forms: first it tries importing the new version (with `keras.src`), and if that fails, it tries importing the old version (with `keras`)

Without this change, TF Addons cannot be imported with the latest `keras-nightly` pip package.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New Tutorial
- [ ] Additional Testing

# Checklist:

# How Has This Been Tested?

*  Ran the official TensorFlow models with the updated TensorFlow Addons package, confirmed it worked correctly.

### DIFF
--- a/tensorflow_addons/optimizers/discriminative_layer_training.py
+++ b/tensorflow_addons/optimizers/discriminative_layer_training.py
@@ -21,8 +21,14 @@ import tensorflow as tf
 from tensorflow_addons.optimizers import KerasLegacyOptimizer
 from typeguard import typechecked
 
-from keras import backend
-from keras.utils import tf_utils
+try:
+    # New versions of Keras require importing from `keras.src` when
+    # importing internal symbols.
+    from keras.src import backend
+    from keras.src.utils import tf_utils
+except ImportError:
+    from keras import backend
+    from keras.utils import tf_utils
 
 
 @tf.keras.utils.register_keras_serializable(package="Addons")
--- a/tensorflow_addons/optimizers/discriminative_layer_training.py
+++ b/tensorflow_addons/optimizers/discriminative_layer_training.py
@@ -18,15 +18,16 @@ from typing import List, Union
 
 import tensorflow as tf
 
+from packaging.version import Version
 from tensorflow_addons.optimizers import KerasLegacyOptimizer
 from typeguard import typechecked
 
-try:
+if Version(tf.__version__).release >= Version("2.13").release:
     # New versions of Keras require importing from `keras.src` when
     # importing internal symbols.
     from keras.src import backend
     from keras.src.utils import tf_utils
-except ImportError:
+else:
     from keras import backend
     from keras.utils import tf_utils
 
--- a/tensorflow_addons/utils/test_utils.py
+++ b/tensorflow_addons/utils/test_utils.py
@@ -27,7 +27,12 @@ from tensorflow_addons import options
 from tensorflow_addons.utils import resource_loader
 
 if Version(tf.__version__) >= Version("2.9"):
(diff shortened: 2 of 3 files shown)


reedwm:
CC @bhack

tensorflow_addons/optimizers/discriminative_layer_training.py
```diff
@@ -21,8 +21,14 @@
 from tensorflow_addons.optimizers import KerasLegacyOptimizer
 from typeguard import typechecked
 
-from keras import backend
-from keras.utils import tf_utils
+try:
+    # New versions of Keras require importing from `keras.src` when
+    # importing internal symbols.
+    from keras.src import backend
```
sampathweb, line 27: Do we need this line?  Shouldn't `from keras import backend` work since its public API?

reedwm: Below, on line 144, `backend._current_graph()` is called. Since this is an internal symbol, it requires the `keras.src` import. `from keras import backend` will only allow public symbols to be accessed on `backend`, while `from keras.src import backend` allows all symbols from `backend.py` to be accessed.

sampathweb: Thanks for the details.  Sounds good.

sampathweb (dismissed):

bhack:
Are you using try/except as it is faster instead of checking the version?

reedwm:
> Are you using try/except as it is faster instead of checking the version?

No, I use try-except because the issue only affects recent `keras-nightly` versions, and users with older `keras-nightly` versions need the old import. I don't think there's an easy way to check if the `keras-nightly` commit is newer than a certain commit.

reedwm:
I think, but that version only refers to the stable version. The try-except will work on any given commit of Keras, which allows any version of `keras-nightly` to work. This is a fairly minor use case, so I can switch to a version check if you prefer.

bhack:
Right, It is that keras-nightly is a quite ambiguous concept.

See my old point (2021) at:

https://github.com/keras-team/keras/issues/14867#issuecomment-876002266

I am ok with both. I am only worried that in the future users will be surprised by the wrong fallback import if the env is broken.

/cc @seanpmorgan what do you think?

reedwm:
@bhack I switched to checking `tf.__version__` instead of using a try-except statement. I'm happy to switch it back if you prefer the try-except statement. The `tf.__version__` check means older version of `keras-nightly` will not be supported, but this is fine as users should expect to have to update the nightly TF and Keras packages if they also update TF-Addons.

The issue is causing various things to break, such as the official models, so it would be great if this could get merged quickly. Since @seanpmorgan hasn't responded yet, can you review this without @seanpmorgan's approval?

sampathweb (approved):

bhack:
No problem, I am re-running two failed job. The I'll merge it.

bhack (approved):

reedwm:
Thanks @bhack! Agreed, without CI coverage we cannot make any compatibility guarantees.

Issue #2849, opened 2023-09-13
Cannot build tensorflow addons with TF 2.14


trevor-m:
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 20.04): manylinux2014
- TensorFlow version and how it was installed (source or binary): 2.14-rc1 from source
- TensorFlow-Addons version and how it was installed (source or binary): 0.21.0 from source
- Python version: 3.10
- Is GPU used? (yes/no): yes

**Describe the bug**

Bazel build fails with the following error which looks related to the recent hermetic python changes in TF2.14.
```
#43 71.88 Analyzing: target //:build_pip_pkg (1 packages loaded, 0 targets configured)
#43 72.90 Analyzing: target //:build_pip_pkg (20 packages loaded, 101 targets configured)
#43 73.05 INFO: Repository go_sdk instantiated at:
#43 73.05   /opt/tensorflow/tf-addons/WORKSPACE:32:14: in <toplevel>
#43 73.05   /root/.cache/bazel/_bazel_root/81bbe6448980452f7a50e79b39fc5624/external/org_tensorflow/tensorflow/workspace0.bzl:135:20: in workspace
#43 73.05   /root/.cache/bazel/_bazel_root/81bbe6448980452f7a50e79b39fc5624/external/com_github_grpc_grpc/bazel/grpc_extra_deps.bzl:36:27: in grpc_extra_deps
#43 73.05   /root/.cache/bazel/_bazel_root/81bbe6448980452f7a50e79b39fc5624/external/io_bazel_rules_go/go/private/sdk.bzl:431:28: in go_register_toolchains
#43 73.05   /root/.cache/bazel/_bazel_root/81bbe6448980452f7a50e79b39fc5624/external/io_bazel_rules_go/go/private/sdk.bzl:130:21: in go_download_sdk
#43 73.05 Repository rule _go_download_sdk defined at:
#43 73.05   /root/.cache/bazel/_bazel_root/81bbe6448980452f7a50e79b39fc5624/external/io_bazel_rules_go/go/private/sdk.bzl:117:35: in <toplevel>
[...]

PR #2855, opened 2023-10-18
[WIP] Build against tf2.14 bazel strategy

seanpmorgan:
# Description

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,6 +29,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
+          TF_PYTHON_VERSION: ${{ env.MIN_PY_VERSION }}
           python-version: ${{ env.MIN_PY_VERSION }}
       - name: Build wheels
         run: |
@@ -87,6 +88,7 @@ jobs:
         env:
           OS: ${{ runner.os }}
           PY_VERSION: ${{ matrix.py-version }}
+          TF_PYTHON_VERSION: ${{ matrix.py-version }}
           TF_VERSION: ${{ matrix.tf-version }}
           NIGHTLY_TIME: ${{ steps.author-date.outputs.result }}
           CPU: ${{ matrix.cpu }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,7 +37,7 @@ jobs:
           pip install --default-timeout=1000 -r tools/install_deps/pytest.txt -r tools/install_deps/tensorflow-cpu.txt -r requirements.txt
           bash tools/install_deps/install_bazelisk.sh ./
           python configure.py
-          bazel test -k --test_timeout 300,450,1200,3600 --test_output=errors //tensorflow_addons/...
+          bazel test --action_env PYENV_VERSION=3.9  -k --test_timeout 300,450,1200,3600 --test_output=errors //tensorflow_addons/...
   release-wheel:
     name: Test and build release wheels
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,9 +29,10 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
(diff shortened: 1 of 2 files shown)


bhack:
We have the same issue when I've introduced this in the release PR. It seems that pytest is not installed in the same python env where we are executing tests with bazel.

trevor-m:
Fixes #2849

seanpmorgan:
@angerson we (and other SIGs / downstream consumers) have had issues with the the hermetic python bazel rule. Can you or someone in SIG build take a look at this failing test when time allows:

1. Our workspace `install_deps` call does not properly install the 3 requirements files that we have specified, since we error on `ModuleNotFoundError: No module named 'pytest'` 
2. We need to run a `py_binary()` for `configure.py` using the hermetic python after the install deps are available instead of using the system python install as is down in the test above. 

We tried a few things on the last release... but we want to get ahead of this for upcoming releases when the new workspace strategy may be required  (I see that we're still publishing multiple py docker images for future tf releases though).

georgiyekkert:
@seanpmorgan Why do you pull tensorflow via http_archive and also via pip?  You started doing it not so long ago - https://github.com/tensorflow/addons/commit/1c3a2b2b0c5ca902e056427ec943e326cf6b70c1 
Looking at the code, you should install it only via pip.

seanpmorgan:
> @seanpmorgan Why do you pull tensorflow via http_archive and also via pip? You started doing it not so long ago - [1c3a2b2](https://github.com/tensorflow/addons/commit/1c3a2b2b0c5ca902e056427ec943e326cf6b70c1) Looking at the code, you should install it only via pip.

Because we load the 4 bazel workspace files that are provided since we compile custom CPU and CUDA ops. IIRC when on TF2.9 there was rev in CUDA version and our previous local toolchain broke. We started loading the workspace files from tensorflow to build the same way.

georgiyekkert:
> > @seanpmorgan Why do you pull tensorflow via http_archive and also via pip? You started doing it not so long ago - [1c3a2b2](https://github.com/tensorflow/addons/commit/1c3a2b2b0c5ca902e056427ec943e326cf6b70c1) Looking at the code, you should install it only via pip.
> 
> https://github.com/tensorflow/addons/blob/master/WORKSPACE#L18-L32
> 
> Because we load the 4 bazel workspace files that are provided since we compile custom CPU and CUDA ops. IIRC when on TF2.9 there was rev in CUDA version and our previous local toolchain broke. We started loading the workspace files from tensorflow to build the same way.

There are 3 ways to fix it:
1. Remove pulling tensorflow via http_archive - you are bringing a lot more than you need when you load all TF dependencies. Load only what you need. IMO this is the correct way
2. Since EOL is planned in May 2024, we can patch tensorflow, so it keeps using non-hermetic Python. You can use the same patch as TF-federated - https://github.com/tensorflow/federated/commit/8580528acd9227ff1fe84c9ab95e6a483f2b5dfe#diff-4d440744ea1e145db1f807bb811e815b7e0526db8cc2dfdbba441ef80ceb9e81. This is the easiest way
3. Use the hermetic python in your build. Currently you have errors like "ModuleNotFoundError: No module named 'pytest'" because you are not loading "pip" deps correctly, see https://rules-python.readthedocs.io/en/latest/pypi-dependencies.html#using-third-party-packages-as-dependencies. Each "pip" dependency must be specified explicitly in the BUILD file where needed. E.g. in tensorflow https://github.com/tensorflow/tensorflow/blob/a17ef94e5144fb0fe8447d7ae2e1b32f367a584c/tensorflow/python/distribute/failure_handling/BUILD#L50 https://github.com/tensorflow/tensorflow/blob/a17ef94e5144fb0fe8447d7ae2e1b32f367a584c/tensorflow/python/distribute/failure_handling/failure_handling_util.py#L19. Since you are pulling TF via both http_archive and pip, making configure.py work the same way as it is now might be tricky.

bhack:
@georgiyekkert Can you send a PR on this repo directly or over this PR's source branch? 
Thanks

seanpmorgan:
https://github.com/tensorflow/addons/pull/2856

PR #2856, opened 2023-11-09
Build with TF 2.14

georgiyekkert:
# Description
Create a patch to use non_hermetic python in TF, bump workspace TF version to 2.14.0

### DIFF
--- a/build_deps/tf_dependency/tf.patch
+++ b/build_deps/tf_dependency/tf.patch
@@ -1,0 +1,74 @@
+diff --git tensorflow/tools/toolchains/cpus/aarch64/aarch64_compiler_configure.bzl tensorflow/tools/toolchains/cpus/aarch64/aarch64_compiler_configure.bzl
+index a2bdd6a7eed..ec25c23d8d4 100644
+--- tensorflow/tools/toolchains/cpus/aarch64/aarch64_compiler_configure.bzl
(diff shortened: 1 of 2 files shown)


georgiyekkert:
@bhack PTAL

bhack:
@georgiyekkert Thanks. 
@seanpmorgan Probably it is better to review and merge this before starting the 2.15 PR.

build_deps/tf_dependency/tf.patch
```diff
@@ -0,0 +1,74 @@
+diff --git tensorflow/tools/toolchains/cpus/aarch64/aarch64_compiler_configure.bzl tensorflow/tools/toolchains/cpus/aarch64/aarch64_compiler_configure.bzl
```
bhack, line 1: @seanpmorgan NIP: There could be the risk that this could be invalidated on 2.15 and we need to manually regenerate that patch on new releases.

bhack: As expected this approach will a short life (/cc @MichaelHudgins):
https://github.com/tensorflow/tensorflow/commit/ce4adb5e12eb3ad180aba7162772b51b029b3fcb

seanpmorgan: Yes, but it appears that other repos (e.g. tf/federated) depend on this too so I'm okay with trying this strategy for the couple of releases we have left

seanpmorgan (approved):
LGTM, Thanks!

PR #2857, opened 2023-11-20
Build against TF 2.15

seanpmorgan:


seanpmorgan:
Looks like a header is missing for Windows TF2.15:
https://github.com/tensorflow/tensorflow/issues/61830

@bhack @angerson I think intel builds the windows whls now?

bhack:
Yes I agree to remove it at least if something will not change with this ticket:

https://github.com/tensorflow/tensorflow/issues/59918

I think that the Intel team can release a window version of Addons compatible with their window TF artifacts if requested.

bhack:
I think we could add a string in the Readme.

seanpmorgan:
> I think we could add a string in the Readme.

Added. Hoping that we can just add it back with a patch release but the issue is a few months old now without any sign of fix coming

bhack:
> Added. Hoping that we can just add it back with a patch release but the issue is a few months old now without any sign of fix coming

Yes we could just subscribe to the upstream ticket.

bhack (approved):

PR #2858, opened 2023-12-04
Fix tf-addons for upcoming keras 3 default.

cantonios:
# Description

Keras 3.0 will become default in TF 2.16 (and is currently default in tf-nightly).  This breaks this tf-addons package.  Here we make minimal changes to return functionality in a backward-compatible way.

Fixes # (issue)

None known reported.

## Type of change

- [X] Bug fix

# Checklist:

- [X] I've properly [formatted my code according to the guidelines](https://github.com/tensorflow/addons/blob/master/CONTRIBUTING.md#coding-style)
    - [] By running Black + Flake8
    - [X] By running pre-commit hooks

# How Has This Been Tested?

Imported with tf-nightly, keras-nightly, tf_keras-nightly, setting both `TF_USE_LEGACY_KERAS=0` and `1`.

### DIFF
--- a/tensorflow_addons/optimizers/lazy_adam.py
+++ b/tensorflow_addons/optimizers/lazy_adam.py
@@ -149,3 +149,6 @@ class LazyAdam(adam_optimizer_class):
         }
 
         return resource_scatter_op(**resource_update_kwargs)
+
+    def get_config(self):
+        return super().get_config()
--- a/tensorflow_addons/rnn/abstract_rnn_cell.py
+++ b/tensorflow_addons/rnn/abstract_rnn_cell.py
@@ -1,0 +1,133 @@
+# Copyright 2023 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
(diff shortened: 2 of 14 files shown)


tensorflow_addons/optimizers/lazy_adam.py
```diff
@@ -149,3 +149,6 @@ def _resource_scatter_operate(self, resource, indices, update, resource_scatter_
         }
 
         return resource_scatter_op(**resource_update_kwargs)
+
+    def get_config(self):
```
qlzh727, line 153: Any reason this is needed?

tensorflow_addons/rnn/abstract_rnn_cell.py
```diff
@@ -0,0 +1,133 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Base class for RNN cells.
+
+Adapted from legacy github.com/keras-team/tf-keras.
```
qlzh727, line 17: What's the reason this is needed? Is it because of the usage of tf.keras.layers?

tensorflow_addons/optimizers/lazy_adam.py
cantonios: Yes, with keras 3 it throws an error that `LazyAdam` doesn't have a `get_config` method.

tensorflow_addons/rnn/abstract_rnn_cell.py
cantonios: The `AbstractRNNCell` only exists in `tf_keras` (it previously existed in keras 2, but doesn't exist in keras 3).

qlzh727 (approved):
Chatted offilne. Most of the duplication/copy here is that they are not available in keras 3.0 if installed.

seanpmorgan (approved):
LGTM thanks for the PR ahead of time

PR #2866, opened 2024-04-19
[WIP] Build against TF2.16

seanpmorgan:
# Description

Build against TF2.16


seanpmorgan:
Breaking on build_deps patch for the new version:
https://github.com/tensorflow/addons/pull/2856

Looks like federated did not update theirs yet either:
https://github.com/tensorflow/federated/blob/main/third_party/tensorflow/python_toolchain.patch

@georgiyekkert not sure if you're still working on this project, but would you be able to submit a final patch so we can make our last release?

bhack:
@seanpmorgan It seems that TFio updated to 2.16.1 without the patch:
https://github.com/tensorflow/io/blob/master/WORKSPACE

bhack:
@seanpmorgan It is also mandatory to support cuda 12.x for the new TF. As we are approaching at the end of the month let me know what we want to do here cause I suppose there is still enough work required to prepare a release.

seanpmorgan:
Closing in favor of #2870 which utilizes the same build patch we've been using.

PR #2870, opened 2024-05-27
[WIP] Build tf216  (Keras3)

seanpmorgan:
# Description
Looks like an overwhelming amount of breaking changes on the new Keras3 default. To me, aligning with Keras3 is out of scope for the final release unless someone from the TF team wants to assist

@bhack I'm in agreement that this would require pretty major overhaul and we don't want to encourage Keras3 usage of this library... so I think we keep TFA 0.23 as the final release.

https://github.com/tensorflow/addons/actions/runs/9260858538/job/25475309267?pr=2870


supercharleszhu:
> # Description
> Looks like an overwhelming amount of breaking changes on the new Keras3 default. To me, aligning with Keras3 is out of scope for the final release unless someone from the TF team wants to assist
> 
> @bhack I'm in agreement that this would require pretty major overhaul and we don't want to encourage Keras3 usage of this library... so I think we keep TFA 0.23 as the final release.
> 
> https://github.com/tensorflow/addons/actions/runs/9260858538/job/25475309267?pr=2870

Hi @seanpmorgan we are interested in contributing and making tf addons compatible with tf_keras with 2.16 (instead of keras 3.0). Wondering whether you have any ideas on what is the remaining todo items?

